PR #18, opened 2018-11-05
Added home button with zoom to layers extent functionality

sklencar:
The button is temporally set to MainPanel with broken design. Design fix has been omitted on purpose for later work.

### DIFF
--- a/app/loader.cpp
+++ b/app/loader.cpp
@@ -14,6 +14,7 @@
  ***************************************************************************/
 
 #include "loader.h"
+#include "qgsvectorlayer.h"
 #include <QDebug>
 
 Loader::Loader(QObject* parent):QObject(parent)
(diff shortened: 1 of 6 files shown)


app/loader.cpp
```diff
@@ -35,3 +36,19 @@ void Loader::load(const QString& filePath) {
         qDebug() << " ******** SKIPPED ";
     }
 }
+
+void Loader::zoomToDesign(QgsQuickMapSettings *mapSettings)
```
wonder-sk: Rename to zoomToProject

app/loader.cpp
```diff
@@ -35,3 +36,19 @@ void Loader::load(const QString& filePath) {
...
+{
+    if (!mapSettings) {
+        qDebug() << "Cannot zoom to layers extent, mapSettings is not defined";
+        return;
+    }
+
+    const QVector<QgsVectorLayer*> layers = mProject.layers<QgsVectorLayer*>();
```
wonder-sk: do not consider just vector layers

app/loader.cpp
```diff
@@ -35,3 +36,19 @@ void Loader::load(const QString& filePath) {
...
+    QgsRectangle extent;
+    for (const QgsVectorLayer* layer: layers) {
+        QgsRectangle layerExtent = layer->extent();
+        extent.combineExtentWith(layerExtent);
+    }
+    mapSettings->setExtent(extent);
```
wonder-sk: add a bit of extra margin (e.g. 5%) around the extent for nicer result

app/qml/MainPanel.qml
```diff
@@ -54,6 +55,21 @@ Item {
...
+            source: "home.svg"
+            fillMode: Image.PreserveAspectFit
+            clip:true
+            anchors.verticalCenter: parent.verticalCenter
+
+            MouseArea {
+                anchors.fill: parent
+                onClicked: {
+                    __loader.zoomToDesign(mainPanel.mapSettings)
+                }
```
wonder-sk: Better than adding mapSettings as a property here to MainPanel (it does not need information about mapSettings) let's add another signal and do the zoom() call in main.qml

PR #19, opened 2018-11-05
Feature/support map themes

sklencar:
Themes are listed in separate panel for now, onClicked layers are reloaded with assigned theme. Button is similar to "ProjectButton" and design fixes have been omitted due to future work.


app/mapthemesmodel.cpp
```diff
@@ -0,0 +1,106 @@
...
+    return;
+
+  mMapThemes = mapThemes;
+  emit mapThemesChanged();
+}
+
+void MapThemesModel::applyTheme( const QString &name )
+{
+    QgsLayerTree* root = mProject->layerTreeRoot();
+    QgsLayerTreeModel* model = new QgsLayerTreeModel(root);
```
wonder-sk: don't forget to delete the model afterwards. Or even better, have it on the stack (`QgsLayerTreeModel model(root);`)

app/mapthemesmodel.cpp
```diff
@@ -0,0 +1,106 @@
...
+        endResetModel();
+        emit mapThemesChanged();
+    }
+}
+
+QVariant MapThemesModel::data( const QModelIndex& index, int role ) const
+{
+    int row = index.row();
+    if (row < 0 || row >= mMapThemes.count())
+        return QVariant("");
```
wonder-sk: Better to return invalid variant (`QVariant()`) here than passing an empty string.

PR #30, opened 2018-11-12
Feature/project screen

sklencar:
Project screen according mobile design. Font and elements size is currently not properly calculated for scaling for all displays. More love needed for having proper size ratio for text according reference design.

screenshot_2018-11-12-10-00-24-909_uk co lutraconsulting


app/qml/main.qml
```diff
@@ -183,9 +185,10 @@ ApplicationWindow {
 
     OpenProjectPanel {
         id: openProjectPanel
-        height: window.height
-        width: QgsQuick.Utils.dp * 600
-        edge: Qt.LeftEdge
+        height: window.height - mainPanel.height
```
wonder-sk: I think it should take the whole window height. Different pages will have different content of their toolbar, so I would suggest that each page has its own toolbar (otherwise having one toolbar shared for all pages will quickly get difficult)

app/qml/OpenProjectPanel.qml
```diff
@@ -1,68 +1,250 @@
 import QtQuick 2.7
 import QtQuick.Controls 2.2
+import QtQuick.Controls.Styles 1.4
```
wonder-sk: How is this import used?

app/qml/OpenProjectPanel.qml
```diff
@@ -1,68 +1,250 @@
...
+                height: projectMenuButtons.height
+                id: button1
+                anchors.bottom: parent.bottom
+                anchors.bottomMargin: 0
+                anchors.top: parent.top
+                anchors.topMargin: 0
+                background: Rectangle {
+                    anchors.fill: parent
+                    color: "transparent"
+                }
```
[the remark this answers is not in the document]
sklencar: If color is not overwritten, its set to white. If TabButton background is not overwritten, its set to some dark grey color (could depend on qtquickcontrols2.conf)

PR #56, opened 2018-11-23
Feature/default survey layer

### LINKED ISSUE #37: Default survey layer selection
It should be possible to have an option to pick the default survey layer - when one is active, Record button would skip asking for the survey layer.

Suggested two ways of setting default survey layer:
1. tap and hold record button
2. in settings page

It should be also possible to unset default survey layer, so that record button would ask for survey layer again.

sklencar:
Default survey layer can be changed on hold of RecordBtn or in Settings panel
This setting is not persistent, so it resets after restart of app or changing a project (for now)
closes #37


app/layersmodel.h
```diff
@@ -52,16 +53,20 @@ class LayersModel : public QAbstractListModel
...
     void layersChanged();
+    void defaultLayerIndexChanged();
 
   public slots:
     void reloadLayers(); //when project file changes, reload all layers, etc.
 
   private:
     QgsProject* mProject;
     QList<QgsMapLayer*> mLayers; // all layers
+    int mDefaultLayerIndex;
```
wonder-sk: The variable is uninitialized... Initialize it either here or in the default constructor

app/qml/ActiveLayerPanel.qml
```diff
@@ -6,12 +6,32 @@ import "."  // import InputStyle singleton
...
     property string activeLayerName: __layersModel.data(__layersModel.index(activeLayerIndex), LayersModel.Name)
+    property string defaultLayerName: __layersModel.data(__layersModel.index(__layersModel.defaultLayerIndex), LayersModel.Name)
+    property string title: "Survey Layer"
 
     signal layerSettingChanged()
 
+    function openPanel(state) {
+        activeLayerPanel.state = state
+        if (state === "setup") {
+
```
wonder-sk: Is the empty body here intentional? can we remove it?

app/qml/ActiveLayerPanel.qml
```diff
@@ -22,6 +42,22 @@ Drawer {
         color: InputStyle.clrPanelMain
     }
 
+    Item {
```
wonder-sk: do we need this sub-item to define states? can't be put the states to the root ActiveLayerPanel item?

app/qml/main.qml
```diff
@@ -23,8 +23,23 @@ ApplicationWindow {
                 )
     }
 
+    function saveRecordedFeature(pair) {
+        if (pair.valid) {
+            highlight.featureLayerPair = pair
+            featurePanel.show_panel(pair, "Add", "form")
+        } else {
+            popup.text = "Recording feature is not valid"
```
wonder-sk: also wrap into qsTr()?

app/qml/ActiveLayerPanel.qml
```diff
@@ -22,6 +42,22 @@ Drawer {
...
+        id: stateManager
+        states: [
+            State {
+                name: "setup"
+                PropertyChanges { target: layerPanel; title: "Default survey layer"}
```
wonder-sk: qsTr() here and in the title of the other state?

app/qml/ActiveLayerPanel.qml
```diff
@@ -22,6 +42,22 @@ Drawer {
...
+            }
+
+           ,State {
```
wonder-sk: Just for better aesthetic impression - can we have the comma right after the closing brace above? :)

app/qml/ActiveLayerPanel.qml
```diff
@@ -22,6 +42,22 @@ Drawer {
...
+                name: "record"
+                PropertyChanges { target: layerPanel; title: "Survey layer" }
```
wonder-sk: If we have just two states here and we only change title, maybe it would be easier to just use `title: state == "setup" ? "default survey layer" : "survey layer"` instead of using `PropertyChanges` but I don't have a strong opinion about that...

app/qml/ActiveLayerPanel.qml
sklencar: The panel itself i a drawer which is inherited from Popup which is inherited from QtObject. None of those supports state/states property.

PR #68, opened 2018-11-27
Bug/stuck highlight

### LINKED ISSUE #61: Highlight stays visible when preview panel is closed
1. click on a feature - preview panel is shown with feature info
2. click somewhere else on the map where nothing is shown - preview panel will get closed but the highlight of the feature is still there.

### LINKED ISSUE #64: No trace of recorded feature when recording linestrings
When I start recording a new feature for a linestring layer, there is no highlight behind the GPS position that would indicate what has been recorded. Only when I finish recording, it appears.

Second problem: when I clicked "back" button in the feature form, the highlight stays there (it should have been removed).

sklencar:
closes #64 closes #61


app/qml/main.qml
```diff
@@ -158,6 +158,17 @@ ApplicationWindow {
         layer: activeLayerPanel.activeVectorLayer
     }
 
+    // Highlighting a new feature while digitizing
+    Connections {
+        target: digitizing.recordingFeatureModel
+        onFeatureLayerPairChanged: {
+            if (!highlight.visible) {
+                highlight.visible = true
+            }
```
wonder-sk: why not just "visible = true" instead of "if (!visible) visible = true" ?

PR #91, opened 2018-12-03
Bug/preview edit mode

### LINKED ISSUE #76: Click on preview panel vs click on edit button
There does not seem to be a difference between clicking the edit button in preview panel and anywhere else in the preview panel - both show the feature panel in the same state. If I click a previously added feature, I always get feature panel in editing mode (buttons delete/save/add photo), if I click any other layer, I always get read-only feature panel (with just edit button).

sklencar:
closes #76

### DIFF
--- a/app/qml/PreviewPanel.qml
+++ b/app/qml/PreviewPanel.qml
@@ -15,6 +15,7 @@ Item {
     property variant previewFields: []
 
     signal contentClicked()
+    signal editClicked()
 
     MouseArea {
         anchors.fill: parent
@@ -58,6 +59,15 @@ Item {
                         anchors.left: titleText.right
                         anchors.right: parent.right
 
+                        MouseArea {
+                            anchors.fill: parent
+                            propagateComposedEvents: true
+                            onClicked: {
+                                previewPanel.editClicked()
+                                mouse.accepted = false
+                            }
+                        }
+
                         Image {
                             id: icon
                             anchors.fill: parent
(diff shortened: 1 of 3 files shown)


app/qml/PreviewPanel.qml
```diff
@@ -92,6 +102,12 @@ Item {
                 anchors.top: header.bottom
                 anchors.bottom: parent.bottom
 
+                MouseArea {
+                    anchors.fill: parent
+                    onClicked: contentClicked()
+                    z: content.z + 1
+                }
```
wonder-sk: why do we need this here when there is already a mouse area for the full size (around line 20) ?

app/qml/PreviewPanel.qml
```diff
@@ -58,6 +59,15 @@ Item {
...
                         anchors.right: parent.right
 
+                        MouseArea {
+                            anchors.fill: parent
+                            propagateComposedEvents: true
+                            onClicked: {
+                                previewPanel.editClicked()
+                                mouse.accepted = false
+                            }
+                        }
```
wonder-sk: I would prefer to avoid the tricks with propagateComposedEvents... In fact when clicked on edit icon, we only want editClicked() to be triggered, not both editClicked() and contentClicked(). Then in onEditClicked() handler we can simply add a line with `stateManager.state = "form"` to switch to full-size window. This gives us also more flexibility for the future and the code would be easier to understand.

PR #110, opened 2018-12-07
GPS marker position animation

### LINKED ISSUE #78: GPS marker movement with animation
It would be nice if we moved the GPS marker position with some animation so that it does not "jump" around

sklencar:
Some improvements could be done on timing or transition type (currently linear is used). However with small position changes different effect is not much visible. Feel free to suggest any changes.

closes #78

### DIFF
--- a/app/qml/PositionMarker.qml
+++ b/app/qml/PositionMarker.qml
@@ -11,7 +11,7 @@ Item {
     property QgsQuick.PositionKit positionKit
     property color baseColor: InputStyle.highlightColor
     property bool withAccuracy: true
-    // animated position
+    // animated position in screen coords
     property real animatedPosX
     property real animatedPosY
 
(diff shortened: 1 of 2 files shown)


app/qml/main.qml
```diff
@@ -238,6 +239,9 @@ ApplicationWindow {
         target: mapCanvas.mapSettings
         onExtentChanged: {
             scaleBar.visible = true
+            var pos = mapCanvas.mapSettings.coordinateToScreen( positionKit.projectedPosition)
+            positionMarker.animatedPosX = pos.x
+            positionMarker.animatedPosY = pos.y
```
wonder-sk: Rather than configuring some internal properties of position marker here, this should go to PositionMarker implementation so that all the positioning code is in one place

app/qml/PositionMarker.qml
```diff
@@ -11,6 +11,42 @@ Item {
...
+        animateX.restart()
+
+        animateY.from = animatedPosY
+        animateY.to = posNext.y
+        animateY.restart()
+    }
+
+    Component.onCompleted: {
+        animatedPosX =  positionKit.projectedPosition.x
+        animatedPosY = positionKit.projectedPosition.y
```
wonder-sk: shouldn't this use coordinates transformed to screen? like in the code on canvas extents change?

app/qml/PositionMarker.qml
```diff
@@ -11,6 +11,42 @@ Item {
...
+    }
+
+    NumberAnimation {
+        id: animateX
+        target: positionMarker
+        properties: "animatedPosX"
+        duration: 1000
+        easing {type: Easing.Linear}
```
wonder-sk: I would suggest to use a shorter duration e.g. 500 so that the marker isn't constantly moving and `InOutQuad` easing curve for more pleasant animation

app/qml/PositionMarker.qml
```diff
@@ -11,6 +11,42 @@ Item {
     property QgsQuick.PositionKit positionKit
     property color baseColor: InputStyle.highlightColor
     property bool withAccuracy: true
+    // animated position
+    property real animatedPosX
+    property real animatedPosY
```
wonder-sk: would be good to note in comments whether these are screen coordinates or map coordinates

PeterPetrik:
@sklencar can you rebase this with latest master?

sklencar:
Obsolete PR, closing.
Can be reopen and updated later if needed.

PR #118, opened 2018-12-11
Mergin/list all projects

### LINKED ISSUE #117: [Mergin] List all projects

sklencar:
Listing input projects from mergin with basic auth - the token has to be defined in config.pri
closes #117


app/merginapi.cpp
```diff
@@ -0,0 +1,73 @@
...
+      mMerginProjects = parseProjectsData(data);
+    }
+    else {
+        // TODO propagate error
+        QString message = QStringLiteral("Network API error: %1(): %2").arg("listProjects", r->errorString());
+        qDebug("%s", message.toStdString().c_str());
+    }
+
+
+    mModel->resetProjects(mMerginProjects);
```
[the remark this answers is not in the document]
sklencar: Can handling be done only in cpp through MerginApi signal with parameter - ProjectList or is it better to send empty signal and handle it in QML?

app/merginapi.h
```diff
@@ -1,34 +1,38 @@
...
 signals:
     void listProjectsFinished();
+    void networkErrorOccurred(QString errorMessage);
 
 private slots:
     void listProjectsReplyFinished();
 
 private:
-    ProjectList parseProjectsData( QByteArray data );
+    ProjectList parseProjectsData( const QByteArray data );
```
wonder-sk: const QByteArray &

app/merginapi.cpp
```diff
@@ -43,15 +53,26 @@ void MerginApi::listProjectsReplyFinished()
...
+        QtAndroid::runOnAndroidThread([errMsg] {
+            QAndroidJniObject javaString = QAndroidJniObject::fromString(errMsg);
+            QAndroidJniObject toast = QAndroidJniObject::callStaticObjectMethod("android/widget/Toast", "makeText",
+                                                                                "(Landroid/content/Context;Ljava/lang/CharSequence;I)Landroid/widget/Toast;",
+                                                                                QtAndroid::androidActivity().object(),
+                                                                                javaString.object(),
+                                                                                jint(1));
+            toast.callMethod<void>("show");
+        });
+#endif
```
wonder-sk: This should go to a separate utility class e.g. AndroidUtils, as a method showToast(QString text)

Here we should only set our internal error (mLastError) - and somewhere else whoever listens to networkErrorOccurred() should check whether last error is empty, and show the toast (on Android) / message box (otherwise)... right now on desktop build one wouldn't get any warning at all

PR #122, opened 2018-12-18
Mergin/download project

sklencar:
Downloading a project from Mergin.
Behavior after downloading a project has not been discussed yet, so after download, a project is not loaded, just added into "My projects" list. Downloading is visualise with busyIndicator, no other indications (e.g. project is already downloaded, downloaded project is out of date).
Added cache file for listed project.

closes #119

PR #123, opened 2018-12-21
Mergin/download_project_data_stream

### LINKED ISSUE #119: [Mergin] Download a project

### LINKED ISSUE #121: [Mergin] Update project

sklencar:
Downloading project from data stream reading chunks and parsing files. Data handling implementation is similar to one in Mergin. Same signal is emitted for download/update project. If download/update request got data, obsolate files (extra files + modified files) are deleted. 
MerginApi is using a new instance of Mergin where "updated" value has been added for listing projects. Projects list is cached in a local file (.cachedProjects.txt) which is updated after any project list request or successful download/update request. Project status heavily depends on this file and list of Projects from the server. If is missing, mergin projects appears as with "noVersion" projectStatus otherwise server and local updated values are compared to get projectStastus. According projectStatus, the app sends either donwload or update requests.

closes #119 closes #121 
This PR makes #122 downloading project with zip out of date.

### DIFF
--- a/app/main.cpp
+++ b/app/main.cpp
@@ -240,11 +240,12 @@ int main(int argc, char *argv[])
 #ifdef MERGIN_TOKEN
   merginToken.append(STR(MERGIN_TOKEN));
 #endif
-  MerginApi* ma=  new MerginApi(QString("https://mergin.dev.cloudmappin.com"), merginToken );
-  engine.rootContext()->setContextProperty( "__merginApi", ma );
+
+  std::unique_ptr<MerginApi> ma =  std::unique_ptr<MerginApi>(new MerginApi(QString("https://mergin.dev.cloudmappin.com"), dataDir, merginToken ));
+  engine.rootContext()->setContextProperty( "__merginApi", ma.get() );
 
   // Create mergin projects model
-  MerginProjectModel mpm( ma );
+  MerginProjectModel mpm;
   engine.rootContext()->setContextProperty( "__merginProjectsModel", &mpm );
 
   // Connections
@@ -252,6 +253,9 @@ int main(int argc, char *argv[])
   QObject::connect(&loader, &Loader::projectReloaded, &lm, &LayersModel::reloadLayers);
   QObject::connect(&loader, &Loader::projectReloaded, &mtm, &MapThemesModel::reloadMapThemes);
   QObject::connect(&mtm, &MapThemesModel::reloadLayers, &lm, &LayersModel::reloadLayers);
+  QObject::connect(ma.get(), &MerginApi::downloadProjectFinished, &mpm, &MerginProjectModel::downloadProjectFinished);
+  QObject::connect(ma.get(), &MerginApi::downloadProjectFinished, &pm, &ProjectModel::addProject);
+  QObject::connect(ma.get(), &MerginApi::listProjectsFinished, &mpm, &MerginProjectModel::resetProjects);
 
 #ifdef ANDROID
   engine.rootContext()->setContextProperty( "__appwindowvisibility", "Maximized");
--- a/app/merginapi.cpp
+++ b/app/merginapi.cpp
@@ -5,27 +5,134 @@
 #include <QJsonObject>
 #include <QJsonArray>
 #include <QDate>
+#include <QByteArray>
+#include <QSet>
 
-
-MerginApi::MerginApi(const QString &root, QByteArray token, QObject *parent)
+MerginApi::MerginApi(const QString &root, const QString& dataDir, QByteArray token, QObject *parent)
   : QObject (parent)
   , mApiRoot(root)
+  , mDataDir(dataDir + "/downloads/")
+  , mCacheFile(".projectsCache.txt")
   , mToken(token)
 {
+    QObject::connect(this, &MerginApi::merginProjectsChanged,this, &MerginApi::cacheProjects);
 }
 
 void MerginApi::listProjects()
 {
-    mMerginProjects.clear();
+    if (mToken.isEmpty()) {
+        emit networkErrorOccurred( "Auth token is invalid", "Mergin API error: listProjects" );
+        return;
+    }
+
     QNetworkRequest request;
     // projects filtered by tag "input_use"
     QUrl url(mApiRoot + "/v1/project?tags=input_use");
-
     request.setUrl(url);
     request.setRawHeader("Authorization", QByteArray("Basic " + mToken));
 
     QNetworkReply *reply = mManager.get(request);
     connect(reply, &QNetworkReply::finished, this, &MerginApi::listProjectsReplyFinished);
+}
+
+void MerginApi::downloadProject(QString projectName)
+{
+    if (mToken.isEmpty()) {
+        emit networkErrorOccurred( "Auth token is invalid", "Mergin API error: downloadProject" );
+    }
+
+    QNetworkRequest request;
+    QUrl url(mApiRoot + "/v1/project/download/" + projectName);
+
+    if (mPendingRequests.contains(url)) {
+        QString errorMsg = QStringLiteral("Download request for %1 is already pending.").arg(projectName);
+        qDebug() << errorMsg;
+        emit networkErrorOccurred( errorMsg, "Mergin API error: downloadProject" );
+        return;
+    }
+
+    request.setUrl(url);
+    request.setRawHeader("Authorization", QByteArray("Basic " + mToken));
+
+    QNetworkReply *reply = mManager.get(request);
+    mPendingRequests.insert(url, projectName);
+    connect(reply, &QNetworkReply::finished, this, &MerginApi::downloadProjectReplyFinished);
+}
+
+void MerginApi::updateProject(QString projectName)
+{
+
+    if (mToken.isEmpty()) {
+        emit networkErrorOccurred( "Auth token is invalid", "Mergin API error: projectInfo" );
+    }
+
+    QNetworkRequest request;
+    QUrl url(mApiRoot + "/v1/project/" + projectName);
+
+    request.setUrl(url);
+    request.setRawHeader("Authorization", QByteArray("Basic " + mToken));
+
+    QNetworkReply *reply = mManager.get(request);
+    connect(reply, &QNetworkReply::finished, this, &MerginApi::updateInfoReplyFinished);
+
+}
+
+void MerginApi::downloadProjectFiles(QString projectName, QByteArray json)
+{
+    if (mToken.isEmpty()) {
+        emit networkErrorOccurred( "Auth token is invalid", "Mergin API error: fetchProject" );
+        return;
+    }
+
+    QNetworkRequest request;
+    QUrl url(mApiRoot + "/v1/project/fetch/" + projectName);
+    request.setUrl(url);
+    request.setRawHeader("Authorization", QByteArray("Basic " + mToken));
+    request.setRawHeader("Content-Type", "application/json");
+    request.setRawHeader("Accept", "application/json");
+    mPendingRequests.insert(url, projectName);
+
+    QNetworkReply *reply = mManager.post(request, json);
+    connect(reply, &QNetworkReply::finished, this, &MerginApi::downloadProjectReplyFinished);
+}
+
+ProjectList MerginApi::updateMerginProjectList(ProjectList serverProjects)
+{
+    QHash<QString, QDateTime> projectUpdates;
+    for (std::shared_ptr<MerginProject> project: mMerginProjects) {
+        projectUpdates.insert(project->name, project->updated);
+    }
+
+    for (std::shared_ptr<MerginProject> project: serverProjects) {
+        if (projectUpdates.contains(project->name)) {
+            QDateTime localUpdate = projectUpdates.value(project->name);
+            project->updated = localUpdate;
+            project->status = getProjectStatus(project->updated, project->serverUpdated);
+        }
+    }
+    return serverProjects;
+}
+
+void MerginApi::setUpdateToProject(QString projectName)
+{
+    for (std::shared_ptr<MerginProject> project: mMerginProjects) {
+        if (projectName == project->name) {
+            project->updated = project->serverUpdated;
+            emit merginProjectsChanged();
+            return;
+        }
+    }
+}
+
+void MerginApi::deleteObsoleteFiles(QString projectPath)
+{
+    if (!mObsoleteFiles.value(projectPath).isEmpty()) {
+        for (QString filename: mObsoleteFiles.value(projectPath)) {
+            QFile file (projectPath + filename);
+            file.remove();
+        }
+       mObsoleteFiles.remove(projectPath);
+    }
 }
 
 ProjectList MerginApi::projects()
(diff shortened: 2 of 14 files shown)


sklencar:
Added not-smart upload which downloads files which have been added, modified on the server. Deletes files which are extra (not matching with server files).

app/main.cpp
```diff
@@ -271,8 +271,9 @@ int main(int argc, char *argv[])
 #ifdef MERGIN_TOKEN
   merginToken.append(STR(MERGIN_TOKEN));
 #endif
-  MerginApi* ma=  new MerginApi(QString("https://mergin.dev.cloudmappin.com"), merginToken );
-  engine.rootContext()->setContextProperty( "__merginApi", ma );
+
+  std::shared_ptr<MerginApi> ma =  std::make_shared<MerginApi>(QString("https://mergin.dev.cloudmappin.com"), dataDir, merginToken );
```
wonder-sk: why shared_ptr - wouldn't be unique_ptr good enough here?

app/merginapi.cpp
```diff
@@ -5,29 +5,101 @@
 #include <QJsonObject>
 #include <QJsonArray>
 #include <QDate>
+#include <QByteArray>
+#include <QSet>
 
+#ifdef ANDROID
+#include <QtAndroid>
+#include <QAndroidJniObject>
+#endif
```
wonder-sk: should not be needed here anymore?

app/merginapi.cpp
```diff
@@ -5,29 +5,101 @@
...
+
+    request.setUrl(url);
+    request.setRawHeader("Authorization", QByteArray("Basic " + mToken));
+
+    QNetworkReply *reply = mManager.get(request);
+    mPendingRequests.insert(url, projectName);
+    connect(reply, &QNetworkReply::finished, this, &MerginApi::downloadProjectReplyFinished);
+}
+
+void MerginApi::getInfoAndFetch(QString projectName)
```
wonder-sk: can we rename this to something easier to understand? e.g. updateProject()

app/merginapi.h
```diff
@@ -5,37 +5,75 @@
...
-     Q_OBJECT
+    Q_OBJECT
 public:
-    explicit MerginApi(const QString& root, QByteArray token, QObject* parent = nullptr );
+    explicit MerginApi(const QString& root, const QString& dataDir, QByteArray token, QObject* parent = nullptr );
     ~MerginApi() = default;
     Q_INVOKABLE void listProjects();
+    Q_INVOKABLE void downloadProject(QString projectName);
+    Q_INVOKABLE void getInfoAndFetch(QString projectName);
+    void fetchProject(QString projectName, QByteArray json);
```
wonder-sk: I think this should be a private function

app/merginapi.h
```diff
@@ -5,37 +5,75 @@
...
 class MerginApi: public QObject {
-     Q_OBJECT
+    Q_OBJECT
 public:
-    explicit MerginApi(const QString& root, QByteArray token, QObject* parent = nullptr );
+    explicit MerginApi(const QString& root, const QString& dataDir, QByteArray token, QObject* parent = nullptr );
     ~MerginApi() = default;
     Q_INVOKABLE void listProjects();
+    Q_INVOKABLE void downloadProject(QString projectName);
+    Q_INVOKABLE void getInfoAndFetch(QString projectName);
```
wonder-sk: can you please add some docs on the above three methods - what they do, that they are non-blocking and what signals to expect in the client code...

app/merginapi.cpp
```diff
@@ -42,18 +114,122 @@ void MerginApi::listProjectsReplyFinished()
...
+void MerginApi::downloadProjectReplyFinished()
+{
+
+    QNetworkReply* r = qobject_cast<QNetworkReply*>(sender());
+    if (!r) {
+        emit networkErrorOccurred( "No reply received.", "Mergin API error: downloadProject" );
+        qDebug() << r->errorString();
+        mPendingRequests.remove(r->url());
+        return;
+    }
```
wonder-sk: I don't think it can ever happen that there's no reply. And even if it could, it would crash inside the block because of calls to r->url() but r is null ptr. I would suggest to replace this by Q_ASSERT(r);

app/merginapi.cpp
```diff
@@ -42,18 +114,122 @@ void MerginApi::listProjectsReplyFinished()
...
+    r->deleteLater();
+}
+
+void MerginApi::projectInfoReplyFinished()
+{
+    QNetworkReply* r = qobject_cast<QNetworkReply*>(sender());
+    if (!r) {
+        emit networkErrorOccurred( "No reply received.", "Mergin API error: projectInfo" );
+        return;
+    }
```
wonder-sk: Same as above - let's just replace by Q_ASSERT(r);

app/merginapi.cpp
```diff
@@ -5,29 +5,101 @@
...
+
+    request.setUrl(url);
+    request.setRawHeader("Authorization", QByteArray("Basic " + mToken));
+
+    QNetworkReply *reply = mManager.get(request);
+    connect(reply, &QNetworkReply::finished, this, &MerginApi::projectInfoReplyFinished);
+
+}
+
+void MerginApi::fetchProject(QString projectName, QByteArray json)
```
wonder-sk: can we rename to something like downloadProjectFiles() ? I was having trouble understanding the difference between downloadProject() and fetchProject() - from the name they seem to do the same thing.

app/merginapi.cpp
```diff
@@ -42,18 +114,122 @@ void MerginApi::listProjectsReplyFinished()
...
+        emit networkErrorOccurred( "No reply received.", "Mergin API error: downloadProject" );
+        qDebug() << r->errorString();
+        mPendingRequests.remove(r->url());
+        return;
+    }
+
+    if (r->error() == QNetworkReply::NoError)
+    {
+        QString projectName("temp");
+        if (mPendingRequests.contains(r->url())) {
```
wonder-sk: could it happen that mPendingRequests does not contain the url? If not, let's just do `QString projectName = mPendingRequests.value(r->url());`

app/merginapi.cpp
```diff
@@ -42,18 +114,122 @@ void MerginApi::listProjectsReplyFinished()
...
+            projectName = mPendingRequests.value(r->url());
+        }
+        QString projectDir = mDataDir + projectName;
+        QFile file(projectDir);
```
wonder-sk: "file" variable does not seem to be used anywhere

app/merginapi.cpp
```diff
@@ -42,18 +114,122 @@ void MerginApi::listProjectsReplyFinished()
...
+          Q_ASSERT( v.isArray() );
+          QJsonArray vArray = v.toArray();
+
+          QSet<QString> localFiles = listFiles(projectPath);
+          for ( auto it = vArray.constBegin(); it != vArray.constEnd(); ++it)
+          {
+            QJsonObject projectInfoMap = it->toObject();
+            QString serverChecksum = projectInfoMap.value("checksum").toString();
+            QString path = projectInfoMap.value("path").toString();
+            QString localChecksum = QString::fromStdString(getChecksum(projectPath + path).toStdString());
```
wonder-sk: no need to go through std::string - you can use e.g. QString::fromLatin1() to convert from QByteArray to QString

app/merginapi.cpp
```diff
@@ -64,13 +240,213 @@ ProjectList MerginApi::parseProjectsData(const QByteArray &data)
...
+bool MerginApi::cacheProjectsData(const QByteArray &data)
+{
+    QFile file(mDataDir + "projectsCache.txt");
+    createPathIfNotExists(mDataDir + "projectsCache.txt");
+    if (!file.open(QIODevice::WriteOnly)) {
+        return false;
+    }
+
+    QTextStream stream(&file);
+    stream << data;
```
wonder-sk: no need to use QTextStream here... you can simply do file.write(data)

app/merginapi.cpp
```diff
@@ -64,13 +240,213 @@ ProjectList MerginApi::parseProjectsData(const QByteArray &data)
...
+void MerginApi::handleDataStream(QNetworkReply* r, QString projectDir)
+{
+    // Read content type from reply's header
+    QByteArray contentType;
+    QString contentTypeString;
+    QList<QByteArray> headerList = r->rawHeaderList();
+    // QByteArray.compare was introduced in Qt 5.12; String conversion needed for Android build
+    QString headString;
+    foreach(QByteArray head, headerList) {
+        headString = QString::fromStdString(head.toStdString());
```
wonder-sk: again - simply use QString::fromLatin1() - also below

app/merginapi.cpp
```diff
@@ -64,13 +240,213 @@ ProjectList MerginApi::parseProjectsData(const QByteArray &data)
...
+        if (headString.compare("Content-Type") == 0) {
```
wonder-sk: better to use simply `headString == QStringLiteral("Content-Type")`

app/merginapi.cpp
```diff
@@ -64,13 +240,213 @@ ProjectList MerginApi::parseProjectsData(const QByteArray &data)
...
+    QString boundary;
+    QRegularExpression re;
+    re.setPattern("[^;]+; boundary=(?<boundary>.+)");
+    QRegularExpressionMatch match = re.match(contentTypeString);
+    if (match.hasMatch()) {
+        boundary = match.captured("boundary");
+    }
+
+    // plus some safety trashold
+    int boundarySize = boundary.length() + 8;
```
wonder-sk: why + 8 ?

app/merginapi.cpp
```diff
@@ -64,13 +240,213 @@ ProjectList MerginApi::parseProjectsData(const QByteArray &data)
...
+
+        // Write rest of chunk to file
+        if (!activeFilePath.isEmpty()) {
+            saveFile(data.left(data.size() - boundarySize), activeFilePath, true);
+        }
+        data = data.remove(0, data.size() - boundarySize);
+    }
+}
+
+bool MerginApi::saveFile(const QByteArray &data, QString filePath, bool append)
```
wonder-sk: I see the saveFile() is always called with append=true - do we need that argument?

app/merginapi.cpp
```diff
@@ -64,13 +240,213 @@ ProjectList MerginApi::parseProjectsData(const QByteArray &data)
...
+    QString activeFilePath;
+
+    while (true) {
+        QByteArray chunk = r->read(CHUNK_SIZE);
+        if (chunk.isEmpty()) {
+            // End of stream - write rest of data to active file
+            if (!activeFilePath.isEmpty()) {
+                QRegularExpressionMatch endMatch = endPattern.match(data);
+                int tillIndex = data.indexOf(endMatch.captured(0));
+                saveFile(data.left(tillIndex), activeFilePath, true);
```
wonder-sk: This way of iteratively opening file, appending some data and closing it again is not very efficient - with larger files it may be slowing down things a lot. I would suggest to keep the file open, write the chunks and then close it once for all.

app/merginapi.cpp
```diff
@@ -64,13 +240,213 @@ ProjectList MerginApi::parseProjectsData(const QByteArray &data)
...
+        while (boundaryMatch.hasMatch()) {
+            if (!activeFilePath.isEmpty()) {
+                int tillIndex = data.indexOf(boundaryMatch.captured(0));
+                saveFile(data.left(tillIndex), activeFilePath, true);
+                activeFilePath = "";
+            }
+
+            // delete previously written data with next boundary part
+            int tillIndex = data.indexOf(boundaryMatch.captured(0)) + boundaryMatch.captured(0).length();
+            data = data.remove(0, tillIndex); // String != QByteArray
```
wonder-sk: what does the comment "String != QByteArray" mean here?

app/merginapi.h
```diff
@@ -5,37 +5,75 @@
 #include <QNetworkAccessManager>
 #include <QEventLoop>
 #include <memory>
+#include <QFile>
+
+enum ProjectStatus
+{
+  NoVersion = Qt::UserRole + 1,
```
wonder-sk: why start with Qt::UserRole + 1 ?

app/merginapi.h
```diff
@@ -5,37 +5,75 @@
...
+    void notify(QString message);
 
 private slots:
     void listProjectsReplyFinished();
+    void downloadProjectReplyFinished(); // download + fetch
+    void projectInfoReplyFinished();
 
 private:
     ProjectList parseProjectsData( const QByteArray &data );
+    void makeToast(const QString &errorMessage, const QString &additionalInfo);
```
wonder-sk: This function does not seem to exist

app/merginprojectmodel.cpp
```diff
@@ -22,6 +21,12 @@ QVariant MerginProjectModel::data( const QModelIndex& index, int role ) const
     {
     case Name: return QVariant(project->name);
     case ProjectInfo: return QVariant(project->info);
+    case Status: {
+        if (project->status == ProjectStatus::OutOfDate) return QVariant("outOfDate");
+        if (project->status == ProjectStatus::UpToDate) return QVariant("upToDate");
+        if (project->status == ProjectStatus::NoVersion) return QVariant("noVersion");
```
wonder-sk: ProjectStatus::Pending value seems to be unhandled here

app/projectsmodel.h
```diff
@@ -54,9 +54,15 @@ class ProjectModel : public QAbstractListModel
     int rowCount(const QModelIndex &parent = QModelIndex()) const;
 
     QString dataDir() const;
+signals:
+    void projectsChanged();
+
+    public slots:
+        void addProject(QString projectFolder, QString projectName);
```
wonder-sk: formatting seems off

app/merginprojectmodel.cpp
```diff
@@ -49,12 +62,28 @@ int MerginProjectModel::rowCount(const QModelIndex &parent) const {
...
+
+void MerginProjectModel::downloadProjectFinished(QString projectFolder, QString projectName)
+{
+    Q_UNUSED(projectFolder);
+    for (std::shared_ptr<MerginProject> project: mMerginProjects) {
+        if (project->name == projectName) {
+            beginResetModel();
+            project->pending = false;
+            project->status = ProjectStatus::UpToDate;
+            endResetModel();
```
wonder-sk: no need to use [begin/end]ResetModel() here: we are not resetting the whole model, just modifying one item's content. For that it's best to use `emit dataChanged(index, index)` with the index of the changed project. This will trigger more optimized view update (it will not need to delete and recreate all items)

app/merginprojectmodel.cpp
```diff
@@ -49,12 +62,28 @@ int MerginProjectModel::rowCount(const QModelIndex &parent) const {
...
+            emit merginProjectsChanged();
```
wonder-sk: Is this signal used anywhere? I can see a signal with the same name is used in MerginApi, but can't see any code that would be using this signal. If not used, best to remove it completely.

app/merginapi.h
```diff
@@ -5,37 +5,109 @@
...
+};
+Q_ENUMS( ProjectStatus )
 
 struct MerginProject {
     QString name;
     QStringList tags;
-    QString info;
+    QString created;
+    QString updated; // local updated
+    QString serverUpdated;
```
wonder-sk: All these created, updated, serverUpdated members are in fact date/time variables, right? Why not use QDateTime to store the content?

app/merginapi.h
```diff
@@ -5,37 +5,109 @@
...
+    bool pending = false;
```
wonder-sk: pending means that we have a pending download/update request in MerginApi for this project right? Can we add a short comment about what does it mean?

app/merginapi.h
```diff
@@ -5,37 +5,109 @@
...
+     * when a respond is received, parses project json, writes it to a cache text file and sets mMerginProjects.
+     * Eventually emits listProjectsFinished on which ProjectPanel (qml component) updates content.
+     * If listing has been successful, updates cached merginProjects list.
+     * @param
+     */
     Q_INVOKABLE void listProjects();
+
+    /**
+     * Sends non-blocking GET request to the server to download a project with a given name. On downloadProjectReplyFinished,
+     * when a respond is received, parses data-stream and creates files. Eventually emits downloadProjectFinished on which
```
wonder-sk: s/respond/response/

(also in the comment for updateProject)

app/merginapi.h
```diff
@@ -5,37 +5,109 @@
...
+    void deleteObsoleteFiles(QString projectName);
 
     QNetworkAccessManager mManager;
     QString mApiRoot;
     ProjectList mMerginProjects;
+    QString mDataDir;
+    QString mCacheFile;
     QByteArray mToken;
+    QHash<QUrl, QString>mPendingRequests;
+    QHash<QString, QSet<QString>> mObsolateFiles;
```
wonder-sk: s/Obsolate/Obsolete/

app/merginapi.cpp
```diff
@@ -5,29 +5,136 @@
...
+
+    QNetworkReply *reply = mManager.post(request, json);
+    connect(reply, &QNetworkReply::finished, this, &MerginApi::downloadProjectReplyFinished);
+}
+
+ProjectList MerginApi::updateMerginProjectList(ProjectList serverProjects)
+{
+    QHash<QString, QString> projectUpdates;
+    for (std::shared_ptr<MerginProject> project: mMerginProjects) {
+        projectUpdates.insert(project.get()->name, project.get()->updated);
```
wonder-sk: use `project->name` instead of `project.get()->name`  ... shared_ptr also overrides `->` operator so it "feels" like working with a raw pointer.

same thing in for-cycle below and in setUpdateToProject

app/merginapi.cpp
```diff
@@ -64,13 +279,224 @@ ProjectList MerginApi::parseProjectsData(const QByteArray &data)
...
+
+        // Write rest of chunk to file
+        if (!activeFilePath.isEmpty()) {
+            saveFile(data.left(data.size() - boundarySize), activeFilePath, false);
+        }
+        data = data.remove(0, data.size() - boundarySize);
+    }
+}
+
+bool MerginApi::saveFile(const QByteArray &data, QString filePath, bool closeFile)
```
wonder-sk: This approach with closeFile argument will not work as expected. The local `QFile file` is closed in its destructor anyway. You need to e.g. pass QFile by reference to this function (and do not reopen it if it is already open!)

app/merginapi.cpp
```diff
@@ -379,43 +378,44 @@ void MerginApi::handleDataStream(QNetworkReply* r, QString projectDir)
...
-        dataString = QString::fromStdString(data.toStdString());
+        dataString = QString::fromLatin1(data.data(), data.size());
         QRegularExpressionMatch boundaryMatch = boundaryPattern.match(dataString);
 
         while (boundaryMatch.hasMatch()) {
             if (!activeFilePath.isEmpty()) {
                 int tillIndex = data.indexOf(boundaryMatch.captured(0));
-                saveFile(data.left(tillIndex), activeFilePath, true);
-                activeFilePath = "";
+                QFile activeFile(activeFilePath);
```
wonder-sk: This defines "activeFile" variable in the local block and thus shadows the other "activeFile" from the outer scope!

PR #134, opened 2019-01-30
Mergin/auth

sklencar:
Added login form to Input to authenticate to Mergin. Currently there is no auth service to generate token and validate auth input data, all its done within Input. Auth request is triggered if there is missing token in a request or auth fails.
Login name is added to setting panel where a user can log out on click.


app/qml/SettingsPanel.qml
```diff
@@ -57,6 +57,23 @@ Popup {
...
+                color: InputStyle.clrPanelMain
+                text: qsTr("Login")
+                text2: __merginApi.username ? __merginApi.username : "(none)"
+
+                MouseArea {
+                    anchors.fill: parent
+                    onClicked: {
+                        if (__merginApi.username) {
+                            __merginApi.username = ""
+                            usernameField.text2 = __merginApi.username ? __merginApi.username : "(none)"
```
wonder-sk: If the signals are working correctly, this should not be needed?

app/qml/SettingsPanel.qml
```diff
@@ -57,6 +57,23 @@ Popup {
...
+            PanelItem {
+                id: usernameField
+                color: InputStyle.clrPanelMain
+                text: qsTr("Login")
+                text2: __merginApi.username ? __merginApi.username : "(none)"
+
+                MouseArea {
+                    anchors.fill: parent
+                    onClicked: {
+                        if (__merginApi.username) {
```
wonder-sk: use hasValidToken() ?

app/merginapi.cpp
```diff
@@ -40,7 +40,7 @@ void MerginApi::listProjects()
 void MerginApi::downloadProject(QString projectName)
 {
     if (mToken.isEmpty()) {
-        emit networkErrorOccurred( "Auth token is invalid", "Mergin API error: downloadProject" );
+        emit authRequested();
```
wonder-sk: missing return here?

app/merginapi.cpp
```diff
@@ -65,7 +65,7 @@ void MerginApi::updateProject(QString projectName)
 {
 
     if (mToken.isEmpty()) {
-        emit networkErrorOccurred( "Auth token is invalid", "Mergin API error: projectInfo" );
+        emit authRequested();
```
wonder-sk: missing return here too?

app/merginapi.cpp
```diff
@@ -111,10 +111,30 @@ void MerginApi::uploadProject(QString projectName)
...
+void MerginApi::authorize(QString username, QString password)
+{
+    QString concatenated = username + ":" + password;
+    mToken = concatenated.toLocal8Bit().toBase64();
+    mUsername = username;
+    emit usernameChanged();
+    emit authChanged();
+}
+
+void MerginApi::logoutRequested()
```
wonder-sk: let's rename this to something like clearAuth()  ... logoutRequested sounds like a signal (similar to authRequested)

app/merginapi.cpp
```diff
@@ -212,6 +232,19 @@ void MerginApi::deleteObsoleteFiles(QString projectPath)
...
+    return mUsername;
+}
+
+void MerginApi::setUsername(const QString &value)
+{
+    mUsername = value;
+    if (mUsername.isEmpty()) {
+        mToken.clear();
+    }
+}
```
wonder-sk: Do we need to keep a separate mUsername in addition to mToken? I would suggest to either: 1. keep just token and decode username from it when needed -or- 2. keep username+password and generate token from them when needed. The second option probably makes more sense...

app/merginapi.h
```diff
@@ -39,6 +39,7 @@ typedef QList<std::shared_ptr<MerginProject>> ProjectList;
 
 class MerginApi: public QObject {
     Q_OBJECT
+    Q_PROPERTY(QString username READ username WRITE setUsername NOTIFY usernameChanged)
```
wonder-sk: I don't think username should be writable property with its own signal (authChanged signal should be enough)

app/merginapi.h
```diff
@@ -39,6 +39,7 @@ typedef QList<std::shared_ptr<MerginProject>> ProjectList;
...
 public:
     explicit MerginApi(const QString& root, const QString& dataDir, QByteArray token, QObject* parent = nullptr );
```
wonder-sk: we should get rid of passing token to MerginApi constructor. Better if clients of MerginApi create it without token and then call authorize()

app/qml/AuthPanel.qml
```diff
@@ -0,0 +1,249 @@
+import QtQuick 2.7
+import QtQuick.Controls 2.0
+import QtQuick.Layouts 1.3
+import QtGraphicalEffects 1.0
+import QgsQuick 0.1 as QgsQuick
+
+Drawer {
```
wonder-sk: As shown on the screenshot, the auth panel does not work for me at all... There are some tweaks of z property and I don't understand the logic there. Why do we need to explicitly modify Z order? Maybe there's a better base Item than Drawer that would avoid the need to do those tweaks? Drawer is meant to be opened/closed by users, but what we actually want is just a simple Item that covers the whole screen...

app/qml/AuthPanel.qml
```diff
@@ -0,0 +1,249 @@
+import QtQuick 2.7
+import QtQuick.Controls 2.0
+import QtQuick.Layouts 1.3
+import QtGraphicalEffects 1.0
+import QgsQuick 0.1 as QgsQuick
```
wonder-sk: please add this here: `import "."  // import InputStyle singleton`

PR #145, opened 2019-02-04
Feature/project handling by folder

### LINKED ISSUE #116: Internal app files created in user folder
The app creates INPUT/resources directory with qgis.db and srs.db files. These files are internal for the use of qgis libraries and they should not be in such location. They should go to app's private folder - worth checking QgsApplication::qgisSettingsDirPath() directory and fixing it by either passing the correct folder to QgsApplication constructor or in QgsApplication::init()

### LINKED ISSUE #133: Confusion with Project and folder names
There should be a better way of displaying Projects name under ALL PROJECTS. Currently, it displays folder name and confuses users.

sklencar:
closes #133


app/main.cpp
```diff
@@ -99,50 +99,76 @@ static void setEnvironmentQgisPrefixPath() {
   qDebug() << "QGIS_PREFIX_PATH: " << ::getenv("QGIS_PREFIX_PATH");
 }
 
+static bool cpDir(const QString &srcPath, const QString &dstPath)
+{
+    QDir parentDstDir(QFileInfo(dstPath).path());
+    QFile::setPermissions(dstPath, QFile::ReadUser|QFile::WriteUser|QFile::ReadOwner|QFile::WriteOwner);
```
wonder-sk: Before the permissions were getting set only for files, now they are only being set for directories... why?

app/main.cpp
```diff
@@ -99,50 +99,76 @@ static void setEnvironmentQgisPrefixPath() {
...
-      }
-      if (!QFile(destFilePath).exists())
-      {
-        bool res = f.copy(destFilePath);
-        qDebug("copying [%d] %s", res, destFilePath.toLatin1().data());
-        // by default the permissions are set as readable for everyone - and not writable!
-        res = QFile::setPermissions(destFilePath, QFile::ReadUser|QFile::WriteUser|QFile::ReadOwner|QFile::WriteOwner);
-        qDebug("chmod [%d]", res);
-      }
+      cpDir(assetsBasePath + "/qgis-data/" + dataFolder, pkgPath + "/" + dataFolder);
```
wonder-sk: Can we simplify this - instead of the for cycle, why not just one command cpDir(assetsBasePath + "/qgis-data", pkgPath) ?

app/main.cpp
```diff
@@ -99,50 +99,76 @@ static void setEnvironmentQgisPrefixPath() {
...
+  demoProjectNames << "survey" << "test_project";
+
+  foreach (const QString& projectName, demoProjectNames)
+  {
+      QDir projectDir(dataPath + "/" + projectName);
+      // if projectFolder exists, do not copy
+      if (projectDir.exists()) continue;
+
+      cpDir(assetsBasePath + "/demo-projects/" + projectName, dataPath + "/" + projectName);
+  }
```
wonder-sk: Same comment as above - can we have this without for cycle, just cpDir(assetsBasePath + "/demo-projects", dataPath) ?

app/main.cpp
```diff
@@ -153,9 +179,6 @@ static void init_qgis(const QString& qgisDataPath)
 #ifdef ANDROID
   // QGIS plugins on Android are in the same path as other libraries
   QgsApplication::setPluginPath( QApplication::applicationDirPath() );
-  QgsApplication::setPkgDataPath(qgisDataPath);
```
wonder-sk: we don't need to set the data path? is it already set to the correct directory?

app/projectsmodel.cpp
```diff
@@ -30,34 +30,61 @@ ProjectModel::ProjectModel(const QString &dataDir, QObject* parent)
...
     }
-    std::sort(mProjectFiles.begin(), mProjectFiles.end());
+
+    ProjectFile project;
+    if (!foundProjects.isEmpty()) {
+        project = foundProjects.at(0);
+        if (foundProjects.length() > 1) {
+            project.info = "invalid project";
+            project.isValid = false;
+        }
```
wonder-sk: why do we set the first project as invalid?

sklencar: Projects files are read per folder. If there is more than one *.qgs project file (*.qgz is still not supported yet), the first ProjectFile is taken, marked as invalid and added to projectList, others are ignored. Its because of avoiding duplicity of projects.

app/main.cpp
sklencar: PkgDataPath is set by default to:` /data/user/0/uk.co.lutraconsulting/files/share`
Since srs.db is copied there and by default read from `pkgDataPath() + QStringLiteral( "/resources/srs.db" ); ` or from tempFolder, where is copied afterwards from pkgDataPath.

app/main.cpp
sklencar: Seems like if files (demo projects or resources) are copied to destination in internal/external memory, its not necessary to set permissions. Its different with system folders.

app/main.cpp
sklencar: Eventually pkgData default value is overwritten since there is losing permission issue when copying resources files to system pkg path.

app/main.cpp
```diff
@@ -153,9 +156,7 @@ static void init_qgis(const QString& qgisDataPath)
...
-#else
-  Q_UNUSED(qgisDataPath);
```
wonder-sk: missing #else with q_unused

app/main.cpp
```diff
@@ -195,9 +196,10 @@ int main(int argc, char *argv[])
   // Set/Get enviroment
   QString dataDir = getDataDir();
   setEnvironmentQgisPrefixPath();
-  expand_assets_data(dataDir);
 
-  init_qgis(dataDir);
+  init_qgis(dataDir + "/qgis-data");
+  expand_pkg_data( app.pkgDataPath() );
```
wonder-sk: QgsApplication:: instead of app.

app/merginapi.cpp
```diff
@@ -12,7 +12,7 @@
 MerginApi::MerginApi(const QString &root, const QString& dataDir, QObject *parent)
   : QObject (parent)
   , mApiRoot(root)
-  , mDataDir(dataDir + "/downloads/")
+  , mDataDir(dataDir + "/")
```
wonder-sk: let's use "projects" sub-dir

PR #147, opened 2019-02-06
Documentation about Installing Input with docker and building Input with Qt Creator

edips:
I tried to install OSGeo4A for Ubuntu and Manjaro, I got errors installing some recipes like geos. And installing QGIS from the source is difficult for Ubuntu because of many dependencies. I think installing Input with docker is easier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,19 @@ docker cp 71eef3e8038a:/opt/android-ndk /home/apps/android
 docker cp 71eef3e8038a:/opt/android-sdk /home/apps/android
 ```
 - The folder copied from docker has root permission, so we cannot compile any app in Qt Creator with root files. Change these folder's permissions to your user name.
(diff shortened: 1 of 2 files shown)


wonder-sk:
Hi @edips thanks for your pull request!

I see you have removed a portion of the existing developer documentation which is still valid - for example the instructions for desktop builds for Linux or Mac OS. Also you have removed a part of config.pri.default which is used for desktop builds... It would be better to just have your additions for docker-based build and keep all the original content.

wonder-sk:
As this branch is in conflict with master, I have posted your documentation to a new file:
https://github.com/lutraconsulting/input/blob/master/docs/developers/android_build_with_docker.md

Thanks for your contribution!

edips:
Hi @wonder-sk You are welcome. Thanks for the great project also. Sorry it was my first pull request in github. I will edit my documentation according to your suggestion.

PR #152, opened 2019-02-08
Added permissions requirements from code

sklencar:
Added docs about publishing the app to Google play.

### DIFF
--- a/app/androidutils.cpp
+++ b/app/androidutils.cpp
@@ -39,7 +39,6 @@ bool AndroidUtils::isAndroid() const
 void AndroidUtils::requirePermissions()
 {
     checkAndAcquirePermissions("android.permission.WRITE_EXTERNAL_STORAGE");
-    checkAndAcquirePermissions("android.permission.CAMERA");
 }
 
 bool AndroidUtils::checkAndAcquirePermissions( const QString &permissionString )
--- a/app/androidutils.cpp
+++ b/app/androidutils.cpp
@@ -35,3 +35,24 @@ bool AndroidUtils::isAndroid() const
     return false;
 #endif
 }
+
+void AndroidUtils::requirePermissions()
+{
+    checkAndAcquirePermissions("android.permission.WRITE_EXTERNAL_STORAGE");
+    checkAndAcquirePermissions("android.permission.CAMERA");
+}
+
+bool AndroidUtils::checkAndAcquirePermissions( const QString &permissionString )
+{
+  QtAndroid::PermissionResult r = QtAndroid::checkPermission( permissionString );
+  if ( r == QtAndroid::PermissionResult::Denied )
+  {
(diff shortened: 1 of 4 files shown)


app/androidutils.cpp
```diff
@@ -35,3 +35,24 @@ bool AndroidUtils::isAndroid() const
     return false;
 #endif
 }
+
+void AndroidUtils::requirePermissions()
+{
+    checkAndAcquirePermissions("android.permission.WRITE_EXTERNAL_STORAGE");
+    checkAndAcquirePermissions("android.permission.CAMERA");
```
wonder-sk: I think we should only ask for camera permission the first time the user is going to take a picture... otherwise it is kind of scare to give an application access to camera immediately when it starts :-)

app/androidutils.cpp
```diff
@@ -35,3 +35,24 @@ bool AndroidUtils::isAndroid() const
     return false;
 #endif
 }
+
+void AndroidUtils::requirePermissions()
+{
+    checkAndAcquirePermissions("android.permission.WRITE_EXTERNAL_STORAGE");
```
wonder-sk: on my Android 6 phone things worked without being asked for this permission... is it possible to do Android version check and only ask on Android >= 8 ?

publishing.md
```diff
@@ -0,0 +1,30 @@
...
+It should prompt that "Certificate already exists in keystore under alias . Do you still want to add it? [no]:"
+6. Type 'y' and press enter. You will receive a confirmation message.
+7. For subsequent builds sign the app using the same process as in (2).
+
+Note: Make sure you have updated `android:versionName="x.y.z"` and `android:versionCode="v"` in `AndroidManifest.xml` before any release update. 
+Also check if you have target SDK set at least to version 26 (Android 8): `android:targetSdkVersion="26"`.
+
+## Internal test release (Google Play Console )
+To make an internal test release, open the Google Play Console -> Release management -> Internal test track byb clickin on "Manage" button.
+1. Upload the apk you created in release mode with updated `android:versionCode="v"`. You can
```
wonder-sk: unfinished sentence? You can...

app/androidutils.cpp
sklencar: For my Android 6 and 7, it has to be there otherwise data folder is not accessible. Have you properly uninstall the app? There was no such issue with SDK 21 though. However, a new problem has been introduce on startup. If permission are accepted, on the first lunch of app there are missing demo projects and even all downloaded projects are tagged as invalid. From second and further lunch it is all ok.

app/androidutils.cpp
sklencar: Camera permission has been removed. It's automatically requested when PhotoPanel is init, thus when featureForm with externalResource field is init. Would be better to send a signal only if a new picture is about to be taken.

PR #166, opened 2019-02-18
Feature/custom mergin link

### LINKED ISSUE #160: Move link to mergin as a part of settings
Mergin server should have its own setting entry, so user can change the link to the instance.

### LINKED ISSUE #162: Provide a link to Mergin sign up page
Under all projects, it is confusing for users where to go, if they already don't have a login

sklencar:
Added
- option to change Mergin URL in AuthPanel
- default Mergin URL in config
- link for signing up

closes #162 closes #160


app/qml/MerginSettingDialog.qml
```diff
@@ -0,0 +1,153 @@
+import QtQuick 2.7
+import QtQuick.Controls 2.0
+import QtQuick.Layouts 1.3
+import QtQuick.Dialogs 1.2
+import QtGraphicalEffects 1.0
+import QgsQuick 0.1 as QgsQuick
+
+Dialog {
```
wonder-sk: Hmm with the desktop build I can't really set the new URL - pressing Enter does not trigger anything and buttons are missing too. This is what I see:

image

app/config.pri.default
```diff
@@ -1,5 +1,8 @@
 BASE_DIR=</home/folder>/Projects/quick
 
+# Default Mergin URL - used with /auth/signup
+DEFAULT_MERGIN_URL=<MERGIN URL>
```
wonder-sk: Let's hardcode the default mergin URL in MerginApi to https://public.cloudmergin.com/

It could be e.g. `static QString defaultApiRoot() { return "...."; }` in MerginApi

app/main.cpp
```diff
@@ -249,8 +249,11 @@ int main(int argc, char *argv[])
   engine.rootContext()->setContextProperty( "__appSettings", &as );
 
   // Create mergin api
-  //std::unique_ptr<MerginApi> ma =  std::unique_ptr<MerginApi>(new MerginApi(QString("https://mergin.dev.cloudmappin.com"), projectDir ));
-  std::unique_ptr<MerginApi> ma =  std::unique_ptr<MerginApi>(new MerginApi(QString("https://dev.dev.cloudmergin.com/"), projectDir ));
+  QString defaultMerginUrl;
+#ifdef DEFAULT_MERGIN_URL
+  defaultMerginUrl=STR(QGIS_QUICK_DATA_PATH)
```
wonder-sk: does this work even without semicolon? but best to use hardcoded value for default URL without config.pri

app/merginapi.cpp
```diff
@@ -11,14 +11,18 @@
...
 {
     QObject::connect(this, &MerginApi::syncProjectFinished,this, &MerginApi::setUpdateToProject);
     QObject::connect(this, &MerginApi::merginProjectsChanged,this, &MerginApi::cacheProjects);
     QObject::connect(this, &MerginApi::authChanged,this, &MerginApi::saveAuthData);
 
+    QSettings settings;
+    settings.beginGroup("Input/");
+    settings.setValue("apiRootDefault", root);
+    settings.endGroup();
+
```
wonder-sk: let's just initialize mApiRoot to MerginApi::defaultApiRoot() here to make it simpler

app/qml/AuthPanel.qml
```diff
@@ -213,8 +216,78 @@ Item {
                         elide: Text.ElideRight
                     }
                 }
+
+                Button {
+                    id: singInButton
```
wonder-sk: signUpButton?

app/qml/MerginSettingDialog.qml
```diff
@@ -0,0 +1,153 @@
...
+                    }
+
+                    TextField {
+                        id: linkField
+                        text: merginLinkDialog.linkFieldText
+                        height: fieldHeight
+                        font.pixelSize: InputStyle.fontPixelSizeNormal
+                        color: merginLinkDialog.fontColor
+                        placeholderText: qsTr("Mergin instance")
+                        inputMethodHints: Qt.ImhNoPredictiveText | Qt.ImhNoAutoUppercase
```
wonder-sk: Unrelated, but can we have this hint also for username in auth panel? I have seen that on mobile it tries to start with upper case

app/merginapi.cpp
sklencar: Since defualtApiRoot is hardcoded, this part is obsolete. Anyway, wouldn't be better to load apiRoot from QSettings and only load default if its empty?

wonder-sk: Ah right - let's just drop this.

I see in loadAuthData() we already load apiRooot - that's good. Yes let's load api root from qsettings and only use default when it was not set yet

app/qml/MerginSettingDialog.qml
sklencar: Hopefully fixed by proper sizing of a dialog and better key handling - however textField has to have focus to listen on accept. All in new commit.

PR #178, opened 2019-02-25
Added digitizing period setting - 

### LINKED ISSUE #176: Add an option for GPS frequency when recording path
It will be good to have an option in the Settings, so that users can set the intervals for which the GPS creates a vertex on the recorded path.

sklencar:
Each time after this period a point will be added to recordedPoints according the current position of QgsQuickPositionKit while recording.

closes #176

### DIFF
--- a/app/appsettings.h
+++ b/app/appsettings.h
@@ -11,7 +11,7 @@ class AppSettings: public QObject {
     Q_PROPERTY(QString defaultProjectName READ defaultProjectName NOTIFY defaultProjectChanged)
     Q_PROPERTY(QString defaultLayer READ defaultLayer WRITE setDefaultLayer NOTIFY defaultLayerChanged)
     Q_PROPERTY(bool autoCenterMapChecked READ autoCenterMapChecked WRITE setAutoCenterMapChecked NOTIFY autoCenterMapCheckedChanged)
-    Q_PROPERTY(int digitizingPeriod READ digitizingPeriod WRITE setDigitizingPeriod NOTIFY digitizingPeriodChanged)
+    Q_PROPERTY(int lineRecordingInterval READ lineRecordingInterval WRITE setLineRecordingInterval NOTIFY lineRecordingIntervalChanged)
     Q_PROPERTY(int gpsAccuracyTolerance READ gpsAccuracyTolerance WRITE setGpsAccuracyTolerance NOTIFY gpsAccuracyToleranceChanged)
 
 public:
@@ -34,8 +34,8 @@ public:
     int gpsAccuracyTolerance() const;
     void setGpsAccuracyTolerance(int gpsAccuracyTolerance);
 
-    int digitizingPeriod() const;
-    void setDigitizingPeriod(int digitizingPeriod);
+    int lineRecordingInterval() const;
+    void setLineRecordingInterval(int lineRecordingInterval);
 
 signals:
     void defaultProjectChanged();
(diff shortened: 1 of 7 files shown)


app/digitizingcontroller.cpp
```diff
@@ -141,10 +153,17 @@ void DigitizingController::onPositionChanged()
...
   fixZ(&point);
 
-  mRecordedPoints.append( point );
+  if (mDigitizingPeriodTimer.isActive()) {
+      mRecordedPoints.last().setX(point.x());
+      mRecordedPoints.last().setY(point.y());
+  } else {
+      mRecordedPoints.append( point );
+      mDigitizingPeriodTimer.start( mDigitizingPeriod * 1000 );
+  }
```
wonder-sk: Does this work correctly? I am worried that if the QTimer gets deactivated around the time of new gps position coming, we may be getting inconsistent recording (it could be +/- one second). I think it would be better if we just kept QDateTime of the last position (from QDateTime::currentDateTime()) and then check if secsTo() is >= than our interval. The QTimer class is more suited for cases when we want to trigger an action on its timeout() signal...

app/appsettings.h
```diff
@@ -11,6 +11,7 @@ class AppSettings: public QObject {
     Q_PROPERTY(QString defaultProjectName READ defaultProjectName NOTIFY defaultProjectChanged)
     Q_PROPERTY(QString defaultLayer READ defaultLayer WRITE setDefaultLayer NOTIFY defaultLayerChanged)
     Q_PROPERTY(bool autoCenterMapChecked READ autoCenterMapChecked WRITE setAutoCenterMapChecked NOTIFY autoCenterMapCheckedChanged)
+    Q_PROPERTY(int digitizingPeriod READ digitizingPeriod WRITE setDigitizingPeriod NOTIFY digitizingPeriodChanged)
```
wonder-sk: Can we rename this to e.g. lineRecordingInterval? So it is clear it only relates to lines and it is consistent with GUI

saberraz: How about VertexRecordingInterval, in case we want to add polygon capturing to Input?

app/digitizingcontroller.cpp
```diff
@@ -141,11 +152,17 @@ void DigitizingController::onPositionChanged()
...
     return;
 
+
   QgsPoint point = mPositionKit->position();
   fixZ(&point);
 
-  mRecordedPoints.append( point );
-
+  if (mLastTimeRecorded.addSecs(mLineRecordingInterval) <= QDateTime().currentDateTime()) {
+      mLastTimeRecorded = QDateTime().currentDateTime();
```
wonder-sk: better to use QDateTime::currentTime() here and on the line above

app/digitizingcontroller.cpp
```diff
@@ -141,11 +152,17 @@ void DigitizingController::onPositionChanged()
...
+      mRecordedPoints.append( point );
+  } else {
```
wonder-sk: probably worth checking there's at least one point to avoid a possible crash

app/digitizingcontroller.h
```diff
@@ -68,6 +73,8 @@ private slots:
   QVector<QgsPoint> mRecordedPoints;  //!< for recording of linestrings
   QgsQuickAttributeModel *mRecordingModel = nullptr;  //!< to be used for highlight of feature being recorded
   QgsQuickMapSettings *mMapSettings = nullptr;
+  int mLineRecordingInterval = 3; // in seconds
+  QDateTime mLastTimeRecorded;
```
wonder-sk: I think we will also need to reset mLastTimeRecorded when starting recording, so if someone uses 30s interval, stops recording and then starts it again,  we would not wait and immediately record the first position.

app/digitizingcontroller.cpp
sklencar: QTime::currentTime() ?

Issue #127, opened 2019-01-16
Adding feature - freehand


saberraz:
Currently only GPS location can be recorded as a node/point.

wonder-sk:
Adding/editing of points can be now done also manually on the map

PR #181, opened 2019-02-28
Record/add point

sklencar:
First phase of manually editing/recording of features. 

related #127

### DIFF
--- a/app/qml/RecordCrosshair.qml
+++ b/app/qml/RecordCrosshair.qml
@@ -1,5 +1,5 @@
-import QtQuick 2.12
-import QtGraphicalEffects 1.12
+import QtQuick 2.7
+import QtGraphicalEffects 1.0
 import QgsQuick 0.1 as QgsQuick
 import lc 1.0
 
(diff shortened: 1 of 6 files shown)


app/qml/RecordCrosshair.qml
```diff
@@ -0,0 +1,27 @@
+import QtQuick 2.12
+import QtGraphicalEffects 1.12
```
wonder-sk: Was getting QML error due to the version being too high. Had to change to
```
import QtQuick 2.7
import QtGraphicalEffects 1.0
```

app/qml/RecordToolbar.qml
```diff
@@ -0,0 +1,79 @@
...
+        anchors.fill: parent
+
+        Item {
+            width: parent.width/parent.children.length
+            height: parent.height
+
+            MainPanelButton {
+                id: gpsSwitchBtn
+                width: mainPanel.itemSize
+                text: qsTr("Gps")
```
wonder-sk: GPS

PR #182, opened 2019-03-01
Record/edit point

sklencar:
Edit geometry for point features + new crosshair

related #127


app/digitizingcontroller.cpp
```diff
@@ -193,3 +193,23 @@ QgsQuickFeatureLayerPair DigitizingController::lineFeature()
 
   return QgsQuickFeatureLayerPair(f, featureLayerPair().layer());
 }
+
+QgsPoint DigitizingController::mapCoords(QgsQuickFeatureLayerPair pair)
```
wonder-sk: Please rename to make it clear this is meant to work with point features - e.g. pointFeatureMapCoordinates

app/digitizingcontroller.cpp
```diff
@@ -193,3 +193,23 @@ QgsQuickFeatureLayerPair DigitizingController::lineFeature()
...
+      return QgsPoint();
+
+    QgsPointXY res = mMapSettings->mapSettings().layerToMapCoordinates(pair.layer(), QgsPoint(pair.feature().geometry().asPoint()));
+    return QgsPoint(res);
+}
+
+QgsQuickFeatureLayerPair DigitizingController::changePointGeometry(QgsQuickFeatureLayerPair pair, QgsPoint point)
+{
+    QgsPointXY layerPoint = mMapSettings->mapSettings().mapToLayerCoordinates(pair.layer(), QgsPointXY(point.x(), point.y()));
+    QgsPoint* mapPoint = new QgsPoint( layerPoint );
```
wonder-sk: It is confusing to call it mapPoint when in fact it is using layer coordinate system.... maybe use layerPointXY and layerPoint variables?

app/digitizingcontroller.cpp
```diff
@@ -193,3 +193,23 @@ QgsQuickFeatureLayerPair DigitizingController::lineFeature()
...
+    fixZ(mapPoint);
+    QgsGeometry geom( mapPoint );
+
+    pair.featureRef().setGeometry(geom);
+    return pair;
```
wonder-sk: Why do we return the feature pair here if we get it as an argument and it is not modified in the function body?

app/qml/FeaturePanel.qml
```diff
@@ -198,10 +202,14 @@ Drawer {
             y: parent.height - height
             state: featurePanel.formState
             visible: !featurePanel.isReadOnly
+            isFeatureLine: (!!featurePanel.feature.layer) && digitizing.hasLineGeometry(featurePanel.feature.layer)
```
wonder-sk: why do we need double bang here? (!!featurePanel.feature.layer)

app/qml/FeatureToolbar.qml
```diff
@@ -102,12 +104,12 @@ Item {
             MainPanelButton {
 
                 width: toolbar.itemSize
-                text: qsTr("Add photo")
-                imageSource: "add_photo.svg"
-                disabled: true
+                text: qsTr("Edit geometry")
+                imageSource: "edit.svg"
+                disabled: isFeatureLine
```
wonder-sk: what if the feature is polygon?

app/qml/MainPanelButton.qml
```diff
@@ -25,10 +25,10 @@ Rectangle {
     MouseArea {
         anchors.fill: parent
         onClicked: {
-            root.activated()
+            if (!disabled) root.activated()
         }
         onPressAndHold: {
-            root.activatedOnHold()
+            if (!disabled) root.activatedOnHold()
```
wonder-sk: Rather than doing explicit checks for disabled property, how about we just disable the whole MouseArea when the button is disabled?

app/qml/RecordCrosshair.qml
```diff
@@ -1,27 +1,17 @@
...
-        anchors.centerIn: parent
-        width: root.lineWidth
-        height: root.lineLength
-        color: "black"
-        opacity: root.childrenOpacity
+        height: root.size
+        width: height
+        source: "crosshair.svg"
+        sourceSize.width: width
+        sourceSize.height: heightß
```
wonder-sk: there's a weird character at the end of the line

app/qml/main.qml
```diff
@@ -23,6 +23,9 @@ ApplicationWindow {
             },
             State {
                 name: "record"
+            },
+            State {
+                name: "edit"
```
wonder-sk: can you please add a comment to each state regarding its purpose?

app/qml/main.qml
```diff
@@ -287,7 +327,12 @@ ApplicationWindow {
...
-        onRemoveClicked: stateManager.state = "view"
+        onRemoveClicked: {
+            if (stateManager.state === "edit") {
+                featurePanel.show_panel(featurePanel.feature, "Edit", "form")
+            }
+            stateManager.state = "view"
+        }
     }
 
     RecordCrosshair {
```
wonder-sk: Right now the crosshair is below the GPS location marker - I think it should be on top of it.

app/digitizingcontroller.cpp
sklencar: The pair's feature geometry is modified above. In https://github.com/lutraconsulting/input/pull/182/files#diff-dffcb24829ceef0df628841dea895e27R86, featurePanel.feature is given as the first parameter - its a recent/current feature pair. This modified pair is set to featurePanel.feature and save fn is called. It wouldnt work without assigning modified feature pair.

wonder-sk: Ahh right I thought it was just a reference...

PR #222, opened 2019-03-22
Feature/better choice survey layer

### LINKED ISSUE #218: Better choice of survey layer
The idea is to have an extra row with survey layer name shown above the main toolbar when recording is started.

By clicking on the layer name the popup appears with list of available survey layers, allowing change of the survey layer.

The choice of the survey layer is kept for the next time user starts recording.

The "default survey layer" setting can be removed from the settings panel - it is in fact configured using this new method.

sklencar:
closes #218


sklencar:
Obsolate due to #223

PR #249, opened 2019-04-01
Test - create and remove project

sklencar:


app/test/testmerginapi.cpp
```diff
@@ -61,11 +62,59 @@ void TestMerginApi::testDownloadProject()
...
+
+    bool containsTestProject = false;
+    for (std::shared_ptr<MerginProject> project: projects) {
+        if (project->name == mProjectName) containsTestProject = true;
+    }
+    Q_ASSERT(containsTestProject);
+    qDebug() << "TestMerginApi::testCreateProject PASSED";
+}
+
+void TestMerginApi::testDeleteProject()
```
wonder-sk: I suggest to merge the create+delete project tests together into one test function. (so there isn't just testCreateProject() method that leaves behind a test). The unified create+delete test should also check that the project does not exist already before creating it (so we are sure that create project works as expected)

app/test/testmerginapi.cpp
```diff
@@ -61,11 +62,59 @@ void TestMerginApi::testDownloadProject()
...
+    mApi->deleteProject(mProjectName);
+    spy.wait(1000);
+
+    ProjectList projects = getProjectList();
+    bool containsTestProject = false;
+    for (std::shared_ptr<MerginProject> project: projects) {
+        if (project->name == mProjectName) containsTestProject = true;
+    }
+    Q_ASSERT(!containsTestProject);
+    qDebug() << "TestMerginApi::testDeleteProject PASSED";
```
wonder-sk: Please add also standalone tests in separate methods to 1. try to create a project that already exists (should fail), 2. try to delete a project that does not exist (should fail).

app/merginapi.h
```diff
@@ -51,9 +51,9 @@ class MerginApi: public QObject {
      * when a response is received, parses project json, writes it to a cache text file and sets mMerginProjects.
      * Eventually emits listProjectsFinished on which ProjectPanel (qml component) updates content.
      * If listing has been successful, updates cached merginProjects list.
-     * @param
+     * @param withFilter If true, applies "input" tag in request.
      */
-    Q_INVOKABLE void listProjects();
+    Q_INVOKABLE void listProjects(bool withFilter = false);
```
wonder-sk: can we rename `withFilter` to something like `onlyWithInputTag` so that it's easier to understand what kind of filter is applied? Or even better, make it `QString filterTag = QString()` so that we can test with other tags in future...

app/merginapi.cpp
```diff
@@ -261,6 +309,41 @@ void MerginApi::saveAuthData()
...
 
+void MerginApi::createProjectFinished()
+{
+    QNetworkReply* r = qobject_cast<QNetworkReply*>(sender());
+    Q_ASSERT(r);
+
+    if (r->error() == QNetworkReply::NoError)
+    {
+        emit notify("Project created");
+        emit projectCreated();
```
wonder-sk: Would it make sense to pass project name in the signal as well to make it consistent with delete project signal?

sklencar: Additionally, it would be nice to connect it with api's project list and have newly created project listed there without fetching a list again (Yes, similarly to delete project)

wonder-sk: yeah but at this point we don't want to have create/delete project functionality in Input - something people should do in web client / qgis plugin...

app/test/testmerginapi.cpp
```diff
@@ -61,11 +67,136 @@ void TestMerginApi::testDownloadProject()
...
+    QList<QVariant> arguments = spy2.takeFirst();
+    QVERIFY(arguments.at(0).type() == QVariant::String);
+    QVERIFY(arguments.at(1).type() == QVariant::String);
+
+    QCOMPARE(arguments.at(1), QStringLiteral("Mergin API error: createProject"));
+
+    //Clean created project
+    QSignalSpy spy3(mApi, SIGNAL(networkErrorOccurred(QString, QString)));
+    mApi->deleteProject(mProjectName);
+    spy3.wait(1000);
```
wonder-sk: Shouldn't we wait for project deleted signal here instead of network error?

app/test/testmerginapi.cpp
```diff
@@ -61,11 +67,136 @@ void TestMerginApi::testDownloadProject()
...
+
+void TestMerginApi::testCreateProjectTwice()
+{
+    qDebug() << "TestMerginApi::testCreateProjectTwice START";
+    mProjectName = "TEMPORARY_TEST_PROJECT";
+    bool containsTestProject = false;
+    ProjectList projects = getProjectList();
+    for (std::shared_ptr<MerginProject> project: projects) {
+        if (project->name == mProjectName) containsTestProject = true;
+    }
```
wonder-sk: Since we use this piece of code multiple times, let's put it into a function we can reuse, e.g. `bool hasProject(projectName, projects)`

app/test/testmerginapi.h
```diff
@@ -30,8 +35,10 @@ class TestMerginApi: public QObject
     QString mApiRoot;
     QString mUsername;
     QString mPassword;
-
     QString mProjectName;
+    QEventLoop mEventLoop;
```
wonder-sk: not needed?

Issue #264, opened 2019-04-04
Unable to upload edits to Mergin from Input app


Dukestep:
I have downloaded a project from Mergin and made some changes to the attributes of layers in the project and deleted some geometries.

Following these edits, I wish to upload the changes back to Mergin. From what I see in the user documentation, I should see a cloud icon with an up arrow for that project in the "All Projects" tab. However, the icon never changes and instead display a green check mark.

Am I missing something really obvious here? 

I am using the latest version of Input.

saberraz:
It should definitely display the cloud with arrow up. If not, it is a bug.
Let me see if I can reproduce it....

saberraz:
OK...I managed to reproduce the issue for one my projects.
I had to delete the project under MY PROJECT and re-download it for the upload to work after change.

Please make a copy of data (/sdcard/INPUT/projects) if you want to delete it. We will look into this issue and release a fix soon.

Dukestep:
Not sure I follow... Tried to reproduce your steps but I didn't get it to work. I'll wait for the next release, see if it solves my issue and follow up with you! If the release doesn't resolve my issue, I don't mind adding you as a user on the project so that you can debug on your end with my project files.

saberraz:
if you could add me to the project (saber), I can have a look too. But, the issue is likely to be with the local cache not being updated, hence the upload button does not appear.

As a work around:
1- Connect your phone to PC using USB
2- Open file explorer, and go to your mobile folder
3- Under INPUT > projects, select your project and copy it to your PC.
4- Delete .projectsCache.txt
5- Upload the files on your PC to mergin
6- In Input app, open MY PROJECTS and delete the problematic project
7- Under ALL PROJECTS, download the project
8- Hopefully things should work as expected

Dukestep:
Followed the work around and still wasn't able to get the upload icon to show... I've added you to the project (Etienne) if you want to dig around and confirm that your fix may work for my case!

saberraz:
hi @Dukestep I managed to replicate your issue.

Then I downloaded your project:
1- renamed it to get rid so there is no _ at the start of the project
2- Got rid of the qgis project back up file.
3- Ran `ogrinfo -sql "VACUUM" filegeo.gpkg`
4- Created a new project (sync_issue) which all seems to be working fine now.

I have shared the project with you...so you can test it too.

Dukestep:
Confirming that your fix resolved the issue!

[...]

PR #272, opened 2019-04-16
Modification time issue

sklencar:
Get latest modification datetime out of all files rather than from only project's folder.
In case of shapefiles, project's folder has the same mod timestamp although shapefile has been changed.

possibly closing #264


PeterPetrik:
code looks good. How it is with automatic tests?

sklencar:
No tests yet regarding this functionality, not even for uploading changes to Mergin server :/
I am +1 for covering MerginApi with tests.

PR #292, opened 2019-05-06
Mergin project panel added

sklencar:
New panel with 4 tabs listing project according origin of the list with separate search.
Relies on a new Mergin API version, therefore changes in authentication + added namespace.

Some previews:
Project list
Screenshot 2019-05-02 at 17 07 00

Search
Screenshot 2019-05-02 at 17 07 17

Login Page
Screenshot 2019-05-02 at 17 08 54

User page
Screenshot 2019-05-02 at 17 07 46


app/merginapi.h
```diff
@@ -83,16 +90,17 @@ class MerginApi: public QObject
...
-     * @param projectName Name of project to upload.
+     * \param projectName Name of project to upload.
      */
     Q_INVOKABLE void uploadProject( const QString &projectName );
 
     /**
     * Currently no auth service is used, only "username:password" is encoded and asign to mToken.
-    * @param username
-    * @param password
+    * \param username
```
PeterPetrik: add some docs? otherwise it is empty string in doxygen

app/inputsearchmodel.h
```diff
@@ -0,0 +1,36 @@
+/***************************************************************************
```
PeterPetrik: I do not know, all logic is in `MerginProjectModel` anyways, so I still do not understand why we need new class just as a storage of one string. 

Can you just remove input search model class alltogether?

PR #297, opened 2019-05-10
PositionMarker direction from compass

sklencar:
Simple implementation of using compass direction if available

### DIFF
--- a/app/qml/PositionMarker.qml
+++ b/app/qml/PositionMarker.qml
@@ -13,14 +13,15 @@ Item {
     property color baseColor: InputStyle.highlightColor
     property bool withAccuracy: true
     property int interval: 100 // Interval of direction marker updates in ms
-    property real trashold: 2 // trashold used to minimized direction marker updates (otherwise shaking)
-    property real azimuth: 0
+    property real threshold: 3 // threshold used to minimized direction marker updates (otherwise shaking)
+    property real direction
 
     Timer {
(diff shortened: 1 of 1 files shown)


app/qml/PositionMarker.qml
```diff
@@ -11,7 +12,23 @@ Item {
     property QgsQuick.PositionKit positionKit
     property color baseColor: InputStyle.highlightColor
     property bool withAccuracy: true
+    property int interval: 100 // Interval of direction marker updates in ms
+    property real trashold: 2 // trashold used to minimized direction marker updates (otherwise shaking)
```
PeterPetrik: you mean threshold ?

app/qml/PositionMarker.qml
```diff
@@ -32,12 +49,12 @@ Item {
         id: direction
         source: "gps_direction.svg"
         fillMode: Image.PreserveAspectFit
-        rotation: positionKit.direction
```
PeterPetrik: lets stick with the terminology and create positionMarker.direction

I would prefer if ALL logic is inside the onTriggered() method  and the rest of the code would just use positionMarker.direction instead of positionKit.direction 

logic could be 
```
if (compass.reading)
  dir = compass.reading.azimuth
else 
  dir = positionKit.direction

if (abs(new - old) > 2) positionMarker.direction  = dir 
```

PR #308, opened 2019-05-18
Use QtQuick.Shapes based feature highlight

### LINKED ISSUE #36: Use QtQuick.Shapes for highlights
Map highlights (for selected feature) for linestrings/points using QgsQuickFeatureHighlight are using OpenGL functionality to render thick lines/points which is generally very very poorly looking and full of graphical errors.

Since Qt 5.10 there is QtQuick.Shapes module in Qt which should render things with much better quality and quickly at the same time (rendering done on GPU).

Once QGIS support Qt 5.10 we can migrate that improvement back to QgsQuickFeatureHighlight

### LINKED ISSUE #120: Invalid polygons don't get highlighted
A polygon with a small self-intersection failed to highlight on identify... ideally we should be able to detect invalid geometries and fix them so that the highlight works.

### LINKED ISSUE #194: Identified points are not highlighted
When identifying a point feature, it does not highlight it with red buffer.
It works fine with lines, but I think the red buffer is rendered below the point.

### LINKED ISSUE #214: Change rubberband color
Add it to settings, so users can change the color

wonder-sk:
Removes multiple issues with rendering of highlight:
- points highlight often not working on mobile devices
- impossible to customize point highlight beside a square
- really ugly rendering of linestrings
- invalid polygons would not show up due to failed tessellation

Requires Qt >= 5.10

Closes #36
Closes #120
Closes #194
Closes #214


app/input.pro
```diff
@@ -138,9 +138,9 @@ RESOURCES += \
 
 TRANSLATIONS +=
 
-lupdate_only {
-    SOURCES += qml/*.qml
-}
+#lupdate_only {
```
sklencar: Why commented out? Can it be removed?

app/qml/Highlight.qml
```diff
@@ -0,0 +1,176 @@
...
+
+import QgsQuick 0.1 as QgsQuick
+
+Item {
+  id: highlight
+
+  // color for line geometries
+  property color lineColor: "black"
+  // width for line geometries
+  property real lineWidth: 4
```
sklencar: Small note about default values - even if this and properties below (related to any width/height) overwritten in main.qml while init the component, maybe consider adding QgsQuick.Utils.dp or remove default values (hmm, maybe could be interesting when extracted to similar app)?

app/input.pro
wonder-sk: originally i didn't want to include it, but it doesn't hurt: we currently don't support translations so lupdate is useless - and I figured out that with this code enabled, qt creator is running qmake every time a .qml file is modified, slowing me down :)

app/qml/Highlight.qml
wonder-sk: fixed in 6a50bbf

PR #312, opened 2019-05-20
Feature/reverted project storage

sklencar:
To enable use project's folder (and its structure) as it was before updating API.

### DIFF
--- a/app/merginapi.cpp
+++ b/app/merginapi.cpp
@@ -229,7 +229,6 @@ void MerginApi::createProject( const QString &projectNamespace, const QString &p
   QJsonDocument jsonDoc;
   QJsonObject jsonObject;
   jsonObject.insert( QStringLiteral( "name" ), projectName );
-  jsonObject.insert( QStringLiteral( "namespace" ), projectNamespace );
   jsonObject.insert( QStringLiteral( "public" ), false );
   jsonDoc.setObject( jsonObject );
   QByteArray json = jsonDoc.toJson( QJsonDocument::Compact );
--- a/app/merginapi.cpp
+++ b/app/merginapi.cpp
@@ -11,7 +11,7 @@
 
 MerginApi::MerginApi( const QString &dataDir, QObject *parent )
   : QObject( parent )
-  , mDataDir( dataDir + '/' )
+  , mDataDir( dataDir )
   , mCacheFile( QStringLiteral( ".projectsCache.txt" ) )
 {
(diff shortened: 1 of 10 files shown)


app/merginapi.cpp
```diff
@@ -706,15 +703,18 @@ void MerginApi::downloadProjectReplyFinished()
   QNetworkReply *r = qobject_cast<QNetworkReply *>( sender() );
   Q_ASSERT( r );
 
-  QString projectName = mPendingRequests.value( r->url() );
+  QString projectFullName = mPendingRequests.value( r->url() );
+  QStringList parts = projectFullName.split( "/" );
+  QString projectNamespace = parts.at( 0 );
+  QString projectName = parts.at( 1 );
```
wonder-sk: let's also put the namespace/project splitting code into a function - it should also return bool whether splitting was successful, the individual namespace and project names could be passed out through QString&

app/merginapi.h
```diff
@@ -21,6 +21,8 @@ Q_ENUMS( ProjectStatus )
 struct MerginProject
 {
   QString name; //projectNamespace + projectName
+  QString projectNamespace;
+  QString projectDir;
```
wonder-sk: projectDir does not seem to be used, is it?

app/merginapi.h
```diff
@@ -21,6 +21,8 @@ Q_ENUMS( ProjectStatus )
 struct MerginProject
 {
   QString name; //projectNamespace + projectName
```
wonder-sk: the comment does not seem to be right - it is just project name without namespace...?

app/test/testmerginapi.cpp
```diff
@@ -68,27 +68,30 @@ void TestMerginApi::testDownloadProject()
...
 
   QVERIFY( spy.wait( 5000 ) );
   QCOMPARE( spy.count(), 1 );
 
   ProjectList projects = mMerginProjectModel->projects();
   Q_ASSERT( !mMerginProjectModel->projects().isEmpty() );
-  projectName = mMerginProjectModel->projects().at( 0 )->name;
+
+  bool downloadSuccessful = mProjectModel->containsProject( projectName );
+  Q_ASSERT( downloadSuccessful );
```
wonder-sk: tests should use QVERIFY instead of Q_ASSERT - the latter aborts the whole thing while the former just aborts the one test method and gives a nice error message

PR #313, opened 2019-05-20
Feature/check mergin version

sklencar:
Screenshot 2019-05-20 at 14 32 48


app/merginapi.cpp
```diff
@@ -551,6 +626,22 @@ int MerginApi::diskUsage() const
   return mDiskUsage;
 }
 
+void MerginApi::pingMergin()
+{
+  if ( mApiVersionStatus == MerginApiStatus::PASSED ) return;
+
+  setApiVersionStatus( MerginApiStatus::UNKNOWN );
+  emit apiIncompatibilityOccured( QStringLiteral( "Reset Mergin version" ), QStringLiteral( "pingMergin request about to send" ) );
```
wonder-sk: why do we emit the signal here?

app/merginapistatus.h
```diff
@@ -0,0 +1,23 @@
...
+    Q_GADGET
+  public:
+    explicit MerginApiStatus();
+
+    enum VersionStatus
+    {
+      UNKNOWN, // unchecked
+      NOT_FOUND, // cannot be checked due to network/wrong url
+      PASSED,
+      FAILED
```
wonder-sk: please document what does this mean (supposedly the version test has failed). Maybe also better to rename to something like "incompatible version" as there are more ways to fail? (e.g. "not found" is also a kind of failure)

app/merginapi.cpp
sklencar: To avoid having blank loading while waiting for ping. Maybe we can have different notification message or different graphic while input sends ping request.
Screenshot 2019-05-20 at 16 40 56

wonder-sk:
Suggestions:
- add "in progress" state to the status enum - this should simplify GUI update logic too
- add some kind of "try again" button when in "not found" or "incompatible version" state to force the check again

PR #318, opened 2019-05-27
Feature/cache per project

sklencar:
Added cache file per projects

### DIFF
--- a/app/merginapi.cpp
+++ b/app/merginapi.cpp
@@ -324,26 +324,25 @@ void MerginApi::uploadProjectFiles( const QString &projectNamespace, const QStri
 
 ProjectList MerginApi::updateMerginProjectList( const ProjectList &serverProjects )
 {
-  // List of downloaded projects;
-  QHash<QString, std::shared_ptr<MerginProject>> projectUpdates;
+  QHash<QString, std::shared_ptr<MerginProject>> downloadedProjects;
   for ( std::shared_ptr<MerginProject> project : mMerginProjects )
   {
     if ( !project->projectDir.isEmpty() )
     {
-      projectUpdates.insert( getFullProjectName( project->projectNamespace, project->name ), project );
+      downloadedProjects.insert( getFullProjectName( project->projectNamespace, project->name ), project );
     }
   }
 
-  if (projectUpdates.isEmpty()) return serverProjects;
+  if ( downloadedProjects.isEmpty() ) return serverProjects;
 
   for ( std::shared_ptr<MerginProject> project : serverProjects )
   {
     QString fullProjectName = getFullProjectName( project->projectNamespace, project->name );
-    if ( projectUpdates.contains( fullProjectName ) )
+    if ( downloadedProjects.contains( fullProjectName ) )
     {
-      project->projectDir = projectUpdates.value( fullProjectName ).get()->projectDir;
-      QDateTime localUpdate = projectUpdates.value( fullProjectName ).get()->updated.toUTC();
-      project->lastSync = projectUpdates.value( fullProjectName ).get()->lastSync.toUTC();
+      project->projectDir = downloadedProjects.value( fullProjectName ).get()->projectDir;
+      QDateTime localUpdate = downloadedProjects.value( fullProjectName ).get()->updated.toUTC();
+      project->lastSync = downloadedProjects.value( fullProjectName ).get()->lastSync.toUTC();
       QDateTime lastModified = getLastModifiedFileDateTime( project->projectDir );
       project->updated = localUpdate;
       project->status = getProjectStatus( project->updated, project->serverUpdated, project->lastSync, lastModified );
@@ -524,8 +523,9 @@ ProjectList MerginApi::parseAllProjectsData()
     if ( info.exists() )
     {
       std::shared_ptr<MerginProject> project = parseProjectData( mDataDir + folderName, mCacheFile );
-      if (project) {
-         projects << project;
+      if ( project )
+      {
+        projects << project;
       }
     }
   }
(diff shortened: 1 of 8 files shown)


app/merginapi.cpp
```diff
@@ -504,20 +512,53 @@ void MerginApi::pingMerginReplyFinished()
...
   }
 }
 
+void MerginApi::projectDeletedOnPath( const QString &projectDir )
+{
+  for ( std::shared_ptr<MerginProject> project : mMerginProjects )
+  {
+    if ( project->projectDir == mDataDir + projectDir )
+    {
+      projectDeleted( getFullProjectName( project->projectNamespace, project->name ) );
```
wonder-sk: looks like we have a double for-cycle here going through list of mergin projects :-O

how about we get rid of MerginApi::projectDeleted() and simplify the code?

wonder-sk: you have added new function getProject() - why not use it in projectDeletedOnPath() ?

app/merginapi.cpp
```diff
@@ -594,6 +635,66 @@ bool MerginApi::extractProjectName( const QString &sourceString, QString &projec
...
+    if ( projectFullName == getFullProjectName( project->projectNamespace, project->name ) )
+    {
+      return project;
+    }
+  }
+
+  return std::shared_ptr<MerginProject>();
+}
+
+QString MerginApi::forceCreateDir( QString path )
```
wonder-sk: I would suggest to rename this function to something easier to understand, e.g. findUniqueProjectDirectorName( dir, name ). I would remove mkdir() calls, so it only comes up with good new directory name. The caller function would first get unique dir name, then call mkdir() itself.

Also please add docs what the function does.

app/merginapi.cpp
```diff
@@ -12,17 +12,18 @@
 MerginApi::MerginApi( const QString &dataDir, QObject *parent )
   : QObject( parent )
   , mDataDir( dataDir )
-  , mCacheFile( QStringLiteral( ".projectsCache.txt" ) )
+  , mCacheFile( QStringLiteral( "/.mergin/mergin.json" ) )
```
wonder-sk: Couldn't this be a static const string? That shouldn't change in different MerginApi instances...

app/projectsmodel.cpp
```diff
@@ -97,6 +100,10 @@ void ProjectModel::addProjectFromPath( QString path )
   if ( projectExistsAt >= 0 )
     mProjectFiles.removeAt( projectExistsAt );
 
+  std::shared_ptr<MerginProject> cachedProject = MerginApi::parseProjectData( path, mCachedName );
```
wonder-sk: I don't like the idea that ProjectModel needs to know the relative path to mergin metadata file - it should not be necessary - let's remove it

app/projectsmodel.h
```diff
@@ -35,7 +35,7 @@ class ProjectModel : public QAbstractListModel
   public:
     enum Roles
     {
-      Name = Qt::UserRole + 1,
+      Name = Qt::UserRole + 1, // name of *.gsq project file
```
wonder-sk: .qgs ?

app/merginapi.cpp
```diff
@@ -594,6 +635,66 @@ bool MerginApi::extractProjectName( const QString &sourceString, QString &projec
...
+      return project->projectDir;
+    }
+  }
+  return QStringLiteral();
+}
+
+QString MerginApi::cacheFile() const
+{
+  return mCacheFile;
+}
```
wonder-sk: this api call should not be really needed...

app/merginapi.cpp
```diff
@@ -1025,77 +1119,80 @@ ProjectList MerginApi::parseProjectsData( const QByteArray &data, bool dataFromS
...
-    projectMap.insert( QStringLiteral( "writers" ), writers );
-    array.append( projectMap );
+    projectMap.insert( QStringLiteral( "created" ), project->created.toString( Qt::ISODateWithMs ) );
+    projectMap.insert( QStringLiteral( "updated" ), project->updated.toString( Qt::ISODateWithMs ) );
+    projectMap.insert( QStringLiteral( "lastSync" ), project->lastSync.toString( Qt::ISODateWithMs ) );
+    projectMap.insert( QStringLiteral( "name" ), project->name );
+    projectMap.insert( QStringLiteral( "namespace" ), project->projectNamespace );
+
+    doc.setObject( projectMap );
+    cacheData( doc.toJson(), projectDir + "/" + mCacheFile );
```
wonder-sk: can we rename mCacheFile to something like mMetadataFile ? (or with "s" prefix instead of "m" if it will be static)
That file does not really cache anything, just keeps project metadata - best to avoid possible confusion

app/merginapi.cpp
```diff
@@ -1025,77 +1119,80 @@ ProjectList MerginApi::parseProjectsData( const QByteArray &data, bool dataFromS
...
+    return true;
   }
-  doc.setArray( array );
-  cacheProjectsData( doc.toJson() );
+  return false;
+}
+
+bool MerginApi::cacheData( const QByteArray &data, const QString &path )
```
wonder-sk: I would rename this to something like writeProjectMetadata()

app/merginapi.cpp
```diff
@@ -1025,77 +1119,80 @@ ProjectList MerginApi::parseProjectsData( const QByteArray &data, bool dataFromS
...
       }
+      p.serverUpdated = updated;
       result << std::make_shared<MerginProject>( p );
     }
   }
   return result;
 }
 
-bool MerginApi::cacheProjectsData( const QByteArray &data )
+std::shared_ptr<MerginProject> MerginApi::parseProjectData( const QString &projectPath, const QString &cacheName )
```
wonder-sk: I would rename this to readProjectMetadata() - and get rid of the cacheName argument

app/merginapi.h
```diff
@@ -206,33 +202,61 @@ class MerginApi: public QObject
...
+    * \param QString path
+    */
+    QString findUniqueProjectDirectorName( QString path );
+    /**
+    * Clears projects metadata.
+    * \param project std::shared_ptr<MerginProject> to access certain project.
+    */
+    void clearProject( std::shared_ptr<MerginProject> project );
+    QNetworkReply *getProjectInfo( const QString &projectFullName );
+    QHash<QString, MerginProject> mTempMerginProjects;
```
wonder-sk: please add a comment why it is necessary to have this hash / when it is used.

PR #326, opened 2019-06-06
Feature/logging

### LINKED ISSUE #323: Keep sync log file
In order to make debugging of sync issues easier, let's keep a log file for all uploads and downloads:
- each log item with a timestamp and project name
- log important network events (started request, finished request)
- log important sync logic information (e.g. detected 2 changed files, 1 added file)

Log file can be simple text file that is always just getting appended to inside INPUT data directory.

sklencar:
Feel free to comment

closes #323

### DIFF
--- a/app/inpututils.h
+++ b/app/inpututils.h
@@ -2,6 +2,7 @@
 #define INPUTUTILS_H
 
 #include <QObject>
+#include "merginapi.h"
 #include "qgsquickfeaturelayerpair.h"
 #include "qgsquickmapsettings.h"
 
@@ -50,7 +51,16 @@ class InputUtils: public QObject
 
     static void log( const QString &msg, const QString &info = QStringLiteral() );
 
+    static void setLogFilename( const QString &value );
+
+    static QString filesToString( QList<MerginFile> files );
+
+    static QString sLOG_FILE;
+
+
   private:
+    static QString &LOG_FILE();
+
     static void appendLog( const QByteArray &data, const QString &path );
 };
 
--- a/app/inpututils.h
+++ b/app/inpututils.h
@@ -5,6 +5,8 @@
 #include "qgsquickfeaturelayerpair.h"
 #include "qgsquickmapsettings.h"
 
+extern QString LOG_FILE;
+
 class InputUtils: public QObject
 {
     Q_OBJECT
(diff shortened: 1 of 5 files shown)


app/merginapi.cpp
```diff
@@ -876,7 +885,11 @@ void MerginApi::updateInfoReplyFinished()
   r->deleteLater();
   mPendingRequests.remove( url );
   jsonDoc.setArray( fileArray );
-  downloadProjectFiles( projectFullName, jsonDoc.toJson( QJsonDocument::Compact ) );
+  QString info = QString( "STARTED - removed: %1, updated: %2, added: %3, renamed: %4" )
+                 .arg( files.value( "added" ).size() ).arg( files.value( "updated" ).size() )
+                 .arg( files.value( "removed" ).size() ).arg( files.value( "renamed" ).size() );
```
wonder-sk: Rather than creating info string and passing it to the next function to log it, why not just log it directly here?

And it would be good to also log the individual file names...?

app/merginapi.cpp
```diff
@@ -786,12 +791,14 @@ void MerginApi::downloadProjectReplyFinished()
...
       emit notify( QStringLiteral( "Download successful" ) );
     }
+    InputUtils::log( r->url().toString(), QStringLiteral( "FINISHED" ) );
   }
   else
   {
     qDebug() << r->errorString();
     emit syncProjectFinished( projectDir, projectFullName, false );
     emit networkErrorOccurred( r->errorString(), QStringLiteral( "Mergin API error: downloadProject" ) );
+    InputUtils::log( r->url().toString(), QStringLiteral( "FAILED" ) );
```
wonder-sk: Worth adding reply's error string so that we know why it failed...

app/main.cpp
```diff
@@ -235,6 +237,7 @@ int main( int argc, char *argv[] )
   // Set/Get enviroment
   QString dataDir = getDataDir( IS_TEST );
   QString projectDir = dataDir + "/projects/";
+  LOG_FILE = projectDir + ".logs";
```
wonder-sk: Would be better to have the log filename defined in InputUtils - and have a function to set the log filename path... e.g. InputUtils::setLogFilename( ... )

app/inpututils.h
```diff
@@ -5,6 +5,8 @@
 #include "qgsquickfeaturelayerpair.h"
 #include "qgsquickmapsettings.h"
 
+extern QString LOG_FILE;
```
wonder-sk: better to define inside InputUtils as a static member rather than global variable

app/merginapi.cpp
sklencar: +1 for logging more info.
I also dont like passing extra parameter, but I think its better to log after a request is sent otherwise it would be logged but request wont be sent due to auth validation or api check. It might not happen often in this stage of data sync since it should validate and check api version while listing projects.

PR #341, opened 2019-06-26
Feature/ignore files

### LINKED ISSUE #340: Do not sync temp files
Exclude all the temporary files (e.g. spatialite, sqlite, geopackage, ~qgs files) when syncing both ways.
Sometimes, those files are 0 kb and the sync fails in Input.

sklencar:
closes #340

### DIFF
--- a/app/merginapi.h
+++ b/app/merginapi.h
@@ -257,7 +257,7 @@ class MerginApi: public QObject
     ProjectStatus getProjectStatus( std::shared_ptr<MerginProject> project, const QDateTime &lastMod );
     QDateTime getLastModifiedFileDateTime( const QString &path );
     int getProjectFilesCount( const QString &path );
-    bool isInIgnore( QFileInfo info );
+    bool isInIgnore( const QFileInfo &info );
     QByteArray getChecksum( const QString &filePath );
     QSet<QString> listFiles( const QString &projectPath );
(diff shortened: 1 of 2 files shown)


app/merginapi.h
```diff
@@ -256,6 +257,7 @@ class MerginApi: public QObject
     ProjectStatus getProjectStatus( std::shared_ptr<MerginProject> project, const QDateTime &lastMod );
     QDateTime getLastModifiedFileDateTime( const QString &path );
     int getProjectFilesCount( const QString &path );
+    bool isInIgnore( QFileInfo info );
```
wonder-sk: it's a good habit to use `const QFileInfo & info` when passing as a function argument

app/merginapi.h
```diff
@@ -319,7 +321,8 @@ class MerginApi: public QObject
     QHash<QString, QSet<QString>> mObsoleteFiles;
     QHash<QString, QString> mTransactions; // projectFullname -> transactionUUID
     QHash<QString, QNetworkReply *> mOpenConnections; // related to upload
-    QSet<QString> mIgnoreFiles = QSet<QString>() << "gpkg-shm" << "gpkg-wal" << "qgs~" << "qgz~";
+    QSet<QString> mIgnoreExtensions = QSet<QString>() << "gpkg-shm" << "gpkg-wal" << "qgs~" << "qgz~" << "pyc" << ".swap";
```
wonder-sk: swap instead of .swap

PR #342, opened 2019-06-26
Upload tests

sklencar:
Cancel and separate test with modification of a file.

### DIFF
--- a/app/test/testmerginapi.cpp
+++ b/app/test/testmerginapi.cpp
@@ -19,11 +19,12 @@ TestMerginApi::TestMerginApi( MerginApi *api, MerginProjectModel *mpm, ProjectMo
   initTestCase();
 
   testListProject();
-  testDownloadProject();
+  //testDownloadProject();
   testCreateProjectTwice();
   testDeleteNonExistingProject();
   testCreateDeleteProject();
   testUploadProject();
+  testPushChangesOfProject();
 
   cleanupTestCase();
   qDebug() << QString( "TestMerginApi - PASSED: %1/%2" ).arg( passedTests ).arg( runTests );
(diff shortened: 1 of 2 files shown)


app/test/testmerginapi.cpp
```diff
@@ -259,6 +250,52 @@ void TestMerginApi::testUploadProject()
...
+void TestMerginApi::testPushChangesOfProject()
+{
+  qDebug() << "TestMerginApi::testPushChangesOfProject START";
+  runTests++;
+
+  QString projectName = TestMerginApi::TEST_PROJECT_NAME;
+  QString projectNamespace = mUsername;
+  std::shared_ptr<MerginProject> project = prepareTestProject();
+
+  QDateTime serverT0 = project->serverUpdated;
```
wonder-sk: unused?

app/test/testmerginapi.cpp
```diff
@@ -308,6 +345,23 @@ void TestMerginApi::initTestProject()
...
+  QString projectNamespace = mUsername;
+  ProjectList projects = getProjectList();
+
+  std::shared_ptr<MerginProject> project = mApi->getProject( MerginApi::getFullProjectName( projectNamespace, projectName ) );
+  if ( project->clientUpdated < project->serverUpdated && project->serverUpdated > project->lastSyncClient.toUTC() )
+  {
+    // Fake same version on client of what is on the server
+    project->clientUpdated = project->serverUpdated;
+    project->projectDir = mProjectModel->dataDir() + projectName;
+  }
```
wonder-sk: Not sure if I understand this part - why do we set local version to the server version - is the project supposed to already exist on the server? From the name prepareTestProject() I expected we would be creating a new one...

sklencar: The `TestMerginApi::TEST_PROJECT_NAME` project is created in `initTestCase`. In current upload test functions, we strictly depend on existence of the project. However, these functions doesnt care, what is really upload on the server, just want to do kind of "push --force" (without downloading current version). Therefore `project->clientUpdated = project->serverUpdated;`, where we pretend, the test project is in the latest version to skip step of downloading whatever has been pushed to the project before.

Issue #320, opened 2019-05-29
Nicer empty projects list


wonder-sk:
If there are no local projects, the user only sees an empty white screen. It would be useful to have a short note saying something like "No projects found. See [how to create a project] and [how to download it] to your device."  - with two links pointing to our documentation.

wonder-sk:
For now the links could be these:
- create project - https://github.com/lutraconsulting/input/blob/master/docs/users/project_config.md
- download - https://github.com/lutraconsulting/input/blob/master/docs/users/data_sync.md

PR #350, opened 2019-06-28
Feature/project list ux #331

### LINKED ISSUE #331: Better UX when listing of projects failed
When I had no internet connection and tried to list projects, an error message was briefly shown and the wait indicator kept turning. It would be better to stop the wait indicator and show in the middle of the page something like "Failed to download list of projects. Please check your internet connection." - and have a big Retry button underneath to download projects again.

sklencar:
Screenshot 2019-06-28 at 13 58 53

closes #331 
related also to: #320

### DIFF
--- a/app/merginapi.cpp
+++ b/app/merginapi.cpp
@@ -547,6 +547,7 @@ void MerginApi::pingMerginReplyFinished()
   else
   {
     InputUtils::log( r->url().toString(), QStringLiteral( "FAILED - %1" ).arg( r->errorString() ) );
+    msg = r->errorString();
   }
   r->deleteLater();
   emit pingMerginFinished( apiVersion, msg );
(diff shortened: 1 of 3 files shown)


edips:
When I tried the apk, contour QGIS project opening and zooming a little slow. And when selecting a contour polyline, the app freezes. But edit panel appears for other projects. The data is here: https://we.tl/t-2oFoHsVD4U
For project panel, a custom file dialog can be preferable like below:
Screenshot_2019-06-28-22-33-57-086_net sf pentobi

wonder-sk:
@edips 
1. geometries in those projects are very complex - tens of thousands of vertices per geometry. Generation of the highlight is not very optimized so it looks like the app has frozen, but if you wait for several seconds it eventually shows the highlight and unfreezes. Something to address later...
2. project loading - actually we find it easier with the existing project list, so the user does not need to go through the files on the device manually with a file dialog.

Generally if you have some problems, it is best to create a new issue on github for each rather than posting to an unrelated pull request...

sklencar:
Close due to  git somehow omitted last commit.
See instead #358

PR #355, opened 2019-07-01
Feature/refactoring parse and compare

### LINKED ISSUE #347: Refactor parseAndCompareProjectFiles()
- The function should return a struct/class to make it easier and less error-prone to access the results.
- It should not take QNetworkReply as an argument - that will make it easier to test
- add unit testing of the function to test different cases: 1. no changes, 2. a single changed file, 3. added file, 4. removed file, 5. combination of added/changed/removed files.
- unit test should also run getProjectStatus() and check that result from it is consistent with what parseAndCompareProjectFiles() says

sklencar:
closes #347

### DIFF
--- a/app/merginapi.cpp
+++ b/app/merginapi.cpp
@@ -1125,26 +1125,13 @@ void MerginApi::updateInfoReplyFinished()
 
     for ( MerginFile file : diff.added )
     {
-      qreal rawNoOfChunks = qreal( file.size ) / UPLOAD_CHUNK_SIZE;
-      int noOfChunks = qCeil( rawNoOfChunks );
-      file.chunks = generateChunkIds( noOfChunks ); // doesnt really matter whats there, only how many chunks are expected
+      file.chunks = generateChunkIdsForSize( file.size ); // doesnt really matter whats there, only how many chunks are expected
       filesToDownload << file;
     }
 
     for ( MerginFile file : diff.modified )
     {
-      qreal rawNoOfChunks = qreal( file.size ) / UPLOAD_CHUNK_SIZE;
-      int noOfChunks = qCeil( rawNoOfChunks );
-      file.chunks = generateChunkIds( noOfChunks ); // doesnt really matter whats there, only how many chunks are expected
-      filesToDownload << file;
-    }
-
-    // renamed is handled as new file
-    for ( MerginFile file : diff.renamed )
-    {
-      qreal rawNoOfChunks = qreal( file.size ) / UPLOAD_CHUNK_SIZE;
-      int noOfChunks = qCeil( rawNoOfChunks );
-      file.chunks = generateChunkIds( noOfChunks ); // doesnt really matter whats there, only how many chunks are expected
+      file.chunks = generateChunkIdsForSize( file.size ); // doesnt really matter whats there, only how many chunks are expected
       filesToDownload << file;
     }
 
@@ -1199,13 +1186,9 @@ void MerginApi::uploadInfoReplyFinished()
     jsonArray.append( modified );
     filesToUpload.append( diff.modified );
 
-    QJsonObject removed = prepareUploadChangesJSON( diff.removed );
+    QJsonObject removed = prepareUploadChangesJSON( diff.removed, true );
     jsonArray.append( removed );
     // removed not in filesToUpload
-
-    QJsonObject renamed = prepareUploadChangesJSON( diff.renamed );
-    jsonArray.append( renamed );
-    filesToUpload.append( diff.renamed );
 
     r->deleteLater();
     mPendingRequests.remove( url );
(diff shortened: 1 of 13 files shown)


app/merginapi.h
```diff
@@ -47,6 +47,11 @@ struct MerginProject
   QList<int> writers;
 };
 
+struct ProjectDiff
+{
+  QHash<QString, QList<MerginFile>> changes;
```
wonder-sk: For ProjectDiff structure I was expecting that added, modified, removed, (renamed) files would be individual member variables so that we do not need a QHash...

app/test/testmerginapi.cpp
```diff
@@ -380,6 +438,22 @@ void TestMerginApi::copyTestProject()
   qDebug() << "TestMerginApi::copyTestProject DONE";
 }
 
+void TestMerginApi::verifyDiff( const ProjectDiff &diff, const QString &key )
+{
+  for ( QString k : diff.changes.keys() )
+  {
+    QList<MerginFile> files = diff.changes.value( k );
+    if ( key == key )
```
wonder-sk: always true?

app/test/testmerginapi.cpp
```diff
@@ -380,6 +438,22 @@ void TestMerginApi::copyTestProject()
   qDebug() << "TestMerginApi::copyTestProject DONE";
 }
 
+void TestMerginApi::verifyDiff( const ProjectDiff &diff, const QString &key )
```
wonder-sk: what should this do actually? :) check whether "added" or "removed" or "updated" is not empty? I would suggest to have a simpler test like QVERIFY( diff.updated.contains( file_expected_to_be_modified ) );

app/merginapi.h
```diff
@@ -128,16 +133,26 @@ class MerginApi: public QObject
...
+    static std::shared_ptr<MerginProject> readProjectMetadataFromPath( const QString &projectPath,
+        const QString &metadataFile = MerginApi::sMetadataFile );
 
     // Test functions
     void createProject( const QString &projectNamespace, const QString &projectName );
     void deleteProject( const QString &projectNamespace, const QString &projectName );
     void clearTokenData();
 
+    // Production and Test functions (therefore not private)
+    ProjectDiff compareProjectFiles( const QList<MerginFile> &origin, const QList<MerginFile> &current );
```
wonder-sk: please add a docstring

app/merginapi.cpp
```diff
@@ -1076,49 +1096,70 @@ void MerginApi::updateInfoReplyFinished()
...
+    }
+
+    // renamed is handled as new file
+    for ( MerginFile file : diff.renamed )
+    {
+      qreal rawNoOfChunks = qreal( file.size ) / UPLOAD_CHUNK_SIZE;
+      int noOfChunks = qCeil( rawNoOfChunks );
+      file.chunks = generateChunkIds( noOfChunks ); // doesnt really matter whats there, only how many chunks are expected
+      filesToDownload << file;
+    }
```
wonder-sk: let's kill the "renamed" section

app/merginapi.cpp
```diff
@@ -1127,75 +1168,68 @@ void MerginApi::uploadInfoReplyFinished()
...
-  r->deleteLater();
-  mPendingRequests.remove( url );
-  mFilesToUpload.insert( projectFullName, filesToUpload );
+    QJsonObject removed = prepareUploadChangesJSON( diff.removed );
+    jsonArray.append( removed );
+    // removed not in filesToUpload
+
+    QJsonObject renamed = prepareUploadChangesJSON( diff.renamed );
+    jsonArray.append( renamed );
+    filesToUpload.append( diff.renamed );
```
wonder-sk: let's kill handling of renamed

app/merginapi.cpp
```diff
@@ -1127,75 +1168,68 @@ void MerginApi::uploadInfoReplyFinished()
...
+    QJsonObject modified = prepareUploadChangesJSON( diff.modified );
+    jsonArray.append( modified );
+    filesToUpload.append( diff.modified );
 
-  r->deleteLater();
-  mPendingRequests.remove( url );
-  mFilesToUpload.insert( projectFullName, filesToUpload );
+    QJsonObject removed = prepareUploadChangesJSON( diff.removed );
+    jsonArray.append( removed );
+    // removed not in filesToUpload
```
wonder-sk: Removed files probably should be passed to the server just as a list of file names, rather than having full file info?

app/merginapi.cpp
```diff
@@ -1504,40 +1429,46 @@ std::shared_ptr<MerginProject> MerginApi::readProjectMetadataFromPath( const QSt
...
-  p.version = projectMap.value( QStringLiteral( "version" ) ).toString();
+    for ( auto it = vArray.constBegin(); it != vArray.constEnd(); ++it )
+    {
+      QJsonObject merginFileInfo = it->toObject();
+      // Include metadata of file from server in temp project's file
+      MerginFile merginFile;
+      merginFile.checksum = merginFileInfo.value( QStringLiteral( "checksum" ) ).toString();;
+      merginFile.path = merginFileInfo.value( QStringLiteral( "path" ) ).toString();;
+      merginFile.size = merginFileInfo.value( QStringLiteral( "size" ) ).toInt();;
+      merginFile.mtime =  QDateTime::fromString( merginFileInfo.value( QStringLiteral( "mtime" ) ).toString(), Qt::ISODateWithMs ).toUTC();;
```
wonder-sk: double semicolons here and in lines above

PR #362, opened 2019-07-02
Addded a proper error msg from the server reply

### LINKED ISSUE #361: Add server error cause in logs
If something fails, mergin server also returns cause of the error as the response body. It is important to log also the server error messages to better diagnose the issues.

sklencar:
closes #361

### DIFF
--- a/app/merginapi.cpp
+++ b/app/merginapi.cpp
@@ -453,8 +453,10 @@ void MerginApi::createProjectFinished()
   }
   else
   {
-    InputUtils::log( r->url().toString(), QStringLiteral( "FAILED - %1" ).arg( r->errorString() ) );
-    emit networkErrorOccurred( r->errorString(), QStringLiteral( "Mergin API error: createProject" ) );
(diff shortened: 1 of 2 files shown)


app/merginapi.cpp
```diff
@@ -878,14 +899,17 @@ void MerginApi::listProjectsReplyFinished()
...
+    InputUtils::log( r->url().toString(), QStringLiteral( "FAILED - %1" ).arg( message ) );
     mMerginProjects.clear();
 
-    if ( r->errorString() == QLatin1String( "Host requires authentication" ) )
+    int status = r->attribute( QNetworkRequest::HttpStatusCodeAttribute ).toInt();
+    if ( status == 401 || status == 400 )
     {
-      emit authRequested();
+      emit authFailed();
+      emit notify( serverMsg );
```
wonder-sk: why changing from authRequested to authFailed+notify ? does it fix a a bug?

sklencar: The code is actually obsolate, will be removed.

PR #364, opened 2019-07-04
Feature/cancel_download

### LINKED ISSUE #333: Cancel download of a project
It should be possible to cancel download (pull) of a project (currently only possible to cancel upload)

sklencar:
Added cancel download functionality
Added test for a new functionality
Fixed bugs related with refactored upload

closes #333

### DIFF
--- a/app/test/testmerginapi.cpp
+++ b/app/test/testmerginapi.cpp
@@ -9,6 +9,7 @@
 #include "inpututils.h"
 
 const QString TestMerginApi::TEST_PROJECT_NAME = "TEMPORARY_TEST_PROJECT";
+const QString TestMerginApi::TEST_PROJECT_NAME_DOWNLOAD = TestMerginApi::TEST_PROJECT_NAME + "_DOWNLOAD";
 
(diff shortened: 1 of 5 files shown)


app/test/testmerginapi.cpp
```diff
@@ -114,6 +116,83 @@ void TestMerginApi::testDownloadProject()
...
+  QList<QVariant> arguments = spy3.takeFirst();
+  QVERIFY( arguments.at( 2 ).toBool() );
+
+  // Remove the whole project
+  QDir( projectDir ).removeRecursively();
+  QFileInfo info( projectDir );
+  QDir dir( projectDir );
+  QCOMPARE( info.size(), 0 );
+  QVERIFY( dir.isEmpty() );
+  mApi->clearProject( p );
```
wonder-sk: It would be good to have a method in MerginApi to remove the whole project and so ProjectModel::deleteProject() would also use that API call. This would avoid the need to have projectDeletedOnPath() signal + slot in MerginApi. Not needed to address now, though...

PR #375, opened 2019-07-15
Bug/project list order

### LINKED ISSUE #367: Order of projects
Projects in Home tab are currently quite randomly ordered. I would suggest this ordering:
- first projects that are not from mergin (listed without namespace)
- then projects from mergin ("namespace/project"): first ordered by namespace, then in case of the same namespace by project name
- sorting is not case sensitive

Also when a new project is downloaded, it should be placed at the correct place (nowadays it is placed at the bottom).

sklencar:
closes #367

### DIFF
--- a/app/projectsmodel.h
+++ b/app/projectsmodel.h
@@ -47,6 +47,36 @@ class ProjectModel : public QAbstractListModel
     };
     Q_ENUMS( Roles )
 
+    explicit ProjectModel( const QString &dataDir, QObject *parent = nullptr );
+    ~ProjectModel();
+
+    Q_INVOKABLE QVariant data( const QModelIndex &index, int role ) const override;
+    Q_INVOKABLE QModelIndex index( int row ) const;
+    Q_INVOKABLE int rowAccordingPath( QString path ) const;
+    Q_INVOKABLE void deleteProject( int row );
+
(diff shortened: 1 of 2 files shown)


app/projectsmodel.h
```diff
@@ -88,7 +88,24 @@ class ProjectModel : public QAbstractListModel
 
       bool operator < ( const ProjectFile &str ) const
```
wonder-sk: can we rename "str" to "other" ? it is common naming scheme for operators and it is easier to follow the code then...

app/projectsmodel.h
```diff
@@ -88,7 +88,24 @@ class ProjectModel : public QAbstractListModel
...
+        if ( !projectNamespace.isEmpty() && str.projectNamespace.isEmpty() )
+        {
+          return false;
+        }
+        if ( projectNamespace.isEmpty() && !str.projectNamespace.isEmpty() )
+        {
+          return true;
+        }
+
+        if ( ( projectNamespace.isEmpty() && str.projectNamespace.isEmpty() ) || projectNamespace.toLower() == str.projectNamespace.toLower() )
```
wonder-sk: is it necessary to have also the check for isEmpty() here? doesn't the second check cover that too?

PR #376, opened 2019-07-15
Test improvements

wonder-sk:
A bunch of cleanups of mergin-related tests - see individual commits

### DIFF
--- a/app/test/testmerginapi.cpp
+++ b/app/test/testmerginapi.cpp
@@ -53,10 +53,9 @@ void TestMerginApi::initTestCase()
 
   mUsername = username;  // keep for later
 
-  // figure out the directory with our test data (in this git repo)
-  QDir dir = QFileInfo( __FILE__ ).dir();
-  QVERIFY( dir.cd( "../../test/test_data" ) );
-  mTestDataPath = dir.absolutePath();
+  QDir testDataDir( STR( INPUT_TEST_DATA_DIR ) );  // #defined in input.pro
+  mTestDataPath = testDataDir.canonicalPath();  // get rid of any ".." that may cause problems later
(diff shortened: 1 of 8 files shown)


app/test/testmerginapi.cpp
```diff
@@ -11,96 +11,106 @@
...
 
 /**
  * Download project from a scratch using fetch endpoint.
  */
 void TestMerginApi::testDownloadProject()
 {
-  qDebug() << "TestMerginApi::testDownloadProject START";
-  runTests++;
+  // create the project on the server (the content is not important)
+  QString projectName = "mobile_demo_mod";
```
sklencar: The name is not relevant, but anyway I would rename it to value of TEST_PROJECT_NAME.

app/test/testmerginapi.cpp
```diff
@@ -110,41 +120,49 @@ void TestMerginApi::testDownloadProject()
...
-
-  qDebug() << "TestMerginApi::testDownloadProject PASSED";
-  passedTests++;
+  std::shared_ptr<MerginProject> project = mApi->getProject( MerginApi::getFullProjectName( projectNamespace, projectName ) );
+  QVERIFY( project );
+  QCOMPARE( project->name, projectName );
+  QCOMPARE( project->projectNamespace, projectNamespace );
+  QCOMPARE( project->projectDir, mApi->projectsPath() + projectName );  // assuming no name clash that would require renaming of the dir
+  QCOMPARE( project->version, "v2" );  // for some unknown reason, first version in Mergin is v2 :-)
+  QCOMPARE( project->filesCount, 2 );  // .qgs + .txt file
```
sklencar: const variable of no of files also for line 137

Issue #380, opened 2019-07-16
Photo filenames


mostlyAtNight:
I wish to have (in the office) a QGIS project with a folder under that location called 'Photos' (for photos to sit in).  I want to send multiple surveyors out into the field with Input to collect photos, then, when they come back, I wish to copy/paste the photo point features they've created into my main, office-based layer which contains all my photos.

I do not wish to simply make this office-based project available via mergin as the photos folder is going to get very large.  

So I face two problems:

1. Filenaming: currently photos created by Input look like this: `IMG_00000001.jpg`

If two surveyors are out on site, they will end up taking photos with duplicated filenames.  Also with only one surveyor, there's a high lilihood of dup filenames clashing with those in the folder in the office.  I recommend that instead, Input bases photo filenames on a combination of timestamp and user ID.  This way you should never have clashing filenames and you also know who took the photo.

2. File location: photos are always placed in the project root - you should be able to specify a destination subfolder (a) where Input will place the photos.  This could be set by the user to be the same naming convention (Photos) as the office-based project so a point's attribute value would be something like 'Photos/20190707101022-Pete.jpg' - this way when it's pasted into the office-based project, the file location is valid.

mostlyAtNight:
The sub-path to stick photos should ideally be taken from here:

![image](https://user-images.githubusercontent.com/884954/61291361-3cfbaa00-a7c6-11e9-94f5-1db49471587f.png)

PR #393, opened 2019-07-29
QgsQuick external resource widget handling

sklencar:
Changes required by QgsQuick changes (related to external widget and photopanel)

Can be closed after:  https://github.com/qgis/QGIS/pull/30929

closes: #380

### DIFF
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -270,6 +270,37 @@ bool InputUtils::cpDir( const QString &srcPath, const QString &dstPath )
   return true;
 }
 
+QString InputUtils::renameFile( const QString &srcPath, const QString &prefix )
+{
+  QFile file( srcPath );
+
+  if ( file.exists() )
+  {
+    QFileInfo info( srcPath );
+    QString newFilename;
+    QString currTime = QDateTime::currentDateTime().toUTC().toString( "yyyy-MM-dd_hh_mm_ss.zzz" );
+    if ( !prefix.isEmpty() )
+    {
+      newFilename = QString( "%1_%2.%3" ).arg( prefix ).arg( currTime ).arg( info.suffix() );
+    }
+    else
+    {
+      newFilename = QString( "%1.%2" ).arg( currTime ).arg( info.suffix() );;
+    }
+    QString newPath( info.absolutePath() + "/" + newFilename );
+
+    if ( file.rename( newPath ) ) {
+        return newPath;
+    } else {
+        return QString();
+    }
+
+    if ( file.rename( newPath ) ) return newPath; else return QString();
+  }
+
+  return QString();
+}
+
 void InputUtils::log( const QString &msg, const QString &info )
 {
   QString logFilePath;
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -11,7 +11,7 @@
 #include <QFileInfo>
 
 QString InputUtils::sLogFile = QStringLiteral();
-static const QString DATE_TIME_FORMAT = QStringLiteral( "YYMMDD-HHMMSS" );
+static const QString DATE_TIME_FORMAT = QStringLiteral( "yymmdd-hhmmss" );
 
 InputUtils::InputUtils( QObject *parent ): QObject( parent )
 {
@@ -271,22 +271,13 @@ bool InputUtils::cpDir( const QString &srcPath, const QString &dstPath )
   return true;
 }
 
-QString InputUtils::renameWithUniqueName( const QString &srcPath, const QString &prefix, const QDateTime &dateTime )
+QString InputUtils::renameWithDateTime( const QString &srcPath, const QDateTime &dateTime )
 {
   if ( QFile::exists( srcPath ) )
   {
     QFileInfo info( srcPath );
-    QString newFilename;
     QString timestamp = ( dateTime.isValid() ) ? dateTime.toString( DATE_TIME_FORMAT ) : QDateTime::currentDateTime().toString( DATE_TIME_FORMAT );
-
-    if ( !prefix.isEmpty() )
-    {
-      newFilename = QString( "%1_%2.%3" ).arg( prefix ).arg( timestamp ).arg( info.suffix() );
-    }
-    else
-    {
-      newFilename = QString( "%1.%2" ).arg( timestamp ).arg( info.suffix() );;
-    }
+    QString newFilename = QString( "%1.%2" ).arg( timestamp ).arg( info.suffix() );;
     QString newPath( info.absolutePath() + "/" + newFilename );
 
     if ( QFile::rename( srcPath, newPath ) ) return newPath; else return QString();
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -11,6 +11,7 @@
 #include <QFileInfo>
 
 QString InputUtils::sLogFile = QStringLiteral();
+static const QString DATE_TIME_FORMAT = QStringLiteral( "YYMMDD-HHMMSS" );
 
 InputUtils::InputUtils( QObject *parent ): QObject( parent )
 {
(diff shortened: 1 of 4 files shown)


app/inpututils.cpp
```diff
@@ -270,6 +270,40 @@ bool InputUtils::cpDir( const QString &srcPath, const QString &dstPath )
   return true;
 }
 
+QString InputUtils::renameFile( const QString &srcPath, const QString &prefix )
```
wonder-sk: Can we have a better name for this function? From the current name I would have thought that it takes source and destination path and just does rename without extra magic :)

app/inpututils.cpp
```diff
@@ -270,6 +270,40 @@ bool InputUtils::cpDir( const QString &srcPath, const QString &dstPath )
...
+{
+  QFile file( srcPath );
+
+  if ( file.exists() )
+  {
+    QFileInfo info( srcPath );
+    QString newFilename;
+    QString currTime = QDateTime::currentDateTime().toUTC().toString( "yyyy-MM-dd_hh_mm_ss.zzz" );
```
wonder-sk: I think we can skip milliseconds, no need for so much precision :) Also I would simplify the filenames to something like YYMMDD-HHMMSS

Finally, I would use local time instead of UTC... I think that's more natural...

app/inpututils.cpp
```diff
@@ -270,6 +270,40 @@ bool InputUtils::cpDir( const QString &srcPath, const QString &dstPath )
...
+    if ( file.rename( newPath ) )
+    {
+      return newPath;
+    }
+    else
+    {
+      return QString();
+    }
+
+    if ( file.rename( newPath ) ) return newPath; else return QString();
```
wonder-sk: not sure why this is here again...?

app/inpututils.cpp
```diff
@@ -270,6 +270,40 @@ bool InputUtils::cpDir( const QString &srcPath, const QString &dstPath )
   return true;
 }
 
+QString InputUtils::renameFile( const QString &srcPath, const QString &prefix )
+{
+  QFile file( srcPath );
+
+  if ( file.exists() )
```
wonder-sk: Just a nitpick, I find it easier to read with the static version `QFile::exists( srcPath )`  ... no need to look back what is 'file' :)

app/inpututils.cpp
wonder-sk: I would suggest to have time also as an argument to this function. The default value could be invalid time (`QDateTime()`). With invalid time we would use current time. This way it's easier to write unit tests for such function (deterministic).

app/inpututils.cpp
```diff
@@ -270,6 +270,40 @@ bool InputUtils::cpDir( const QString &srcPath, const QString &dstPath )
...
+    {
+      newFilename = QString( "%1_%2.%3" ).arg( prefix ).arg( currTime ).arg( info.suffix() );
+    }
+    else
+    {
+      newFilename = QString( "%1.%2" ).arg( currTime ).arg( info.suffix() );;
+    }
+    QString newPath( info.absolutePath() + "/" + newFilename );
+
+    if ( file.rename( newPath ) )
```
wonder-sk: Also a nitpick - this could be also done as a static method `QFile::rename(oldpath, newpath)`

app/qml/ExternalResourceBundle.qml
```diff
@@ -39,17 +40,35 @@ Item {
...
-                fileName = __inputUtils.getFileName(imagePath)
-            }
-            if (!QgsQuick.Utils.fileExists(homePath, fileName)) {
-                __inputUtils.copyFile(imagePath, homePath + "/" + fileName)
-            }
-            externalResourceHandler.itemWidget.valueChanged(fileName, false)
+        property var confirmImage: function confirmImage(itemWidget, prefixToRelativePath, value) {
+          var newPath = __inputUtils.renameFile(prefixToRelativePath + "/" + value, customPrefix)
+          if (newPath) {
+            externalResourceHandler.itemWidget = itemWidget
```
wonder-sk: do we need to do the association of itemWidget, then some calls and then set itemWidget to undefined again?

app/qml/ExternalResourceBundle.qml
```diff
@@ -39,17 +40,35 @@ Item {
...
+            var newCurrentValue = QgsQuick.Utils.getRelativePath(newPath, prefixToRelativePath)
+            externalResourceHandler.itemWidget.image.currentValue = newCurrentValue
+            externalResourceHandler.itemWidget.valueChanged(newCurrentValue, newCurrentValue === "" || newCurrentValue === null)
             externalResourceHandler.itemWidget = undefined
+          }
+        }
+
+        property var imageSelected: function imageSelected(imagePath) {
```
wonder-sk: How do we know externalResourceHandler.itemWidget is set already here? And why do we reset it to undefined at the end? I am a bit confused with the logic here...

It would be also good to add a bit of documentation what confirmImage() and imageSelected() are meant to do so one does not need to dig into QGIS code...

app/qml/FeaturePanel.qml
```diff
@@ -260,6 +260,9 @@ Drawer {
         }
     }
 
-    ExternalResourceBundle {id: externalResourceBundle}
+    ExternalResourceBundle {
+      id: externalResourceBundle
+      customPrefix: __merginApi.username
+    }
```
wonder-sk: Hmm I would skip using mergin usernames in the panel. It's adding something unrelated - people may not use mergin at all...  My preference is not to have username added to image file paths - having timestamps should be good enough to avoid conflicts.

sklencar: Agree that timestamp should be good enough to avoid conflicts. However, similar approach with using username/ID has been mentioned in issue description and therefore has been implemented. If a user is not log into Mergin, only timestamp will be used anyway. 

I am easy with either keeping or removing custom prefix.

app/qml/ExternalResourceBundle.qml
sklencar: Removing reference is just kind of good will to avoid possible conflicts in case of multiple external source fields. However, this shouldn't happen at all since reference is set to bundle every time when value is going to be changed. Therefore this "cleanup" can be removed.
 
There is even a case, when a user choose image, cancel a filedialog and the whole workflow is at its end, but reference hasnt been removed.

app/inpututils.cpp
```diff
@@ -11,6 +11,7 @@
 #include <QFileInfo>
 
 QString InputUtils::sLogFile = QStringLiteral();
+static const QString DATE_TIME_FORMAT = QStringLiteral( "yymmdd-hhmmss" );
```
wonder-sk: months vs minutes...?

app/inpututils.cpp
```diff
@@ -270,6 +271,21 @@ bool InputUtils::cpDir( const QString &srcPath, const QString &dstPath )
...
+QString InputUtils::renameWithDateTime( const QString &srcPath, const QDateTime &dateTime )
+{
+  if ( QFile::exists( srcPath ) )
+  {
+    QFileInfo info( srcPath );
+    QString timestamp = ( dateTime.isValid() ) ? dateTime.toString( DATE_TIME_FORMAT ) : QDateTime::currentDateTime().toString( DATE_TIME_FORMAT );
+    QString newFilename = QString( "%1.%2" ).arg( timestamp ).arg( info.suffix() );;
+    QString newPath( info.absolutePath() + "/" + newFilename );
+
+    if ( QFile::rename( srcPath, newPath ) ) return newPath; else return QString();
```
wonder-sk: `else return QString()` could be removed for brevity

app/qml/ExternalResourceBundle.qml
```diff
@@ -7,13 +7,25 @@ import "."  // import InputStyle singleton
 Item {
 
     property alias handler: externalResourceHandler
+    property string customPrefix;
```
wonder-sk: can we remove the custom prefix altogether?

app/qml/ExternalResourceBundle.qml
```diff
@@ -39,17 +51,42 @@ Item {
...
+          var newPath = __inputUtils.renameWithDateTime(prefixToRelativePath + "/" + value)
+          if (newPath) {
+            externalResourceHandler.itemWidget = itemWidget
+            var newCurrentValue = QgsQuick.Utils.getRelativePath(newPath, prefixToRelativePath)
+            externalResourceHandler.itemWidget.image.currentValue = newCurrentValue
+            externalResourceHandler.itemWidget.valueChanged(newCurrentValue, newCurrentValue === "" || newCurrentValue === null)
+          }
+        }
+
         property var imageSelected: function imageSelected(imagePath) {
```
wonder-sk: Can we have a description for imageSelected() too? I guess it's called when an image from gallery is picked and the logic is to copy it to the project directory?

app/qml/ExternalResourceBundle.qml
wonder-sk: I am mainly interested why we need to temporarily assign the widget to to the handler at all... instead of just doing something like:
```
itemWidget.image.currentValue = newCurrentValue
itemWidget.valueChanged(...)
```
If it is necessary to assign it, then it's ok, but it would be good to have a comment why - and the reset to 'undefined' is a good idea.

sklencar: For most actions, there is no need for that. However, while choosing an image from a gallery, there are callbacks either on ok and cancel, when a externalHandler has to have a reference to a correct field's widget to finish a procedure. Just wanted to have all the same at the beginning. 

Going to remove assigning a widget to the handler if not needed.

PR #396, opened 2019-07-29
Ping url changed to get proper Mergin API version.

sklencar:
Compatible with current Mergin public/dev instance.


wonder-sk:
After discussions this will be handled in a different manner... closing this PR

PR #398, opened 2019-07-30
Loading indicators

### LINKED ISSUE #321: Loading project screen
Sometimes a project takes some time to load and the app may look frozen during that time (either when starting or when tapping on project name in the list). It would be good to have a page to indicate that a project is being loaded - ideally with a subtle animation. There wouldn't be any button to cancel etc.

### LINKED ISSUE #354: An indicator showing the progress for displaying maps
Usually takes some time to display WMTS/XYZ tiles and there is no feedback from app.

sklencar:
Added loading indicator for rendering
Added loading screen for loading projects - therefore added signals in Loader class.
Since loading projects seems to be blocking GUI for redrawing a new component just before, a short event loop has been added. 

Note: Current implementation shows loading screen even if a loading is super quick which might be annoying in a user-perspective.

closes #321 
closes #354

### DIFF
--- a/app/qml/LoadingIndicator.qml
+++ b/app/qml/LoadingIndicator.qml
@@ -6,13 +6,13 @@ ProgressBar {
     id: loadingIndicator
     anchors.top: parent.top
     value: 0
-    height: 5
-    width: parent.width
     clip: true
 
     property int easingType: Easing.InOutQuad
 
     function toggle(value) {
+      if (loadingIndicatorAnimation.running === value) return
+
       loadingIndicatorAnimation.running = value
       visible = value
     }
@@ -73,6 +73,7 @@ ProgressBar {
         property: "value"
         from: 1
         to: 1
+        duration: 100
         running: true
         loops: Animation.Infinite
         easing.type: easingType
--- a/app/qml/LoadingIndicator.qml
+++ b/app/qml/LoadingIndicator.qml
@@ -2,80 +2,37 @@ import QtQuick 2.7
 import QtQuick.Controls 2.2
 import QtGraphicalEffects 1.0
 
-ProgressBar {
-    id: loadingIndicator
-    anchors.top: parent.top
-    value: 0
-    clip: true
 
-    property int easingType: Easing.InOutQuad
+Rectangle {
+  id: loadingIndicator
+  anchors.top: parent.top
+  color: InputStyle.fontColor
 
-    function toggle(value) {
-      if (loadingIndicatorAnimation.running === value) return
+  Rectangle {
+    id: bar
+    width: parent.width
+    height: parent.height
 
-      loadingIndicatorAnimation.running = value
-      visible = value
-    }
-
-    background: Rectangle {
-        implicitWidth: loadingIndicator.width
-        implicitHeight: loadingIndicator.height
-        border.color: InputStyle.panelBackgroundDarker
-        radius: 0
-    }
-    contentItem: Item {
-        implicitWidth: loadingIndicator.width
-        implicitHeight: loadingIndicator.height
-
-        Rectangle {
-            id: bar
-            width: loadingIndicator.visualPosition * parent.width
-            height: parent.height
-            radius: 0
-        }
-
-        LinearGradient {
-            anchors.fill: bar
-            start: Qt.point(0, 0)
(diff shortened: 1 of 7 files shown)


saberraz:
That looks great @sklencar 
Tested it on a couple of project and it is much clearer that loading is in progress. For very short delays, maybe you can delay it to kick off after a second or two. But to be honest, the loading bar is so subtle that doesn't bother me at all.

app/qml/LoadingIndicator.qml
```diff
@@ -0,0 +1,81 @@
...
+    clip: true
+
+    property int easingType: Easing.InOutQuad
+
+    function toggle(value) {
+      if (loadingIndicatorAnimation.running === value) return
+
+      loadingIndicatorAnimation.running = value
+      visible = value
+    }
```
wonder-sk: Can we simplify this? Instead of toggle() method, just something like "isRunning" property, which will be assigned to "visible" and "loadingIndicatorAnimation.running" ?

app/qml/LoadingIndicator.qml
```diff
@@ -0,0 +1,81 @@
+import QtQuick 2.7
+import QtQuick.Controls 2.2
+import QtGraphicalEffects 1.0
+
+ProgressBar {
+    id: loadingIndicator
+    anchors.top: parent.top
+    value: 0
+    clip: true
```
wonder-sk: why clipping?

app/qml/LoadingIndicator.qml
```diff
@@ -0,0 +1,81 @@
...
+        id: loadingIndicatorAnimation
+        target: loadingIndicator
+        property: "value"
+        from: 1
+        to: 1
+        duration: 100
+        running: true
+        loops: Animation.Infinite
+        easing.type: easingType
+    }
```
wonder-sk: Animation from 1 to 1 ... what is the trick here? :)

app/qml/LoadingIndicator.qml
```diff
@@ -0,0 +1,81 @@
...
+        LinearGradient {
+            anchors.fill: bar
+            start: Qt.point(0, 0)
+            end: Qt.point(0, bar.height)
+            source: bar
+            gradient: Gradient {
+                GradientStop { position: 0.0; color: Qt.rgba(0,0,0,0) }
+                GradientStop { position: 0.5; color: Qt.rgba(1,1,1,0.3) }
+                GradientStop { position: 1.0; color: Qt.rgba(0,0,0,0.05) }
+            }
```
wonder-sk: why two linear gradients applied to "bar" ? don't they get mixed up?

app/qml/LoadingIndicator.qml
```diff
@@ -0,0 +1,81 @@
...
+import QtQuick.Controls 2.2
+import QtGraphicalEffects 1.0
+
+ProgressBar {
+    id: loadingIndicator
+    anchors.top: parent.top
+    value: 0
+    clip: true
+
+    property int easingType: Easing.InOutQuad
```
wonder-sk: would be good to add used colors as properties, so we could use this Item in theory elsewhere too

app/qml/LoadingIndicator.qml
```diff
@@ -0,0 +1,81 @@
...
+
+      loadingIndicatorAnimation.running = value
+      visible = value
+    }
+
+    background: Rectangle {
+        implicitWidth: loadingIndicator.width
+        implicitHeight: loadingIndicator.height
+        border.color: InputStyle.panelBackgroundDarker
+        radius: 0
```
wonder-sk: no need to set radius if it is zero? (also below)

app/qml/ProjectLoadingScreen.qml
```diff
@@ -0,0 +1,32 @@
...
+    fillMode: Image.PreserveAspectFit
+    sourceSize.width: width
+  }
+
+  Text {
+    text: "Project loading ..."
+    anchors.verticalCenterOffset: parent.height/6
+    anchors.horizontalCenter: parent.horizontalCenter
+    anchors.verticalCenter: parent.verticalCenter
+    font.italic: true
```
wonder-sk: can we remove the italic font?

PR #401, opened 2019-08-01
Improve "update before upload" handling in mergin client

### LINKED ISSUE #324: Extract dialog from merginApi
In current implementation, a QMessageBox is invoked when update has to be done before upload. Such handling should be done in different place than in MerginApi.

wonder-sk:
- removes a message box triggered from mergin api (closes #324)
- more precise server version check (using most recent server version instead of cached)
- added a test for this situation

### DIFF
--- a/app/merginapi.cpp
+++ b/app/merginapi.cpp
@@ -7,7 +7,6 @@
 #include <QDate>
 #include <QByteArray>
 #include <QSet>
-#include <QMessageBox>
 #include <QUuid>
 #include <QtMath>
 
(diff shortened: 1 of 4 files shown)


app/merginapi.cpp
```diff
@@ -1196,88 +1168,100 @@ void MerginApi::updateInfoReplyFinished()
...
+  QList<MerginFile> localFiles = getLocalProjectFiles( transaction.projectDir + "/" );
+  MerginProjectMetadata serverProject = MerginProjectMetadata::fromJson( data );
+  MerginProjectMetadata oldServerProject = MerginProjectMetadata::fromCachedJson( transaction.projectDir + "/" + sMetadataFile );
 
-    transaction.totalSize = totalSize;
-    transaction.files = filesToDownload;
+  transaction.projectMetadata = data;
+  transaction.version = serverProject.version;
+  transaction.diff = compareProjectFiles( oldServerProject.files, serverProject.files, localFiles );
+  InputUtils::log( "update", transaction.diff.dump() );
```
sklencar: Any particular reason why log "update " instead url as always?

wonder-sk: we don't have the URL available here (and it's not worth having another argument just for that). I would like to revisit these log contexts later... Over time I would like to see fewer URLs in the log - e.g. only when request is started or reply has finished. Normally we can simply just use project full name for context...

PR #405, opened 2019-08-06
Removed missing libs

sklencar:
Removed ogr and gdal provider include since update in QGIS


PeterPetrik:
already done in the other PR for arm64

PR #434, opened 2019-08-29
Feature/about panel back button

### LINKED ISSUE #432: add back button to about page
impossible on ios to get to previous menu

sklencar:
fixes #432 and warnings

### DIFF
--- a/app/qml/PanelHeader.qml
+++ b/app/qml/PanelHeader.qml
@@ -74,8 +74,3 @@ Rectangle {
     }
 
 }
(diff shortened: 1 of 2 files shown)


app/qml/PanelHeader.qml
```diff
@@ -76,3 +74,8 @@ Rectangle {
     }
 
 }
+
+/*##^## Designer {
```
PeterPetrik: what is this?

PR #439, opened 2019-09-05
Reset default project after loading crashes

### LINKED ISSUE #437: open projects panel after crash
when project crashes app on open (e.g. 
https://github.com/lutraconsulting/input/issues/435 ) when you restart app it crashes immediately (since the project immediately crashes it).

It would be good to have some kind of tmp .running file on disk and if it is present open the main projects menu instead of the project so the app is usable ...

sklencar:
Creating a file during loading project and handling crash on the next run.

closes #437

### DIFF
--- a/app/loader.cpp
+++ b/app/loader.cpp
@@ -21,6 +21,8 @@
 #endif
 #include <QDebug>
 
+const QString Loader::LOADING_FLAG_FILE_PATH = QString( "%1/.input_loading_project" ).arg( QStandardPaths::standardLocations( QStandardPaths::TempLocation ).first() );
+
 Loader::Loader( QObject *parent )
   : QObject( parent )
 {
@@ -63,8 +65,9 @@ bool Loader::load( const QString &filePath )
   }
 
   emit loadingStarted();
-  QFile flagFile( loadingFile() );
+  QFile flagFile( LOADING_FLAG_FILE_PATH );
   flagFile.open( QIODevice::WriteOnly );
+  flagFile.close();
 
   // Give some time to other (GUI) processes before loading a project in the main thread
   QEventLoop loop;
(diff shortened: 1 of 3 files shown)


app/loader.h
```diff
@@ -68,6 +70,8 @@ class Loader: public QObject
     QgsProject *mProject = nullptr;
     QgsQuickPositionKit *mPositionKit = nullptr;
     bool mRecording = false;
+    //! Exists only while loading project with non-empty path
+    QString mLoadingFile = QString( "%1/.input_loading_project" ).arg( QStandardPaths::standardLocations( QStandardPaths::TempLocation ).first() );
```
PeterPetrik: move comment to function "loadingFile" . Also I do not think we need new member variable, path can be created on the fly

app/loader.h
```diff
@@ -40,6 +40,8 @@ class Loader: public QObject
     bool isRecording() const { return mRecording; }
     void setRecording( bool isRecording );
 
+    QString loadingFile() const;
```
PeterPetrik: maybe rename to `loadingProjectProgressFilePath()` and add  some nice comment so no one confuses it with the path to the QGIS project file that loads....

app/main.cpp
```diff
@@ -279,6 +279,15 @@ int main( int argc, char *argv[] )
   QObject::connect( ma.get(), &MerginApi::syncProjectStatusChanged, &mpm, &MerginProjectModel::syncProjectStatusChanged );
   QObject::connect( ma.get(), &MerginApi::reloadProject, &loader, &Loader::reloadProject );
 
+  QFile file( loader.loadingFile() );
```
PeterPetrik: please use some descriptive names for variables, not generic `file`

app/loader.cpp
```diff
@@ -63,6 +63,8 @@ bool Loader::load( const QString &filePath )
   }
 
   emit loadingStarted();
+  QFile flagFile( loadingFile() );
+  flagFile.open( QIODevice::WriteOnly );
```
PeterPetrik: don't you need to immediately close it, so it actually stores/flushes it to the disk?

app/loader.h
sklencar: What about make it static (const), so it can be also accessible from main.cpp (the existence of the file is checked there on init) and it will be created only once?

PeterPetrik: sounds good...

Issue #418, opened 2019-08-21
Restore ios pick from gallery/showing images etc


PeterPetrik:
https://doc.qt.io/qt-5/ios-platform-notes.html#native-image-picker

crashes when FileDialog in ExternalResourceBundle is enabled

PeterPetrik:
fixed with https://github.com/lutraconsulting/input-sdk/commit/a28c329fbf280ea3574475603747fa610ac3003d

PeterPetrik:
also camera

PeterPetrik:
https://github.com/PeterPetrik/QGIS/pull/new/ios_fixes3

PeterPetrik:
```
2019-09-09 10:18:09.098180+0200 Input[833:349700] qrc:/QgsQuick/qgsquickexternalresource.qml:268: TypeError: Value is null and could not be converted to an object
2019-09-09 10:18:12.289901+0200 Input[833:349700] qrc:/QgsQuick/qgsquickphotopanel.qml:20:1: module "QtMultimedia" is not installed
```

PeterPetrik:
qgis change: https://github.com/qgis/QGIS/pull/31640

PeterPetrik:
todo

- [ ] gallery on iOS

saberraz:
Maybe this will help:
https://forum.qt.io/topic/45295/solved-accessing-ios-photo-folder/5

PeterPetrik:
required to submit app to app store:

```Guideline 2.1 - Performance - App Completeness


We discovered one or more bugs in your app when reviewed on iPad running iOS 13.2 on Wi-Fi.

Specifically, adding a photo from the gallery is not functional.```

PR #444, opened 2019-09-09
fix for camera, disable gallery for now [ios]

PeterPetrik:
fix for camera. disable gallery for now on iOS

related to https://github.com/lutraconsulting/input/issues/418


PeterPetrik:
related qgsquick change https://github.com/qgis/QGIS/pull/31640

PR #466, opened 2019-09-18
Overwrite mode in copy function

sklencar:
Copy function was ended after the first problem with copying a file (QFile::copy returns false if a file at destination already exists). Therefore some files might be skipped afterwards. However, in the cases where functions is used, it should be in overwrite mode anyway.

The investigation and change has been invoked by https://github.com/lutraconsulting/qgis-mergin-plugin/issues/50

### DIFF
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -286,8 +286,13 @@ bool InputUtils::cpDir( const QString &srcPath, const QString &dstPath )
     {
       if ( !QFile::copy( srcItemPath, dstItemPath ) )
       {
-        qDebug() << "Cannot copy file " << srcItemPath << " > " << dstItemPath;
-        return false;
+        QFile::remove( dstItemPath );
+        if ( !QFile::copy( srcItemPath, dstItemPath ) )
+        {
+          qDebug() << "Cannot copy file " << srcItemPath << " > " << dstItemPath;
+          return false;
+        }
+
       }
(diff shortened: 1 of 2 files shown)


app/inpututils.cpp
```diff
@@ -286,8 +286,13 @@ bool InputUtils::cpDir( const QString &srcPath, const QString &dstPath )
     {
       if ( !QFile::copy( srcItemPath, dstItemPath ) )
       {
-        qDebug() << "Cannot copy file " << srcItemPath << " > " << dstItemPath;
-        return false;
+        QFile::remove( dstItemPath );
```
wonder-sk: Would be good to have error handling here as well (and the same for other error messages in this function)

app/inpututils.cpp
```diff
@@ -286,8 +286,13 @@ bool InputUtils::cpDir( const QString &srcPath, const QString &dstPath )
...
+        if ( !QFile::copy( srcItemPath, dstItemPath ) )
+        {
+          qDebug() << "Cannot copy file " << srcItemPath << " > " << dstItemPath;
```
wonder-sk: maybe it would be better to put this into the log?

app/inpututils.cpp
```diff
@@ -286,8 +286,13 @@ bool InputUtils::cpDir( const QString &srcPath, const QString &dstPath )
...
+          return false;
```
wonder-sk: how about we do not return immediately and just skip the file? So we will make the best effort to copy all the files...

PR #483, opened 2019-10-08
Changed the way how list of project is fetched

### LINKED ISSUE #449: 'Explore' tab should list projects when not logged in
The 'explore' tab lists only publicly shared projects on Mergin. Right now it requires users to log in before they can list and download public projects. It would be good not to require logging in (lower the entry barrier for new users).

sklencar:
Allow user browse among public projects without authorization.
Eventually using API and applied tags in the request instated of local handling.

closes #449

### DIFF
--- a/app/qml/MerginProjectPanel.qml
+++ b/app/qml/MerginProjectPanel.qml
@@ -11,7 +11,6 @@ Item {
 
   property int activeProjectIndex: -1
   property string activeProjectPath: __projectsModel.data(__projectsModel.index(activeProjectIndex), ProjectModel.Path)
-  property string merginSearchExpression: ""
   property var busyIndicator
 
   property real rowHeight: InputStyle.rowHeightHeader * 1.2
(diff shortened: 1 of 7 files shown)


app/qml/MerginProjectPanel.qml
```diff
@@ -56,7 +55,7 @@ Item {
         if (__merginApi.hasAuthData()) {
           authPanel.visible = false
           // filters should be set already
-          __merginApi.listProjects(searchField.text)
+          __merginApi.listProjects("", "created")
```
wonder-sk: why "created" ? isn't this dependent on the active tab?

sklencar: It should be dependent indeed. However, the same case is `onAuthChanged` when myProjects button is always activated. This was kind of default behaviour to switch to my project first. Have no problem to stay in active tab which makes more sense

PR #487, opened 2019-10-11
Fixed external resource bundle (handler) that mixes feature's values

sklencar:
CurrentValue should be updated with binding, not manually in a handler
Added icon as its all related to one PR [#32175](https://github.com/qgis/QGIS/pull/32175)


wonder-sk:
Looks fine, but let's merge after the QGIS PR is merged and we have a new version of SDK (this PR should include update of SDK version)

PR #489, opened 2019-10-15
Geodiff integration

wonder-sk:
- linking to geodiff
- upload a diff of gpkg files if possible (instead of full file)
- "pull" support: be able to download diffs instead of full files + apply them (and rebase them if needed)
- unit tests


wonder-sk:
The rebase test currently fails due to https://github.com/lutraconsulting/geodiff/issues/29

app/geodiffutils.cpp
```diff
@@ -0,0 +1,96 @@
+#include "geodiffutils.h"
```
PeterPetrik: missing copyright header

app/geodiffutils.cpp
```diff
@@ -0,0 +1,96 @@
...
+  QString uuid = QUuid::createUuid().toString( QUuid::WithoutBraces );
+  QString diffName = filePath + "-diff-" + uuid;
+  QString modifiedPath = projectDir + "/" + filePath;
+  basePath = projectDir + "/.mergin/" + filePath;
+  diffPath = projectDir + "/.mergin/" + diffName;
+  return GEODIFF_createChangeset( basePath.toUtf8(), modifiedPath.toUtf8(), diffPath.toUtf8() );
+}
+
+
+bool GeodiffUtils::hasPendingChanges( const QString &projectDir, const QString &filePath )
```
PeterPetrik: https://github.com/lutraconsulting/MDAL/pull/171/files#diff-300277169f447d7fec0f5c08e9525e23R258

app/geodiffutils.h
```diff
@@ -0,0 +1,46 @@
+#ifndef GEODIFFUTILS_H
```
PeterPetrik: missing copyright header

app/geodiffutils.h
```diff
@@ -0,0 +1,46 @@
...
+    //! Test whether the file has changed according to geodiff compared to the original server version
+    static bool hasPendingChanges( const QString &projectDir, const QString &filePath );
+
+    //! Take JSON changeset summary string and parse it
+    static ChangesetSummary parseChangesetSummary( const QString &json );
+
+    //! Returns JSON with local pending changes o a diffable file
+    static QString diffableFilePendingChanges( const QString &projectDir, const QString &filePath, bool onlySummary );
+
+    //! run geodiff on a local project's file, compare it to locally cached original and create a diff file
```
PeterPetrik: Runs, compares, ...

app/geodiffutils.cpp
wonder-sk: hmm there are more files without the header - let's address that separately

app/geodiffutils.cpp
wonder-sk: ?

app/input.pro
```diff
@@ -175,3 +182,9 @@ QMAKE_CXXFLAGS += -std=c++11
 
 include(android.pri)
 include(ios.pri)
+
+HEADERS += \
```
PeterPetrik: should be in sources.pri

app/geodiffutils.cpp
wonder-sk: ???

PR #522, opened 2019-11-19
fixes #519

### LINKED ISSUE #519: Use of wrong icons in Windows package
Both for installer and the installed app.

saberraz:
this should hopefully bring consistency for icons across all platforms.


wonder-sk:
The icons in the PR are not correct either  :)

This is the correct icon (without the extra white circles):

Screenshot_20191119_204630

Probably the SVG contains some layers that were only used during the design process and were supposed to stay hidden?

PR #538, opened 2019-11-29
Feature/project status panel

### LINKED ISSUE #496: Project status panel
For local projects connected to mergin projects, it would be very useful to have a window where any local changes would be shown: added, removed, modified files. For modified files that are "diffable" (geopackages), we should also display a summary of changes inside the file - total counts of inserted/updated/deleted rows for each changed table (this information is provided by geodiff).

GUI suggestion: in the list of projects, the "trash" icon should be replaced by three dots - that one would open a menu with the following items:
1. Show local changes -- this would open the new panel
2. Sync
3. Remove project

sklencar:
Known issues: 
* QML Menu (three dots menu on ProjectPanel) is not working properly for iOS when popup() is called - position is always set to bottom instead of parent center according the docs.
* Missing iOS native alert window: https://github.com/lutraconsulting/input/issues/539

closes #496 
Screenshot 2019-12-04 at 10 25 13


app/merginapi.cpp
```diff
@@ -420,6 +419,24 @@ void MerginApi::uploadProject( const QString &projectNamespace, const QString &p
   }
 }
 
+
+bool MerginApi::infoProject( const QString &projectFullName )
```
wonder-sk: This is a bit confusing - to have infoProject() and signal infoProjectFinished() .... this operation is not asynchronous, so why not directly return the result? And do we even need this function? There is already MerginApi::getLocalProject() and MerginApi::localProjectChanges() - so why not call those directly from the project status model? The model could have an invokable function that would be called from QML and would populate the model contents.

app/merginprojectstatusmodel.cpp
```diff
@@ -0,0 +1,127 @@
...
+      }
+      else
+      {
+        GeodiffUtils::ChangesetSummary summary = GeodiffUtils::parseChangesetSummary( summaryJson ) ;
+        for ( QString key : summary.keys() )
+        {
+          QString subtext;
+          if ( summary[key].inserts != 0 ) subtext.append( QString( "inserted %1;" ).arg( summary[key].inserts ) );
+          if ( summary[key].updates != 0 ) subtext.append( QString( "updated %1;" ).arg( summary[key].updates ) );
+          if ( summary[key].deletes != 0 ) subtext.append( QString( "deleted %1;" ).arg( summary[key].deletes ) );
```
wonder-sk: the "subtext" does not seem to be used anywhere?

app/merginprojectstatusmodel.cpp
```diff
@@ -0,0 +1,127 @@
...
+  insertIntoItems( projectDiff.localDeleted, ProjectChangelogStatus::Deleted, projectDir );
+
+  for ( QString file : projectDiff.localUpdated )
+  {
+    if ( MerginApi::isFileDiffable( file ) )
+    {
+      QString summaryJson = GeodiffUtils::diffableFilePendingChanges( projectDir, file, true );
+      if ( summaryJson.startsWith( "ERROR" ) )
+      {
+        InputUtils::log( "MerginProjectStatusModel", QString( "Diff summary JSON for %1 in %2 has an error." ).arg( projectDir ).arg( file ) );
```
wonder-sk: could we get also an item for the error in GUI so it is visible to the user that diff went wrong?

app/merginapi.h
```diff
@@ -244,6 +244,12 @@ class MerginApi: public QObject
      */
     Q_INVOKABLE void updateCancel( const QString &projectFullName );
 
+    /**
+     * TODO @vsklencar
+     * \param projectFullName Project's full name to get info about
+     */
+    Q_INVOKABLE bool infoProject( const QString &projectFullName );
```
wonder-sk: can we remove the method if it is unused now?

Issue #553, opened 2019-12-05
Canvas freezes for one point project


edips:
Hello. I tested Input with a point layer project. It includes one point. CRS of the point layer is Pseudo Mercator. When opening first, it can zoom-in, zoom-out. But when I press "Zoom to Project" the map canvas freezes. "Zoom to Layer" works when I add second point. 
Another problem is when I use empty qgs project template, it doesn't zoom-in, zoom-out or extent. Is there any way to fix it? Thanks.

Tested on Android 8.0 with input 0.4.4
[point1.zip](https://github.com/lutraconsulting/input/files/3926363/point1.zip)

saberraz:
Confirmed. I cannot see my position either.

edips:
There can be a feature counter function when loading a project. If there is a one point feature, the current extent can be set to x+10, x-10, y+10, y-10. This function can be used onClicked event for Zoom to Layer action.

edips:
I debugged extent of the point, it was "Empty". After adding this code to [Loader's](https://paste.ubuntu.com/p/jGRMbzSWp4/) `zoomToProject` function "Zoom to Project" and GPS mode work:
```
if(extent.isEmpty()){
     extent.combineExtentWith( QgsRectangle(10.0,10.0,20.0,20.0) );
  }
```

saberraz:
Do you want to do a Pull Request with your changes?

edips:
I have old input repository, I will update it and I will do a Pull Request.

wonder-sk:
Fixed in #560

PR #560, opened 2019-12-06
Fix for one point and empty projects

edips:
Fix for #553

### DIFF
--- a/app/loader.cpp
+++ b/app/loader.cpp
@@ -113,9 +113,9 @@ void Loader::zoomToProject( QgsQuickMapSettings *mapSettings )
     extent.combineExtentWith( layerExtent );
   }
   if ( extent.isEmpty() )
-   {
-     extent.grow( mProject->crs().isGeographic() ? 0.01 : 1000.0 );
-   }
+  {
+    extent.grow( mProject->crs().isGeographic() ? 0.01 : 1000.0 );
+  }
   extent.scale( 1.05 );
   mapSettings->setExtent( extent );
 }
--- a/app/loader.cpp
+++ b/app/loader.cpp
@@ -112,6 +112,9 @@ void Loader::zoomToProject( QgsQuickMapSettings *mapSettings )
     QgsRectangle layerExtent = mapSettings->mapSettings().layerExtentToOutputExtent( layer, layer->extent() );
     extent.combineExtentWith( layerExtent );
   }
+  if(extent.isEmpty()){
+     extent.combineExtentWith( QgsRectangle(10.0,10.0,20.0,20.0) );
+  }
   extent.scale( 1.05 );
   mapSettings->setExtent( extent );
 }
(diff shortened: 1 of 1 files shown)


app/loader.cpp
```diff
@@ -112,6 +112,10 @@ void Loader::zoomToProject( QgsQuickMapSettings *mapSettings )
     QgsRectangle layerExtent = mapSettings->mapSettings().layerExtentToOutputExtent( layer, layer->extent() );
     extent.combineExtentWith( layerExtent );
   }
+  if ( extent.isEmpty() )
+  {
+    extent.combineExtentWith( QgsRectangle( 10.0, 10.0, 20.0, 20.0 ) );
+  }
```
wonder-sk: I am wondering about this arbitrary QgsRectangle... for a project with a single point, wouldn't it be better to use extent.grow(delta) to add some extra space around the point? E.g. delta 1000 for projected CRS and 0.01 for geographic CRS (which should both add ~1km space).

edips: You are welcome. Yes I thought the same. I didn't know about grow method. I thought it would be nice to use layer->crs().bounds().toString(2); but bounds() of CRS looks empty. I tested with WGS84, GRS80. If layer->crs()->isGeographic () 0.01 can be used. For one point projects, point's coordinates can be defined. I will try.

edips: The temporary solution also works for empty projects. The only use case is it enables GPS position, so a user can collect data from the current position. Some user may want to use empty project templates.

edips: Sorry I forgot to mention the issue #553 .

edips: I tried grow method, it works fine. I replaced the code with this:
```
if(extent.isEmpty()){
      if(layers.at(0)->crs().isGeographic()){
        extent.grow(0.01);
      }else{
          extent.grow(1000.0);
      }
  }
```

wonder-sk: Cool - this could crash though with an empty project because of the access to the first layer which could be null :)
How about something like this
```
extent.grow( mProject->crs().isGeographic() ? 0.01 : 1000.0 );
```

edips: That's much better idea. I didn't consider input would open a project without layer since there is nothing to input. 
I tested an empty project with a layer which doesn't have any feature.

wonder-sk:
@edips I think there's a tiny indentation problem with your code and so the CI failed. When it is fixed, I can merge the PR...

(by the way, it is possible to set up a git hook locally on your computer to run the indentation script before commit - to avoid these kind of problems)

edips:
OK @wonder-sk It is the first time I do a pull request with code :) I only did pull request about documentation before. I didn't know there are many checks on GitHub even with a few line of codes :) I will try to be familiar with git tools and git hook.

wonder-sk:
No problem @edips - now the build is green again and ready to merge :-) Thanks for your contribution!

PR #561, opened 2019-12-06
Fix of wrong and not connected choosing mapTheme with a record layer

### LINKED ISSUE #535: Switching between survey layers
When I try to switch between two survey layers (e.g. a polygon and point), the capture mode seems to get stuck in one of them. In my case, when I switch from polygon to point, it still thinks I am in polygon mode and keeps asking me to add more vertices to create the polygon.

sklencar:
Refactored code - moved handling of active layer and map theme to cpp code (models)

Travis build config updated due to error.

closes #535

### DIFF
--- a/app/layersmodel.cpp
+++ b/app/layersmodel.cpp
@@ -73,19 +73,6 @@ void LayersModel::updateActiveLayer( const QString &name )
   setActiveIndex( row );
 }
 
-QString LayersModel::activeLayerName() const
-{
-  return mActiveLayerName;
-}
-
-void LayersModel::setActiveLayerName( const QString &activeLayerName )
-{
-
-  mActiveLayerName = activeLayerName;
-  updateActiveLayer( mActiveLayerName );
-  emit activeLayerNameChanged( mActiveLayerName );
-}
-
 int LayersModel::activeIndex() const
 {
   return mActiveIndex;
@@ -98,7 +85,7 @@ void LayersModel::setActiveIndex( int activeIndex )
     mActiveIndex = activeIndex;
   }
   emit activeIndexChanged();
-  emit activeLayerNameChanged( mLayers.at( mActiveIndex )->name() );
+  emit activeLayerNameChanged( activeLayer()->name() );
 }
 
 QVariant LayersModel::data( const QModelIndex &index, int role ) const
--- a/app/layersmodel.cpp
+++ b/app/layersmodel.cpp
@@ -67,6 +67,27 @@ void LayersModel::reloadLayers()
   }
 }
 
+void LayersModel::updateActiveLayer( const QString &name )
+{
+  int row = rowAccordingName( name, firstNonOnlyReadableLayerIndex() );
+  setActiveIndex( row );
+}
+
+int LayersModel::activeIndex() const
+{
+  return mActiveIndex;
+}
+
+void LayersModel::setActiveIndex( int activeIndex )
+{
+  if ( mActiveIndex != activeIndex )
+  {
+    mActiveIndex = activeIndex;
+  }
+  emit activeIndexChanged();
+  emit activeLayerNameChanged( mLayers.at( mActiveIndex )->name() );
+}
+
 QVariant LayersModel::data( const QModelIndex &index, int role ) const
 {
   int row = index.row();
@@ -190,6 +211,11 @@ int LayersModel::firstNonOnlyReadableLayerIndex() const
   }
 
   return -1;
+}
+
+QgsMapLayer *LayersModel::activeLayer()
+{
+  return mLayers.at( activeIndex() );
 }
 
 int LayersModel::rowCount( const QModelIndex &parent ) const
--- a/app/layersmodel.cpp
+++ b/app/layersmodel.cpp
@@ -63,7 +63,7 @@ void LayersModel::reloadLayers()
     mLayers = allLayers;
     endResetModel();
 
-    emit layersChanged();
+    emit layersReloaded();
   }
 }
 
@@ -71,6 +71,19 @@ void LayersModel::updateActiveLayer( const QString &name )
 {
   int row = rowAccordingName( name, firstNonOnlyReadableLayerIndex() );
   setActiveIndex( row );
+}
+
+QString LayersModel::activeLayerName() const
+{
+  return mActiveLayerName;
+}
+
+void LayersModel::setActiveLayerName( const QString &activeLayerName )
+{
+
+  mActiveLayerName = activeLayerName;
+  updateActiveLayer( mActiveLayerName );
+  emit activeLayerNameChanged( mActiveLayerName );
 }
 
 int LayersModel::activeIndex() const
--- a/app/layersmodel.cpp
+++ b/app/layersmodel.cpp
@@ -69,7 +69,7 @@ void LayersModel::reloadLayers()
 
 void LayersModel::updateActiveLayer( const QString &name )
 {
-  int row = rowAccordingName( name, firstNonOnlyReadableLayerIndex() );
+  int row = rowAccordingName( name, firstWritableLayerIndex() );
   setActiveIndex( row );
 }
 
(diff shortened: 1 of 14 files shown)


app/mapthemesmodel.cpp
```diff
@@ -112,3 +112,17 @@ void MapThemesModel::applyTheme( const QString &name )
...
+  int i = 0;
+  for ( QString theme : mMapThemes )
+  {
+    if ( theme == name )
+    {
+      return i;
+    }
+    i++;
+  }
+  return defaultIndex;
```
wonder-sk: How about using `mMapThemes.indexOf(name)` ?

app/layersmodel.cpp
```diff
@@ -63,10 +63,31 @@ void LayersModel::reloadLayers()
...
+  return mActiveIndex;
+}
+
+void LayersModel::setActiveIndex( int activeIndex )
+{
+  if ( mActiveIndex != activeIndex )
+  {
+    mActiveIndex = activeIndex;
+  }
+  emit activeIndexChanged();
```
PeterPetrik: should we emit only when index is actually changed?

app/layersmodel.cpp
```diff
@@ -63,10 +63,31 @@ void LayersModel::reloadLayers()
...
     endResetModel();
 
-    emit layersChanged();
+    emit layersReloaded();
   }
 }
 
+void LayersModel::updateActiveLayer( const QString &name )
+{
+  int row = rowAccordingName( name, firstNonOnlyReadableLayerIndex() );
```
PeterPetrik: firstNonOnlyReadableLayerIndex() ... interesting ... whatabout firstWritableLayerIndex?

app/layersmodel.cpp
```diff
@@ -192,6 +213,11 @@ int LayersModel::firstNonOnlyReadableLayerIndex() const
   return -1;
 }
 
+QgsMapLayer *LayersModel::activeLayer()
+{
+  return mLayers.at( activeIndex() );
```
PeterPetrik: https://doc.qt.io/qt-5/qlist.html#at

is activeIndex() always in bounds of the mLayers? if yes, add Q_ASSERT, if not , add a condition to return nullptr

app/layersmodel.h
```diff
@@ -50,6 +51,8 @@ class LayersModel : public QAbstractListModel
     Q_INVOKABLE int noOfEditableLayers() const;
     Q_INVOKABLE int firstNonOnlyReadableLayerIndex() const;
 
+    Q_INVOKABLE QgsMapLayer *activeLayer();
```
PeterPetrik: add docs. could be nullprt? ....

app/layersmodel.h
```diff
@@ -59,16 +62,22 @@ class LayersModel : public QAbstractListModel
...
+    void activeLayerNameChanged( const QString &name );
 
   public slots:
     void reloadLayers(); //when project file changes, reload all layers, etc.
+    void updateActiveLayer( const QString &name );
 
   private:
     QgsProject *mProject;
     QList<QgsMapLayer *> mLayers; // all layers
+    int mActiveIndex;
```
PeterPetrik: add default value (=-1?)

app/mapthemesmodel.h
```diff
@@ -42,6 +43,7 @@ class MapThemesModel : public QAbstractListModel
     Q_INVOKABLE QVariant data( const QModelIndex &index, int role ) const override;
     Q_INVOKABLE QModelIndex index( int row, int column = 0, const QModelIndex &parent = QModelIndex() ) const override;
     Q_INVOKABLE void applyTheme( const QString &name );
+    Q_INVOKABLE int rowAccordingName( QString name, int defaultIndex = -1 ) const;
```
PeterPetrik: I do not understand the name of the function. either replace name or write a docs

app/main.cpp
```diff
@@ -294,23 +295,30 @@ int main( int argc, char *argv[] )
...
-  QObject::connect( &loader, &Loader::projectReloaded, &mtm, &MapThemesModel::reloadMapThemes );
-  QObject::connect( &mtm, &MapThemesModel::reloadLayers, &lm, &LayersModel::reloadLayers );
   QObject::connect( ma.get(), &MerginApi::syncProjectFinished, &pm, &ProjectModel::addProject );
   QObject::connect( ma.get(), &MerginApi::listProjectsFinished, &mpm, &MerginProjectModel::resetProjects );
   QObject::connect( ma.get(), &MerginApi::syncProjectStatusChanged, &mpm, &MerginProjectModel::syncProjectStatusChanged );
   QObject::connect( ma.get(), &MerginApi::reloadProject, &loader, &Loader::reloadProject );
 
+  QObject::connect( &loader, &Loader::projectReloaded, &mtm, &MapThemesModel::reloadMapThemes );
+  QObject::connect( &mtm, &MapThemesModel::mapThemesReloaded, &as, &AppSettings::reloadDefaultMapTheme );
+  QObject::connect( &as, &AppSettings::reloadDefaultMapThemeSignal, &mtm, &MapThemesModel::updateMapTheme );
```
PeterPetrik: you do not need to name `reloadDefaultMapThemeSignal` with signal suffix

app/main.cpp
```diff
@@ -294,23 +295,30 @@ int main( int argc, char *argv[] )
...
+  QObject::connect( &mtm, &MapThemesModel::mapThemeChanged, &as, &AppSettings::setDefaultMapTheme );
+  QObject::connect( &mtm, &MapThemesModel::mapThemeChanged, &lm, &LayersModel::reloadLayers );
```
PeterPetrik: if you change mapThemeChanged you need to reload all layers?

app/main.cpp
```diff
@@ -294,23 +295,30 @@ int main( int argc, char *argv[] )
...
+
+  QObject::connect( &lm, &LayersModel::layersReloaded, &as, &AppSettings::reloadDefaultLayer );
+  QObject::connect( &as, &AppSettings::reloadDefaultLayerSignal, &lm, &LayersModel::updateActiveLayer );
```
PeterPetrik: remove suffix signal

app/main.cpp
sklencar: List of map layers depends on mapTheme

app/main.cpp
sklencar: reloadDefaultMapThemeSignal and reloadDefaultMapTheme are related (both in AppSettings). Basically the idea is to retrieve name of default map theme and to avoid introducing AppSetting instance to the model therefore the model can be independent.

Same for the retrieving defaultLayer name in layer's model.

The signal suffix or a whole signal name can be replaced though (getDefault* and sendDefault* to keep the relation obvious)

app/layersmodel.cpp
sklencar: Signal is used in main.qml to update the RecordToolbar. Consider a use case when mapTheme has changed, layers in the layers model have changed, (default/active) index has not changed, but actual layer where the index is pointing at is different from the previous state.

A new signal can be introduced and used for updating the RecordToolbar instead + added docs.

PeterPetrik: i would emit only when index is changed, and for the testcase you describe I imaging there is signal in maptheme that is emitted already?

app/main.cpp
PeterPetrik: if you have getDefault* and sendDefault*, why it cannot be one signal that is connected directly here? If you want to send string in argument list, it does not matter, or does it?

app/main.cpp
PeterPetrik: reload means they are really fetched again from data provider? maybe I am just confused with naming.

app/layersmodel.h
```diff
@@ -41,34 +42,44 @@ class LayersModel : public QAbstractListModel
...
     QHash<int, QByteArray> roleNames() const override;
 
     QList<QgsMapLayer *> layers() const;
 
-    int defaultLayerIndex() const;
-    void setDefaultLayerIndex( int index );
+    int activeIndex() const;
+    /**
+     * Sets layer if index is in layers list length range.
+     * @param activeIndex index of the layder from mLayers
```
PeterPetrik: we use "\" notation over "@" in doxy comments

app/mapthemesmodel.h
```diff
@@ -50,16 +53,30 @@ class MapThemesModel : public QAbstractListModel
     QList<QString> mapThemes() const;
     void setMapThemes( const QList<QString> &mapThemes );
 
+    int activeThemeIndex() const;
+    /**
+     * Sets and applies mapTheme if index is in mapThemes list length range.
+     * @param activeThemeIndex index of the theme from the list of themes
```
PeterPetrik: use \ for doxy comments

PR #562, opened 2019-12-11
Fixing overflowing issue for Active Layer Panel

edips:
Active layer panel's layers are overflowing when dragging down for projects that has many layers. I added `clip:true `for list view. Another issue is when dragging down, the panel sometimes closes interactively. I turned it off and added `PanelHeader` with back button.
new

### DIFF
--- a/app/loader.cpp
+++ b/app/loader.cpp
@@ -113,9 +113,9 @@ void Loader::zoomToProject( QgsQuickMapSettings *mapSettings )
     extent.combineExtentWith( layerExtent );
   }
   if ( extent.isEmpty() )
-   {
-     extent.grow( mProject->crs().isGeographic() ? 0.01 : 1000.0 );
-   }
+  {
+    extent.grow( mProject->crs().isGeographic() ? 0.01 : 1000.0 );
+  }
   extent.scale( 1.05 );
   mapSettings->setExtent( extent );
 }
--- a/app/loader.cpp
+++ b/app/loader.cpp
@@ -112,6 +112,9 @@ void Loader::zoomToProject( QgsQuickMapSettings *mapSettings )
     QgsRectangle layerExtent = mapSettings->mapSettings().layerExtentToOutputExtent( layer, layer->extent() );
     extent.combineExtentWith( layerExtent );
   }
+  if(extent.isEmpty()){
+     extent.combineExtentWith( QgsRectangle(10.0,10.0,20.0,20.0) );
+  }
   extent.scale( 1.05 );
   mapSettings->setExtent( extent );
 }
--- a/app/loader.cpp
+++ b/app/loader.cpp
@@ -112,10 +112,10 @@ void Loader::zoomToProject( QgsQuickMapSettings *mapSettings )
     QgsRectangle layerExtent = mapSettings->mapSettings().layerExtentToOutputExtent( layer, layer->extent() );
     extent.combineExtentWith( layerExtent );
   }
-  if( extent.isEmpty() )
-  {
-    extent.grow( mProject->crs().isGeographic() ? 0.01 : 1000.0 );
-  }
+  if ( extent.isEmpty() )
+   {
+     extent.grow( mProject->crs().isGeographic() ? 0.01 : 1000.0 );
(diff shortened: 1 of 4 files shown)


edips:
I forgot `color: InputStyle.panelBackgroundLight` Sould I do a pull request again for this?

PeterPetrik:
Hi @edips, thank you for your pull request. 

@sklencar can you please have a look and do a review? thanks

app/loader.cpp
```diff
@@ -112,6 +112,10 @@ void Loader::zoomToProject( QgsQuickMapSettings *mapSettings )
     QgsRectangle layerExtent = mapSettings->mapSettings().layerExtentToOutputExtent( layer, layer->extent() );
     extent.combineExtentWith( layerExtent );
   }
+  if ( extent.isEmpty() )
```
sklencar: Doesn't seem to be related to overflow issue with ActiveLayerPanel to me. If it is not, please, remove it from PR.

app/qml/ActiveLayerPanel.qml
```diff
@@ -25,34 +25,25 @@ Drawer {
...
-            color: InputStyle.fontColor
-            font.pixelSize: InputStyle.fontPixelSizeTitle
-            font.bold: true
-            horizontalAlignment: Text.AlignLeft
-            verticalAlignment: Text.AlignVCenter
-        }
-
-        layer.enabled: true
-        layer.effect: Shadow {}
+    PanelHeader {
```
sklencar: Generally looks good, having swipe allowed on panel with list of items is not a good idea. However got two remarks:
- would be great if the panel closes on click outside as it is, so just the swiping will be disabled. Back button can be kept.
- Very similar panel is for MapThemes either with the same issue, would be nice to have them both as much similar as can be. Could you please update the MapThemePanel as well or at least create an issue? Many Thanks

app/loader.cpp
edips: Yes it isn't related. Sorry I don't know GitHub well :) I already did pull request for this cpp file before. I didn't know GitHub would consider it as a change although I changed QML file only.

app/qml/ActiveLayerPanel.qml
edips: Thank you too for this great open source project. 
- I agree with you about the panel closes on click outside. Another problem is back button doesn't work when `interactive: false` I couldn't find a way for this. When interactive is true, clicking outside works but dragging down the layers area closes the panel too. `edge: Qt.LeftEdge` can work but I think it doesn't look better than `edge: Qt.BottomEdge` When `edge: Qt.LeftEdge` is active the panel is top of the screen.
- Yes I will test and update the pull request.
Thank you.

edips:
MapThemesPanel updated same as ActiveLayerPanel. Added property to visible of Back text to PanelHeader. I think Header with back symbol + back text don't seem good compared to back symbol for MapThemesModel. I can redo changes about back text visibility property if you want. Here are two pictures with back text and without back text:
Screenshot_2020-01-11-13-08-53-401_uk co lutraconsulting

app/qml/MapThemePanel.qml
```diff
@@ -11,33 +11,25 @@ Drawer {
...
-        layer.enabled: true
-        layer.effect: Shadow {}
+    PanelHeader {
+          id: header
+          height: InputStyle.rowHeightHeader
+          width: parent.width
+          color: InputStyle.panelBackgroundLight
+          rowHeight: InputStyle.rowHeightHeader
+          titleText: qsTr("Map Themes")
+          //backTextVisible: false
```
sklencar: Text will be set to invisible after merge

app/qml/ActiveLayerPanel.qml
sklencar: I see, lets go with non-interactive panel to fix your problem. It seems ok now. Thank you!

app/loader.cpp
sklencar: yes, no worries, seems just like a leftover. However could be nice to unstage/reset the files/changes not related to PR.

PR #570, opened 2020-01-15
Added writersnames to project metadata

### LINKED ISSUE #569: Support new access
hasWriteAccess functions is checking access according user ID although it has been changed on Mergin and only usernames suppose to be use instead.

Update the function to use mainly usernames with temporary back-support of IDs

sklencar:
According changes in the Merign server due to getting rid of Ids

closes #569


PeterPetrik:
stale PR, closing for now

PR #589, opened 2020-03-09
Feature/compass direction

### LINKED ISSUE #557: Wrong compass indicator
Input displays the wrong direction for its compass.

sklencar:
Suggestions:
* trigger calculation of direction immediately after user (device) orientation has been changed? 

closes #557


app/qml/PositionMarker.qml
```diff
@@ -11,7 +12,62 @@ Item {
     property QgsQuick.PositionKit positionKit
     property color baseColor: InputStyle.highlightColor
     property bool withAccuracy: true
+    property int interval: 200 // Interval of direction marker updates in ms
+    property real threshold: 3 // threshold used to minimized direction marker updates (otherwise shaking)
```
wonder-sk: what are the units of the treshold - degrees?

app/inpututils.h
```diff
@@ -55,6 +57,8 @@ class InputUtils: public QObject
 
     Q_INVOKABLE void showNotification( const QString &message );
 
+    Q_INVOKABLE qreal groundSpeedFromSource( QgsQuickPositionKit *positionKit );
```
wonder-sk: Please add a comment what it does and it what units is the value

app/qml/PositionMarker.qml
```diff
@@ -11,7 +12,62 @@ Item {
...
+    property real speedLimit: 15  // Over speed limit, directions depends on direction of movement
+    property real direction: -1
+    property real groundSpeed: -1
```
wonder-sk: does this need to be a property? if not petter to just have it as a local "var" in onTriggered

app/qml/PositionMarker.qml
```diff
@@ -11,7 +12,62 @@ Item {
...
 
+    Timer {
+      interval: positionMarker.interval; running: true; repeat: true
+      onTriggered: {
+        positionMarker.groundSpeed = __inputUtils.groundSpeedFromSource(positionKit)
+        if (groundSpeed > 0) {
+          var newDirection = compass.reading && groundSpeed < speedLimit ? compass.reading.azimuth : positionKit.direction
+        } else {
+          newDirection = compass.reading ? compass.reading.azimuth  : -1
+        }
```
wonder-sk: Can we rewrite this condition so it's easier to read? e.g.
```
var newDirection = -1  // start with unknown direction
if (groundSpeed >= speedLimit)
  newDirection = positionKit.direction
else if (compass.reading)
  newDirection = compass.reading.azimuth
```

app/qml/PositionMarker.qml
```diff
@@ -11,7 +12,62 @@ Item {
...
+
+        var deltaAbsDirection = Math.abs(positionMarker.direction - newDirection)
```
wonder-sk: This does not seem to handle wrapping around 360 degrees - e.g. angles 358 and 0 should give delta = 2

wonder-sk: Shouldn't we account for compass.userOrientation before calculating delta?

app/qml/PositionMarker.qml
```diff
@@ -11,7 +12,62 @@ Item {
...
+        if (deltaAbsDirection > positionMarker.threshold) {
+          positionMarker.direction = newDirection + compass.userOrientation
```
wonder-sk: Do we need to apply compass.userOrientation adjustment also if ground speed is used? Because that did not seem to be needed before...

app/qml/PositionMarker.qml
```diff
@@ -11,7 +12,62 @@ Item {
...
+          }
+      }
+
+    Compass {
+      id: compass
+      active: true
+      axesOrientationMode: Sensor.AutomaticOrientation
+      skipDuplicates: true
+      bufferSize: 1
+      dataRate: 1
```
wonder-sk: Do we need to set this? The documentation says "Setting this property is not portable and can cause conflicts with other applications. Check with the sensor backend and platform documentation for any policy regarding multiple applications requesting a data rate." -- I would rather keep it default unless there's a need to change it

app/qml/PositionMarker.qml
```diff
@@ -11,7 +12,62 @@ Item {
...
+            }
+          }
+      }
+
+    Compass {
+      id: compass
+      active: true
+      axesOrientationMode: Sensor.AutomaticOrientation
+      skipDuplicates: true
+      bufferSize: 1
```
wonder-sk: Documentation says "By default, the buffer size is 1, which means no buffering." -- do we need to set this?

app/qml/PositionMarker.qml
```diff
@@ -32,14 +88,16 @@ Item {
...
         source: "gps_direction.svg"
         fillMode: Image.PreserveAspectFit
-        rotation: positionKit.direction
+        rotation: positionMarker.direction
         transformOrigin: Item.Bottom
         width: positionMarker.size * 2
         height: width
         smooth: true
-        visible: positionKit.hasPosition && positionKit.direction >= 0
+        visible: positionKit.hasPosition
```
wonder-sk: why remove `&& positionKit.direction >= 0` ... if we have no speed and no compass reading... we would be showing direction=-1 ? Probably better not to make it invisible

app/qml/PositionMarker.qml
```diff
@@ -11,7 +12,62 @@ Item {
...
+        } else {
+          newDirection = compass.reading ? compass.reading.azimuth  : -1
+        }
+
+        var deltaAbsDirection = Math.abs(positionMarker.direction - newDirection)
+        if (deltaAbsDirection > positionMarker.threshold) {
+          positionMarker.direction = newDirection + compass.userOrientation
+        }
+       }
+    }
```
wonder-sk: Would it be possible to move all this sensor reading logic to C++? I would prefer to use QML only for GUI stuff. So ideally the QML should just read "direction" property from a C++ object (which would call a "directionChanged" signal whenever needed).

app/qml/PositionMarker.qml
sklencar: yes, can be local. used for debug in main.qml.

app/digitizingcontroller.h
```diff
@@ -66,6 +66,9 @@ class DigitizingController : public QObject
     bool manualRecording() const;
     void setManualRecording( bool manualRecording );
 
+    qreal direction() const;
+    void setDirection( const qreal &direction );
```
wonder-sk, line 70: new functions but no implementation in .cpp?

app/positiondirection.h
```diff
@@ -0,0 +1,58 @@
...
+  private:
+    //! any direction value in degrees that is < -180 is not valid.
+    qreal MIN_INVALID_DIRECTION = -180.1;
+    qreal mDirection = MIN_INVALID_DIRECTION;
+    QgsQuickPositionKit *mPositionKit = nullptr;
+#ifdef MOBILE_OS
+    QOrientationSensor *mOrientationSensor = nullptr;
+    QCompass *mCompass = nullptr;
+    QTimer mTimer;
+    const qreal mDirectionTrahsold = 3; //! in degrees.
```
wonder-sk, line 51: trahsold -> threshold :-)
but maybe better to use something like mUpdateMinAngleDelta ?

app/positiondirection.h
```diff
@@ -0,0 +1,58 @@
...
+    const qreal mSpeedLimit = 4.16;  //! 4.16 m/s ~= 15km/h. Over speed limit, directions depends on direction of movement.
+    //! Returns difference of angle values in degrees.
+    qreal angleBetween( qreal d1, qreal d2 );
```
wonder-sk, line 54: What are the expected input and output values? (0,360) ? (-180, 180) ?

app/positiondirection.h
```diff
@@ -0,0 +1,58 @@
...
+    void updateDirection();
+    void setUserOrientation();
+#endif
+
+  private:
+    //! any direction value in degrees that is < -180 is not valid.
+    qreal MIN_INVALID_DIRECTION = -180.1;
+    qreal mDirection = MIN_INVALID_DIRECTION;
+    QgsQuickPositionKit *mPositionKit = nullptr;
+#ifdef MOBILE_OS
```
wonder-sk, line 47: Is it possible to get it working without these #ifdefs for MOBILE_OS ? Is it that some platforms do not have QOrientationSensor and QCompass classes defined at all? If they just don't return sensible results, we could just keep the sensor/compass as null pointers and not start the timer...

app/positiondirection.h
```diff
@@ -0,0 +1,58 @@
...
+#include <QTimer>
+
+#ifdef MOBILE_OS
+#include <QOrientationSensor>
+#include <QCompass>
+#endif
+
+#include "qgsquickpositionkit.h"
+
+class PositionDirection : public QObject
```
wonder-sk, line 14: Some basic comment for the class would be great (e.g. that it uses speed from GPS with higher speed or from compass+orientation when available, that it may have no valid direction at all, that it uses a minimal angle for updates of direction)

app/qml/PositionMarker.qml
```diff
@@ -2,16 +2,25 @@ import QtQuick 2.3
...
 import "."
+import lc 1.0
 
 Item {
     id: positionMarker
     property int size: InputStyle.rowHeightHeader/2
     property QgsQuick.PositionKit positionKit
     property color baseColor: InputStyle.highlightColor
     property bool withAccuracy: true
+    property real direction: -1
```
wonder-sk, line 16: is this used anywhere?

app/qml/PositionMarker.qml
```diff
@@ -2,16 +2,25 @@ import QtQuick 2.3
...
 
+    onPositionKitChanged: positionDirection.positionKit = positionMarker.positionKit
+
+    PositionDirection {
+      id: positionDirection
+      onDirectionChanged: direction.rotation = positionDirection.direction
```
wonder-sk, line 22: Is this needed? the "direction" item has "rotation" set to positionDirection.direction below

app/qml/PositionMarker.qml
```diff
@@ -2,16 +2,25 @@ import QtQuick 2.3
...
 
 Item {
     id: positionMarker
     property int size: InputStyle.rowHeightHeader/2
     property QgsQuick.PositionKit positionKit
     property color baseColor: InputStyle.highlightColor
     property bool withAccuracy: true
+    property real direction: -1
 
+    onPositionKitChanged: positionDirection.positionKit = positionMarker.positionKit
```
wonder-sk, line 18: can't we just set `positionKit: positionMarker.positionKit` in the `positionDirection` item definition? that would be cleaner...

app/positiondirection.h
```diff
@@ -0,0 +1,58 @@
...
+{
+    Q_OBJECT
+
+    Q_PROPERTY( qreal direction READ direction NOTIFY directionChanged )
+    Q_PROPERTY( QgsQuickPositionKit *positionKit READ positionKit WRITE setPositionKit NOTIFY positionKitChanged )
+    Q_PROPERTY( const qreal MIN_INVALID_DIRECTION READ minInvalidDirection NOTIFY minInvalidDirectionChanged )
```
wonder-sk, line 20: can we have a "bool" property hasDirection() instead of exporting MIN_INVALID_DIRECTION ? It would be much cleaner (the NOTIFY would be upon directionChanged signal)

app/positiondirection.h
sklencar: QOrientationSensor and QCompass are from Qt.Sensors which have been added only to iOS and Android build. The Win build is kept untouched (cannot test it anyway) as well as desktop. We can add Qt.Sensors to all builds and make a code more OS independent (with ignoring functionality if sensors are not available). Dont have strong idea on this.

PR #600, opened 2020-03-21
Do not forget pending status of a project

### LINKED ISSUE #371: Sync progress of a project temporarily lost
It seems pending mode of a project in the mergin project list is always forgotten after refresh
Write tests.

### LINKED ISSUE #565: When switching apps, Input loses the status of the project
When I am in a project and switch to another app, after I come back, Input has already shut down the project and takes me to another screen.

### LINKED ISSUE #582: When uploading large files get interrupted, the sync will not resume
`server replied: BAD REQUEST. Another process is running. Please try later.`

To reproduce:
download a simple project, then add a bunch of images (so it's a couple of MB to upload) - and while uploading, try to break it somehow e.g. jump out of projects list and back, out of the application and back etc - and see if you would be allowed to press sync button again instead of getting progress bar there with a cancel button

sklencar:
List of projects has been whole replaced regardless pending state of some of the projects.

closes #582 closes #565 closes #371

### DIFF
--- a/app/merginprojectmodel.cpp
+++ b/app/merginprojectmodel.cpp
@@ -90,25 +90,35 @@ int MerginProjectModel::rowCount( const QModelIndex &parent ) const
 void MerginProjectModel::resetProjects( const MerginProjectList &merginProjects )
 {
   beginResetModel();
-  mMerginProjects.clear();
+  ProjectList newProjects;
+
   for ( MerginProjectListEntry entry : merginProjects )
   {
-    std::shared_ptr<MerginProject> project = std::make_shared<MerginProject>();
-    project->projectNamespace = entry.projectNamespace;
-    project->projectName = entry.projectName;
-    project->serverUpdated = entry.serverUpdated;
+    QString fullProjectName = MerginApi::getFullProjectName(entry.projectNamespace, entry.projectName );
+    std::shared_ptr<MerginProject> currentMerginProject = findProjectByFullName( fullProjectName );
+    if (currentMerginProject && currentMerginProject->pending) {
+        newProjects << currentMerginProject;
+    } else {
+        std::shared_ptr<MerginProject> project = std::make_shared<MerginProject>();
+        project->projectNamespace = entry.projectNamespace;
+        project->projectName = entry.projectName;
+        project->serverUpdated = entry.serverUpdated;
 
-    // figure out info from local projects (projectDir etc)
-    LocalProjectInfo localProject = mLocalProjects.projectFromMerginName( entry.projectNamespace, entry.projectName );
-    if ( localProject.isValid() )
-    {
-      project->projectDir = localProject.projectDir;
-      project->status = localProject.status;
-      // TODO: what else to copy?
+        // figure out info from local projects (projectDir etc)
+        LocalProjectInfo localProject = mLocalProjects.projectFromMerginName( entry.projectNamespace, entry.projectName );
+        if ( localProject.isValid() )
+        {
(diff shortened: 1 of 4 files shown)


app/merginprojectmodel.cpp
```diff
@@ -90,25 +90,38 @@ int MerginProjectModel::rowCount( const QModelIndex &parent ) const
...
-    LocalProjectInfo localProject = mLocalProjects.projectFromMerginName( entry.projectNamespace, entry.projectName );
-    if ( localProject.isValid() )
+    QString fullProjectName = MerginApi::getFullProjectName( entry.projectNamespace, entry.projectName );
+    std::shared_ptr<MerginProject> currentMerginProject = findProjectByFullName( fullProjectName );
+    if ( currentMerginProject && currentMerginProject->pending )
     {
-      project->projectDir = localProject.projectDir;
-      project->status = localProject.status;
-      // TODO: what else to copy?
+      newProjects << currentMerginProject;
```
wonder-sk: Could you please add a comment why we are doing this? I guess we don't want to loose the "pending" and "progress" information, anything else?

I am wondering if it would not be cleaner to still make new MerginProject like before, just have these two attributes (pending+progress) copied from the old project - just like how we copy stuff from mLocalProjects. This way it would be more explicit why we do it... what do you think?

sklencar: Don't have preferences. The other solutions would bring an update of server version, if it would have increased in a meanwhile..or even deleted (not be present anymore). So +1.

However, just realised the issue is more complex, so WIP

app/merginprojectmodel.cpp
```diff
@@ -107,8 +109,17 @@ void MerginProjectModel::resetProjects( const MerginProjectList &merginProjects
       // TODO: what else to copy?
     }
 
+    if ( pendingProjects.contains( fullProjectName ) )
+    {
+
+      TransactionStatus projectTransaction = pendingProjects.value( fullProjectName );
+      project->progress = projectTransaction.transferedSize / projectTransaction.totalSize;
+      project->pending = project->progress >= 0;
```
wonder-sk: shouldn't this be `project->pending = true` if it is listed within pendingProjects ?

app/merginprojectmodel.cpp
```diff
@@ -150,6 +174,8 @@ void MerginProjectModel::syncProjectStatusChanged( const QString &projectFullNam
   project->pending = progress >= 0;
   project->progress = progress >= 0 ? progress : 0;
 
+  //if project->pending()
+
```
wonder-sk: some unwanted update?

app/merginprojectmodel.h
```diff
@@ -82,8 +83,10 @@ class MerginProjectModel: public QAbstractListModel
   private:
 
     int findProjectIndex( const QString &projectFullName );
+    std::shared_ptr<MerginProject> findProjectByFullName( const QString &projectFullName );
 
     ProjectList mMerginProjects;
+    ProjectList mMerginPendingProjects;
```
wonder-sk: unused?

app/merginprojectmodel.h
```diff
@@ -82,8 +83,10 @@ class MerginProjectModel: public QAbstractListModel
   private:
 
     int findProjectIndex( const QString &projectFullName );
+    std::shared_ptr<MerginProject> findProjectByFullName( const QString &projectFullName );
```
wonder-sk: this function does not seem to exist?

PR #615, opened 2020-04-16
Fix #613 + fix unit tests + unnecessary qml imports

### LINKED ISSUE #613: Sync failures due to broken basefile
It seems there is a bug in MerginFile::fromJsonObject() - e.g. when updating from v3 to v4, it is expected that we will get references to diffs for v2-v3 and v3-v4, so we ignore the first one. But it could happen that there is no update in v2-v3, and we end up ignoring v3-v4, ending up with broken basefiles.

Another related problem is that even when we find out that basefile was not updated correctly with diffs (for any reason), we ignore that fact and end up with basefiles.

wonder-sk:
This fixes all errors in mergin tests.

Also it seems we do not use any QML objects from QtLocation and QtSensors - can we remove them? (less stuff will need to be packaged in the final .apk)


app/qml/main.qml
```diff
@@ -7,7 +7,6 @@ import QtQuick.Controls 2.2
 import QtMultimedia 5.8
 import QtQml.Models 2.2
 import QtPositioning 5.8
-import QtLocation 5.8
```
PeterPetrik, line 10: note that in iOS the resolving of the QtQuick import is not working properly. So everything that is imported from QgisQuick library MUST be also included in the mail.qml

wonder-sk: Sure but I can run Input even without this line (so it does not seem to be needed), moreover this import has been added very recently (1-2 weeks ago)

wonder-sk: (I wanted to say I do not have that module installed on my linux system at all)

PeterPetrik:
can your run scripts/check_all.bash to fix circleci?

PR #625, opened 2020-04-25
fix windows location request

### LINKED ISSUE #624: Windows App hangs on launch when Location is not enabled
When the application is launched on windows, when user has disabled location in the OS Settings, the application hangs, it sits in Task Manager but there is nothing shown visually. 

Workaround is to allow Location in Settings before launching the application 

https://support.microsoft.com/en-us/help/4468240/windows-10-location-service-and-privacy

PeterPetrik:
fix #624


PeterPetrik:
neither of these commits helps...

PR #628, opened 2020-04-27
Increased version to 0.6.0

sklencar:
Only in version.pri


PeterPetrik:
why this is merged? it misses change in https://github.com/lutraconsulting/input/blob/master/scripts/version.cmd

PR #630, opened 2020-04-27
Fix of stuck context menu at the bottom [iOS]

### LINKED ISSUE #540: [StatusPanel] Context menu position
Context menu in the StatusPanel is always stick with the bottom of the screen. Seems like the component cannot get right position from parent item.
IMG_A0FDDAEB7459-1

sklencar:
Sets y-offset of context menu. Takes into consideration relative y value of the parent item to show it either below or above the related item if there is no space.

Tested on iOS and Android with long project list.
closes #540 

Examples of context menu related to `gallery_test`. 
IMG_0023
IMG_0024

### DIFF
--- a/app/qml/MerginProjectPanel.qml
+++ b/app/qml/MerginProjectPanel.qml
@@ -441,8 +441,7 @@ Item {
         id: contextMenu
         height: (projectNamespace && projectName) ? menuItemHeight * 2 : menuItemHeight
         width:Math.min( parent.width, 300 * QgsQuick.Utils.dp )
-        modal: true
-        dim: false
+        leftMargin: Math.max(parent.width - (width + projectsPanel.rowHeight), 0)
 
         MenuItem {
           height:  (projectNamespace && projectName) ? contextMenu.menuItemHeight : 0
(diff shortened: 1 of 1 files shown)


saberraz:
@sklencar apart from being in the wrong place on the iOS, the buttons could not be activated. Could you check, if you tap on Delete, it actually deletes the project?

sklencar:
Yes, on both iOS and Android its possible to delete a project by clicking on Delete button- acting regularly as its deletes it from a list and a storage.

Now, it appears at the top of parent element which is project item from the list (one rectangle) and its blocking the view on the name of the project. Should we move it below (or above in case of end of the list and end screen) to the related project item?

iOS and Android seems to have native delete dialog at the bottom of the screen (tried on deleting a photo from gallery), thats also an option, but thats basically what we wanted to fix on iOS

wonder-sk:
> Now, it appears at the top of parent element which is project item from the list (one rectangle) and its blocking the view on the name of the project. Should we move it below (or above in case of end of the list and end screen) to the related project item?

yes please move it below so we can see the name of the project to which the menu is related to + it would be good to move the menu to the right.

PR #651, opened 2020-05-07
INPUT_TEST was evaluated before it was defined

wonder-sk:
(and causing build error)

followup to #634


wonder-sk:
#652 replaces this PR

PR #691, opened 2020-05-20
Fixed translation strings

sklencar:


app/qml/ExternalResourceBundle.qml
```diff
@@ -165,7 +165,7 @@ Item {
         id: fileDialog
         title: qsTr( "Open Image" )
         visible: false
-        nameFilters: [ qsTr( "Image files (*.gif *.png *.jpg)" ) ]
+        nameFilters: [ qsTr( "Image files" ) + " (*.gif *.png *.jpg)" ]
```
PeterPetrik, line 168: can your revert this one? it is already translated in the previous form so we will have to change it now...

sklencar: Then the functionality heavily depends on translation which could or could not work.

wonder-sk: That's fine - we hope translators know what they're doing :-)

For example a string "There are %1 goats" is also better to be kept as a single string rather than splitting it to avoid possible translation mistakes with %1

Issue #659, opened 2020-05-11
Show all user's mergin projects


wonder-sk:
Sometimes users get confused when their mergin project does not show up in Input - this could be e.g. if they do not have a .qgs/.qgz project file inside it. It would be better to show those projects anyway and either do not allow download (with a sensible message) or to download them but do not allow opening...

wonder-sk:
We also need to properly handle the case with multiple .qgs/.qgz files within a single mergin project - that's also invalid...

PR #717, opened 2020-05-22
Show all user projects

tomasMizera:
#659  Added fetch of all projects from Mergin. Project can also be downloaded, but not opened.

Project without QGIS project file looks like this now:
image

There was also minor bug fix when `accept` signal is called twice and found string not being in translatable group.


saberraz:
The problem is, the project is not valid:
 - if you have no qgs/qgs file
 - if you have more than one qgs/qgz files

I think the wording can be still misleading.

wonder-sk:
Good point - let's also deal with the case when there are multiple .qgs/.qgz projects...

PR #733, opened 2020-05-26
Add different text for multiple projects

tomasMizera:
An option to make error messages more readable, not only: _invalid project_

image

### DIFF
--- a/app/localprojectsmanager.cpp
+++ b/app/localprojectsmanager.cpp
@@ -15,7 +15,7 @@
 #include <QDir>
 #include <QDirIterator>
 
-static QString findQgisProjectFile( const QString &projectDir )
+static QString findQgisProjectFile( const QString &projectDir, QString &err )
 {
   QList<QString> foundProjectFiles;
   QDirIterator it( projectDir, QStringList() << QStringLiteral( "*.qgs" ) << QStringLiteral( "*.qgz" ), QDir::Files, QDirIterator::Subdirectories );
@@ -28,10 +28,17 @@ static QString findQgisProjectFile( const QString &projectDir )
   {
     return foundProjectFiles.first();
   }
-  else if ( foundProjectFiles.count() > 1 ) // multiple projects
-    return QString( "ERR_MULTIPLE_PROJECTS" );
-  else // no projects
-    return QString( "ERR_NO_PROJECTS" );
+  else if ( foundProjectFiles.count() > 1 )
+  {
+    // error: multiple project files found
+    err = QString( "ERR_MULTIPLE_PROJECTS" );
+  }
+  else
+  {
+    // no projects
+    err = QString( "ERR_NO_PROJECTS" );
+  }
+  return QString();
 }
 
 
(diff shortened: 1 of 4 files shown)


app/projectsmodel.cpp
```diff
@@ -48,14 +48,18 @@ void ProjectModel::findProjectFiles()
     projectFile.projectName = project.projectName;
     projectFile.projectNamespace = project.projectNamespace;
     QDateTime created = fi.created().toUTC();   // TODO: why UTC ???
-    if ( !project.qgisProjectFilePath.isEmpty() )
```
PeterPetrik, line 51: I was thinking if LocalProjectInfo has new function `setInvalid(const std::string& reason)` or something similar, we can pass the error reason this way. this implementation does not set isValid == false, but the project is invalid. Also it uses projectFilePath as way how to pass the error around, which would not be required...

wonder-sk: Good suggestion - I am easy, also happy to merge it as is... your call Peter - change it?

PeterPetrik: if it is not too much struggle i would change it, my concern is that isValid() function with the current implementation does not say truth. and Tomas wants to hear the truth...

tomasMizera:
@PeterPetrik @wonder-sk I changed the error propagation, what do you think about it now? `isValid` in `LocalProjectInfo` stays the same (as we discussed). This solution is up for discussion, attribute `projectError` needs to be reset on project sync - this bug is already on board.

app/localprojectsmanager.cpp
```diff
@@ -28,6 +28,16 @@ static QString findQgisProjectFile( const QString &projectDir )
   {
     return foundProjectFiles.first();
   }
+  else if ( foundProjectFiles.count() > 1 )
+  {
+    // error: multiple project files found
+    err = QString( "ERR_MULTIPLE_PROJECTS" );
```
PeterPetrik, line 34: just put the translated error message directly here?

app/localprojectsmanager.cpp
```diff
@@ -28,6 +28,16 @@ static QString findQgisProjectFile( const QString &projectDir )
...
+  }
+  else
+  {
+    // no projects
+    err = QString( "ERR_NO_PROJECTS" );
```
PeterPetrik, line 39: just put the translated error message directly here?

app/localprojectsmanager.h
```diff
@@ -28,10 +28,17 @@ struct LocalProjectInfo
 {
   bool isValid() const { return !projectDir.isEmpty(); }
 
+  void setprojectFileError( const QString error ) { projectError = error; }
```
PeterPetrik, line 31: remove function

app/localprojectsmanager.cpp
```diff
@@ -109,8 +122,11 @@ void LocalProjectsManager::updateProjectStatus( const QString &projectDir )
 void LocalProjectsManager::addMerginProject( const QString &projectDir, const QString &projectNamespace, const QString &projectName )
 {
   LocalProjectInfo project;
+  QString err;
   project.projectDir = projectDir;
-  project.qgisProjectFilePath = findQgisProjectFile( projectDir );
+  project.qgisProjectFilePath = findQgisProjectFile( projectDir, err );
```
PeterPetrik, line 127: send there directly `project. qgisProjectError`

app/localprojectsmanager.h
```diff
@@ -28,10 +28,17 @@ struct LocalProjectInfo
...
+
+  bool isShowable() const { return projectError.isEmpty(); }
+
   QString projectDir;  //!< full path to the project directory
 
   QString qgisProjectFilePath;  //!< path to the .qgs/.qgz file (or empty if not have exactly one such file)
 
+  QString projectError; //!< If project is invalid, projectError carry more information why
```
PeterPetrik, line 39: qgisProjectError

app/localprojectsmanager.cpp
tomasMizera: Sure, we can.

tomasMizera: maybe not

PR #749, opened 2020-06-10
Fix of selecting project after download

### LINKED ISSUE #578: Problem opening a project immediately after download
To reproduce the issue:
1- Open Input
2- Open a project
3- Download another project
4- Switch to Home
5- Try to open the newly downloaded project...(or tap anywhere on the Home page)
6- It keeps opening the project from step 2

The work-around is to close the app and then Home screen becomes responsive.

sklencar:
Issue caused by not updated currently selected project index.

closes #578

### DIFF
--- a/app/qml/MerginProjectPanel.qml
+++ b/app/qml/MerginProjectPanel.qml
@@ -61,6 +61,16 @@ Item {
   }
 
   Connections {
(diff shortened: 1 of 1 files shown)


tomasMizera:
Looking good :+1:

PR #753, opened 2020-06-12
Inconsistent back button on Android

### LINKED ISSUE #551: Inconsistent behaviour of Back button
There are several issues with android "back" button...

It is not always consistent with in-app "back" button in the top-left corner of pages:

A: Click on a feature in map, then click again the popup to open "feature" panel, then press Back button
Issue: Nothing happens
Expected: Go back to map panel like when the in-app "back" button is pressed

B: Go to Settings > About and then press Back
Issue: it takes you to the map
Expected: Go back to Settings panel like when the in-app "back" button is pressed

When in "map" panel, android "back" button should have the meaning of going one step back. Cases where this is not correct:

C: Press Record button to go to Record mode, then press Back
Issue: Nothing happens
Expected: recording is canceled as if the Cancel button in toolbar got pressed

D: Press Record button to go to Record mode, then click on survey layer name (light green area) to open a popup with active layer selection, then press Back
Issue: Nothing happens
Expected: it should go back to the Record mode with no switch to the active layer

Finally, there's a bit of random behavior:

E: when in main "map" panel and press android back button, sometimes it works (minimizes the app) and sometimes it does nothing.

### LINKED ISSUE #751: "More" menu sometimes takes 2 clicks to open
"More" button in app map view takes 2 clicks to open if it was opened before

tomasMizera:
#551  Most of the problems are fixed now, however, there are two left: 

 1. feature form comes from qgis quick and eats active focus, so we need to change qgis quick to return focus (e.g. field text edit or calendar)
 2. `MessageDialog` qml type does not support close by Android back button (`MessageDialog` ~ system dialog). Solution might be to create custom dialog and handle focus.

It is a lot of changes but basically, now it works like this:

 - If no project is loaded, back button from main project lists (downloaded projects, my projects, shared with me and public projects) closes the app, other windows (user page, auth page, ..) first returns to home screen and then closes app
 - If a project is loaded, back no longer leads to home (downloaded projects list), but to the map of loaded project, closing app is back is hit and only map is showed

Closes #551 
> Also fixes #751


saberraz:
Very nice clean up @tomasMizera !

tomasMizera:
Thanks @saberraz, I'll create 2 separate tickets for the mentioned problems

PeterPetrik:
@saberraz you have tested it and works ok?

app/qml/AuthPanel.qml
```diff
@@ -46,17 +46,18 @@ Item {
...
   }
 
+  id: root
+  focus: true
+
   Keys.onReleased: {
+
     if (event.key === Qt.Key_Back || event.key === Qt.Key_Escape) {
-      event.accepted = true
+      event.accepted = true;
```
PeterPetrik, line 55: why is ; needed if there is newline?

app/qml/FeaturePanel.qml
```diff
@@ -43,6 +43,17 @@ Drawer {
         PropertyAnimation { properties: "height"; easing.type: Easing.InOutQuad }
     }
 
+    Item {
+      id: backHandler
+      focus: true
+      Keys.onReleased: {
+        if (event.key === Qt.Key_Back || event.key === Qt.Key_Escape) {
+          // needs to be here to close form
```
PeterPetrik, line 51: I think the comment does not give any extra info, please remove it

app/qml/MainPanel.qml
```diff
@@ -28,13 +28,20 @@ Item {
...
 
     Rectangle {
         anchors.fill: parent
         color: InputStyle.clrPanelBackground
         opacity: InputStyle.panelOpacity
     }
 
+    Keys.onReleased: {
+      if (event.key === Qt.Key_Back || event.key === Qt.Key_Escape) {
+        // if main panel has focus - nothing else is opened, back button closes app
```
PeterPetrik, line 41: not sure, this function does nothing, maybe you can remove it all? or it has missing function call?

app/qml/AuthPanel.qml
tomasMizera: not needed, I was just use to place it even in `js`

app/qml/MainPanel.qml
tomasMizera: True, it does nothing now, however I kept it there for "future" use case. It is last place we handle the back button before app is closed. But we do not need it, I can remove it.

saberraz:
Excellent. I have tested the build and the behaviours are much more predictable! Apart from the points you have mentioned in the PR and https://github.com/lutraconsulting/input/issues/551#issuecomment-630645100 the rest are fine. I suggest to merge and we can have other tickets to handle those.

Issue #579, opened 2020-02-06
Refresh the map after sync


saberraz:
It is not clear what has changed after sync. It would be good to refresh the map after each sync for the added/deleted features to show up immediately, without needing the user to zoom in/out or pan.

gubuntu:
in my experience, one has to open another project after sync and then re-open the one you're working on, to force read of all data and project updates. I agree the whole project should auto-reload after sync

saberraz:
Addressed in 0.6.4

PR #754, opened 2020-06-12
Forced reloading project after its sync

### LINKED ISSUE #542: Project reloading workflows
If I update a project on the desktop I should follow this workflow:

1. Save the project in QGIS
2. Synchronise the project in Mergin plugin in QGIS
3. Go to My projects in Input and press the dual arrow icon to sync
4. Go to home and open the project

However if the project is already open in Input, the updated project is not shown when you return to the main, the fresh project is not shown, you need to hard quite Input, restart it and then the updated project is shown.

Proposed fix: Unload the project before you sync if the project being synced matches the active project.

sklencar:
After sync, its called reloadProject which tried to force reload current project if its the one which has been synced. Does nothing if synced project haven't been opened.

closes #542

### DIFF
--- a/app/loader.cpp
+++ b/app/loader.cpp
@@ -58,6 +58,11 @@ void Loader::setRecording( bool isRecordingOn )
 
 bool Loader::load( const QString &filePath )
 {
+  return forceLoad( filePath, false );
+}
+
+bool Loader::forceLoad( const QString &filePath, bool force )
+{
   qDebug() << "Loading " << filePath;
   // Just clear project if empty
   if ( filePath.isEmpty() )
@@ -80,7 +85,7 @@ bool Loader::load( const QString &filePath )
   loop.exec();
 
   bool res = true;
-  if ( mProject->fileName() != filePath )
+  if ( mProject->fileName() != filePath || force )
   {
     res = mProject->read( filePath );
 
@@ -102,7 +107,7 @@ bool Loader::reloadProject( QString projectDir )
 {
   if ( mProject->homePath() == projectDir )
   {
-    return load( mProject->fileName() );
+    return forceLoad( mProject->fileName(), true );
   }
   return false;
 }
--- a/app/loader.cpp
+++ b/app/loader.cpp
@@ -72,7 +72,8 @@ bool Loader::forceLoad( const QString &filePath, bool force )
     return true;
   }
 
-  emit loadingStarted();
+  if ( !force )
+    emit loadingStarted();
   QFile flagFile( LOADING_FLAG_FILE_PATH );
   flagFile.open( QIODevice::WriteOnly );
   flagFile.close();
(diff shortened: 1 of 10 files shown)


app/loader.cpp
```diff
@@ -80,7 +85,7 @@ bool Loader::load( const QString &filePath )
   loop.exec();
 
   bool res = true;
-  if ( mProject->fileName() != filePath )
+  if ( mProject->fileName() != filePath || force )
```
tomasMizera, line 88: Isn't this condition always true?
```
if ( not the same || is the same)
```

tomasMizera: Maybe we would not need the `force` flag

app/loader.h
```diff
@@ -89,6 +89,9 @@ class Loader: public QObject
     LayersModel &mLayersModel;
     AppSettings &mAppSettings;
 
+    //! Reloads project; if force us is true, reloads mProject on given path, otherwise used only for loading a new projects.
```
tomasMizera, line 92: typo - us?

```suggestion
    //! Reloads project; if force is true, reloads mProject on given path, otherwise used only for loading a new projects.
```

PeterPetrik: iu

tomasMizera: @PeterPetrik You did that to me! :grin:

wonder-sk:
Shouldn't we reload the project only if the .qgs/.qgz file was modified? Or do we need to reload the project every time?

sklencar:
Consider a use case when only gpkg is updated and synced on Input.
Related to general reloading of the project with the loading splash screen - it can be done smarted and with validity check

wonder-sk:
Related ticket - #579

wonder-sk:
Can we have the logic like this upon mergin project update:
- .qgz/.qgs file was not changed -> canvas gets re-rendered, to make sure that any added/modified/removed features are visible
- .qgz/.qgs file was changed -> full project reload (because anything may have changed)

app/loader.cpp
sklencar: Force flag is not always true, only if a project has been synced and diff contains project file as Martin suggested. It has been added to extend and not change functionality of the original method and its used only after synced (not always though).

tomasMizera: I understand it is not always true, but the flag is basically being set to true when `mProject->homePath() == projectDir` and here is a comparison of something similiar, so if we moved the flag condition here, it would be: `mProject->fileName() != filePath || mProject->homePath() == projectDir` and that is what I thought is always true. But I sense that there is something I am missing.

sklencar: I think its mergeable. However, its good though that there is kind of extra step between `syncFinsihed` signal and `load` function, where local manager checks if synced project is currently active project. However, without that extra step, the condition will be always True as Tomas said, which we dont want to have. So the condition (extra step) must be there and cannot be omitted and send it directly to as a parameter to load function, otherwise any project after sync would be loaded.

PR #763, opened 2020-06-23
Set correct language when multiple system languages are in use

tomasMizera:
Adding English translation fixes this issue because if English was second language and e.g. Polish as third, it would not choose English because no such file existed. Thus it choosed Polish.


PeterPetrik:
how this input_en.ts file is created?

tomasMizera:
I created it with Qt Linguist. However we do not need to keep track of all texts (we do not need to "translate" them to english everytime we add new text). We only need that the .qm file is generated and imported to .qrc and that is now done so it should work fine :)

PeterPetrik:
the thing is that if we have this ts and qm there, there must be up to date, so there must be managed by transfixex->qm script

app/i18n/input_i18n.qrc
```diff
@@ -7,6 +7,7 @@
     <file>input_it.qm</file> 
     <file>input_es.qm</file> 
     <file>input_tr.qm</file> 
-    <file>input_fr.qm</file> 
+    <file>input_fr.qm</file>
+    <file>input_en.qm</file>
```
PeterPetrik, line 11: also to info.plist? https://github.com/lutraconsulting/input/blob/master/docs/developers/translations.md#how-to-add-language

tomasMizera: Hm, I am not sure, there is `en` already.

tomasMizera:
I updated the mentioned `release` script

PR #793, opened 2020-07-14
Refactoring of cpp layers models

### LINKED ISSUE #783: Get all identifiable layers
Changes requested in this ticket:

- ~~change `FeaturesModel` to not respond to map theme change (persist data)~~
- ~~refactor `LayersModel` so that we have one abstract layers model and then inherit another two models (one for active layer selection - editable layers; second for layers selection in attribute table - identifiable layers)~~
- ~~chage `FeaturesModel` to get info from new model (with identifiable layers)~~
- ~~add fetching of icons to BE `FeaturesModel` (from previous choosing specific icon in FE js)~~

EDIT:
after discussion, these steps should be taken:

 - use `QgsMapLayerModel` and `QgsMapLayerProxyModel` instead of custom models
 - move logic form UI components to models
 - refactor `LayersModel` so that it inherits from mentioned qgs layer model(s), not `QAbstractListModel` 
 - try to have only one layers model with option to filter between layers
 - move information about `mapSettings` away from layers model
 - move code handling active layer away from layers model
 - stayed: add fetching of icons to BE `FeaturesModel` (from previous choosing specific icon in FE js)

tomasMizera:
- `QgsMapLayerModel` and `QgsMapLayerProxyModel` used instead of `QAbstract...` models
- We now have 2 classes for layers models: `LayersModel` that inherits `QgsMapLayerModel` - it holds all layers in the project and serves as a source model to `LayersProxyModel` that inherits `QgsMapLayerProxyModel` and is instantiated two times - (1) recording layers choosing (2) browsing layers in attribute table. Both have different filters
- Decoupled `mapSettings` from layers model to `Loader`
- New class `ActiveLayer` handles active layer (setting, reading and notifying): decoupled from `LayersModel`
- Fetching of icons to BE in FeaturesModel ( from previous choosing specific icon in FE )

closes #783 

Disclaimer!
I did not intentionally delete @PeterPetrik 's licence header :slightly_smiling_face: :slightly_smiling_face:  I just deleted entire `LayersModel` and created new ~ thus with new header. 

It is probably good not to compare 1 to 1 old and new `LayersModel` because it was rewritten completely. Better to use split view probably.


app/activelayer.cpp
```diff
@@ -0,0 +1,63 @@
...
+{
+  if ( mLayer )
+    return mLayer->name();
+
+  return QString();
+}
+
+void ActiveLayer::activeMapThemeChanged()
+{
+  setActiveLayer( nullptr );
```
wonder-sk, line 48: Instead of resetting active layer on theme change, can we check whether the previously active layer is present in the new map theme and do nothing if it is? If a user switches themes from time to time, it would be nice if the survey layer stayed the same if possible...

app/activelayer.h
```diff
@@ -0,0 +1,64 @@
...
+ * \see LayersProxyModel for methods to convert layer name / index to layer.
+ */
+class ActiveLayer : public QObject
+{
+    Q_OBJECT
+
+    Q_PROPERTY( QString layerId READ layerId )
+    Q_PROPERTY( QString layerName READ layerName )
+    Q_PROPERTY( QgsMapLayer *layer WRITE setActiveLayer READ layer NOTIFY activeLayerChanged )
+    Q_PROPERTY( QgsVectorLayer *vectorLayer READ vectorLayer )
```
wonder-sk, line 32: I think all these properties should have `NOTIFY activeLayerChanged` - otherwise when used in QML, Qt will not know that the value has changed...

app/activelayer.h
```diff
@@ -0,0 +1,64 @@
...
+#include <QObject>
+#include "qgsmaplayer.h"
+#include "qgsvectorlayer.h"
+
+
+/**
+ * @brief The ActiveLayer class holds information about current active layer for recording.
+ * It supports setting active layer by layer and offers information about layer.
+ *
+ * \see LayersProxyModel for methods to convert layer name / index to layer.
```
wonder-sk, line 23: this note is probably not relevant anymore?

app/activelayer.h
```diff
@@ -0,0 +1,64 @@
...
+#define ACTIVELAYER_H
+
+#include <QObject>
+#include "qgsmaplayer.h"
+#include "qgsvectorlayer.h"
+
+
+/**
+ * @brief The ActiveLayer class holds information about current active layer for recording.
+ * It supports setting active layer by layer and offers information about layer.
```
wonder-sk, line 21: It would be useful to add a note that it is registered as `__activeLayer` object in QML ?

app/featuresmodel.cpp
```diff
@@ -57,28 +56,21 @@ QVariant FeaturesModel::data( const QModelIndex &index, int role ) const
...
-    case FeatureId:
-      return QVariant( feat.feature().id() );
-    case Description:
-      return QVariant( QString( "description" ) );
-    case GeometryType:
-      return QVariant( feat.feature().geometry().type() );
-    default:
-      return QVariant();
+    case FeatureId: return QVariant( feat.feature().id() );
+    case Description: return QVariant( QString( "description" ) );
```
wonder-sk, line 60: Let's simply use "Feature ID %1" .arg( feat.feature().id() ) ... it will be better than just "description" :-)

app/layersmodel.cpp
```diff
@@ -13,126 +7,35 @@
...
-  emit activeIndexChanged();
-
-  QgsMapLayer *layer = activeLayer();
-  if ( layer )
-    return layer->name();
-  else
-    return QString();
-}
+  QgsVectorLayer *vectorLayer;
+  QgsMapLayer *layer;
```
wonder-sk, line 26: IMHO it's better to declare local variables at the same time they are first initialized... this way I need to check further down to make sure they don't stay uninitialized and it's two lines of code more :-)

app/layersmodel.cpp
```diff
@@ -13,126 +7,35 @@
...
-QVariant LayersModel::data( const QModelIndex &index, int role ) const
-{
-  int row = index.row();
-  if ( row < 0 || row >= mLayers.count() )
-    return QVariant( "" );
+  layer = layerFromIndex( index );
+  if ( !layer || !layer->isValid() ) return QVariant();
 
-  QgsMapLayer *layer = mLayers.at( row );
+  vectorLayer = qobject_cast<QgsVectorLayer *>( layer );
```
wonder-sk, line 31: shall we check for null pointer in case `layer` is not QgsVectorLayer ?

app/layersmodel.h
```diff
@@ -17,81 +11,33 @@
...
-#include <QList>
-#include <QSet>
+#include <QObject>
 
-class QgsQuickMapSettings;
-class QgsMapLayer;
-class QgsProject;
+#include "qgsmaplayermodel.h"
+#include "qgsmaplayer.h"
+#include "qgsvectorlayer.h"
```
wonder-sk, line 18: Whenever possible, I prefer to keep #includes outside of header files and only use forward declarations... for example do we need to #include "qgsvectorlayer.h" here?

app/layersproxymodel.h
```diff
@@ -0,0 +1,78 @@
...
+
+#include <QObject>
+
+#include "qgsmaplayerproxymodel.h"
+#include "qgsmaplayermodel.h"
+#include "qgsvectorlayer.h"
+#include "qgsmaplayer.h"
+#include "qgsproject.h"
+#include "qgslayertree.h"
+#include "qgslayertreelayer.h"
```
wonder-sk, line 22: some #includes could be surely moved to .cpp and/or replaced by forward declarations?

app/loader.cpp
```diff
@@ -263,3 +303,22 @@ QList<QgsExpressionContextScope *> Loader::globalProjectLayerScopes( QgsMapLayer
   scopes << QgsExpressionContextUtils::layerScope( layer );
   return scopes;
 }
+
+void Loader::setActiveLayerFromName( QString layerName ) const
+{
+  mLayersProxyModel.invalidate();
+  mActiveLayer.setActiveLayer(
+    mLayersProxyModel.layerFromName( layerName )
+  );
```
wonder-sk, line 312: I am wondering why we need layers proxy model here - why not just go through QgsProject's map layers and use the one with matching name?

app/loader.cpp
```diff
@@ -263,3 +303,22 @@ QList<QgsExpressionContextScope *> Loader::globalProjectLayerScopes( QgsMapLayer
...
+
+QString Loader::loadIconFromLayer( QgsMapLayer *layer )
+{
+  if ( !layer )
+    return QString();
+
+  return mLayersProxyModel.data(
+           mLayersProxyModel.index( mLayersProxyModel.indexFromLayer( layer ), 0 ),
+           LayersModel::IconSourceRole
+         ).toString();
```
wonder-sk, line 323: Again, not sure why we need a model here... can't we just check layer's geometry type and return the icon path directly without model?

I think ideally Loader should be free from any model-specific code ...

app/main.cpp
```diff
@@ -243,13 +246,16 @@ static void init_proj( const QString &pkgPath )
...
-  qmlRegisterUncreatableType<LayersModel>( "lc", 1, 0, "LayersModel", "" );
   qmlRegisterUncreatableType<MapThemesModel>( "lc", 1, 0, "MapThemesModel", "" );
   qmlRegisterUncreatableType<Loader>( "lc", 1, 0, "Loader", "" );
   qmlRegisterUncreatableType<AppSettings>( "lc", 1, 0, "AppSettings", "" );
   qmlRegisterUncreatableType<MerginApiStatus>( "lc", 1, 0, "MerginApiStatus", "MerginApiStatus Enum" );
   qmlRegisterUncreatableType<MerginProjectStatusModel>( "lc", 1, 0, "MerginProjectStatusModel", "Enum" );
   qmlRegisterUncreatableType<FeaturesModel>( "lc", 1, 0, "FeaturesModel", "" );
+  qmlRegisterUncreatableType<LayersProxyModel>( "lc", 1, 0, "BrowseDataLayersModel", "" );
+  qmlRegisterUncreatableType<LayersModel>( "lc", 1, 0, "LayersModel", "" );
+  qmlRegisterUncreatableType<LayersProxyModel>( "lc", 1, 0, "LayersProxyModel", "" );
```
wonder-sk, line 257: Why do we register LayersProxyModel class twice? (as `LayersProxyModel` and as `BrowseDataLayersModel`)

app/activelayer.h
tomasMizera: I put `NOTIFY` only to the one that actually change the active layer - `layer`; others only read the active layer, not changing it so I thought we do not need it there. Is that true or should I add them?

app/activelayer.cpp
tomasMizera: sure, good idea

app/featuresmodel.cpp
tomasMizera: I wanted to change this in #796, but sure I can change it here

app/layersmodel.cpp
tomasMizera: Well, yes and no. At first I had a check there, but it is not always true that when `vectorLayer` is null, we want to return nothing. 
There is a case when Raster layer looks for icon source and it would then return empty qvariant. That is why I moved the check to the cases in switch that use vectorLayer.

app/activelayer.h
tomasMizera: I added this note now, it was not there before :grin:

app/activelayer.h
wonder-sk: As far as I know, if you don't add NOTIFY to a property, QML will think it is constant. So if you use layerName somewhere in QML code and then the active layer changes, then QML will still use the old layer name.

app/activelayer.h
wonder-sk: Ah ok... I have misread it, sorry

app/layersmodel.cpp
wonder-sk: Ah ok I see - I thought there was a chance of null pointer dereference... but looking again, there isn't :-)

app/loader.cpp
tomasMizera: I understand the reason but .. would then more components be handling the same thing?

wonder-sk: well... it seems to me simpler to do just get layer from name using QgsProject::mapLayersByName() than going through some extra proxy model class... It would make the Loader design simpler if it knew nothing about some models and proxy models

tomasMizera: Cool, I did not know about `mapLayersByName` function :)

app/activelayer.h
tomasMizera: updated

app/loader.cpp
tomasMizera: updated

PeterPetrik:
@tomasMizera can you please rebase with the dev_0.7.0 branch ?

tomasMizera:
> @tomasMizera can you please rebase with the dev_0.7.0 branch ?

@PeterPetrik Done

PeterPetrik:
hmm, this was merge, better to do rebase (`git rebase -i origin/dev_0.7.0`) since merge does nasty history

tomasMizera:
hmm, I did `git rebase dev_0.7.0`. First I pulled `dev..` and then in feature branch used `git rebase dev..`. Isn't it the same?

app/loader.h
```diff
@@ -75,19 +109,24 @@ class Loader: public QObject
...
     QList<QgsExpressionContextScope *> globalProjectLayerScopes( QgsMapLayer *layer );
     QgsProject *mProject = nullptr;
     QgsQuickPositionKit *mPositionKit = nullptr;
     bool mRecording = false;
 
     MapThemesModel &mMapThemeModel;
-    LayersModel &mLayersModel;
     AppSettings &mAppSettings;
+    ActiveLayer &mActiveLayer;
+    LayersProxyModel &mLayersProxyModel;
```
wonder-sk, line 128: Now that we don't use the proxy model, let's remove it from Loader...

tomasMizera: Sure!

tomasMizera: done

PeterPetrik:
> hmm, I did `git rebase dev_0.7.0`. First I pulled `dev..` and then in feature branch used `git rebase dev..`. Isn't it the same?

ok, np

https://imgs.xkcd.com/comics/git_2x.png

saberraz:
Very nice... Great work @tomasMizera

PR #799, opened 2020-07-21
fix ios CI

PeterPetrik:


PeterPetrik:
autotests are failing due to problem in deployment of test.dev.cloudmergin.com @varmar05 should look at it

PR #820, opened 2020-07-28
Feature/mergin variables #776

### LINKED ISSUE #776: Add mergin related variables to Input
Same as for  QGIS plugin: 
https://github.com/lutraconsulting/qgis-mergin-plugin/issues/111

sklencar:
closes #776

### DIFF
--- a/app/localprojectsmanager.cpp
+++ b/app/localprojectsmanager.cpp
@@ -51,6 +51,16 @@ LocalProjectInfo LocalProjectsManager::projectFromDirectory( const QString &proj
   return LocalProjectInfo();
 }
 
+LocalProjectInfo LocalProjectsManager::projectByProjectFilePath( const QString &projectFilePath ) const
+{
+  for ( const LocalProjectInfo &info : mProjects )
+  {
+    if ( info.qgisProjectFilePath == projectFilePath )
+      return info;
+  }
+  return LocalProjectInfo();
+}
+
 LocalProjectInfo LocalProjectsManager::projectFromMerginName( const QString &projectFullName ) const
 {
   for ( const LocalProjectInfo &info : mProjects )
--- a/app/localprojectsmanager.cpp
+++ b/app/localprojectsmanager.cpp
@@ -51,7 +51,7 @@ LocalProjectInfo LocalProjectsManager::projectFromDirectory( const QString &proj
   return LocalProjectInfo();
 }
 
-LocalProjectInfo LocalProjectsManager::projectByProjectFilePath( const QString &projectFilePath ) const
+LocalProjectInfo LocalProjectsManager::projectFromProjectFilePath( const QString &projectFilePath ) const
 {
   for ( const LocalProjectInfo &info : mProjects )
   {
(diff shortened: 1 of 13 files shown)


app/variablesmanager.cpp
```diff
@@ -0,0 +1,41 @@
+#include "variablesmanager.h"
+
+#include "qgsexpressioncontextutils.h"
+
+VariablesManager::VariablesManager( MerginApi *merginApi, QObject *parent )
+  : QObject( parent )
+  , mMerginApi( merginApi )
+{
+  QgsExpressionContextUtils::setGlobalVariable( QStringLiteral( "mergin_url" ),  mMerginApi->apiRoot() );
```
wonder-sk, line 9: Is it always guaranteed that authChanged() is called? e.g. if user opens Input with a project loaded without going to Projects page, is the username going to be set?

app/variablesmanager.cpp
```diff
@@ -0,0 +1,41 @@
...
+}
+
+void VariablesManager::apiRootChanged()
+{
+  QgsExpressionContextUtils::setGlobalVariable( QStringLiteral( "mergin_url" ),  mMerginApi->apiRoot() );
+}
+
+void VariablesManager::authChanged()
+{
+  QgsExpressionContextUtils::setGlobalVariable( QStringLiteral( "mergin_username" ),  mMerginApi->merginUserName() );
```
wonder-sk, line 22: What if user logs out - or fails to log in - will be the username set?

app/variablesmanager.cpp
```diff
@@ -0,0 +1,41 @@
...
+}
+
+void VariablesManager::merginProjectChanged( QgsProject *project )
+{
+  QString filePath = project->fileName();
+  LocalProjectInfo info = mMerginApi->localProjectsManager().projectByProjectFilePath( filePath );
```
wonder-sk, line 28: This can return invalid project info if it is not mergin project... I guess we want to return early - and probably also clear project variables if there were any left from the previous project?

app/variablesmanager.h
```diff
@@ -0,0 +1,24 @@
+#ifndef VARIABLESMANAGER_H
+#define VARIABLESMANAGER_H
+
+#include <QObject>
+#include "merginapi.h"
+#include "qgsproject.h"
+
+class MerginApi;
+
+class VariablesManager : public QObject
```
wonder-sk, line 10: Could you please add doxygen comments to the class about the variables that it sets?

app/variablesmanager.h
```diff
@@ -0,0 +1,24 @@
+#ifndef VARIABLESMANAGER_H
```
wonder-sk, line 1: Missing copyright header here and in cpp

app/localprojectsmanager.cpp
```diff
@@ -51,6 +51,16 @@ LocalProjectInfo LocalProjectsManager::projectFromDirectory( const QString &proj
   return LocalProjectInfo();
 }
 
+LocalProjectInfo LocalProjectsManager::projectByProjectFilePath( const QString &projectFilePath ) const
```
wonder-sk, line 54: Can we rename to projectFrom.... for consistent naming with other functions

app/variablesmanager.cpp
sklencar: if a user logs out, authChanged is called anyway. If a user wants to log in, but fails, a user had to be log out before and therefore username is empty and will be empty after it fails.

Question - Use case:  logged in user who suddenly lost connection. Then a user wants to reload list of project and but fails to authorised (authFailed signal) - do we want to keep username or not?

PeterPetrik:
ios build failure expected, the certificate is changed on dev branch and this branch is not rebased.

saberraz:
@sklencar could you list the variables here so that I can add them to the documentation?

PR #825, opened 2020-07-30
Attr table search logic #795

### LINKED ISSUE #795: Search bar logic
Two possible approaches (TBD):
 - Locator from QGIS
 - Custom search function

So far only search in attributes visible for user

tomasMizera:
Closes #795 

- Search bar logic for searching within features.
- So far we allow search in text and number attributes. 
- name of found attribute together with its value is shown next to description text


tomasMizera:
Rebased on current dev

PR #827, opened 2020-08-03
Fix of multiplying connections

### LINKED ISSUE #824: Mergin API call for billing called multiple times
on sign in there is multiple call of `/v1/plan` endpoint

sklencar:
Input Purchasing backend is created at the very beginning. However it is cleaned and recreated once again after merginStatus has been changed - also called at the beginning while pinging mergin and getting its status. Each call of `createBackend` resets `mBackend` pointer and creates also new connections to already existing mMerginApi connections and new connections to reseted mBackend pointer. The call of the function caused multiplying of connections and therefore calling some (even server) functions multiple times.

This is simple fix of disconnecting existing connection on clean.  
closes #824

### DIFF
--- a/app/purchasing.cpp
+++ b/app/purchasing.cpp
@@ -398,6 +398,22 @@ void Purchasing::onFetchPurchasingPlansFinished()
 
 void Purchasing::clean()
 {
+  if ( mBackend )
+  {
+    disconnect( mBackend.get(), &PurchasingBackend::planRegistrationSucceeded, this, &Purchasing::onPlanRegistrationSucceeded );
+    disconnect( mBackend.get(), &PurchasingBackend::planRegistrationFailed, this, &Purchasing::onPlanRegistrationFailed );
+
+    disconnect( mBackend.get(), &PurchasingBackend::transactionCreationSucceeded, this, &Purchasing::onTransactionCreationSucceeded );
+    disconnect( mBackend.get(), &PurchasingBackend::transactionCreationFailed, this, &Purchasing::onTransactionCreationFailed );
+  }
+
+  if ( mMerginApi )
+  {
+    disconnect( mMerginApi, &MerginApi::apiRootChanged, this, &Purchasing::onMerginServerChanged );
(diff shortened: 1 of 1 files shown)


app/purchasing.cpp
```diff
@@ -398,6 +398,22 @@ void Purchasing::onFetchPurchasingPlansFinished()
...
+    disconnect( mBackend.get(), &PurchasingBackend::transactionCreationSucceeded, this, &Purchasing::onTransactionCreationSucceeded );
+    disconnect( mBackend.get(), &PurchasingBackend::transactionCreationFailed, this, &Purchasing::onTransactionCreationFailed );
+  }
+
+  if ( mMerginApi )
+  {
+    disconnect( mMerginApi, &MerginApi::apiRootChanged, this, &Purchasing::onMerginServerChanged );
+    disconnect( mMerginApi, &MerginApi::apiSupportsSubscriptionsChanged, this, &Purchasing::onMerginServerChanged );
+    disconnect( mMerginApi, &MerginApi::apiVersionStatusChanged, this, &Purchasing::onMerginServerStatusChanged );
+  }
```
wonder-sk, line 415: rather than connecting to MerginApi in createBackend() and disconnecting later, it is better to connect in the constructor once and then never disconnect.

app/purchasing.cpp
```diff
@@ -398,6 +398,22 @@ void Purchasing::onFetchPurchasingPlansFinished()
...
 void Purchasing::clean()
 {
+  if ( mBackend )
+  {
+    disconnect( mBackend.get(), &PurchasingBackend::planRegistrationSucceeded, this, &Purchasing::onPlanRegistrationSucceeded );
+    disconnect( mBackend.get(), &PurchasingBackend::planRegistrationFailed, this, &Purchasing::onPlanRegistrationFailed );
+
+    disconnect( mBackend.get(), &PurchasingBackend::transactionCreationSucceeded, this, &Purchasing::onTransactionCreationSucceeded );
+    disconnect( mBackend.get(), &PurchasingBackend::transactionCreationFailed, this, &Purchasing::onTransactionCreationFailed );
+  }
```
wonder-sk, line 408: If we delete mBackend, all connections would be removed, so we should not need to disconnect them explicitly

PR #836, opened 2020-08-21
Fix/map theme from project

### LINKED ISSUE #789: Inconsistent layer visibility between QGIS and Input when there is a map theme
See:
saber/map_theme_bug

Image from QGIS:
image

Image from Input:
Screenshot_20200713-194711_Input

sklencar:
Changed behaviour of selecting active map theme. Now it is changed according project. If a current project theme is not found either in project's collection or ThemeModel, active index is set to -1.

Updating active index has to be done after any map theme reloading - even if a model hasn't really changed (use case - having two same projects with very same themes. One project differs with the other with selected theme, so active index has to be changed after project loading even if themes in model haven't changed).

closes #789

### DIFF
--- a/app/appsettings.cpp
+++ b/app/appsettings.cpp
@@ -18,7 +18,6 @@ AppSettings::AppSettings( QObject *parent ): QObject( parent )
   settings.beginGroup( mGroupName );
   QString path = settings.value( "defaultProject", "" ).toString();
   QString layer = settings.value( "defaultLayer/"  + path, "" ).toString();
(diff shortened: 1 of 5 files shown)


tomasMizera:
What if we merged this into `dev_0.7` to avoid future conflicts when merging dev and master?

saberraz:
Agree. It is not a critical bug and we can have it on 0.7 branch for the time being.

PeterPetrik:
@sklencar can you please rebase for dev_0.7 branch?

PR #844, opened 2020-08-31
Apple fixes2

### LINKED ISSUE #802: Create test for purchasing
Add Tests for purchasing based on TestingPurchasingBackend

### LINKED ISSUE #806: Transaction pending when ios dialog cancelled
If you try to buy subscription and you cancel the dialog with “sigh in to the app store” application stays in “Transaction pending…” state

PeterPetrik:
fixes #806 
fixed #802

### DIFF
--- a/app/merginapi.cpp
+++ b/app/merginapi.cpp
@@ -752,6 +752,7 @@ void MerginApi::authorizeFinished()
       mUserAuth->clearTokenData();
       emit authFailed();
       InputUtils::log( "auth", QStringLiteral( "FAILED - invalid JSON response" ) );
+      qDebug() << data;
       emit notify( "Internal server error during authorization" );
     }
   }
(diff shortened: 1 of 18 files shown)


app/merginapi.cpp
```diff
@@ -739,21 +737,22 @@ void MerginApi::authorizeFinished()
...
     else
     {
-      mUserAuth->setUsername( QString() );
-      mUserAuth->setPassword( QString() );
+      whileBlocking( mUserAuth )->setUsername( QString() ); //clearTokenData emits the authChanged
+      whileBlocking( mUserAuth )->setPassword( QString() ); //clearTokenData emits the authChanged
       mUserAuth->clearTokenData();
       emit authFailed();
       InputUtils::log( "auth", QStringLiteral( "FAILED - invalid JSON response" ) );
+      qDebug() << data;
```
sklencar, line 755: do we want qDebug?

PeterPetrik: i will fix in the following PR

PeterPetrik: i think it is useful, since i spend significant time finding out that response from server is 301 - moved pernamently and trying to debug it all to see what going on is difficutt

PR #852, opened 2020-09-10
Attr table m3 value relation

tomasMizera:
ValueRelation widget change in feature form. Now we can choose a relation from list view with option to search in features instead of combobox.

If referenced layer has 5 and more features, feature form opens this new page, otherwise (for four and less) it opens combobox. 
We also got rid of our custom `features model` and moved it to qgsquick.

Showcase:

This change depends on https://github.com/qgis/QGIS/pull/38665


tomasMizera:
rebased

tomasMizera:
Page title now set by field name

PR #870, opened 2020-09-15
Added mergin username and local version to a conflict filename

### LINKED ISSUE #787: Better handle failures of diff-based pull
When pull of a diffable file fails for some reason (e.g. if the schema has changed), a conflict copy of the local file is created (e.g. for `x.gpkg` Input creates `x.gpkg_conflict`). It is however not possible to figure out what was the original version of the file. It would be good to choose a filename that would contain more information including base version and possibly also username:  e.g. `x.gpkg_conflict_martin_v213`

sklencar:
Note: Do not merge without passing tests!
closes #787

### DIFF
--- a/app/test/testmerginapi.cpp
+++ b/app/test/testmerginapi.cpp
@@ -1329,7 +1329,6 @@ void TestMerginApi::writeFileContent( const QString &filename, const QByteArray 
 QByteArray TestMerginApi::readFileContent( const QString &filename )
 {
   QFile f( filename );
-  qDebug() << "!!!Opening file " << filename;
   Q_ASSERT( f.exists() );
   Q_ASSERT( f.open( QIODevice::ReadOnly ) );
   QByteArray data = f.readAll();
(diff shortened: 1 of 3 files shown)


app/test/testmerginapi.cpp
```diff
@@ -805,9 +805,11 @@ void TestMerginApi::testConflictRemoteUpdateLocalUpdate()
   uploadRemoteProject( mApi, mUsername, projectName );
 
   // verify the result: the server version should be in test1.txt
-  // and the local version should go to test1.txt_conflict
+  // and the local version should go to test1.txt_conflict_<username>_<version>
+  LocalProjectInfo info = mApi->localProjectsManager().projectFromDirectory( projectDir );
+  QString conflictFilename = filename + "_conflict_" + mUsername + "_v" + info.localVersion;
```
wonder-sk, line 810: Better to use fixed expected version in the tests rather than querying it from the local project info: e.g.
`... + mUsername + "_v123";`

app/test/testmerginapi.cpp
```diff
@@ -1326,6 +1329,8 @@ void TestMerginApi::writeFileContent( const QString &filename, const QByteArray
 QByteArray TestMerginApi::readFileContent( const QString &filename )
 {
   QFile f( filename );
+  qDebug() << "!!!Opening file " << filename;
```
wonder-sk, line 1332: do we want to keep this?

wonder-sk (approved):

PR #871, opened 2020-09-16
Register form adjusted for smaller screens or virtual keyboard

### LINKED ISSUE #863: Not possible to scroll in registration form when keyboard is opened
When one wants to register -> clicks on Username field -> keyboard is opened what is correct, however username field goes out of current screen extent and it is not possible to scroll up. So user won't see the username he is typing.

sklencar:
Added scroll view and simplified calculation of height/margins of the register form. The height of Qt.InputMethod and keyboard rectangle was used using implicit changed signal to change form's height/margins. However, after init of the virtual keyboard, its size wasnt being changed anyway. More promising way is to use visible property.
 
Fixed resizing of the login form with same pattern - the whole form was lifted when keyboard was shown, but it wasnt put back afterwards.
closes #863

### DIFF
--- a/app/qml/AuthPanel.qml
+++ b/app/qml/AuthPanel.qml
@@ -93,16 +93,8 @@ Item {
     LoginForm {
       id: loginForm
       visible: !warningMsgContainer.visible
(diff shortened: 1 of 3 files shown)


PeterPetrik:
any change of attaching screenshot?

wonder-sk (approved):

PR #873, opened 2020-09-16
Fix reprojection when datum transformations are set in project

wonder-sk:
Input was ignoring custom datum transformations set in QGIS projects which may have been causing subtle shifts in on-the-fly reprojection when the default coordinate operation was different from the preferred one.

### DIFF
--- a/app/loader.cpp
+++ b/app/loader.cpp
@@ -156,6 +156,16 @@ void Loader::setMapSettingsLayers() const
   }
 
   mMapSettings->setLayers( allLayers );
+  mMapSettings->setTransformContext( mProject->transformContext() );
+}
+
+// Huh, it looks like we have forgotten to add implementation of QgsQuickMapSettings::setTransformContext()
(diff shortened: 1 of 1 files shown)


PeterPetrik:
I will need to do qgis-sdk update anyways, can you create QGIS PR, i would not put this workaround here...

PeterPetrik:
@wonder-sk are you sure this is required, the context is automatically set when you call QgsQuickMapSettings::setProject, see https://github.com/qgis/QGIS/blob/e7e08fd5045b4fbbec5078ed30de644dabd5ce41/src/quickgui/qgsquickmapsettings.cpp#L55

wonder-sk:
Yes, it is needed and it fixes a bug that was reported to me. Probably the QgsQuickMapSettings::setProject() is called on app startup with QgsProject::instance() when project is still empty, and then never called again (because we use the singleton)

PeterPetrik:
QGIS PR: https://github.com/qgis/QGIS/pull/39010

PR #874, opened 2020-09-17
Translation fixes

mostlyAtNight:


app/qml/ActiveLayerPanel.qml
```diff
@@ -14,7 +14,6 @@ import QgsQuick 0.1 as QgsQuick
 import "."  // import InputStyle singleton
 
 Drawer {
-    property string title: qsTr("Survey Layer")
```
mostlyAtNight, line 17: @PeterPetrik - you mentioned to remove this - is this what you meant?

app/qml/main.qml
```diff
@@ -28,7 +28,7 @@ ApplicationWindow {
     width:  __appwindowwidth
     height: __appwindowheight
     visibility: __appwindowvisibility
-    title: qsTr("Input")
```
mostlyAtNight, line 31: @PeterPetrik - is this a valid way of making untraslatable?

PeterPetrik: I would just remove qsTr and make a comment "/* do not translate, not visible in app */

app/localprojectsmanager.cpp
```diff
@@ -196,12 +196,12 @@ QString LocalProjectsManager::findQgisProjectFile( const QString &projectDir, QS
   else if ( foundProjectFiles.count() > 1 )
   {
     // error: multiple project files found
-    err = tr( "Multiple QGIS project files" );
+    err = tr( "Found multiple QGIS project files - there should be only one." );
```
PeterPetrik, line 199: TODO add link to docs?

app/localprojectsmanager.cpp
```diff
@@ -196,12 +196,12 @@ QString LocalProjectsManager::findQgisProjectFile( const QString &projectDir, QS
...
   }
   else
   {
     // no projects
-    err = tr( "Missing QGIS project file" );
+    err = tr( "Failed to find a QGIS project file - please add one." );
```
PeterPetrik, line 204: TODO add link to docs? or not needed?

app/merginprojectstatusmodel.cpp
```diff
@@ -93,7 +93,7 @@ void MerginProjectStatusModel::infoProjectUpdated( const ProjectDiff &projectDif
 
         ProjectStatusItem item;
         item.status = ProjectChangelogStatus::Message;
-        item.text =  tr( "Unable to determine changes" );
+        item.text =  tr( "Failed to determine changes. More details." );
```
PeterPetrik, line 96: TODO add help link

app/qml/ActiveLayerPanel.qml
PeterPetrik: yes, it is not used anywhere

PeterPetrik (commented):
best to have a list of help pages we need

app/qml/TextHyperlink.qml
```diff
@@ -0,0 +1,25 @@
...
+
+  text: "<style>a:link { color: " + InputStyle.highlightColor
+        + "; text-decoration: underline; }</style>" + richText
+  color: InputStyle.fontColor
+  onLinkActivated: Qt.openUrlExternally(link)
+  textFormat: Text.RichText
+  wrapMode: Text.WordWrap
+  font.pixelSize: InputStyle.fontPixelSizeNormal
+  font.bold: true
+}
```
wonder-sk, line 25: I would suggest to have `Item` with a `text` property and `Text` item inside... otherwise it's easy for caller to use `text` property by default instead of `richText`

app/qml/MerginProjectPanel.qml
```diff
@@ -601,7 +601,7 @@ Item {
         MainPanelButton {
           id: exploreBtn
           width: toolbar.itemSize
-          text: qsTr("Explore")
+          text: qsTr("Public")
```
wonder-sk, line 604: This will be confusing is we rename as we are using the term `Explore` in mergin web and plugin

PeterPetrik (approved):

PR #878, opened 2020-09-17
Forwarding logs from geodiff

### LINKED ISSUE #688: Report problems from geodiff
Currently any errors/warnings reported from geodiff are lost - Input should use the geodiff logger API calls to redirect output to the Input log, so it can be investigated more easily why a geodiff action has failed.

sklencar:
Currently set logging level to `info`. 
Just an idea: it might be useful to set log level according type of build (production, dev, debug,..)
closes #688

### DIFF
--- a/app/geodiffutils.cpp
+++ b/app/geodiffutils.cpp
@@ -109,3 +109,17 @@ bool GeodiffUtils::applyDiffs( const QString &src, const QStringList &diffFiles 
   }
   return true;
 }
+
+void GeodiffUtils::log( GEODIFF_LoggerLevel level, const char *msg )
+{
+  QString prefix;
+  switch ( level )
+  {
+    case LevelError: prefix = "GEODIFF error: "; break;
+    case LevelWarning: prefix = "GEODIFF warning: "; break;
+    case LevelInfo: prefix = "GEODIFF info: "; break;
+    case LevelDebug: prefix = "GEODIFF debug: "; break;
+    default: break;
+  }
+  InputUtils::log( prefix, msg );
+}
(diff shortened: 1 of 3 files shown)


app/geodiffutils.cpp
```diff
@@ -109,3 +109,17 @@ bool GeodiffUtils::applyDiffs( const QString &src, const QStringList &diffFiles
...
+
+void GeodiffUtils::log( GEODIFF_LoggerLevel level, const char *msg )
+{
+  QString prefix;
+  switch ( level )
+  {
+    case LevelError: prefix = "GEODIFF error: "; break;
+    case LevelWarning: prefix = "GEODIFF warning: "; break;
+    case LevelInfo: prefix = "GEODIFF info: "; break;
+    case LevelDebug: prefix = "GEODIFF debug: "; break;
```
wonder-sk, line 121: Other prefixes used in InputUtils::log() do not end up with colon+space, so it would be good to remove them here for consistency

app/merginapi.cpp
```diff
@@ -53,6 +53,8 @@ MerginApi::MerginApi( LocalProjectsManager &localProjects, QObject *parent )
 
   loadAuthData();
   GEODIFF_init();
+  GEODIFF_setLoggerCallback( &GeodiffUtils::log );
+  GEODIFF_setMaximumLoggerLevel( GEODIFF_LoggerLevel::LevelInfo );
```
wonder-sk, line 57: Just checked, python client uses LevelDebug all the time, let's use it here as well - better more logging than not enough :-)

wonder-sk (approved):

PR #915, opened 2020-10-08
Fix unfinished download files 

### LINKED ISSUE #884: Unfinished project download may leave incomplete directory
Download of a project may get interrupted, even without getting a signal about the failure of a request.

This will leave a half-populated directory and may cause further trouble later.

We should have a mechanism that will ensure that half-downloaded projects are never shown in the list - and any half-downloaded project data get pruned. For example, we could download to a dir with a special prefix, and directories with such prefix would be never shown (+would get deleted when Input starts)

TODO: find out when exactly this happens (app gets to background? app is forcefully quit?)

tomasMizera:
What has changed:

- Added temp file to `.mergin` folder to indicate (first time) download in progress
- File is removed once the project is successfully downloaded, but is kept within the project if application was closed during download ~ unsuccessful download
- If this file is found during startup check, given project is considered invalid
- Also we now clear the temp folder of given project when it should be downloaded for the first time in case there was previous unsuccessful attempt to download and some chunks were left here

Closes #884

### DIFF
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -268,6 +268,16 @@ QString InputUtils::logFilename()
   return sLogFile;
 }
 
+bool InputUtils::createFile( const QString &filePath )
+{
+  QFile newFile( filePath );
+  if ( !newFile.open( QIODevice::WriteOnly ) )
+    return false;
+
+  newFile.close();
+  return true;
+}
+
 QString InputUtils::filesToString( QList<MerginFile> files )
 {
   QStringList resultList;
@@ -446,6 +456,12 @@ QString InputUtils::renameWithDateTime( const QString &srcPath, const QDateTime 
   }
 
   return QString();
+}
+
+QString InputUtils::downloadInProgressFilePath( const QString &projectDir )
+{
+  // Use first 11 characters of sha-256 of text "Input" to create unique file name
+  return projectDir + "/.mergin/.project-36ecb4f8669.downloading";
 }
 
 void InputUtils::showNotification( const QString &message )
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -268,7 +268,7 @@ QString InputUtils::logFilename()
   return sLogFile;
 }
 
-bool InputUtils::createFile( const QString &filePath )
+bool InputUtils::createEmptyFile( const QString &filePath )
 {
   QFile newFile( filePath );
   if ( !newFile.open( QIODevice::WriteOnly ) )
@@ -460,8 +460,7 @@ QString InputUtils::renameWithDateTime( const QString &srcPath, const QDateTime 
 
 QString InputUtils::downloadInProgressFilePath( const QString &projectDir )
 {
-  // Use first 11 characters of sha-256 of text "Input" to create unique file name
-  return projectDir + "/.mergin/.project-36ecb4f8669.downloading";
+  return projectDir + "/.mergin/.project.downloading";
 }
 
 void InputUtils::showNotification( const QString &message )
--- a/app/localprojectsmanager.cpp
+++ b/app/localprojectsmanager.cpp
@@ -11,6 +11,7 @@
 
 #include "merginapi.h"
 #include "merginprojectmetadata.h"
+#include "inpututils.h"
 
 #include <QDir>
 #include <QDirIterator>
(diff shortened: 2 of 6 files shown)


tomasMizera:
@mostlyAtNight this PR adds new string in app when project failed to download, it is:
`tr( "Download failed, remove and download project again" );`
Is it ok?

app/inpututils.cpp
```diff
@@ -448,6 +458,12 @@ QString InputUtils::renameWithDateTime( const QString &srcPath, const QDateTime
   return QString();
 }
 
+QString InputUtils::downloadInProgressFilePath( const QString &projectDir )
+{
+  // Use first 11 characters of sha-256 of text "Input" to create unique file name
+  return projectDir + "/.mergin/.project-36ecb4f8669.downloading";
```
PeterPetrik, line 464: not sure if I understand why the hash should be in filename at all? why not something like `.project.downloading`

app/localprojectsmanager.cpp
```diff
@@ -184,12 +185,21 @@ QString LocalProjectsManager::findQgisProjectFile( const QString &projectDir, QS
...
     foundProjectFiles << it.filePath();
   }
-  if ( foundProjectFiles.count() == 1 )
+
+  if ( QFile::exists( InputUtils::downloadInProgressFilePath( projectDir ) ) )
+  {
+    // if this is a mergin project and file indicating download in progress is still there
+    // download failed or copying from .temp to project dir failed (app was probably closed meanwhile)
+
+    err = tr( "Download failed, remove and download project again" );
```
PeterPetrik, line 200: I suggest "Corrupted project directory"

app/inpututils.cpp
tomasMizera: Well, I just wanted the file to be unique as much as possible :grin: , so that geopackages that are in this folder would not have the same name.

app/localprojectsmanager.cpp
tomasMizera: Good idea, I will talk to @mostlyAtNight

tomasMizera:
rebased

app/merginapi.cpp
```diff
@@ -1590,10 +1603,19 @@ void MerginApi::startProjectUpdate( const QString &projectFullName, const QByteA
...
+
     // project has not been downloaded yet - we need to create a directory for it
     transaction.projectDir = createUniqueProjectDirectory( projectName );
     transaction.firstTimeDownload = true;
 
+    // create file indicating first time download in progress
+    QString downloadInProgressFilePath = InputUtils::downloadInProgressFilePath( transaction.projectDir );
+    createPathIfNotExists( downloadInProgressFilePath );
+    InputUtils::createFile( downloadInProgressFilePath );
+    Q_ASSERT( QFile::exists( downloadInProgressFilePath ) );
```
wonder-sk, line 1617: I would rather log such thing instead of having assert (which will do nothing in release build and therefore we won't get any hint something went wrong)

app/merginapi.cpp
```diff
@@ -1407,6 +1415,11 @@ void MerginApi::finalizeProjectUpdate( const QString &projectFullName )
   {
     QString projectNamespace, projectName;
     extractProjectName( projectFullName, projectNamespace, projectName );
+
+    // remove download in progress file
+    if ( !QFile::remove( InputUtils::downloadInProgressFilePath( transaction.projectDir ) ) )
+      InputUtils::log( "sync " + projectFullName, QStringLiteral( "Failed to remove download in progress file for project name %1" ).arg( projectName ) );
```
wonder-sk, line 1421: I would prefer to leave the context string just as `"sync"` like elsewhere - and put the full project name to the arg()

app/inpututils.cpp
```diff
@@ -268,6 +268,16 @@ QString InputUtils::logFilename()
   return sLogFile;
 }
 
+bool InputUtils::createFile( const QString &filePath )
```
wonder-sk, line 271: Maybe rename to createEmptyFile() so it's more clear

app/inpututils.cpp
wonder-sk: as Peter suggests I would also go with a simpler naming :-)

wonder-sk (approved):

PR #917, opened 2020-10-09
REBASED: Wrapped all MerginProjectPanel related pages into stack view

### LINKED ISSUE #906: ios subscription form opens keyboard
Screenshot 2020-09-30 at 18 35 03

sklencar:
There were several issues with current implementation of pages - components from pages were sometimes overlapping to each other. This fixes uses StackView to simplifying paging and key (press) handling.

Generally it might be better to refactor all pages similarly and use single stackView component for that.

TOFIX: Currently "Subscription plans" always open browser instead of showing Subscription page.

NOTE: such change needs to be properly tested. It has been tested on Android and iOS, but it could happen that some use cases were not covered.

closes #906

### DIFF
--- a/app/qml/MerginProjectPanel.qml
+++ b/app/qml/MerginProjectPanel.qml
@@ -180,6 +180,7 @@ Item {
           projectsPanel.forceActiveFocus()
         }
         onRegistrationFailed: stackView.pending = false
+        onRegistrationSucceeded: stackView.pending = false
       }
 
 
--- a/app/qml/MerginProjectPanel.qml
+++ b/app/qml/MerginProjectPanel.qml
@@ -27,730 +27,768 @@ Item {
   property bool showMergin: false
   property real panelMargin: InputStyle.panelMargin
 
+
   function openPanel() {
-    homeBtn.activated()
     projectsPanel.visible = true
-  }
-
-  function getStatusIcon(status, pending) {
-    if (pending) return "stop.svg"
-
-    if (status === "noVersion") return "download.svg"
-    else if (status === "outOfDate") return "sync.svg"
-    else if (status === "upToDate") return "check.svg"
-    else if (status === "modified") return "sync.svg"
-
-    return "more_menu.svg"
-  }
-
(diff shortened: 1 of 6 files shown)


tomasMizera:
Great! I am very happy to see this :star:

PeterPetrik:
thanks for this efford, I imagine it is better to release 0.7.0 without it, after proper review&test

sklencar:
Definitely agree to keep it for the later release.

app/qml/MerginProjectPanel.qml
```diff
@@ -27,730 +27,769 @@ Item {
...
+        }
+      }
+
+      Connections {
+        target: __merginApi
+        onListProjectsFinished: {
+          stackView.pending = false
+        }
+        onListProjectsFailed: {
+          reloadList.visible = true
```
tomasMizera, line 148: Do we still need to set visibility of pages even with stackview?

tomasMizera (commented):
The changes are looking good! I found some issues when I ran it on Android:

- back button on Android closes app when app has no active project and sign in page is opened
- toolbar is not visible in sign in or sign up page (not sure if this is intended)
- back button on sign up page should probably lead to sign in page? 

Other than that, app was running smoothly and I think this is the way to go for future refactoring.

wonder-sk:
@sklencar could you please resolve the conflicts?

PeterPetrik:
@sklencar can you please rebase so we can merge?

app/qml/MerginProjectPanel.qml
sklencar: Its a button to manually refresh project list if it failed before.

PeterPetrik:
@sklencar can we merge? in the PR description there is TODO, is is resolved? Also are Tomas review items fixed now?

saberraz:
> TOFIX: Currently "Subscription plans" always open browser instead of showing Subscription page.

I checked with the latest build and it is not the case in iOS anymore.

Issue #916, opened 2020-10-08
Android back button closes app in account view


tomasMizera:
To reproduce:
- do not have any active project and be logged in
- start Input
- click on profile icon
- click android back button

Expected:
- app is back to projects list

Experienced:
- app is closed

Regression bug.

tomasMizera:
Should be addressed in #917. Stopped fixing, if problem remains, will be fixed after 917 is merged.

tomasMizera:
While working on this bug I found 2 other issues that are addressed in #918

PR #918, opened 2020-10-09
Update: Android back button #916

tomasMizera:
While working on #916 I found 2 other issues that this PR addresses, original bug should be fixed by refactoring in #917. 

**One with log panel**: Hitting back on android in log panel was not consistent with back button in header. (android back leaded immediately to map view while in-app back led to settings panel)

**Second issue**: Wrong component received focus when application started without active project.

These changes should not conflict with #917


PeterPetrik (approved):

tomasMizera:
Since #917 is marked as `Frozen` for this release, I added fix of this bug here, co this closes #916

PR #942, opened 2020-10-19
Fixed RegisterForm after successful registration

### LINKED ISSUE #934: Getting stuck in the Sign up page
Also fix the toast message!

Screenshot_20201015-161355_Input

sklencar:
closes #934

### DIFF
--- a/app/merginapi.cpp
+++ b/app/merginapi.cpp
@@ -808,7 +808,7 @@ void MerginApi::registrationFinished()
   {
     InputUtils::log( "register", QStringLiteral( "Success" ) );
     emit registrationSucceeded();
-    QString msg = tr( "Registration successful.\n You should now be able to sign in.");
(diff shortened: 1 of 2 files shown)


app/merginapi.cpp
```diff
@@ -826,7 +826,7 @@ void MerginApi::registrationFinished()
     {
       // the self-registration is not allowed on the server
       emit registrationFailed();
-      emit notify( tr( "New registrations are not allowed on the selected Mergin server.%1Please check with your administrator." ).arg( "<br/>" ) );
+      emit notify( tr( "New registrations are not allowed on the selected Mergin server.\n Please check with your administrator." ).arg( "<br/>" ) );
```
wonder-sk, line 829: These two changes will break the translations... (and the second one has forgotten `.arg(...)`)
How about we just change the arg(...) content to have `\n` inside?

wonder-sk (approved):

Issue #938, opened 2020-10-19
Restore purchase option should be iOS only


wonder-sk:
in "my account" page

and it should be only shown if you already have an active subsciption

wonder-sk:
see also #937

PR #943, opened 2020-10-19
Restoring a purchase can be done without active subscription

sklencar:
related to #938

### DIFF
--- a/app/qml/AccountPage.qml
+++ b/app/qml/AccountPage.qml
@@ -219,7 +219,7 @@ Page {
 
     Text {
       id: textRestore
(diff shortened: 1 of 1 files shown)


wonder-sk:
Hmm from what I understand from @PeterPetrik 's comments, this should be enabled when user _does not_ have a subscription...? If I read the code correctly, it will be displayed now also if the user already has a subscription?

sklencar:
Pasting a discussion here: 
`It should be iOS only, but also when you do not have active subscription. It is primary used when for example you buy apple subscription, but MERGIN server rejects your receipt so you can re-send it to the server to verify/accept it again`

Should we also consider a case when a user has subscription, but tried to upgrade it and it fails? Or rather should any upgrade be done through web?

PeterPetrik:
it should be enabled all the time on iOS

PeterPetrik (approved):

PR #983, opened 2020-10-28
Password field improvements

### LINKED ISSUE #543: An option to view password while typing
when logging into Mergin from Input:
image

### LINKED ISSUE #931: Reset password visibility when leaving the sign in page
When I enable password visibility (using the padlock) on the Sing in page, it remains visible until I tap on it again, even if I leave the app or change the page.

For security reason, it is better to set it back to invisible as soon as user leaves the page.

sklencar:
Screenshot 2020-10-28 at 17 53 38

closes #931 
closes #543


wonder-sk (commented):
Sorry the tickets were probably not clear:
- the lock icon should stay on the left
- the eye icon should be on the right - like it is common in various web services

Feel free to extract the password widget code to a new QML file if it is copy-paste code in multiple places...

app/qml/RegistrationForm.qml
```diff
@@ -161,138 +162,39 @@ Rectangle {
...
-      opacity: password.focus ? 1 : 0.6
+      y: fieldHeight - height
+      opacity: passwordField.password.focus ? 1 : 0.6
       width: registerForm.width - fieldHeight / 2
       anchors.horizontalCenter: parent.horizontalCenter
     }
 
-    Row {
+    PasswordField {
+      id: passwordConfirmField
```
tomasMizera, line 184: We should probably change placeholder text to "Repeat password" or something similar so that there are not two times fields with text "Password"

app/qml/components/PasswordField.qml
```diff
@@ -0,0 +1,99 @@
...
+
+Row {
+  id: root
+  spacing: 0
+
+  property var fontColor
+  property var bgColor
+  property var password: password
+
+  //property function clear: function() {password.text = ""}
```
tomasMizera, line 18: Should this be commented out?

app/qml/components/PasswordField.qml
```diff
@@ -0,0 +1,99 @@
...
+    id: password
+    width: loginForm.width - iconContainer.width - visibilityIconContainer.width
+    height: fieldHeight
+    font.pixelSize: InputStyle.fontPixelSizeNormal
+    color: root.fontColor
+    placeholderText: qsTr("Password")
+    echoMode: TextInput.Password
+    inputMethodHints: Qt.ImhNoPredictiveText | Qt.ImhNoAutoUppercase
+    font.capitalization: Font.MixedCase
+    //text: root.text
```
tomasMizera, line 55: Lets uncomment or remove :)

tomasMizera (approved):
Looking good, tested also on Android. Good job

app/qml/RegistrationForm.qml
tomasMizera: It was "Confirm password" before

PeterPetrik:
@sklencar can you please adress review comments so we can merge?

app/qml/RegistrationForm.qml
```diff
@@ -161,138 +162,40 @@ Rectangle {
...
-        inputMethodHints: Qt.ImhNoPredictiveText | Qt.ImhNoAutoUppercase
-        font.capitalization: Font.MixedCase
-
-        background: Rectangle {
-          color: root.bgColor
-        }
-      }
+      fontColor: root.fontColor
+      bgColor: root.bgColor
+      placeholder: 'Confirm password'
```
wonder-sk, line 189: qsTr ?

app/qml/components/PasswordField.qml
```diff
@@ -0,0 +1,95 @@
...
+import "../" // import InputStyle singleton
+
+Row {
+  id: root
+  spacing: 0
+
+  property var fontColor
+  property var bgColor
+  property var password: password
+  property string placeholder: 'Password'
```
wonder-sk, line 15: qsTr ?

app/qml/components/PasswordField.qml
```diff
@@ -0,0 +1,95 @@
...
+    }
+  }
+
+  TextField {
+    id: password
+    width: loginForm.width - iconContainer.width - visibilityIconContainer.width
+    height: fieldHeight
+    font.pixelSize: InputStyle.fontPixelSizeNormal
+    color: root.fontColor
+    placeholderText: qsTr(root.placeholder)
```
wonder-sk, line 48: Here I think qsTr does not work - it needs a constant string otherwise the tool that extracts translatable strings can't extract it

app/qml/components/PasswordField.qml
wonder-sk, line 15: can we rename the property to `placeholderText` to make it consistent with other items like TextField from Qt

PR #988, opened 2020-10-29
Add guide lines and polygons when recording new features

### LINKED ISSUE #309: Digitizing rubber band issues
Line/polygon recording:
- when the first point is added, there is no visual feedback that something has happened. we should temporarily put a point marker there (can be removed as soon as the user adds second point)
- when moving crosshair, there should be a line segment going to the last point (maybe a dashed line) so user knows what is going on

tomasMizera:
Enriches UI/UX experience while recording new linestrings and polygons by adding leading line / polygon as can be seen below:

<table>
  <tr>
    <th>Capturing Linestring</th>
    <th>Capturing Polygon</th>
  <tr>
    <td></td>
    <td></td>
  </tr>
 </table>

Closes #309

### DIFF
--- a/app/qml/Highlight.qml
+++ b/app/qml/Highlight.qml
@@ -15,6 +15,8 @@ import QgsQuick 0.1 as QgsQuick
 Item {
   id: highlight
 
+  signal positionChanged()
+
   // color for line geometries
   property color lineColor: "black"
   // width for line geometries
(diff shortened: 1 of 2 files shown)


saberraz:
I presume the helper disappears when in "streaming" mode?

saberraz:
I tried the APK:
1- The guide lines are very helpful and definitely addresses the issue
2- Switching to "Streaming" mode (pressing and holding GPS button) crashed my session

PeterPetrik (approved):

PeterPetrik:
@saberraz is crash fixed? if yes, we can merge

PR #1004, opened 2020-11-16
qgsquick in input

PeterPetrik:


PeterPetrik:
@varmar05 not sure if test server does not need restart?

@tomasMizera input crashes, but the qgsquick code is with all recent changes, but inputapp without, could it be the crash reason?

tomasMizera:
Hi @PeterPetrik I would need to see the log. However, it should not but to be sure, I will build master versions of Input and QGIS to see.

wonder-sk (approved):
:tada:

Issue #1021, opened 2020-11-20
Not possible to add a feature to non-spatial form


saberraz:
To replicate:
- Open lutraconsulting\test_forms
- Browse data > reftable
- Add feature

And the form is not editable.

saberraz:
I guess it is the same issue: but existing features within the non-spatial form cannot be edited.

PR #1027, opened 2020-11-23
change condition for widget loading

tomasMizera:
Reason: Sometimes config do not provide field type (passes empty string). In these scenarios we were showing textfield by default. However, in recent changes we have introduced check for undefined widget type (due to another bug), but empty string was resolved as false in condition too and field would not get generated.


PeterPetrik:
@tomasMizera is this fixing some bug?

saberraz:
Fixes https://github.com/lutraconsulting/input/issues/1021

I can confirm this now has fixed the issue.

PR #1048, opened 2020-12-03
Paging project list

### LINKED ISSUE #1043: Use paginated endpoint to list projects
Let's use the newly available API endpoint that limits number of returned projects (the existing endpoint will be removed).

The new request should look like this:

`https://public.cloudmergin.com/v1/project/paginated?page=1&per_page=50`

The request returns also the total count, so we know if we may need to get more pages or not.

To make things simple, I would suggest that if count is more than the number of items we already have, at the bottom of the list there would be a button "Load more projects" that would fetch another 50 projects and append them to the current list.

sklencar:
Listing all projects has been replaced with listing with paging functionality.

Note: `NonProjectItem` project status has been introduced to easily distinguish real project items and fetch button item in a model (and view). This approach of introducing an extra item in the list has been agreed before the implementation.

UPDATED:
 - fixed sorting by name 
 - fixed jumping at the beginning after fetching new project items
 - partially fixed searching

Known issues:
- Inconsistent project list when a content of a page has been changed on server in a meanwhile of browsing - user already fetched a batch of projects, a project has been deleted on Mergin and therefore another project is lost since order of project has been changed. Reloading the whole page resolves the issue. (https://github.com/lutraconsulting/input/issues/1050)
- Search functionality for explore has been updated to be able search project by name (for time-being, see https://github.com/lutraconsulting/input/issues/1080). Currently filtering of `my` and `shared` projects are done on client side by the project model. Therefore search functionality is performed only on those projects which have been already fetched and stored in the model. 
Screenshot 2020-12-03 at 12 23 17

closes #1043

### DIFF
--- a/app/localprojectsmanager.h
+++ b/app/localprojectsmanager.h
@@ -18,7 +18,8 @@ enum ProjectStatus
   NoVersion,  //!< the project is not available locally
   UpToDate,   //!< both server and local copy are in sync with no extra modifications
   OutOfDate,  //!< server has newer version than what is available locally (but the project is not modified locally)
-  Modified    //!< there are some local modifications in the project that need to be pushed (note: also server may have newer version)
+  Modified,    //!< there are some local modifications in the project that need to be pushed (note: also server may have newer version)
+  Invalid
 };
 Q_ENUMS( ProjectStatus )
 
--- a/app/localprojectsmanager.h
+++ b/app/localprojectsmanager.h
@@ -19,7 +19,7 @@ enum ProjectStatus
   UpToDate,   //!< both server and local copy are in sync with no extra modifications
   OutOfDate,  //!< server has newer version than what is available locally (but the project is not modified locally)
   Modified,    //!< there are some local modifications in the project that need to be pushed (note: also server may have newer version)
-  Invalid      //!< only for mock projects, acts like a hook to enable extra functionality for models working with projects .
+  NonProjectItem      //!< only for mock projects, acts like a hook to enable extra functionality for models working with projects .
 };
 Q_ENUMS( ProjectStatus )
 
--- a/app/localprojectsmanager.h
+++ b/app/localprojectsmanager.h
@@ -19,7 +19,7 @@ enum ProjectStatus
   UpToDate,   //!< both server and local copy are in sync with no extra modifications
   OutOfDate,  //!< server has newer version than what is available locally (but the project is not modified locally)
   Modified,    //!< there are some local modifications in the project that need to be pushed (note: also server may have newer version)
-  Invalid
+  Invalid      //!< only for mock projects, acts like a hook to enable extra functionality for models working with projects .
 };
 Q_ENUMS( ProjectStatus )
 
--- a/app/merginprojectmodel.h
+++ b/app/merginprojectmodel.h
@@ -71,8 +71,16 @@ class MerginProjectModel: public QAbstractListModel
     int rowCount( const QModelIndex &parent = QModelIndex() ) const override;
 
     //! Updates list of projects with synchronization progress if a project is pending
-    //! @param isFirstPage If true clears current model, othewise merginProjects will be appended.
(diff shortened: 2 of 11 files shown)


app/qml/ProjectDelegateItem.qml
```diff
@@ -160,4 +165,15 @@ Rectangle {
             anchors.bottom: parent.bottom
         }
     }
+
+    // Additional item
+    DelegateButton {
+      visible: itemContainer.isAdditional
```
tomasMizera, line 171: Is it not possible to just stick a button on the end of the listview? Clicking on the button would fetch new entries and fill a model with them. We would then not need a new project status nor any other reserved projects name.

app/localprojectsmanager.h
```diff
@@ -18,7 +18,8 @@ enum ProjectStatus
   NoVersion,  //!< the project is not available locally
   UpToDate,   //!< both server and local copy are in sync with no extra modifications
   OutOfDate,  //!< server has newer version than what is available locally (but the project is not modified locally)
-  Modified    //!< there are some local modifications in the project that need to be pushed (note: also server may have newer version)
+  Modified,    //!< there are some local modifications in the project that need to be pushed (note: also server may have newer version)
+  Invalid      //!< only for mock projects, acts like a hook to enable extra functionality for models working with projects .
```
tomasMizera, line 22: If we would stick with new project status maybe we could rename `Invalid` to something like `ReservedName` or something like that. Invalid looks to me like one of real project states.

tomasMizera (changes requested):
Good job. I like how simple you edited the model :+1: 

While testing this PR on my device I came across some bugs:

- Search is not working in _explore_

- When more projects are fetched, application jumps visually to the first project of the list. This is a big problem on Android because scrolling is slow, especially if this is like a 10th page loaded and you need to scroll all 10 pages in order to opt for 11th (and then scroll again for 12th, ..... )

- Projects are no longer sorted alphabetically (_my_, _shared_ nor _explore_). This is not that big of a problem when user has just a few projects, it only make finding my/shared project less user friendly. On the other hand, this issue together with previous ones makes _explore_ basically unusable. Lets say I want to find project lutraconsulting/test_forms. Since projects are not sorted, I have no clue where it could be, so I am just stuck with scrolling and checking each project if it is not the one I am looking for. Search is not working and scrolling is slow.
Solution to this problem probably needs a help from Mergin, to return projects sorted alphabetically.


In the future, I think the best solution would be not using `Fetch more` button at all but pre-fetch projects when user is almost in the end of the list.

tomasMizera:
I have looked at Mergin API and I think that we do not need any change in Mergin in order to get sorted list of (public) projects. What if we simply add sorting into query, possibly with search string. Would it be too slow?

app/qml/ProjectDelegateItem.qml
sklencar: Of course, there are multiple ways how to implement such button. Some of the ideas were discussed with Martin and we agreed on introducing an extra item which seems to be the easiest and fastest solution.

app/merginprojectmodel.h
```diff
@@ -43,6 +43,8 @@ class MerginProjectModel: public QAbstractListModel
 {
     Q_OBJECT
     Q_PROPERTY( QString searchExpression READ searchExpression WRITE setSearchExpression )
+    Q_PROPERTY( int expectedProjectCount READ expectedProjectCount NOTIFY expectedProjectCountChanged )
```
tomasMizera, line 46: Looks like this property is never used. It is only in `updateModel` but we do not store it so I guess this can be removed?

app/qml/ProjectDelegateItem.qml
tomasMizera: Well, for me it seems as two different things and joining them can make some troubles in future (like checking on many places if the project in list is actually a project not some other button). But if you guys are ok with that, I am too :))

tomasMizera (commented):
Search in `Explore` works nice now, good job.

I will approve just let me know if the `expectedProjectCount` is or is not used

sklencar:
Manual tests have been modified accordingly.

tomasMizera (approved):

PR #1061, opened 2020-12-10
Allows to migrate/detach project to/from Mergin

### LINKED ISSUE #614: Allow create and upload new projects
If users have a (n orphan) project on their device not linked to a Mergin server project, we need to offer an option to upload and create a new project (if the namespace is available) on the Mergin server.

sklencar:
closes #614


app/merginapi.cpp
```diff
@@ -1101,6 +1118,37 @@ void MerginApi::pingMergin()
...
+  if ( projectInfo.isValid() )
+  {
+    QDir merginProjectDir( projectInfo.projectDir + "/.mergin" );
+    merginProjectDir.removeRecursively();
+  }
+  // Update localProjects (updating mMerginProjects can be omitted since it is updated on listing projects)
+  mLocalProjects.resetMerginInfo( projectNamespace, projectName );
+  mLocalProjects.reloadProjectDir();
+
+  emit notify( QStringLiteral( "Project detached from Mergin" ) );
```
tomasMizera, line 1148: Translate string `tr ()`

tomasMizera: Also messages from server (like `Project XYZ already exists!`) are probably not being translated but showed to user in english, maybe we could change that in future

tomasMizera (changes requested):
Looking good, good job!

Some notes (not needed to change now, may be discussed and changed in future):
 - Detach is visible and clickable from GUI, is this intended?
 - If user detaches a project, it is not possible to upload it back again - because project with such name already exists on mergin. Maybe we can in future ask for desired project name?
 - Since detaching _cannot_ be undone, what about some confirmation window to not detach project by accident?

After talk with @sklencar, detach should not be visible and clickable from GUI, so I am setting this PR as `request changes`.

sklencar:
Detached functionality is hidden in gui. It would be good to discuss what to do with orphaned projects at some point.

Good point that we probably don't have support for getting translated text from the server. Also good topic to discuss - I have just opened an issue to not forget about it.

tomasMizera (commented):
I confirm that detach is hidden. 

One UX note: after clicking on Upload to Mergin, there is no visual feedback that project is being uploaded, neither a popup that project was successfully uploaded to Mergin. @sklencar I suggest creating an issue for that / adding just a toast after successful upload for now to not block this big PR with UX

tomasMizera (approved):

app/qml/MerginProjectPanel.qml
```diff
@@ -463,6 +463,24 @@ Item {
...
             }
+
+            MenuItem {
+              height: contextMenu.isMerginProject ? 0: contextMenu.menuItemHeight
+              ExtendedMenuItem {
+                  height: parent.height
+                  rowHeight: parent.height
+                  width: parent.width
+                  contentText: contextMenu.isMerginProject ? qsTr("Detach from Mergin") : qsTr("Upload to Mergin")
+                  imageSource: contextMenu.isMerginProject ? InputStyle.detachIcon : InputStyle.uploadIcon
```
wonder-sk, line 474: Better to remove the "Detach from Mergin" text and the detach icon to avoid confusion?

wonder-sk:
If there is no feedback after one presses upload button, that's something we'll need to sort out - there should be the same kind of feedback like when doing ordinary sync of a project, including a cancel button...

wonder-sk:
(but we can do it separately on another PR - so please just file a ticket for it)

sklencar:
An issue related to notification has been created: https://github.com/lutraconsulting/input/issues/1086
Conflicts resolved

PR #1065, opened 2020-12-13
fix crash on opening the project

### LINKED ISSUE #1063: Unable to open project without any editable layers
It is impossible to use EPSG:2959. Input crashes everytime trying to open the project.

See: https://public.cloudmergin.com/projects/saber/crash

PeterPetrik:
fix #1063


saberraz:
It is still crashing with the APK from the PR.

tomasMizera (approved):
Looking good, good catch!

It should work even with QgsMapLayer* instead of void* as you mentioned, however I am fine with this :+1:

PR #1066, opened 2020-12-15
Clear search text

### LINKED ISSUE #1038: Not possible to clear Search with x
Issue is related to A7 of the test in https://github.com/lutraconsulting/input-manual-tests/issues/2

When tapping x, it does not clear the search bar in iOS.

tomasMizera:
There is a new function to clear text of textfield. We now use that one instead of hard setting empty string to textfield.

Notice: there is something wrong either in Input or Qt (I could not get rid of) yelling this error when one clicks on **x** to clear text:
`W Input   : QMetaObject::invokeMethod: No such method QQuickContentItem::inputMethodQuery(Qt::InputMethodQuery,QVariant)` 

This closes #1038 however above message stays and should be looked at in future.

### DIFF
--- a/app/qml/SearchBar.qml
+++ b/app/qml/SearchBar.qml
@@ -60,7 +60,7 @@ Rectangle {
     
     TextField {
       id: searchField
-      width: parent.width
+      width: parent.width - iconContainer.width
       height: InputStyle.rowHeight
       font.pixelSize: InputStyle.fontPixelSizeNormal
       color: root.fontColor
@@ -92,9 +92,8 @@ Rectangle {
     Item {
       id: iconContainer
       height: searchField.height
-      width: root.iconSize
+      width: searchField.height + root.panelMargin
       anchors.right: parent.right
-      anchors.rightMargin: root.panelMargin
       
       Image {
         id: cancelSearchBtn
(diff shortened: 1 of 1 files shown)


tomasMizera:
rebased.

app/qml/SearchBar.qml
```diff
@@ -105,23 +105,22 @@ Rectangle {
...
+        source: cancelSearchBtn
+        color: root.fontColor
+      }
 
       MouseArea {
-        x: cancelSearchBtn.x - cancelSearchBtn.width
-        height: cancelSearchBtn.height * 3
-        width: root.iconSize * 3
+        anchors.fill: iconContainer
+        z: 100
```
sklencar, line 117: It should be on the most top level indeed, but I think it is not necessary.

sklencar:
Looks good, fix works. Only a small comment.

app/qml/SearchBar.qml
PeterPetrik: i am paranoid

wonder-sk: @PeterPetrik unless we really need z=100, I would suggest to remove it. Explicit setting of z order can seriously mess up things (e.g. there is another panel on top of a panel with search bar, and this item may get precedence over some other item that is clearly on top of it...

PeterPetrik: removed on master in https://github.com/lutraconsulting/input/commit/eff33712e245df402a5cb05135c82707b9ac577a

PR #1075, opened 2020-12-18
Remove automatic permissions from Qt and remove RECORD_AUDIO permission

### LINKED ISSUE #1069: consider removal of android.permission.RECORD_AUDIO
error: "Anomalous Permission Usage"

>Users prefer apps that request fewer permissions and requesting unnecessary permissions can affect your app's visibility on the Play Store. If these permissions aren't necessary, you may be able to use alternative methods in your app and request fewer permissions. If they are, we recommend providing an explanation to users of why you need the permissions.

tomasMizera:
Removed automatic expand of permissions from Qt to get rid of RECORD_AUDIO permission which we do not use.

Previously we have used these permissions:

In Android app settings it looked like this:

After this change, **Microphone** permission is not there, however, user can use speech-to-text feature provided by keyboard (it is up for a virtual keyboard to opt for RECORD_AUDIO permission).

Let's wait for CI builds to see if everything works properly (I lost splash screen, not sure if it is only problem with my local build or not)

Resolves #1069

### DIFF
--- a/app/android/AndroidManifest.xml
+++ b/app/android/AndroidManifest.xml
@@ -86,8 +86,21 @@
 
     <!-- The following comment will be replaced upon deployment with default permissions based on the dependencies of the application.
          Remove the comment if you do not require these default permissions. -->
-    <!-- %%INSERT_PERMISSIONS -->
+    <!-- <insert double percentage signs to automatically insert permissions from Qt>INSERT_PERMISSIONS -->
+
+    <!-- Storage permissions -->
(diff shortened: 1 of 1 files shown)


tomasMizera:
p.s. sorry for the branch name :man_facepalming: forgot to create new for this ticket

PeterPetrik:
@tomasMizera can you update a single input-manual-test to check to fill note with speech-to-text?

PeterPetrik (approved):

tomasMizera:
@PeterPetrik sure, updated TC04:A14

Issue #1049, opened 2020-12-03
Crash when opening the NAD83 UTM project


edips:
The project works on Input 0.6.2 without crash but it doesn't work for versions 0.7.0 and 0.7.2. This project opens first time without problem, when I open Input again, it clashes on armv8 device. I also tested 0.7.0 version on armv7 device, this project never opened and crashed. I sent crash report with Android. The crash report is very long, I couldn't copy paste. There is some meaningful keywords in the report:
```
Abroad message: 'terminating with uncaught exception of type QgsCsException'
'/lib/arm64/libqgis_core_arm64-v8a.so (QgsCoordinateTransform::TransformDirection) const 172)'
/lib/arm64/libqgs_quickarm64-v8a.so (QgsQuickUtils::transformPoint(QgsCoordinateReferenceSystem const&, QgsCoordinateReferenceSysyem const&, QgsCoordinateTransformContext const&, QgsPointXY const&)+76)
(QgsQuickPositionKit::updateProjectedPosition()+124)
 (QgsQuickPositionKit::onMapSettingsUpdated()+124)
QgsQuickMapSettings::onReadProject(QDomDocument const&)+552
QgsProject::readProjectFile(QDomDocument const&)
QgsProject::readProjectFile(QString const&, QFlags<QgsProject::ReadFlag>)
Loader::forceLoad(QString const&, bool)
(Loader::qt_metacall(QMetaObject(QMetaObject::Call, int; void**)))
```

I compiled Input Master with Qt Creator. this project works without crash on Ubuntu but it crashes on Android. This project worked on Android when I disabled initialization of proj. I think there will be problem about coordinate system without initialization of Proj library. Other projects work without problem.

[contours.zip](https://github.com/lutraconsulting/input/files/5635757/contours.zip)

saberraz:
Indeed, I am experiencing the same issue.

PeterPetrik:
test project; https://public.cloudmergin.com/projects/lutraconsulting/test_grid_shift

PeterPetrik:
probably fixed by https://github.com/qgis/QGIS/pull/40482

edips:
[...]

PR #1087, opened 2021-01-10
Using Qt Positioning instead of QgsQuickPositionKit to prevent crash for grid shift projects 

edips:
This pull request is just testing an alternative solution for grid shift projects at #1049 . When I try the app with these modifications, it didn't crash with grid shift. I couldn't modify QgsPositionKit because compiling QgsQuick with cmake is complex for me :)


PeterPetrik:
Thanks for looking into this. I am afraid we need to use qgspositionkit & qgsquick. I am wondering why you cannot compile it. ATM we have the local copy of qgsquick as part of this repository. so you just need to use qgis_core from the SDK and the qgsquick is build with input. Have you seen docs https://github.com/lutraconsulting/input/blob/master/docs/developers/linux.md ? they should be up to date (or if not correctl we can have a chat to set it for you?)

edips:
You are welcome. When there is no way about PositionKit, it may be thought as an alternative. I implemented this way to my Surveying Calculator on GitHub, including compass direction and marker. I would improve the pull request if you thought to try it.  Thank you for documentation link. There isn't problem building QgsQuick when I compile Input from docker. When I build it from my Ubuntu laptop, CMAKE gives error like Proj, and some other libraries aren't set. I think it is about setting environment variables. I'm going to try again according to Linux documentation and Docker file workflow.

PR #1103, opened 2021-01-19
Project wizard

### LINKED ISSUE #1044: Create project wizard
For novice users it would be great if they could create a simple project to try Input.

sklencar:
Implementation of project wizard to locally create a new project according specification. Allows to create a project with point gpkg layer. Three fields are predefined with option to add or remove a field. To simply workflow, user defined a name and widget type (with default simple configuration). Accordingly field type is set.

Feel free to comment. I think there are some things to improve, but Id rather share some ideas before spending some more time on it.

Manual test updated.

NOTE: Please use `Squash and merge`  option.

Possible enhancement:
* support for more widget combined with more field types (e.g real + slider) - from discussion, now out of scope of this PR
* better handling of having two fields with same names
* floating "Create project" button with reserved space bellow the project list - currently overlapping the list.

Known issues:
* wrong iOS keyboard popup while typing field names
* On desktop build (at least not printed on Android), I got following error probably related to permissions. Note that adding write permissions to gpkg file does not work with QFile::setPermissions.
```
ERROR 1: sqlite3_exec(CREATE TABLE "Survey" ( "fid" INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, "geom" POINT, "Date" DATETIME, "Notes" TEXT, "Photo" TEXT)) failed: attempt to write a readonly database
ERROR 1: sqlite3_exec(CREATE TABLE gpkg_extensions (table_name TEXT,column_name TEXT,extension_name TEXT NOT NULL,definition TEXT NOT NULL,scope TEXT NOT NULL,CONSTRAINT ge_tce UNIQUE (table_name, column_name, extension_name))) failed: attempt to write a readonly database
ERROR 1: sqlite3_exec(CREATE TABLE "ogr_empty_table" ( "fid" INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, "geom" GEOMETRY)) failed: attempt to write a readonly database
```
Inspite of it, project seems to be valid.

https://user-images.githubusercontent.com/6735606/105025225-5a5da400-5a4d-11eb-8ae9-38de3f31d1a8.mp4

closes #1044

### DIFF
--- a/app/main.cpp
+++ b/app/main.cpp
@@ -369,7 +369,7 @@ int main( int argc, char *argv[] )
   MerginProjectStatusModel mpsm( localProjects );
   InputHelp help( ma.get(), &iu );
   FieldsModel fm;
-  ProjectWizard pw;
+  ProjectWizard pw( projectDir, &fm );
 
   // layer models
   LayersModel lm;
@@ -386,6 +386,7 @@ int main( int argc, char *argv[] )
   QObject::connect( &app, &QCoreApplication::aboutToQuit, &loader, &Loader::appAboutToQuit );
   QObject::connect( ma.get(), &MerginApi::syncProjectFinished, &pm, &ProjectModel::syncedProjectFinished );
   QObject::connect( ma.get(), &MerginApi::projectDetached, &pm, &ProjectModel::findProjectFiles );
+  QObject::connect( &pw, &ProjectWizard::projectCreated, &pm, &ProjectModel::addLocalProject );
   QObject::connect( ma.get(), &MerginApi::listProjectsFinished, &mpm, &MerginProjectModel::updateModel );
   QObject::connect( ma.get(), &MerginApi::syncProjectStatusChanged, &mpm, &MerginProjectModel::syncProjectStatusChanged );
   QObject::connect( ma.get(), &MerginApi::reloadProject, &loader, &Loader::reloadProject );
@@ -486,6 +487,7 @@ int main( int argc, char *argv[] )
   engine.rootContext()->setContextProperty( "__activeLayer", &al );
   engine.rootContext()->setContextProperty( "__purchasing", purchasing.get() );
   engine.rootContext()->setContextProperty( "__fieldsModel", &fm );
+  engine.rootContext()->setContextProperty( "__projectWizard", &pw );
 
 #ifdef MOBILE_OS
   engine.rootContext()->setContextProperty( "__appwindowvisibility", QWindow::Maximized );
--- a/app/main.cpp
+++ b/app/main.cpp
@@ -393,6 +393,7 @@ int main( int argc, char *argv[] )
   QObject::connect( &mtm, &MapThemesModel::mapThemeChanged, &recordingLpm, &LayersProxyModel::onMapThemeChanged );
   QObject::connect( &loader, &Loader::projectReloaded, vm.get(), &VariablesManager::merginProjectChanged );
   QObject::connect( &loader, &Loader::projectWillBeReloaded, &inputProjUtils, &InputProjUtils::resetHandlers );
+  QObject::connect( &pw, &ProjectWizard::notify, &iu, &InputUtils::showNotificationRequested );
   QObject::connect( QgsApplication::messageLog(),
                     static_cast<void ( QgsMessageLog::* )( const QString &message, const QString &tag, Qgis::MessageLevel level )>( &QgsMessageLog::messageReceived ),
                     &iu,
--- a/app/main.cpp
+++ b/app/main.cpp
@@ -227,6 +227,7 @@ void initDeclarative()
   qmlRegisterUncreatableType<ActiveLayer>( "lc", 1, 0, "ActiveLayer", "" );
   qmlRegisterType<DigitizingController>( "lc", 1, 0, "DigitizingController" );
   qmlRegisterType<PositionDirection>( "lc", 1, 0, "PositionDirection" );
+  qmlRegisterType<FieldsModel>( "lc", 1, 0, "FieldsModel" );
 }
 
 #ifdef INPUT_TEST
@@ -368,8 +369,7 @@ int main( int argc, char *argv[] )
   MerginProjectModel mpm( localProjects );
   MerginProjectStatusModel mpsm( localProjects );
   InputHelp help( ma.get(), &iu );
-  FieldsModel fm;
-  ProjectWizard pw( projectDir, &fm );
+  ProjectWizard pw( projectDir );
 
   // layer models
   LayersModel lm;
@@ -487,7 +487,6 @@ int main( int argc, char *argv[] )
   engine.rootContext()->setContextProperty( "__browseDataLayersModel", &browseLpm );
   engine.rootContext()->setContextProperty( "__activeLayer", &al );
   engine.rootContext()->setContextProperty( "__purchasing", purchasing.get() );
-  engine.rootContext()->setContextProperty( "__fieldsModel", &fm );
   engine.rootContext()->setContextProperty( "__projectWizard", &pw );
 
 #ifdef MOBILE_OS
--- a/app/main.cpp
+++ b/app/main.cpp
@@ -50,6 +50,7 @@
 #include "variablesmanager.h"
 #include "inputhelp.h"
 #include "inputprojutils.h"
+#include "fieldsmodel.h"
 
 #ifdef INPUT_TEST
 #include "test/testmerginapi.h"
@@ -366,6 +367,7 @@ int main( int argc, char *argv[] )
   MerginProjectModel mpm( localProjects );
   MerginProjectStatusModel mpsm( localProjects );
   InputHelp help( ma.get(), &iu );
+  FieldsModel fm;
 
   // layer models
   LayersModel lm;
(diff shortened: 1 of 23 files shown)


app/projectwizard.cpp
```diff
@@ -0,0 +1,191 @@
...
+    QgsEditorWidgetSetup setup = getEditorWidget( f, mFieldsModel->findWidgetTypeByFieldName( f.name() ) );
+    l->setEditorWidgetSetup( i, setup );
+  }
+  l->commitChanges();
+
+  return l;
+}
+
+void ProjectWizard::createProject( QString const &projectName )
+{
```
tomasMizera, line 73: Lets check for storage permission first, or better, ask for it when user clicks on "Create Project"

```suggestion
{
if .. InputUtils::hasLocationPermission() 
```

app/qml/MerginProjectPanel.qml
```diff
@@ -604,6 +614,30 @@ Item {
...
+
+        DelegateButton {
+          height: parent.height
+          width: parent.width/2
+          btnWidth: InputStyle.rowHeightHeader * 3
+          btnHeight: parent.height
+          text: qsTr("Create project")
+          anchors.centerIn: parent
+          onClicked: {
+            __fieldsModel.initModel()
```
tomasMizera, line 635: This is probably better place to check for storage permission

PeterPetrik:
wow

PeterPetrik:
@erpas can give us background maps :)

saberraz:
Very impressive, nice work!

saberraz:
@sklencar I would keep the wizard simple in terms of support of other field types. This is meant to get users going with their first project.

app/qml/ProjectWizardPage.qml
```diff
@@ -0,0 +1,203 @@
...
+          onRemoveClicked: __fieldsModel.removeField(index)
+        }
+      }
+
+      Row {
+        id: listButtonContainer
+        width: parent.width
+        Layout.preferredHeight: projectWizardPanel.rowHeight * 0.8
+        Layout.fillWidth: true
+        Button {
```
tomasMizera, line 134: Can we center this button on the screen? Or maybe put it inside toolbar? (not sure which one is better)

app/projectwizard.cpp
```diff
@@ -0,0 +1,191 @@
...
+  QString projectDir = InputUtils::createUniqueProjectDirectory( mDataDir, projectName );
+  QString projectFilepath( QString( "%1/%2.%3" ).arg( projectDir ).arg( projectName ).arg( "qgs" ) );
+  QString gpkgName( QStringLiteral( "data" ) );
+  QString projectGpkgPath( QString( "%1/%2.%3" ).arg( projectDir ).arg( gpkgName ).arg( "gpkg" ) );
+
+  QgsProject project;
+
+  // add background layer
+  QString urlWithParams( "tilePixelRatio=1&type=xyz&url=https://tile.openstreetmap.org/%7Bz%7D/%7Bx%7D/%7By%7D.png&zmax=19&zmin=0" );
```
tomasMizera, line 82: I vote to have some config (header?) file with constants like this. We could also add 'EPSG:3857" and other constants there. What do you think?

app/qml/ProjectWizardPage.qml
```diff
@@ -0,0 +1,203 @@
...
+  signal back
+
+  property real rowHeight: InputStyle.rowHeight
+  property var fontColor: InputStyle.fontColor
+  property var bgColor: InputStyle.clrPanelMain
+  property real panelMargin: 10 * QgsQuick.Utils.dp
+
+  property ListModel items: ListModel {}
+  property var supportedTypes: {
+    var types = __fieldsModel.supportedTypes()
```
tomasMizera, line 22: I do not quite understand this part could you please add comment here what is it doing? I understand you are trying to convert QVariantMap and fill the ListModel with it however, is the `items` property used somewhere? Is it implicitly converting something?

app/qml/ProjectWizardPage.qml
```diff
@@ -0,0 +1,203 @@
...
+    for (var prop in types) {
+      projectWizardPanel.items.append({ "display": types[prop], "widget": prop })
+    }
+
+    projectWizardPanel.items
```
tomasMizera, line 27: Is `return` not needed?

app/qml/ProjectWizardPage.qml
```diff
@@ -0,0 +1,203 @@
...
+        height: parent.height
+        MainPanelButton {
+          id: createProjectBtn
+          width: toolbar.itemSize
+          text: qsTr("Create project")
+          faded: !projectNameField.text
+          imageSource: InputStyle.checkIcon
+
+          onActivated: {
+            if (faded) {
```
tomasMizera, line 193: I suggest this for better UX: 
```suggestion
            if (faded) {
              __InputUtils.play('https://www.youtube.com/watch?v=60ItHLz5WEA')
```

tomasMizera (changes requested):
Great job!
Some notes:
- Please add this to manual tests, 
- I have experienced crash on Android when created project without any attribute and tried to add first feature - application crashed. However I could not reproduce it anymore, it is probably not related to these changes but qgsquick models.

app/qml/ProjectWizardPage.qml
sklencar: Implicit return works in QML, but only while defining property. Of course, `return` statement would work as well. I just prefer to use it without it similarly as I would define one-liner property 
[Small example if you would like](https://pastebin.com/KJ7RPezG)

saberraz:
There is no way to define the layer name, which is good IMO. Can be call it Survey instead of Points.

saberraz:
Can we call everything "Field" and not attribute?
File (19)

wonder-sk:
- In the combo box, the space between the frame and text on the left side is smaller than the space in the text edit box - can we fix that too?
- "project name" edit box does not need any placeholder text - the "project name" label is just above it, making the label duplicate

app/projectwizard.cpp
sklencar: I created const QStrings out of those configuration strings. Id keep them in projectwizard.h since they are only project wizard related and could be possibly variables in the future. However having global constants in one file is quite a good idea.

PeterPetrik:
I think we need to update input-manual-tests as well as input-docs with the change if not done yet

saberraz:
One last request: can we make Create button on the home page similar to Fetch. It will sit at the bottom of the list, instead of floating on top of projects.

app/qml/ProjectWizardPage.qml
tomasMizera: Cool, thanks :)

tomasMizera (changes requested):
Thanks for resolving all the comments. 

There is only one thing I found regarding storage permission:

- If I do not allow storage permission during startup, application now asks for it when I click on `create project` (correct :heavy_check_mark: ). However, when I then allow it and try to create a project, application crashes (:exclamation:). It crashes on this assert https://github.com/lutraconsulting/input/pull/1103/files#diff-344db1d166225401e7e99377d5639b963b4f2ee22014e58826a1349e37542344R57

We should show the same window as when user allows storage permission on any other place. It has id `restartAppDialog`. The reason is that projects are not being reloaded on providing storage permission. I believe we can sort this one out during project models refactoring.

tomasMizera (commented):
Application is no longer crashing due to storage permission :heavy_check_mark: 

However, @sklencar I can not see markers on a map anymore:

sklencar:
Wrong ordering of layer (mentioned above) fixed.

tomasMizera (approved):
Markers are now visible. @wonder-sk I am ok to merge.

app/main.cpp
```diff
@@ -476,6 +482,8 @@ int main( int argc, char *argv[] )
   engine.rootContext()->setContextProperty( "__browseDataLayersModel", &browseLpm );
   engine.rootContext()->setContextProperty( "__activeLayer", &al );
   engine.rootContext()->setContextProperty( "__purchasing", purchasing.get() );
+  engine.rootContext()->setContextProperty( "__fieldsModel", &fm );
```
wonder-sk, line 485: Do we need fields model as a global variable? Can't project wizard create the fields model inside of its QML?

app/projectwizard.cpp
```diff
@@ -0,0 +1,219 @@
...
+#include "qgsvectorfilewriter.h"
+#include "qgsdatetimefieldformatter.h"
+
+ProjectWizard::ProjectWizard( const QString &dataDir, FieldsModel *fieldsModel, QObject *parent )
+  : QObject( parent )
+  , mDataDir( dataDir )
+  , mFieldsModel( fieldsModel )
+{
+
+  mSettings = new QgsMapSettings();
```
wonder-sk, line 16: It seems this is not gettting deleted anywhere....

Best to use std::unique_ptr rather than raw pointers...

app/projectwizard.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+  l->startEditing();
+  l->setCrs( layerCrs );
+  for ( int i = 0; i < l->fields().count(); ++i )
+  {
+    QgsField f = l->fields().at( i );
+    QgsEditorWidgetSetup setup = getEditorWidget( f, mFieldsModel->findWidgetTypeByFieldName( f.name() ) );
+    l->setEditorWidgetSetup( i, setup );
+  }
+  l->commitChanges();
```
wonder-sk, line 69: We should not need to call startEditing() + commitChanges() ... that's only related to editing of layer data (e.g. adding/modifying/removing features), here we are only changing properties of the layer

Issue #1098, opened 2021-01-16
Use GeoPackage for demo projects


saberraz:
Let's not bring back Shapefile from death.

Issue #1096, opened 2021-01-16
Images in the pop-ups for demo projects do not scale well


saberraz:
![File (17)](https://user-images.githubusercontent.com/2649047/104807245-78b96a80-57d5-11eb-99be-a8058c19bf31.jpg)
![File (16)](https://user-images.githubusercontent.com/2649047/104807247-79520100-57d5-11eb-9a32-d74777541d15.jpg)

PR #1107, opened 2021-01-21
Updated demo projects:

mostlyAtNight:
- Prompts are now images
- No longer use shapefiles


mostlyAtNight:
Hi @PeterPetrik ,

Could you please review my changes here - mainly whether these will also apply to the projects on the iOS build?  I am confident of the projects themselves and they were tested on iOS already.  

Thanks!

saberraz:
Resolves https://github.com/lutraconsulting/input/issues/1098 and https://github.com/lutraconsulting/input/issues/1096

PR #1115, opened 2021-01-28
Reuse last entered value

### LINKED ISSUE #798: Respect QGIS setting for last value entered in attribute table
In QGIS > Settings > Options > Digitizing under Feature Creation, there is the option to Reuse last entered attribute values.

By default this option is disabled, but if user has turned it to ON, we should respect it in Input.

tomasMizera:
Added possibility to reuse last entered values while creating new features. This option allows user to faster record new features.

To use this feature:
 - go to settings and check "Save last values"
 - create new feature = you will see checkboxes next to each attribute
 - click some of the checkboxes = these values will be saved for next feature
 - creat feature and try to create another one = selected attributes will be prefilled

Note:
- this also works when switching between multiple layers.
- setting is saved to QSettings and thus even when app is restarted, this check stays the same. (Though saved values are not preserved)
- look of checkboxes will be changed

**EDIT:**

- new UI

https://user-images.githubusercontent.com/22449698/106280790-a9ba8600-623e-11eb-9b53-698bc3c89076.mp4

Resolves #798

### DIFF
--- a/qgsquick/from_qgis/plugin/components/qgsquickcheckboxcomponent.qml
+++ b/qgsquick/from_qgis/plugin/components/qgsquickcheckboxcomponent.qml
@@ -1,0 +1,90 @@
+/***************************************************************************
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+
+import QtQuick 2.8
+import QtQuick.Templates 2.1 as T
+import "../" // import styling singleton
+
+T.CheckBox {
+    id: control
+
+    signal checkboxClicked( var buttonState )
+
+    property var spaceWidth: height / 4.0
+    property var baseSize: height / 5.5
+    property var baseColor: "black"
+
+    implicitHeight: indicator.implicitHeight + topPadding + bottomPadding
+    implicitWidth: indicator.implicitWidth + leftPadding + rightPadding
+
+    leftPadding: spaceWidth
+
+    onClicked: control.checkboxClicked( control.checkState )
+
+    indicator: Rectangle {
+        id: checkboxHandle
+        implicitWidth: baseSize * 2.6
+        implicitHeight: baseSize * 2.6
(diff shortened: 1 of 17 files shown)


saberraz:
Can we rename to "Reuse last value option"? Save has a different meaning. QGIS terminology is: "Reuse last entered attribute values" which is a bit mouthful :)

tomasMizera:
PR ready for review. @saberraz text is changed, suggestions are always welcomed :)

qgsquick/from_qgis/plugin/components/qgsquickcheckboxcomponent.qml
```diff
@@ -0,0 +1,90 @@
...
+import "../" // import styling singleton
+
+T.CheckBox {
+    id: control
+
+    signal checkboxClicked( var buttonState )
+
+    property var spaceWidth: height / 4.0
+    property var baseSize: height / 5.5
+    property var baseColor: "black"
```
sklencar, line 21: PR introduces new styling object for the checkbox widget, which is good. However, its not used, only defined in `qgsquickfeatureformstyling.qml` and redefined for Input in `featurepanel.qml`. Additionally I think import for styling singleton is not needed.

sklencar: Rather use something like `customStyle.checkbox.baseColor` instead of "black".

sklencar (changes requested):
Generally looks good and clean, good work!
However I got just small comment on checkbox styling.

Additionally, there might be some rather rare use-cases where last value doesnt work - Id say its more just point of view or expectations than being actually a bug, e.g:

- if such option should be available for unique field values;
- if last entered value should override default value
In most cases it works nice and those mentioned above are rather just something to write in docs to have clear idea about how it should work.

Also might worth to add it to manual test or possibly write automatic test.

qgsquick/from_qgis/plugin/components/qgsquickcheckboxcomponent.qml
tomasMizera: Thanks for review, I added manual tests and changed the UI things.

tomasMizera:
Qgis PR: https://github.com/qgis/QGIS/pull/41293

wonder-sk:
Looks good to me...

PR #1119, opened 2021-02-01
Initialize demo project only on the first run

### LINKED ISSUE #1109: Can’t delete example project
Since the last update on iOS, the exemple projects « start here! » and « Capturing field data tour » are always present. I can delete them, but when I restart the app, they are here again!

sklencar:
Introducing `isAppInitialized` qsettings variable to see if the app has been already launched or it is the first time run.
Demo projects are copied only on the very first run. This should allow user to delete demo project or kept demo project data from previous runs. 

closes #1109

### DIFF
--- a/app/main.cpp
+++ b/app/main.cpp
@@ -327,6 +327,7 @@ int main( int argc, char *argv[] )
 #ifdef ANDROID
   appBundleDir = dataDir + "/qgis-data";
   demoDir = "assets:/demo-projects";
+  demoDir = dataDir + "/qgis-data/resources/demo-projects";
 #endif
 #ifdef Q_OS_IOS
   appBundleDir = QCoreApplication::applicationDirPath() + "/qgis-data";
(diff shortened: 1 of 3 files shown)


tomasMizera (approved):
LGTM, I like the refactoring of settings.

Maybe in future we could allow users to copy demo projects back.

app/main.cpp
```diff
@@ -343,9 +343,15 @@ int main( int argc, char *argv[] )
 #else
   projDir = dataDir;
 #endif
-
+  AppSettings as;
   InputProjUtils inputProjUtils;
-  copy_demo_projects( demoDir, projectDir );
+
+  // copy demo projects if the app is launched for the first time
+  if ( !as.isAppInitialized() )
```
wonder-sk, line 350: can we use e.g. "demoProjectsCopied" instead of "appInitialized" ? The latter is very generic and it could mean anything - the former would be more descriptive IMHO :-)

wonder-sk (approved):

PR #1147, opened 2021-02-05
Fix missing basefile for a new projects

### LINKED ISSUE #1127: Syncing problem for projects created with wizard
When on tests https://github.com/lutraconsulting/input-manual-tests/issues/25 with Xiaomi Redmi Note Pro 8,

For https://github.com/lutraconsulting/input-manual-tests/issues/3 Test C 

I get syncing problems for projects created with wizard: if edited on another device, the project cannot be synced back on the testing device due to conflicts detected.

Other projects do sync nicely.

sklencar:
Prepare basefile for changeset for a newly added files (mainly for a new project).

By adding migration to Mergin functionality, a new use case has been introduced with it. The only way how to obtain a project that can be synced with Mergin, was to donwload it. After download, a basefile for .gpkg files are copied. However, by migrating a new project to mergin, the basefile preparation was omitted (not needed before though), therefore geodiff couldnt work properly.

closes #1127

### DIFF
--- a/app/merginapi.cpp
+++ b/app/merginapi.cpp
@@ -735,6 +735,10 @@ void MerginApi::createProjectFinished()
         QDir projectDir( info.projectDir );
         if ( projectDir.exists() && !projectDir.isEmpty() )
         {
+
+
+          // TODO @vsklencar create changeset
+          //createChangeset
           uploadProject( projectNamespace, projectName );
         }
       }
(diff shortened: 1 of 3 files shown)


saberraz:
This solves the problem.

wonder-sk (commented):
- copying basefiles for newly added files right in MerginApi::uploadInfoReplyFinished() does not sound like a good idea - if the upload fails, we would end up with some basefiles that should not be there. I think that code should go to e.g. MerginApi::uploadFinishReplyFinished() which also deals with basefiles
- in TransactionStatus we can have a list of newly added diffable files (set in uploadInfoReplyFinished()) and processed in uploadFinishReplyFinished())
- we need a test for this

sklencar:
A new diffable files are contained in `transaction.diff.localAdded`. However we can explicitly include them to `transaction.uploadDiffFiles` or create a new member for transactions e.g `transaction.adddedDiffFiles` as you propose.

PR #1167, opened 2021-02-09
Use PointZ in wizard

### LINKED ISSUE #1166: integrating elevation in automatic projects
Capture d’écran de 2021-02-09 16-02-17

only the x and y are recorded in the geopackage, not the Z

saberraz:
Changed the point type so the Z value will be captured too.

fixes #1166


jancelin:
the elevation is well registered! 
thank you very much!

Capture d'écran 2021-02-09 20:36:14

PeterPetrik (approved):

PR #1081, opened 2021-01-04
Camera permission crash

tomasMizera:
Fixed by opting for a permission when starting a camera, if user do not provide permission camera wont start

+ also removed opting for permission in the application start

Resolves #1078

PR #1185, opened 2021-02-12
Qr code reader

### LINKED ISSUE #408: Possibility to use camera as QR code scanner instead of taking photos
In some cases, users want to scan a QR code and parse the content as a string to the form. It will be good to have an option to use QR scanner feature from the camera, in addition to photo taking.

Not sure, if this will need extending QGIS widget types??

sklencar:
QR code reader implementation using ZXing library.
* Supported multiple platforms
    * full support for Android and iOS,
    * supported also MacOS desktop build, but seems to have bug that it is capturing invalid image for ZXing library,
    * WIN build not supported;
* available for TextEdit widget if an attribute's `name` contains `qrcode` case insensitive. Note that `name` is taken from field.name or overwriten by field.alias if exists.
* added invalid data changed notification
* basic style with simple header and camera overlay

Related manual-test: https://github.com/lutraconsulting/input-manual-tests/issues/28
closes #408 

Known issues
* camera issues
    * blank page when camera is being loaded (affects some devices)
    * `QImage qt_imageFromVideoFrame(const QVideoFrame &) : unsupported pixel format Format_ABGR32` while converting videoOutput to QImage (seems to be Android only).
    * sometimes not QR decoder not responses well

https://user-images.githubusercontent.com/6735606/108194128-8769a900-7116-11eb-8151-dc3987f31e13.MP4


tomasMizera (commented):
Very good job!

I did not go through the whole code yet, though what I have seen so far looks great. I tested it on Android and it works well. I could not get Linux build running but I guess we will be able to sort it out.

Some notes:
 - Permission for camera on Android is now again requested at startup (this was changed in #1081). I think what we want is to opt it when needed - when user clicks on `take a photo` or now qr code icon. It is possible that it comes from ZXing library, but the problem is that it will be forbidden in upcoming version of Android to immediately ask for permissions without any information. 
 - If permission is not granted and user opens QR code scanner, it says something like "Camera service is not operating?" which is not true, user only did not provide permission.
 - Maybe icon for QR Code can be smaller?
 - When user scans invalid (not convertible) value for field, maybe we could display some more user friendly message, now:

.github/workflows/android.yml
```diff
@@ -1,7 +1,7 @@
 name: Android
 on: [push]
 env:
-  SDK_VERSION: android-20
+  SDK_VERSION: android-21
```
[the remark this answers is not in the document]
sklencar: SDK has to be upgraded due to xzing lib

sklencar:
All remarks from notes should be resolved. However I kept the QR code icon unchanged. I would suggest to revise all icons in quick widgets to use its size variable or derive size from a ratio of rowHeight.

tomasMizera (commented):
Based on a phone discussion, please add some more comment to have an idea how this is working. It is working very nice on Android! Thanks for resolving the permission problems, it is now working as expected.

tomasMizera:
Also regarding the manual test, I vote to add it here https://github.com/lutraconsulting/input-manual-tests/issues/5 rather than creating new test case. Thanks!

sklencar:
Remarks resolved + manual test relocated.

I have checked the issue with scanning the same code for multiple fields. The issue seems to be rather linked with the library (or, of course, something else). This use case is possible, but the scanner acts like it has problem to scan it sometimes. Usually when I was moving camera a bit and wait longer time, it eventually scan the code. This could be possibly improved to add some flags to the DecodeHints in decoder (e.g Binarizer which I keep as default since I was having some issues with reading barcodes when i set it to Binarizer::FixedThreshold). In any case, this is something that has to be investigated further.

app/codefilter.cpp
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    QVideoFrame run( QVideoFrame *input,
+                     const QVideoSurfaceFormat &surfaceFormat,
+                     QVideoFilterRunnable::RunFlags flags ) override
+    {
+      Q_UNUSED( surfaceFormat );
+      Q_UNUSED( flags );
+
+      if ( mFilter->decoder()->isDecoding() )
```
tomasMizera, line 33: Check if `mFilter` is a valid pointer

app/codefilter.cpp
```diff
@@ -0,0 +1,91 @@
...
+CodeFilter::CodeFilter()
+{
+  mDecoder = new QRDecoder;
+
+  QObject::connect( mDecoder, &QRDecoder::capturedChanged, this, &CodeFilter::setCapturedData );
+}
+
+QVideoFilterRunnable *CodeFilter::createFilterRunnable()
+{
+  return new QRRunnable( this );
```
tomasMizera, line 64: We should probably save this address to a new attribute and free it in `CodeFilter`'s destructor or somehow destroy QRRunnable instance when `this` (CodeFilter) is deleted

app/codefilter.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+#include <QtMultimedia/qvideoframe.h>
+#include <QImage>
+#include <QOpenGLContext>
+#include <QOpenGLFunctions>
+#include <QDebug>
+
+void processImage( QRDecoder *decoder, const QImage &image )
+{
+  decoder->process( image );
```
tomasMizera, line 13: Check if `decoder` is a valid pointer

app/codefilter.h
```diff
@@ -0,0 +1,44 @@
...
+
+  signals:
+    void capturedDataChanged();
+    void isDecodingChanged( bool isDecoding );
+
+  private slots:
+    void setCapturedData( const QString &capturedData );
+
+  private:
+    QSharedPointer<QMutex> _mutexP;
```
tomasMizera, line 37: Can we remove this? It seems not to be used anywhere

app/qml/CodeReader.qml
```diff
@@ -0,0 +1,112 @@
...
+  Timer {
+    id: codeReaderTimer
+    interval: 1000
+    triggeredOnStart: false
+    repeat: false
+    onRunningChanged: {
+      if (!running) {
+        codeReader.visible = false
+      }
+    }
```
tomasMizera, line 110:
```suggestion
    onTriggered: codeReader.visible = false
```

tomasMizera: This should work too :)

tomasMizera (approved):
Thanks for resolving the notes, however, I unresolved one, I could not find how it was resolved (the one with `Component.onCompleted`)

Other than that there are some notes about memory. After clearing these I am ok to merge so passing to you @wonder-sk 🟢

app/codefilter.cpp
sklencar: From documentation, it seems that destructions is handled. Also added docs to the function.
http://doc.qt.io/qt-5/qabstractvideofilter.html#createFilterRunnable

qgsquick/from_qgis/plugin/qgsquickfeatureform.qml
```diff
@@ -474,6 +497,10 @@ Item {
                return form.loadWidgetFn(widget.toLowerCase())
             else return ''
           }
+
+          Component.onCompleted: {
+            supportDataImport = importDataHandler.supportImportData(Name)
```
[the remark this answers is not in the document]
sklencar: It is simplified to use the function from handler directly to assign `supportDataImport`  attribute. Added also default function to default handler definition.

tomasMizera (approved):

PR #1187, opened 2021-02-16
Add OGR to Linux deps

tomasMizera:
Linux would not compile due to a missing GDAL path.

### DIFF
--- a/app/config.pri.default
+++ b/app/config.pri.default
@@ -43,6 +43,9 @@ unix:!macx:!android {
 
   PROJ_INCLUDE_DIR = $${BASE_DIR}/include
   PROJ_LIB_DIR = $${BASE_DIR}/lib
+
+  # To use OGR API
+  OGR_INCLUDE_DIR = < path to gdal.h like /usr/include/gdal >
 }
 
 macx:!android {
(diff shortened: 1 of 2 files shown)


app/config.pri.default
```diff
@@ -43,6 +43,9 @@ unix:!macx:!android {
 
   PROJ_INCLUDE_DIR = $${BASE_DIR}/include
   PROJ_LIB_DIR = $${BASE_DIR}/lib
+
+  # To use OGR API
+  OGR_INCLUDE_DIR = < path to gdal.h like /usr/include/gdal >
```
PeterPetrik, line 48: where is gdal.h used?

tomasMizera: In ProjectWizard we use `QgsVectorFileWriter` and it is its dependency 
https://github.com/lutraconsulting/input/blob/master/app/projectwizard.cpp#L7

https://github.com/qgis/QGIS/blob/master/src/core/qgsvectorfilewriter.h#L32

PeterPetrik: in that case it should be added to all platforms? (on iOS and Android it is in the same folder as proj, so it works probably, but on mac I am not sure @sklencar )?

PeterPetrik (approved):

app/config.pri.default
sklencar: It has the same folder on MacOS (qgis-deps folder). I have not defined it explicitly since the path is included with proj variables.

PR #1193, opened 2021-02-19
Fix of writing/reading QSettings

sklencar:
closes issue related to not be able to save gspTolerance

### DIFF
--- a/app/appsettings.cpp
+++ b/app/appsettings.cpp
@@ -15,9 +15,7 @@
 AppSettings::AppSettings( QObject *parent ): QObject( parent )
 {
   QSettings settings;
-  settings.setDefaultFormat( QSettings::NativeFormat );
   settings.beginGroup( mGroupName );
-
   QString path = settings.value( "defaultProject", "" ).toString();
   QString layer = settings.value( "defaultLayer/"  + path, "" ).toString();
   bool autoCenter = settings.value( "autoCenter", false ).toBool();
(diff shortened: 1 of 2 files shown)


wonder-sk:
What was the actual issue, why do we need to set NativeFormat? (and which one is the default normally?)

sklencar:
Actually it is super weird, because the default defaultFormat was NativeFormat. Also checked QSettings::scope as well as appname and organization. All seems to be set ok.
Also checked (just briefly) constructors of QSettings and QSettingsPrivate and it uses `static QSettings::Format globalDefaultFormat = QSettings::NativeFormat;`

If needed, I am happy to investigate it more what actually happen when DefaultFormat is set manually and what is difference from an empty constructor. It is not clear to me and of course it raises suspicious what is actually going on.

wonder-sk:
Yes please if you could have a look... it could be that when we initialize QGIS libs they do some magic with QSettings and override the default? (QGIS likes to use .ini format for settings)

sklencar:
Indeed, the problem was elsewhere - caused by messed up qgis init and initialization of AppSettings.
(https://github.com/lutraconsulting/input/pull/1119/files#diff-c1b709b8a90bef4573ddad8a9b2fbc5d5dbc85e4dc759f95b6d9905931fec24fR346)

wonder-sk:
Cool thanks. Can you maybe add a comment to `AppSettings as;` that says it should be created after qgis libs are initialized, so that next time we do not move it by mistake?

tomasMizera (approved):
Good catch!
I do not know whether I understood the problem completely. If it was only about `AppSettings` not storing and reloading saved options, then this is working ✔️ . If anything else needs to be retested, let me know.
 
Maybe it would be a good idea to add manual test for QSettings - if they are actually being saved and reloaded. We can mess up things like this from time to time

wonder-sk (approved):

Issue #755, opened 2020-06-12
Android Back Button Not Working in Qgis Feature Form


tomasMizera:
Qgis feature form from qgisquick eats active focus, we need to change qgisquick to return focus to be able to handle back button properly.

PR #1201, opened 2021-02-25
Relocated save button from Feature form toolbar to header

### LINKED ISSUE #384: Save button obscured
Check this:

image

I've now finished editing attribute data and want to go back to the map (saving my work).  The obvious button that's visible is 'Back' but if I press that, my point is lost (I have done this many times by accident).  To save I have the (1) minimise the KB then (2) hit save.

Perhaps we can put the save button at the top of the page so we have either 'Save' or 'Cancel' there?

### LINKED ISSUE #445: Ask for confirmation before discarding edits
Currently when user taps "Back" button in the form, it will be immediately closed, discarding any pending edits. It would be good to show a confirmation box in case there are any unsaved changes... "Would you like to save changes?" - Yes / No / Cancel.  (Yes/No would close the panel, Cancel would keep the panel open)

sklencar:
* save button turned red if hard constrain are not met (similar to how was it before when save button was disabled in this case)
* added check if a feature has been changed - either if is newly created or edit buffer has uncommitted changes for the feature. Note that saving geometry is committed immediately after geometry edit, therefore was omitted for this function. This also effects attributes changed before geometry has been changed and committed.
* Cancel discard changes
* Save saves all changes
* Back button throws a dialog to choose action

Feel free to comment on how checking uncommitted changes should work or any uix related issues.

Screenshot 2021-02-25 at 14 34 37

closes #384
closed #445

### DIFF
--- a/app/qml/FeaturePanel.qml
+++ b/app/qml/FeaturePanel.qml
@@ -38,6 +38,14 @@ Drawer {
       attributeModel.forceClean();
     }
 
+    function onAboutToClose() {
+      if (featureForm.hasAnyChanges())  {
+        saveChangesDialog.open()
+      } else {
+        featurePanel.visible = false
+      }
+    }
+
     id: featurePanel
     visible: false
     modal: false
@@ -55,7 +63,7 @@ Drawer {
       focus: true
       Keys.onReleased: {
         if (event.key === Qt.Key_Back || event.key === Qt.Key_Escape) {
-          featurePanel.close()
+          featurePanel.onAboutToClose()
         }
       }
     }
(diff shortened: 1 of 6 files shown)


app/qml/FeaturePanel.qml
```diff
@@ -134,10 +142,36 @@ Drawer {
...
+                anchors.right: parent.right
+                anchors.bottom: parent.bottom
+                anchors.top: parent.top
+                anchors.rightMargin: InputStyle.panelMargin // same as back button
+
+                MouseArea {
+                  anchors.fill: parent
+                  onClicked: featureForm.save()
+                }
+            }
```
tomasMizera, line 173: I guess the save button should only be visible when feature form is in edit mode?

```qml
        visible: featureForm.state === "edit"
```

tomasMizera (approved):
Very good job!

I had just some minor notes.

I think this will work even better when we fix #755 (back button losing focus in featureform)

Failing iOS build is due to a new SDK (versions mismatch), should pass when QrCode is merged.

tomasMizera (approved):
🚀  @wonder-sk Fine by me to merge.

Failed iOS check is just the version mismatch.   
Failed QgsQuick check - all UI changes will go to QGIS from master

app/qml/FeaturePanel.qml
```diff
@@ -310,6 +343,29 @@ Drawer {
               }
           }
         }
+
+        MessageDialog {
+          id: saveChangesDialog
+          visible: false
+          title: qsTr( "Uncommited changes" )
```
wonder-sk, line 350:
```suggestion
          title: qsTr( "Unsaved changes" )
```
("uncommitted" sounds too technical)

PR #1208, opened 2021-03-01
Reset old references on project reload

### LINKED ISSUE #1204: Record causes crash
Not sure if this one is already filed (there are more for form models, I think this one is related too), but I finally came across reproducible scenario. 

You need two specific projects: `tomas/test-general` and `tomas/SlovakiaTrees` both on public

Procedure:
 - Open project `test-general` and create new feature
 - Open project `SlovakiaTrees`
 - Hit Record
 - 💥 

Maybe some leftover pointer ..

tomasMizera:
There were leftover references to old project when new one was loaded. This led to application crash on different places.

Resolves #1204

### DIFF
--- a/app/qml/main.qml
+++ b/app/qml/main.qml
@@ -738,6 +738,11 @@ ApplicationWindow {
         target: __loader
         onLoadingStarted: projectLoadingScreen.visible = true
         onLoadingFinished: projectLoadingScreen.visible = false
+        onProjectReloaded: {
+          // clear all previous references to old project
+          highlight.featureLayerPair = null
+          digitizingHighlight.featureLayerPair = null
+        }
     }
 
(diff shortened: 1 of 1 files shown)


sklencar (approved):

app/qml/main.qml
```diff
@@ -738,6 +738,11 @@ ApplicationWindow {
         target: __loader
         onLoadingStarted: projectLoadingScreen.visible = true
         onLoadingFinished: projectLoadingScreen.visible = false
+        onProjectReloaded: {
+          // clear all previous references to old project
```
wonder-sk, line 742: Can you please also add a short note about why we do it (e.g. "if we don't clear references to the previous project, highlights may end up with dangling pointers to map layers and cause crashes")

PR #1209, opened 2021-03-01
Removed "back" text from PanelHeader

sklencar:
Code cleaned from unnecessary references for back text in the header.

### DIFF
--- a/app/qml/PanelHeader.qml
+++ b/app/qml/PanelHeader.qml
@@ -17,9 +17,9 @@ Rectangle {
 
     property real rowHeight
     property string titleText: ""
-    property string backText: qsTr("Back")
+    property string backText: qsTr("Cancel")
     property bool withBackButton: true
-    property bool backTextVisible: true
+    property bool backTextVisible: false
     property bool backIconVisible: true
     property color fontBtnColor: InputStyle.fontColorBright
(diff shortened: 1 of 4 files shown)


app/qml/PanelHeader.qml
```diff
@@ -69,7 +69,7 @@ Rectangle {
             anchors.bottom: parent.bottom
             anchors.top: parent.top
             anchors.leftMargin: backIconVisible ? header.rowHeight/4 : 0
-            visible: ((title.contentWidth + backButton.width * 2) > header.width) || !backTextVisible ? false : true
+            visible: backTextVisible
```
sklencar, line 72: This is something that could look interesting. 
Since we use the text (namely "Cancel")  only for FeaturePanel, but we need the text there because the arrow icon is missing, I just get rid of previous condition which suppose to deal with long text.
However, _Cancel_ still might be a long text in some translations, which will probably end up with clash with a title.

tomasMizera (approved):
Looks good.

Only thing I am thinking about is whether it would not be nicer if the back icon was orange (just as `cancel` text in feature form). What do you think @wonder-sk? It is not in the proposed design. But even if, it can wait. +1 for me to merge

PR #1215, opened 2021-03-02
Fix sed error when running update_all_versions.bash

wonder-sk:
It looks like `sed` on linux does not like the space between `-i` and its value:
```
sed: -e expression #1, char 1: unknown command: `.'
```
This fixes it. @PeterPetrik can you check if the script still works for you on :apple: ?


PeterPetrik:
works thanks

PR #1217, opened 2021-03-03
Feature form tab styling

### LINKED ISSUE #1151: Nicer tab widget
Instead of this

image

Let's have this

image

Notes:
- we should keep in mind that people may have end up with many tabs, or long tab names - there should be a way to scroll through them
- probably it does not need to be completely different tab widget, we just need to restyle it, but keep the behavior (?)
- the same widget could be used also to switch between different tabs in projects list

sklencar:
* border tab has the same color as labels (color picker shows me some weird color from the design image on dropbox - #7A7368)
* from a design, the tab toolbar's height seems to be the same as the one header has. However, now is set to 90% of the rowHeight. This ratio is rather random though, couldn't find any reference about its height.

Generally, fonts have to be adjusted to Rado's design.

closes #1151 
IMG_863AAC526DB3-1

### DIFF
--- a/qgsquick/from_qgis/plugin/qgsquickfeatureformstyling.qml
+++ b/qgsquick/from_qgis/plugin/qgsquickfeatureformstyling.qml
@@ -74,8 +74,8 @@ QtObject {
       property int fontPixelSize: 48 * QgsQuick.Utils.dp
       property real sideMargin: 12 * QgsQuick.Utils.dp // left or right margin for a field's content
       property real outerMargin: 20 * QgsQuick.Utils.dp // left or right margin for a whole field component
-      property int fontPointSize: 16 * QgsQuick.Utils.dp
-      property int labelPointSize: 14 * QgsQuick.Utils.dp
+      property int fontPointSize: 16
+      property int labelPointSize: 14
     }
 
   property QtObject icons: QtObject {
--- a/qgsquick/from_qgis/plugin/qgsquickfeatureformstyling.qml
+++ b/qgsquick/from_qgis/plugin/qgsquickfeatureformstyling.qml
@@ -47,6 +47,8 @@ QtObject {
     property real buttonHeight: height * 0.8
     property real spacing: 0
     property int tabLabelPointSize: 14
+    property real borderWidth: 0 * QgsQuick.Utils.dp
+    property color borderColor: "#999999"
   }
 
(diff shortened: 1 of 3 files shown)


wonder-sk:
> Shall we add also opacity to inactive/disable tabs as it is in the design?

Yes please :-)

Few other bits:
- font is the same for active / inactive tabs (lato bold)
- there's a darker line on top and bottom of the tab widget
- a bit smaller vertical height of the tab widget (like in the design)

sklencar:
Other bits resolved + updated screenshot

qgsquick/from_qgis/plugin/qgsquickfeatureformstyling.qml
```diff
@@ -74,8 +76,8 @@ QtObject {
       property int fontPixelSize: 48 * QgsQuick.Utils.dp
       property real sideMargin: 12 * QgsQuick.Utils.dp // left or right margin for a field's content
       property real outerMargin: 20 * QgsQuick.Utils.dp // left or right margin for a whole field component
-      property int fontPointSize: 16 * QgsQuick.Utils.dp
-      property int labelPointSize: 14 * QgsQuick.Utils.dp
+      property int fontPointSize: 16
+      property int labelPointSize: 14
```
tomasMizera, line 80: True, this should probably not be multiplied by Dp constant, though on Android I feel like the text size is too small then. What if we would rather use pixel sizes? They can be mutliplied and should scale fine on all platforms then.

tomasMizera (commented):
I really like the new form and all these upgrades! 

Is this still WIP PR?

qgsquick/from_qgis/plugin/qgsquickfeatureformstyling.qml
sklencar: Well, pointSize should be more reliable, but it seems to be not. I think it is not really visible from the second screenshot, but also according some forum threads, pointSize doesn't guarantee uniformity for all devices. I think PeterP was dealing with this at the very beginning and we ended up with pixel sizes and dp multiplier.

So, maybe we can ask Rado to give us also pixelSize for all text formats.

tomasMizera (approved):
Sorry, looking closer, font size seems the same as it was before. I am ok to merge :)

PR #1221, opened 2021-03-03
Accuracy warning while recording a feature

### LINKED ISSUE #1192: Show a warning when the accuracy is low
Sometimes users are not aware of the fact their GPS accuracy is low (the yellow dot by the GPS icon is not very prominent).

The suggestion is that when user is in "Record" mode AND the crosshair is centered at the GPS position AND the accuracy is not "green", we should show a warning message on top of the map in the lower part of the screen (just above the toolbar), saying something like ":exclamation: Low GPS position accuracy (XXX m)"

cc @mostlyAtNight

sklencar:
* should be visible while recording a new feature or edit geometry of existing feature
* should be visible if accuracy is none or bellow limit set in settings 
* if there is no gps position, only a toast will be shown as before
* It seems that position kit remembers last position. Therefore in the case when gps services are turned of, the gps warning banner will be shown instead of the toast.
* shown GPS accuracy - precision 1 decimal
 
Banner has still simple design with suggested colors. Would be good to ask Rado for a desing.
This PR also stressed https://github.com/lutraconsulting/input/issues/1225 even more.

closes #1192 
Screenshot 2021-03-09 at 15 35 56


saberraz:
Notes:
- [ ]  The warning will appear only if GPS is used for capturing points (i.e. when user moves the map the warning will disappear)
- [ ] Display the warning when using "streaming"
- [ ] A setting option (below Accuracy threshold) to disable the warning (warning will be **on** by default)
- [ ] Link to documentation for what these stuff mean

app/qml/main.qml
```diff
@@ -164,9 +164,17 @@ ApplicationWindow {
...
+    }
+
     function getGpsIndicatorColor() {
         if (positionKit.accuracy <= 0) return InputStyle.softRed
-        return positionKit.accuracy < __appSettings.gpsAccuracyTolerance ? InputStyle.softGreen : InputStyle.softOrange
+        return isGpsAccuracyLow() ? InputStyle.softOrange : InputStyle.softGreen
+    }
+
+    function showGpsAccuracyWarning() {
+        return (__appSettings.gpsAccuracyWarning && (digitizing.useGpsPoint || !digitizing.manualRecording) && isGpsAccuracyLow() )
```
tomasMizera, line 177: It seems that this is not working. I switched off the `gpsAccuracyWarning` in settings and still can see the message (now it actually does not disappear at all)
ezgif com-gif-maker (3)

app/qml/SettingsPanel.qml
```diff
@@ -174,6 +174,24 @@ Page {
                 }
             }
 
+            PanelItem {
+                height: settingsPanel.rowHeight
+                width: parent.width
+                text: qsTr("Accuracy warning")
```
tomasMizera, line 180: What about `Suppress accuracy warning` and keep it off by default? I think it would be more clear what it does

tomasMizera (changes requested):
The banner looks and works very well 🚀 

One more note:
 - Turn off location service, you will get 2 notifications about no GPS (banner and toast) - this could be handled differently


- Moreover, it shows `-1m` in that case

app/qml/SettingsPanel.qml
sklencar: I am generally easy. However what about `allow accuracy warning` or similar so all the logic doesn't have to be reversed.

tomasMizera: @saberraz @wonder-sk

sklencar:
@saberraz For time-being, the link in the warning takes a user to `https://help.inputapp.io/`. Is there any more specific link we want to have it there instead?

saberraz:
@sklencar could you point it to https://help.inputapp.io/howto/gps_accuracy

app/qml/SettingsPanel.qml
wonder-sk: I would go with `Show accuracy warnings` and have it on by default...

wonder-sk:
I would suggest few more changes:
- move the warning to the top of the screen (as discussed in the last call)
- add some light shadow around the whole rectangle so that it stands out from the map a bit
- use black foreground color and some brigher yellow (let's try e.g. `#ffff99`) - the green-on-orange looks a bit dull to me
- the two different sizes of text look a bit strange to me - probably better use just a single font size
- Instead of "More information [here]" it would be good to have "[Learn more]" link, probably preceded with a sentence giving a bit of clue... e.g. "Please make sure you have good view of the sky. [Learn more]"  -- I'm sure PeteW can then come up with a better copy text :-)

wonder-sk:
@sklencar ah yes can we please keep also the first sentence (`Low GPS position accuracy (XXX m)`) - and below that let's have the second sentence about the view of the sky - sorry if I was unclear with my comments.

Looks like there are also some conflicts now from the other merges - could you please rebase?

sklencar:
Added additional text to banner

PR #1224, opened 2021-03-04
Different feature form header for edit/view

### LINKED ISSUE #1219: Feature form: back button instead of Cancel when not in edit mode
The form should have the usual "back" icon in the top-left corner when editing is not enabled. Only when in editing mode, there should be Cancel/Save buttons.

sklencar:
closes #1219

### DIFF
--- a/app/qml/FeaturePanel.qml
+++ b/app/qml/FeaturePanel.qml
@@ -145,8 +145,8 @@ Drawer {
             fontBtnColor: InputStyle.highlightColor
             rowHeight: InputStyle.rowHeightHeader
             titleText: featurePanel.formState === "Edit" ? qsTr("Edit Feature") : qsTr("Feature")
-            backIconVisible: false
-            backTextVisible: true
+            backIconVisible: !saveButtonText.visible
+            backTextVisible: saveButtonText.visible
 
             onBack: {
               featurePanel.close()


app/qml/FeaturePanel.qml
```diff
@@ -145,8 +145,8 @@ Drawer {
             fontBtnColor: InputStyle.highlightColor
             rowHeight: InputStyle.rowHeightHeader
             titleText: featurePanel.formState === "Edit" ? qsTr("Edit Feature") : qsTr("Feature")
-            backIconVisible: false
-            backTextVisible: true
+            backIconVisible: !saveButtonText.visible
+            backTextVisible: saveButtonText.visible
```
tomasMizera, line 149: Maybe this is more self explanatory to me 

```suggestion
            backIconVisible: featurePanel.formState === "Edit"
            backTextVisible: featurePanel.formState !== "Edit"

```

sklencar: It is maybe a bit clearer, but the condition is a bit more complex than what is suggested:
https://github.com/lutraconsulting/input/blob/1219_fix/app/qml/FeaturePanel.qml#L158
Therefore I would keep it as it is now or make a function out of that condition, so we would have on single place

tomasMizera (approved):

PR #1232, opened 2021-03-05
Project wizard form fixes

### LINKED ISSUE #1220: Create project GUI
Minor UI issues that would be good to solve:

- at least on Android, the combo box has tiny up/down arrows (recently fixed in forms)
- the frame around project name widget, field name and type widgets is very thin, nearly not visible (at least on my phone) - maybe it would be good to make the background of the widgets gray like in forms, for more unified look?
- the "+ Add field" label does not seem to be centered within the button

Screenshot_20210303_113947_uk co lutraconsulting

sklencar:
closes #1220 

Screenshot_20210305-132058

### DIFF
--- a/app/qml/components/FieldRow.qml
+++ b/app/qml/components/FieldRow.qml
@@ -12,7 +12,7 @@ Item {
   signal removeClicked(var index)
 
   property real rowHeight: InputStyle.rowHeightHeader
-  property real iconSize: rowHeight
+  property real iconSize: rowHeight * 0.4
   property color color: InputStyle.fontColor
   property var widgetList: []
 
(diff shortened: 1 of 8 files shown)


tomasMizera (commented):
Looking good ✔️ 

Notes:
 - use `displayText` instead of `text` for attribute names (as discussed) 
 - attribute type combobox seems to have smaller width than attribute name field but not sure if we want that or not


In future it would be great to reuse widgets from qgsquick 🤔 we could separate widgets to not only be used in forms, but also on other places like here. But that is for future :)

app/qml/components/FieldRow.qml
```diff
@@ -21,22 +21,26 @@ Item {
...
       color: fieldDelegate.color
       text: AttributeName
       Layout.fillHeight: true
+      Layout.fillWidth: true
       Layout.preferredWidth: row.itemSize
 
-      onEditingFinished: AttributeName = text
+      onTextChanged: {
+        AttributeName = textField.text
+      }
```
tomasMizera: This did not quite help, I thought:

```suggestion
      onDisplayTextChanged: AttributeName = displayText
```

tomasMizera (commented):
Almost there

PR #1235, opened 2021-03-08
Reset features list model when project is reloaded

tomasMizera:
Another dangling pointer found. `mCurrentLayer` and `mCurrentFeature` not being reset when project gets reloaded.

Added handler for `QgsProject::cleared` signal, is this good approach @wonder-sk?

### DIFF
--- a/qgsquick/from_qgis/qgsquickfeatureslistmodel.cpp
+++ b/qgsquick/from_qgis/qgsquickfeatureslistmodel.cpp
@@ -21,6 +21,7 @@ QgsQuickFeaturesListModel::QgsQuickFeaturesListModel( QObject *parent )
   : QAbstractListModel( parent ),
     mCurrentLayer( nullptr )
 {
+  QObject::connect( QgsProject::instance(), &QgsProject::cleared, this, &QgsQuickFeaturesListModel::emptyData );
 }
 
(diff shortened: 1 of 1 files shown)


qgsquick/from_qgis/qgsquickfeatureslistmodel.cpp
```diff
@@ -21,6 +21,7 @@ QgsQuickFeaturesListModel::QgsQuickFeaturesListModel( QObject *parent )
   : QAbstractListModel( parent ),
     mCurrentLayer( nullptr )
 {
+  QObject::connect( QgsProject::instance(), &QgsProject::cleared, this, &QgsQuickFeaturesListModel::emptyData );
```
wonder-sk, line 24: Cool. It would be worth adding a comment stating this is to avoid dangling pointers of mCurrentLayer and mCurrentFeature (it may not be obvious)

wonder-sk, line 24:
```suggestion
  // avoid dangling pointers to mCurrentLayer/mCurrentFeature when switching projects
  QObject::connect( QgsProject::instance(), &QgsProject::cleared, this, &QgsQuickFeaturesListModel::emptyData );
```

PR #1236, opened 2021-03-08
Not visible value map/ value relation

### LINKED ISSUE #1231: Value Map field doesn't show the selected value in the form
It appears that some fields on my form show empty values but when I open the dropdown there is a selected value.
I think this only happens in the new version (0.9.0).

tomasMizera:
In special scenarios value map and value relation fields did not show text. The reason was that new icons were consuming all the place in widget.

Resolves #1231


saberraz:
Very nice...it works as expected now. Thanks

sklencar:
Nice! It could probably fix also QML warning: `QML ComboBox: Binding loop detected for property "implicitWidth"`

sklencar (approved):

PR #1237, opened 2021-03-08
Range widget ui fixes

### LINKED ISSUE #1023: Range widget UI issues
The range widgets don't look great currently...

The slider bar is too wide, strong color, and the handle is small:
image

The +/- buttons are very narrow:
image

For slider we could use an appearance similar to this (maybe with smaller handle):
image

sklencar:
Would be good to introduce a variable for custom indicator height (checkbox, slider)

Disabled form:
Screenshot 2021-03-08 at 12 34 51

Enabled form:
Screenshot 2021-03-08 at 16 10 09

closes #1023

### DIFF
--- a/qgsquick/from_qgis/plugin/editor/qgsquickrange.qml
+++ b/qgsquick/from_qgis/plugin/editor/qgsquickrange.qml
@@ -72,7 +72,7 @@ Item {
         anchors.fill: parent
         border.color: customStyle.fields.normalColor
         border.width: 1 * QgsQuick.Utils.dp
-        color: fieldItem.enabled ? "#ffffff" : customStyle.fields.backgroundColor
+        color: customStyle.fields.backgroundColor
         radius: customStyle.fields.cornerRadius
       }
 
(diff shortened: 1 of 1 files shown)


qgsquick/from_qgis/plugin/editor/qgsquickrange.qml
```diff
@@ -173,8 +203,18 @@ Item {
...
       }
+
+      handle: Rectangle {
+              x: slider.leftPadding + slider.visualPosition * (slider.availableWidth - width)
+              y: slider.topPadding + slider.availableHeight / 2 - height / 2
+              implicitWidth: slider.height * 0.5
+              implicitHeight: implicitWidth
+              radius: height * 0.5
+              color: "white"
+              border.color: customStyle.fields.backgroundColorInactive
```
tomasMizera, line 216: I would maybe remove border on the handle so that it looks similar to switch widget. What do you think?
image

tomasMizera (approved):
Good job!

- I think we could use bold on the +/- signs since they are text characters.
- For discussion about margins for +/- signs, on left side I would use the same distance as text starts and on the right side maybe the same as combobox icon? @saberraz @wonder-sk what do you think?

Other than that, I am ok to merge

wonder-sk:
> For discussion about margins for +/- signs, on left side I would use the same distance as text starts and on the right side maybe the same as combobox icon?

Sounds good to me!

> I would maybe remove border on the handle so that it looks similar to switch widget. What do you think?

Let's see - Viktor could you please try both and post a screenshot comparing with and without border?

sklencar:
With border + size is bigger as Checkbox indicator about border size:
Screenshot 2021-03-08 at 16 10 09

With border the same size as Checkbox indicator:
Screenshot 2021-03-08 at 16 10 32

Without border the same size as Checkbox indicator:
Screenshot 2021-03-08 at 16 11 01

wonder-sk:
I would go with the first option "With border + size is bigger as Checkbox indicator about border size:"

PR #1251, opened 2021-03-10
Borders of flickable tab component visible only if there is the tab

### LINKED ISSUE #1248: Extra line in feature panel
It seems that in the new version we have an extra line just under the page title which is not supposed to be there...

Screenshot_20210309_192043_uk co lutraconsulting

sklencar:
Screenshot_20210310-131539

Group and Tab panel have not the same height though (not introduced in this PR).

closes #1248

### DIFF
--- a/qgsquick/from_qgis/plugin/qgsquickfeatureform.qml
+++ b/qgsquick/from_qgis/plugin/qgsquickfeatureform.qml
@@ -425,6 +425,7 @@ Item {
       height: form.style.tabs.borderWidth
       anchors.top: flickable.top
       color: form.style.tabs.borderColor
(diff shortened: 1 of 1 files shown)


tomasMizera:
What about group header ( _Location_ here ), should there be lines too?

wonder-sk:
Yes the lines around "Location" group header are correct. But it looks like now the darker lines above/below tabs (Information, Data) are missing :-)

sklencar:
Added screenshot from a clean build. @tomasMizera can you confirm that you have no borders on tab component?

tomasMizera (approved):
@sklencar nope, after clean build I can not reproduce it

PR #1254, opened 2021-03-10
Better selection color

### LINKED ISSUE #1247: Better selection color
In some places we have a green selection background color for the active item, for example:
- projects list - with the current project
- survey layer list - with the currently selected survey layer

According to Rado's design we should use `#9abfa0` instead of `#689d71`

The current selection color is a bit intense :-)

Screenshot_20210309_193616

sklencar:
Just small note about colors which were changed according the ticket: `#9abfa0` used instead of `#689d71`. However, color that was used before was actually `#679d70`. Maybe a color picker or an image itself might cause that small difference, but maybe better to have also color palette from Rado.

Also fixed border style in Layer and Theme panel.

Color changed in:
* record layers panel
Screenshot 2021-03-10 at 12 19 18

* map theme panel
Screenshot 2021-03-10 at 12 18 58
* projects page
Screenshot 2021-03-10 at 12 18 31

closes #1247

### DIFF
--- a/app/qml/LayerList.qml
+++ b/app/qml/LayerList.qml
@@ -27,7 +27,7 @@ ListView {
       Rectangle {
         id: itemContainer
         property color primaryColor: InputStyle.clrPanelMain
-        property color secondaryColor: InputStyle.fontColorBright
+        property color secondaryColor: InputStyle.panelItemHighlight
         width: listView.cellWidth
         height: listView.cellHeight
         visible: height ? true : false
@@ -38,9 +38,12 @@ ListView {
         MouseArea {
           anchors.fill: parent
           onClicked: {
+            activeIndex = index
             listItemClicked( model.layerId )
           }
         }
+
+        Component.onCompleted: if (layerName === __activeLayer.layerName) activeIndex = index
 
         ExtendedMenuItem {
             id: item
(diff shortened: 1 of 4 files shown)


app/qml/LayerList.qml
```diff
@@ -50,7 +53,7 @@ ListView {
             imageSource: iconSource ? iconSource : ""
             overlayImage: false
             highlight: highlightingAllowed && layerId === activeLayerId
-            showBorder: highlightingAllowed ? !__appSettings.defaultLayer || activeIndex - 1 !== index : true
+            showBorder: highlightingAllowed ? !__appSettings.defaultLayer || (listView.activeIndex !== index && listView.activeIndex - 1 !== index) : true
```
tomasMizera, line 56: This is working but it is sort of hacky and we use it all over the place. 

What if we would change it like:
 - in `ExtendedMenuItem` border would be automatically invisible when item is highlighted (no need to set `showBorder` then)
 - pass `index` and `activeIndex` to `ExtendedMenuItem` and apply highlight/border logic there

What do you think @sklencar?

sklencar: I would prefer to think about redesigning it. Borders brings a bit hack indeed, because we want not only hide borders for highlighted item, but also hide borders for an item just above. Thats because otherwise there is a weird short line above highlighted rectangular which doesn't look good.

The condition can be rewritten, of course. However, I think now its quite straightforward how item's index affects borders. Ideally, would be good to get rid of the activeIndex and use only highlight property.

For instance, one of the solutions can be to extend border line to full width and have the row element similar to MerginProjectDelegate element where borders are on full-width. Then the condition can be also simplified as suggested.

tomasMizera: Good points. I agree with that. 
However, it is probably not worth to dig into it now, so I am ok to merge. Good that you found it :)

tomasMizera (approved):

Issue #638, opened 2020-05-01
Access to storage is not available on a fresh install


saberraz:
To replicate the issue:
1- Wipe your data/cache and reset all permissions (as if you have freshly installed your app)
2- Run the app: it should ask you for camera, storage, location...permissions
3- After granting permissions, try to download a project...it appears as invalid project
4- Restarting the app, will fix the issue.

saberraz:
Related to https://github.com/lutraconsulting/input/issues/953

PR #1256, opened 2021-03-11
Fix of downloading a file containing a special character

### LINKED ISSUE #1246: Unable to download a project
I have a project which I can't download in Input - `martin/pg_vt_no_pkg` - happy to share with anyone willing to have a look :-)

When I tap the download button, I only get a toast with strange content (looks like .qgz filename) and that's it - nothing else happens.

sklencar:
The reason was that `+` was converted into a space.
[https://doc.qt.io/qt-5/qurlquery.html#handling-of-spaces-and-plus](https://doc.qt.io/qt-5/qurlquery.html#handling-of-spaces-and-plus)

Simply, `item.filePath.replace("+", "%2b")` was needed. Fixed by using toPercentEncoding() on `file` query parameter.
A test added as well.

NOTES
* maybe better to check if the same issue may occur elsewhere. So far found with search query on explore page (fix included in this PR).
* better notification - server is returning just filename as error message in a case, when filepath is not existing (or filename has been decoded).
[Mergin #638](https://gitlab.cloud.lutraconsulting.co.uk/mergin/mergin/-/issues/638)

closes #1246

### DIFF
--- a/app/merginapi.cpp
+++ b/app/merginapi.cpp
@@ -82,7 +82,7 @@ void MerginApi::listProjects( const QString &searchExpression, const QString &fl
   }
   if ( !searchExpression.isEmpty() )
   {
-    query.addQueryItem( "name", searchExpression );
+    query.addQueryItem( "name", searchExpression.toUtf8().toPercentEncoding() );
   }
   if ( !flag.isEmpty() )
   {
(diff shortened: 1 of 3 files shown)


tomasMizera (approved):
Working for me on both Android and Linux, good catch!

PR #1260, opened 2021-03-12
Write image direction angle to EXIF when capturing pictures on Android

sklencar:
Writes azimuth direction as EXIF GPSImgDirection attribute.

Azimuth is read from sensors and stored with a timestamp. When the camera activity is finished, the proper azimuth value is fetched according the capture photo file timestamp. Note that we use temporary photo file created by a camera, not the copied that is used as value for a field. This suppose to ensure to get a timestamp of the event when user actually hits the capture button.

The azimuth degree takes into consideration device screen rotation and tilt. The value is written with flag `GpsImgDirection`.

LIMITATIONS:
- currently only rear camera is supported for a correct result
- may take sometime ~1s to init sensors at the beginning of capturing

This is not ideal solution, but couldn't get onCapture listener working in reasonable time.

Also contains functions to write EXIF attributes.

### DIFF
--- a/app/android/src/uk/co/lutraconsulting/CameraActivity.java
+++ b/app/android/src/uk/co/lutraconsulting/CameraActivity.java
@@ -33,6 +33,8 @@ import android.provider.MediaStore;
 import android.graphics.Bitmap;
 import android.support.v4.content.FileProvider;
 
+import androidx.exifinterface.media.ExifInterface;
+
 public class CameraActivity extends Activity{
     private static final String TAG = "Camera Activity";
     private static final int CAMERA_CODE = 102;
@@ -46,7 +48,7 @@ public class CameraActivity extends Activity{
 
         targetPath = getIntent().getExtras().getString("targetPath");
         Log.d(TAG, "targetPath: "+ targetPath);
-        
+
         Intent takePictureIntent = new Intent(MediaStore.ACTION_IMAGE_CAPTURE);
         if (takePictureIntent.resolveActivity(getPackageManager()) != null) {
 
@@ -61,7 +63,7 @@ public class CameraActivity extends Activity{
                 Uri photoURI = FileProvider.getUriForFile(this,
                                                           "uk.co.lutraconsulting.fileprovider",
                                                           photoFile);
-                
+
                 takePictureIntent.putExtra(MediaStore.EXTRA_OUTPUT, photoURI);
                 takePictureIntent.putExtra("__RESULT__", "takePictureIntent__RESULT__");
                 startActivityForResult(takePictureIntent, CAMERA_CODE);
@@ -83,7 +85,7 @@ public class CameraActivity extends Activity{
         String currentPhotoPath;
         String timeStamp = new SimpleDateFormat("yyyyMMdd_HHmmss").format(new Date());
         String imageFileName = "JPEG_" + timeStamp + "_";
-        
+
         cameraFile = File.createTempFile(imageFileName,  /* prefix */
                                          ".jpg",         /* suffix */
                                          getCacheDir()      /* directory */
@@ -94,7 +96,14 @@ public class CameraActivity extends Activity{
         Log.d(TAG, "currentPhotoPath: "+currentPhotoPath);
         return cameraFile;
     }
-    
+
+    public static String getExitInfo(String targetPath) {
+        // Create an image file name
+        String info = "trlala";
+
+        return info;
+    }
+
     @Override
     protected void onActivityResult(int requestCode, int resultCode, Intent data) {
         Log.d(TAG, "onActivityResult()");
@@ -104,25 +113,29 @@ public class CameraActivity extends Activity{
 
             Log.d(TAG, "tmp exists: "+cameraFile.exists());
             Log.d(TAG, "tmp path: "+cameraFile.getAbsolutePath());
+            Log.d(TAG, "EXIF test: "+ targetPath);
+
+
             try{
+                readExif(cameraFile.getAbsolutePath());
                 copyFile(cameraFile, new File(targetPath, cameraFile.getName()));
                 if (data == null) {
-                    data = getIntent();                
+                    data = getIntent();
                 }
                 data.putExtra("__RESULT__", cameraFile.getAbsolutePath());
-                setResult(Activity.RESULT_OK, data);   
-                
+                setResult(Activity.RESULT_OK, data);
+
             }catch(IOException e){
                 Intent intent = this.getIntent();
                 if (data == null) {
-                    data = getIntent();                
+                    data = getIntent();
                 }
                 data.putExtra("__RESULT__", e.getMessage());
-                setResult(Activity.RESULT_CANCELED, data);   
+                setResult(Activity.RESULT_CANCELED, data);
             }
-            
+
             // TODO: after copy, verify if is correctly copied and then remove the old one
-            
+
         }
         finish();
     }
@@ -150,8 +163,55 @@ public class CameraActivity extends Activity{
           if (out != null)
             out.close();
         }
-        
-    }    
+
+    }
+
+
+  private void readExif(String src) throws IOException {
+      Log.d(TAG, "readExif: "+src+" to file: ");
+      Uri uri; // the URI you've received from the other app
+      InputStream in = null;
+      try {
+          //in = getContentResolver().openInputStream(uri);
+          in = new FileInputStream(src);
+          ExifInterface exifInterface = new ExifInterface(in);
+          String date = exifInterface.getAttribute("GPSDateStamp");
+          String lat = exifInterface.getAttribute("GPSDestLatitude");
+          String lon = exifInterface.getAttribute("GPSDestLongitude");
+          String lat2 = exifInterface.getAttribute("GPSLatitude");
+          String lon2 = exifInterface.getAttribute("GPSLongitude");
+          String dir = exifInterface.getAttribute("GPSImgDirection");
+          String dir2 = exifInterface.getAttribute("M");
+          String dir3 = exifInterface.getAttribute("T");
+          String dir4 = exifInterface.getAttribute("GPSTrack");
+          String dir5 = exifInterface.getAttribute("GPSDestBearing");
+          String dir6 = exifInterface.getAttribute("GPSBearing");
+          String gpsInfo = exifInterface.getAttribute("GPSSatellites");
+
+          Log.d(TAG, "readExif LAT: " + lat2);
+          Log.d(TAG, "readExif LON: " + lon2);
+          Log.d(TAG, "readExif DIR: " + dir + "," + dir2 + "," + dir3  + "," + dir4 + "," + dir5 + "," + dir6);
+          Log.d(TAG, "readExif INFO: " + gpsInfo );
+
+
+          Log.d(TAG, "readExif REF: " + exifInterface.getAttribute("GPSImgDirectionRef") );
+          Log.d(TAG, "readExif REF: " + exifInterface.getAttribute("GPSTrackRef") );
+          Log.d(TAG, "readExif REF: " + exifInterface.getAttribute("GPSDestBearingRef") );
+
+           Log.d(TAG, "readExif: "+src+" to file: ");
+          // Now you can extract any Exif tag you want
+          // Assuming the image is a JPEG or supported raw format
+      } catch (IOException e) {
+          // Handle any errors
+      } finally {
+          if (in != null) {
+              try {
+                  in.close();
+              } catch (IOException ignored) {}
+          }
+      }
+
+  }
 
 }
 
--- a/app/android/src/uk/co/lutraconsulting/CameraActivity.java
+++ b/app/android/src/uk/co/lutraconsulting/CameraActivity.java
@@ -13,25 +13,14 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
 
-// TODO revise imports
 import android.os.Bundle;
 import android.os.Environment;
 import android.net.Uri;
-import android.app.Dialog;
-import android.app.AlertDialog;
 import android.app.Activity;
 import android.app.AlertDialog.Builder;
-import android.content.DialogInterface;
-import android.content.SharedPreferences;
 import android.content.Context;
 import android.content.Intent;
-import android.widget.LinearLayout;
-import android.widget.Button;
-//import android.view.View.OnClickListener;
-//import android.view.KeyEvent;
-import android.view.LayoutInflater;
 import android.util.Log;
-import android.graphics.drawable.ColorDrawable;
 import android.provider.MediaStore;
 import android.graphics.Bitmap;
 import android.support.v4.content.FileProvider;
@@ -46,7 +35,7 @@ import android.hardware.SensorManager;
 public class CameraActivity extends Activity implements SensorEventListener {
     private static final String TAG = "Camera Activity";
     private static final int CAMERA_CODE = 102;
-    // GPS EXIF TAGS // TODO use correct tags
+    // GPS EXIF TAGS
     private static final String GPS_BEARING_TAG = "GPSDestBearing";
     private static final String GPS_LON_TAG = "GPSDestLongitude";
     private static final String GPS_LAT_TAG = "GPSDestLatitude";
@@ -56,7 +45,8 @@ public class CameraActivity extends Activity implements SensorEventListener {
     private File cameraFile;
 
     // Sensors
-    int SENSOR_DELAY = 1000000; // time in us, so 1000000 is second // suggested: SensorManager.SENSOR_DELAY_NORMAL;
+    int SENSOR_DELAY_MS = 500;
+    int SENSOR_DELAY = SENSOR_DELAY_MS * 1000;
     private SensorManager mSensorManager;
     private Sensor mSensorAccelerometer;
     private Sensor mSensorMagnetometer;
@@ -121,12 +111,6 @@ public class CameraActivity extends Activity implements SensorEventListener {
     protected void onStart() {
         super.onStart();
 
-        // Listeners for the sensors are registered in this callback and
-        // can be unregistered in onStop().
-        //
-        // Check to ensure sensors are available before registering listeners.
-        // Both listeners are registered with a "normal" amount of delay
-        // (SENSOR_DELAY_NORMAL).
         if (mSensorAccelerometer != null) {
             mSensorManager.registerListener(this, mSensorAccelerometer,
                     SENSOR_DELAY);
@@ -135,14 +119,6 @@ public class CameraActivity extends Activity implements SensorEventListener {
             mSensorManager.registerListener(this, mSensorMagnetometer,
                     SENSOR_DELAY);
         }
-    }
-
-    @Override
-    protected void onStop() {
-        super.onStop();
-        // Unregister all sensor listeners in this callback so they don't
-        // continue to use resources when the app is stopped.
-        // mSensorManager.unregisterListener(this);
     }
 
     @Override
@@ -172,7 +148,7 @@ public class CameraActivity extends Activity implements SensorEventListener {
         boolean rotationOK = SensorManager.getRotationMatrix(rotationMatrix,
                 null, mAccelerometerData, mMagnetometerData);
 
-        // Remap the matrix based on current device/activity rotation. // TODO
+        // Remap the matrix based on current device/activity rotation.
         float[] rotationMatrixAdjusted = new float[9];
         rotationMatrixAdjusted = rotationMatrix.clone();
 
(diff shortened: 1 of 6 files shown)


saberraz:
Nice work. Just tested it and works OK on Android and outputs all the info we need. A couple of glitches:
- It took me a couple of attempts to get the photo recorded. It kept coming up with no photo
- When I tried to replace an existing photo in the form, again it took me a couple of times to get the photo replaced

sklencar:
I have also tried to write `GPSRoll`, `GPSPitch` however ExitInterface cannot write it.
They are not officially supported ([see here](https://exiftool.org/TagNames/EXIF.html)), but if a attribute definition is set in EXIF config, it seems that can be used ([ref](https://stackoverflow.com/questions/60341614/adding-to-exifdata-using-undefined-tags))

sklencar:
I have also tried to write `GPSRoll`, `GPSPitch` however ExitInterface cannot write it.
They are not officially supported ([see here](https://exiftool.org/TagNames/EXIF.html)), but if a attribute definition is set in EXIF config, it seems that can be used ([ref](https://stackoverflow.com/questions/60341614/adding-to-exifdata-using-undefined-tags))

app/androidutils.cpp
```diff
@@ -214,6 +226,8 @@ void AndroidUtils::handleActivityResult( int receiverRequestCode, int resultCode
     QString absolutePath = absolutePathJNI.toString();
 
     QString selectedImagePath = "file://" + absolutePath;
+    getExifInfo( absolutePath );
```
tomasMizera, line 229: Do we want to use it all the time? Not only if user sets it in settings or something like that?

tomasMizera, line 229: So we use this all the time for Android? What about setting it on/off in settings or something like that?

tomasMizera (commented):
Wow!

Please rebase master, I can not test it properly.

app/androidutils.cpp
sklencar: Good question. There is a space for improvements such as have a setting for it or check if the value exists by any chance. Now, the function is called unconditionally indeed.

saberraz: I would leave it to be ON by default with no setting to turn it OFF for the time being.

app/android/src/uk/co/lutraconsulting/CameraActivity.java
```diff
@@ -73,85 +100,240 @@ protected void onCreate(Bundle savedInstanceState) {
...
+        double rollDegrees = toDegrees(orientationValues[2]); // angle of rotation about the y axis
+
+        // Store sensor values
+        double normDegrees = adjustDegreesToScreenOrientation(degrees);
+        double cameraAngle = adjustCameraAngleOrientation(pitchDegrees, rollDegrees);
+
+        azimuthData.put(System.currentTimeMillis(), normDegrees);
+        pitchData.put(System.currentTimeMillis(), pitchDegrees);
+        rollData.put(System.currentTimeMillis(), rollDegrees);
+        cameraAngleData.put(System.currentTimeMillis(), cameraAngle);
```
wonder-sk, line 180: I am wondering if ~0.2 ms interval is not too high, witch unnecessarily too many numbers stored. Maybe 5ms could work too? That should be still a plenty of data to use...

app/android/src/uk/co/lutraconsulting/EXIFUtils.java
```diff
@@ -0,0 +1,126 @@
...
+
+        double precision_multiplier = Math.pow(10, DEGREE_PRECISION);
+        double rounded = Math.round(d * precision_multiplier)/precision_multiplier;
+        String[] fraction = Double.toString(rounded).split("\\.");
+
+        int denominator = (int)Math.pow(10, fraction[1].length());
+        int numerator = Integer.parseInt(fraction[0] + "" + fraction[1]);
+
+        return numerator + "/" + denominator;
+    }
```
wonder-sk, line 125: Hmm we probably don't really need to store angles as floating point values - the integer degree values should be just fine...

app/android/src/uk/co/lutraconsulting/EXIFUtils.java
```diff
@@ -0,0 +1,126 @@
...
+        writeExifAttributes(src, attributes);
+    }
+
+    public static void writeExifAttributes(String src, HashMap<String, String> attributes) {
+        Log.d(TAG, "WriteExif: " + src + " to file: ");
+        InputStream in = null;
+        try {
+            ExifInterface exifInterface = new ExifInterface(src);
+            for (String key: attributes.keySet()) {
+                exifInterface.setAttribute(key, attributes.get(key));
```
wonder-sk, line 49: Would it be useful to print keys and values that we store to the log?

app/android/src/uk/co/lutraconsulting/EXIFUtils.java
```diff
@@ -0,0 +1,126 @@
...
+        }
+        return result;
+    }
+
+    // Converts double value to rational string representation
+    private static String calculateRational(double d) {
+
+        double precision_multiplier = Math.pow(10, DEGREE_PRECISION);
+        double rounded = Math.round(d * precision_multiplier)/precision_multiplier;
+        String[] fraction = Double.toString(rounded).split("\\.");
```
wonder-sk, line 119: is toString() guaranteed to include a dot?

app/android/src/uk/co/lutraconsulting/EXIFUtils.java
```diff
@@ -0,0 +1,126 @@
...
+    // GPS EXIF TAGS
+    private static final String GPS_BEARING_TAG = "GPSDestBearing";
+    private static final String GPS_BEARING_REF_TAG = "GPSDestBearingRef";
+    private static final String GPS_LON_TAG = "GPSLongitude";
+    private static final String GPS_LAT_TAG = "GPSLatitude";
+    private static final String GPS_DATE_TAG = "GPSDateStamp";
+    // Experimental TAGS
+    private static final String CAMERA_ELEVATION_TAG = "CameraElevationAngle";
+    private static final String GPS_ROLL_TAG = "GPSRoll";
+    private static final String GPS_PITCH_TAG = "GPSPitch";
```
wonder-sk, line 31: Can we have a brief description of these tags and their expected values, and what does it mean a tag is experimental?

app/android/src/uk/co/lutraconsulting/EXIFUtils.java
```diff
@@ -0,0 +1,126 @@
...
+
+    public static void writeExifGpsDirection(String src, double exifOrientation, double cameraAngle, double pitch, double roll) {
+        HashMap<String, String> attributes = new HashMap<String, String>();
+        attributes.put(CAMERA_ELEVATION_TAG, calculateRational(cameraAngle));
+        attributes.put(GPS_PITCH_TAG, calculateRational(pitch));
+        attributes.put(GPS_ROLL_TAG, calculateRational(roll));
+        attributes.put(GPS_BEARING_TAG, calculateRational(exifOrientation));
+        attributes.put(GPS_BEARING_REF_TAG, "M");
```
wonder-sk, line 39: What does it mean to have "M" here?

app/android/src/uk/co/lutraconsulting/CameraActivity.java
```diff
@@ -73,85 +100,240 @@ protected void onCreate(Bundle savedInstanceState) {
...
+
+        // angle in degree [0 - 360] degree
+        double degrees = toDegrees(orientationValues[0]); // angle of rotation about the -z axis
+        double pitchDegrees = toDegrees(orientationValues[1]); // angle of rotation about the x axis
+        double rollDegrees = toDegrees(orientationValues[2]); // angle of rotation about the y axis
+
+        // Store sensor values
+        double normDegrees = adjustDegreesToScreenOrientation(degrees);
+        double cameraAngle = adjustCameraAngleOrientation(pitchDegrees, rollDegrees);
+
```
wonder-sk, line 176: What if I would use a front-facing vs back-facing camera - is that handled somehow?

app/android/src/uk/co/lutraconsulting/CameraActivity.java
```diff
@@ -73,85 +100,240 @@ protected void onCreate(Bundle savedInstanceState) {
...
+                null, mAccelerometerData, mMagnetometerData);
+
+        // Remap the matrix based on current device/activity rotation.
+        float[] rotationMatrixAdjusted = new float[9];
+        rotationMatrixAdjusted = rotationMatrix.clone();
+
+        // Get the orientation of the device (azimuth, pitch, roll) based
+        // on the rotation matrix. Output units are radians.
+        float orientationValues[] = new float[3];
+        if (rotationOK) {
```
wonder-sk, line 158: What if rotation is not OK? can it happen, and if yes, shouldn't we just skip everything else to make sure we don't use uninitialized values?

app/android/src/uk/co/lutraconsulting/CameraActivity.java
```diff
@@ -73,85 +100,240 @@ protected void onCreate(Bundle savedInstanceState) {
...
+        super.onStart();
+
+        if (mSensorAccelerometer != null) {
+            mSensorManager.registerListener(this, mSensorAccelerometer,
+                    SENSOR_DELAY);
+        }
+        if (mSensorMagnetometer != null) {
+            mSensorManager.registerListener(this, mSensorMagnetometer,
+                    SENSOR_DELAY);
+        }
```
wonder-sk, line 121: Shall we log if accelerometer or magnetometer are not available?

app/android/src/uk/co/lutraconsulting/CameraActivity.java
```diff
@@ -8,45 +8,72 @@
...
         Log.d(TAG, "onCreate()");
         super.onCreate(savedInstanceState);
 
+        // Init sensors
+        mSensorManager = (SensorManager) getSystemService(
+                Context.SENSOR_SERVICE);
+        mSensorAccelerometer = mSensorManager.getDefaultSensor(
+                Sensor.TYPE_ACCELEROMETER);
+        mSensorMagnetometer = mSensorManager.getDefaultSensor(
+                Sensor.TYPE_MAGNETIC_FIELD);
```
wonder-sk, line 72: Are we guaranteed to have permissions to these sensors?

app/android/src/uk/co/lutraconsulting/CameraActivity.java
```diff
@@ -8,45 +8,72 @@
...
+    int SENSOR_DELAY = SensorManager.SENSOR_DELAY_NORMAL; //SENSOR_DELAY_MS * 1000;
+    private SensorManager mSensorManager;
+    private Sensor mSensorAccelerometer;
+    private Sensor mSensorMagnetometer;
+
+    // Current data from accelerometer & magnetometer.  The arrays hold values
+    // for X, Y, and Z.
+    private float[] mAccelerometerData = new float[3];
+    private float[] mMagnetometerData = new float[3];
+    // stores time -> azimuth (degree) for a whole run of the activity
```
wonder-sk, line 55: time in milliseconds?

app/android/src/uk/co/lutraconsulting/CameraActivity.java
```diff
@@ -8,45 +8,72 @@
...
+    private HashMap<Long, Double> azimuthData = new HashMap<Long, Double>();
+    private HashMap<Long, Double> pitchData = new HashMap<Long, Double>();
+    private HashMap<Long, Double> rollData = new HashMap<Long, Double>();
+    private HashMap<Long, Double> cameraAngleData = new HashMap<Long, Double>();
```
wonder-sk, line 59: Using "just" integers for angles should be fine...

app/android/src/uk/co/lutraconsulting/CameraActivity.java
```diff
@@ -73,85 +100,240 @@ protected void onCreate(Bundle savedInstanceState) {
...
+        float roll = orientationValues[2];
+
+        // angle in degree [0 - 360] degree
+        double degrees = toDegrees(orientationValues[0]); // angle of rotation about the -z axis
+        double pitchDegrees = toDegrees(orientationValues[1]); // angle of rotation about the x axis
+        double rollDegrees = toDegrees(orientationValues[2]); // angle of rotation about the y axis
+
+        // Store sensor values
+        double normDegrees = adjustDegreesToScreenOrientation(degrees);
+        double cameraAngle = adjustCameraAngleOrientation(pitchDegrees, rollDegrees);
```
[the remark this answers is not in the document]
sklencar: `adjustDegreesToScreenOrientation` suppose to correct azimuth degree for all device rotations. I assume taking a photo on regular portrait or landscape should give us the same azimuth values.

I will tried to demonstrate it in comments

tomasMizera (approved):
Working good on my Android. 

I can see EXIF location and `GPSDestBearing` tags in taken pictures. (`GPSDestBearing` is there even when I turn off location, great!). I also received a toast when location was turned off and tried to take a photo saying "Turn on your location service in order to capture location data"
I also compared `GPSDestBearing` values with different compass app and values seems to be correct.

It would be good to add manual test.

Since Martin did a good code review, I am approving.
iOS build is failing due to an old app version - nothing with functionality.

sklencar:
On handling rear/back camera:
Camera indent doesn't have information which camera was used. Also, there seems to be no tag in EXIF to see which camera was used. However, camera model and other details can be read from EXIF and can be possibly paired with info  from android's `CameraCharacteristics`.

Note that while checking EXIF on photos taken with Samsung tablet, even `Camera Model Name` is the same for both rear and back camera, so also it is not so straightforward.

Another option is to use custom Camera activity and not not use build-in MediaStore.ACTION_IMAGE_CAPTURE.

app/android/src/uk/co/lutraconsulting/EXIFUtils.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.HashMap;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+import android.util.Log;
+
+public class EXIFUtils
+{
+    private static final String TAG = "EXIF Utils";
+    private static final int DEGREE_PRECISION = 2;
```
wonder-sk, line 21: unused?

app/android/src/uk/co/lutraconsulting/EXIFUtils.java
```diff
@@ -0,0 +1,102 @@
...
+    // GPS EXIF TAGS
+    private static final String GPS_BEARING_TAG = "GPSDestBearing";
+    private static final String GPS_BEARING_REF_TAG = "GPSDestBearingRef";
+    private static final String GPS_LON_TAG = "GPSLongitude";
+    private static final String GPS_LAT_TAG = "GPSLatitude";
+    private static final String GPS_DATE_TAG = "GPSDateStamp";
+
+    public static void writeExifGpsDirection(String src, int bearing, String bearing_ref) {
```
wonder-sk, line 29: The bearing argument - is it in degrees? 0-359 ? 0=North, 90=East, ...?

app/androidutils.cpp
```diff
@@ -73,6 +73,22 @@ bool AndroidUtils::checkAndAcquirePermissions( const QString &permissionString )
   return true;
 }
 
+QString AndroidUtils::getExifInfo( const QString &filePath )
+{
+#ifdef ANDROID
+  QAndroidJniObject jFilePath = QAndroidJniObject::fromString( filePath );
+  QAndroidJniObject jTag = QAndroidJniObject::fromString( "GPSDestBearing" ); // TODO
```
wonder-sk, line 80: todo

app/androidutils.h
```diff
@@ -34,6 +34,9 @@ class AndroidUtils: public QObject
     static void requirePermissions();
     static bool checkAndAcquirePermissions( const QString &permissionString );
 
+    //! Calls getEXIFAttribute from Java
+    static QString getExifInfo( const QString &filePath );
```
wonder-sk, line 38: What does this function really do? Shouldn't we have also configurable attribute name here?

app/androidutils.cpp
```diff
@@ -226,6 +242,8 @@ void AndroidUtils::handleActivityResult( int receiverRequestCode, int resultCode
     QString absolutePath = absolutePathJNI.toString();
 
     QString selectedImagePath = "file://" + absolutePath;
+    getExifInfo( absolutePath );
```
wonder-sk, line 245: Do we need this call here?

app/android/src/uk/co/lutraconsulting/OrientationSensor.java
```diff
@@ -0,0 +1,195 @@
...
+/**
+ * Sensor reads data from gravity and magnetic sensor and calculates azimuth. It is converted to degree and the value is
+ * stored with DATA_FREQUENCY in a map. The calculation takes into account device rotation (screen orientation and tilt).
+ * Implementation taken from:
+ * https://stackoverflow.com/questions/16317599/android-compass-that-can-compensate-for-tilt-and-pitch
+ */
+public class OrientationSensor implements SensorEventListener {
+
+    public final static int SENSOR_UNAVAILABLE = -1;
+    public final static int DATA_FREQUENCY = 500; // in ms
```
wonder-sk, line 22: 500 milliseconds is maybe too high :-) how about 100ms?

app/android/src/uk/co/lutraconsulting/CameraActivity.java
```diff
@@ -83,75 +85,91 @@ private File createImageFile(String targetPath) throws IOException {
...
-            
+
             // TODO: after copy, verify if is correctly copied and then remove the old one
-            
         }
         finish();
     }
 
+    private void extendGPSExifData(long captureTime) {
+        int bearing = getValueByTime(orientationSensor.m_azimuth_data, captureTime);
```
wonder-sk, line 136: what if the sensors are not available for some reason - we should detect it here and not write any direction info as it may be confusing

wonder-sk (commented):
I have also seen `GPSImgDirection` tag somewhere in addition to `GPSDestBearing` - what is the difference and should we use one over the other? For example `QgsImportPhotosAlgorithm` in QGIS only seems to use `GPSImgDirection` ...

sklencar:
So according [this](https://www.w3.org/2003/12/exif/),
`gpsDestBearing` - The bearing to the destination point. The range of values is from 0.00 to 359.99.
`gpsImgDirection` - The direction of the image when it was captured. The range of values is from 0.00 to 359.99.
For both there is also `*_Ref` attribute - 'T' denotes true direction and 'M' is magnetic direction.

From comments above, its better to use gpsImgDirection, especially when QGIS can work with it

app/android/src/uk/co/lutraconsulting/EXIFUtils.java
```diff
@@ -18,21 +18,35 @@
...
-        attributes.put(GPS_BEARING_REF_TAG, bearing_ref);
-        writeExifAttributes(src, attributes);
+    /**
+     * Writes EXIF data related to bering.
+     * @param src Absolute path of a file
+     * @param direction [0-359] degrees (0=North, 90=East, ...)
+     * @param direction_ref Either "M" as Magnetic North or "T" as True North
+     */
+    public static void writeExifGpsDirection(String src, int direction, String direction_ref) {
+        if (direction >= 0 || direction <= 359) {
```
wonder-sk, line 35: this will be always true

app/android/src/uk/co/lutraconsulting/EXIFUtils.java
```diff
@@ -18,21 +18,35 @@
...
     private static final String GPS_LAT_TAG = "GPSLatitude";
     private static final String GPS_DATE_TAG = "GPSDateStamp";
 
-    public static void writeExifGpsDirection(String src, int bearing, String bearing_ref) {
-        HashMap<String, String> attributes = new HashMap<String, String>();
-        attributes.put(GPS_BEARING_TAG, bearing + "/1"); // has to be in rational format
-        attributes.put(GPS_BEARING_REF_TAG, bearing_ref);
-        writeExifAttributes(src, attributes);
+    /**
+     * Writes EXIF data related to bering.
```
wonder-sk, line 29: Suggested edit: Writes image direction to EXIF metadata

app/android/src/uk/co/lutraconsulting/EXIFUtils.java
```diff
@@ -18,21 +18,35 @@
...
+     * @param direction_ref Either "M" as Magnetic North or "T" as True North
+     */
+    public static void writeExifGpsDirection(String src, int direction, String direction_ref) {
+        if (direction >= 0 || direction <= 359) {
+            HashMap<String, String> attributes = new HashMap<String, String>();
+            attributes.put(GPS_DIRECTION_TAG, direction + "/1"); // has to be in rational format
+            attributes.put(GPS_DIRECTION_REF_TAG, direction_ref);
+            writeExifAttributes(src, attributes);
+        } else {
+            Log.d(TAG, "Skipped writing bearing (" + direction + ") - it is out of the range");
```
wonder-sk, line 41: bearing -> image direction ?

wonder-sk (approved):

PR #1274, opened 2021-03-17
Add vertical space before group header

### LINKED ISSUE #1230: Missing vertical space before group title
Before "aldersvariation" there should be vertical space:

image

tomasMizera:
Added vertical space before group header. It is now wrapped in container that has `topMargin` same as fieldName.

Closes #1230

### DIFF
--- a/app/qml/FeaturePanel.qml
+++ b/app/qml/FeaturePanel.qml
@@ -221,7 +221,7 @@ Drawer {
                   property real bottomMargin: 1 * QgsQuick.Utils.dp
                   property real height: 64 * QgsQuick.Utils.dp
                   property color fontColor: InputStyle.fontColor
-                  property int spacing: 0
+                  property int spacing: InputStyle.formSpacing
                   property int fontPixelSize: 24 * QgsQuick.Utils.dp
                 }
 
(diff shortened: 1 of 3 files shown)


sklencar:
Nice and easy. 
However, found another issue with groups and master behaves differently than this PR - so, while having following definition:

Screenshot 2021-03-19 at 13 17 57

On current master, it looks like this:
Screenshot 2021-03-19 at 13 34 36
Whereas on this PR's branch, it looks like this:
Screenshot 2021-03-19 at 13 17 28

Note that field `yr2008` from group `sub_yrs2` is not visible since its hidden by new empty group header. Its probably on discussion if we want to and how to distinguish group items from non-group items.

sklencar (commented):
PR fixes problem by wrapping the whole group. However, it seems that it introduces (or exposes) another problem - see my another comment

tomasMizera:
Sure, I will deal with that, good that you found it!

sklencar:
Previous issues were fixed, great! However found another issue related to an extra margin for a group component - a list item is visible through the extra margin. See on a screen below:

Screenshot 2021-03-22 at 09 54 34

tomasMizera:
Problem resolved. @sklencar I have introduced spacing property back to this listview and subtracted `fieldItemLabel topMargin` by the size of spacing so that form looks the same, but the space is now in `spacing` -> so that it automatically adds space before section header. Section header contains extra space after itself.

Updated preview in description

sklencar (approved):

PR #1292, opened 2021-03-30
Applies default values on update

sklencar:
This fixes updating attributes for fields with default values and when`Apply default value on update` checked.
A new function `QgsQuickAttributeModel::updateDefaultValuesAttributes` is introduced which is called after an attribute has changed.

It is similar to `QgsQuickAttributeModel::resetAttributes` which seems to not be (explicitly) called. I think the idea was to use it in similar manner, however as description says, it should reset all attributes which is not the case now. Since it not used, I suggest to delete it. If not, it would be good to refactor it with `QgsQuickAttributeModel::updateDefaultValuesAttributes` a bit since it also evaluates expression for fields. So far, I kept it untouched before we asses it in the review.

NOTES:
* If a field has such default value definition, it is always updated - even last known value setting is ignored.
* default values are applied when a user changes attribute from a form - so when `valueChanged` signal for each widget in QML is sent. In case of a textWidget, valueChanged is called after typing a single symbol. 

UPDATE:
* Fixed binding warnings (hope all of them)
* Fixed behaviour - now its allowed to edit a field which has default value definition. 
    - Similarly to QGIS, when a new feature is created and field (A) with default value is updated, its allowed to edit it. When a feature is saved, the field (A) has the same value as user type in the field (A).
    - Similarly to QGIS, when an existing feature is edited, field (A) with default value is updated. User can edit also value for field (A), but eventually when a feature is saved and form closes, field (A) value is overwritten with its default value definition.
* Changed behaviour for TextField widget - to have the same behaviour as it is in QGIS for a case, when a user deletes text for a field (B), but field (A) uses value of field (B) for its default value definition. When a user does so in QGIS, value for the field (A) is calculated with an empty string instead of NULL how it was before in Input. Note that most likely NULL values for other widgets may not be handled properly.
    
It seems that there could be multiple use-cases for default values. Would be good to create a complex table/document with expected behaviour for forms and default values - we can copy it from QGIS docs if it does exist. According to the document, we should create tests to cover the cases and avoid possible regression. Also we will see what is still not unsupported. 

Feel free to discuss on behaviours mentioned in NOTES.

IOS build failed due to version.
Purchasing test failed with on test.dev instance (as well as master does).
Added also manual tests: https://github.com/lutraconsulting/input-manual-tests/issues/6
Currently only added a new test for `Forms` tests - we could possibly have an extra test-case for all types of use-cases for default values and last known value. 
 
https://user-images.githubusercontent.com/6735606/113023590-19d58180-9186-11eb-81bf-93c54d6fe9f9.mp4

### DIFF
--- a/qgsquick/from_qgis/attributes/qgsquickattributemodel.cpp
+++ b/qgsquick/from_qgis/attributes/qgsquickattributemodel.cpp
@@ -336,6 +336,47 @@ void QgsQuickAttributeModel::resetAttributes()
   endResetModel();
 }
 
+void QgsQuickAttributeModel::updateDefaultValuesAttributes()
+{
+  if ( !mFeatureLayerPair.layer() )
+    return;
+
+  QgsExpressionContext expressionContext = mFeatureLayerPair.layer()->createExpressionContext();
+  expressionContext.setFeature( mFeatureLayerPair.feature() );
+  QgsFields fields = mFeatureLayerPair.layer()->fields();
+
+  beginResetModel();
+  for ( int i = 0; i < fields.count(); ++i )
+  {
+    QgsDefaultValue defaultDefinition = fields.at( i ).defaultValueDefinition();
+    if ( !defaultDefinition.expression().isEmpty() && defaultDefinition.applyOnUpdate() )
+    {
+      QgsExpression exp( fields.at( i ).defaultValueDefinition().expression() );
+      exp.prepare( &expressionContext );
+      if ( exp.hasParserError() )
+        QgsMessageLog::logMessage( tr( "Default value expression for %1:%2 has parser error: %3" ).arg(
+                                     mFeatureLayerPair.layer()->name(),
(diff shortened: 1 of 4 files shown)


saberraz:
I tested the APK from the PR. This introduces a regression: when a field is set as default using an expression, the values entered by user gets wiped when changing fields.

I have shared saber\default_update with you for investigation.

saberraz:
:clap:  nice work. I have tested it on a couple of projects and it works as expected.

tomasMizera (approved):
Yes, it is now working as in QGIS, good job 🚀 

Maybe it could be good to create a ticket with the bug we discussed (if it is a bug) - when user open created feature, edits anything and saves, values in all fields that have `default value on update` are set to default (no matter if user entered something or not). I understand that it is a bug coming from QGIS, but just to not forget about it (as fix in QGIS might not fix it also automatically for Input). Maybe create ticket for QGIS too if there is not one.

PR #1283, opened 2021-03-22
Android EXIF reader

sklencar:
A new expression function introduce to read EXIF data. Currently only for Android.

Sample project: `lutraconsulting/test_exif` (should have public access on Mergin)
Supported functions: 

- `read_exif('<ABSOLUTE_PATH_TO_IMAGE>', '<EXIF_TAG_STRING>')`
-  `read_exif_img_direction('<ABSOLUTE_PATH_TO_IMAGE>')`
-  `read_exif_latitude('<ABSOLUTE_PATH_TO_IMAGE>')`
-  `read_exif_longitude('<ABSOLUTE_PATH_TO_IMAGE>')`

In case of invalid path, invalid tag or missing value, empty value is returned.
It calls java function from EXIFUtils and formats the result.

Manual test added to https://github.com/lutraconsulting/input-manual-tests/issues/5 

Screenshot_20210323-091027

PR #1298, opened 2021-04-06
Read EXIF for iOS gallery images

sklencar:
EXIF Implementation for iOS using native metadata reader. Currently functional only for images selected from a gallery.

The way how a captured image is saved has been changed. Before, we only fetched image data and wrote to a disk using QImageWriter -  the image was rather raw and contains no other data. That was done for both camera and gallery use cases.

**Image from gallery**
When an image is captured with Camera app and selected from Input through gallery picker, the selected image comes with all metadata including GPS and can be easily read. The copying image into Input's data folder is now done within the activity. 

TODO **Image from Camera**
Captured image is missing some metadata (including GPS) and also URL, so we don't have access to a file, but only to Image data. Therefore, the Location info has to be saved when an image is captured and write GPS data into the image metadata. https://github.com/lutraconsulting/input/issues/1305

NOTE: Need to be rebased after https://github.com/lutraconsulting/input/pull/1283 is merged.

https://user-images.githubusercontent.com/6735606/113991134-7570d080-9852-11eb-8c05-940cd3a0c026.MP4


app/inputexpressionfunctions.cpp
```diff
@@ -38,38 +40,47 @@ QVariant ReadExifImgDirection::func( const QVariantList &values, const QgsExpres
     return QVariant();
 
   return QVariant( resultDouble );
+#elif defined( Q_OS_IOS )
+  QString  result = IOSImagePicker::readExif( filepath, GPS_DIRECTION_TAG );
```
PeterPetrik, line 44: why it is called getExitInfo on android and `readExif `  on iOS? also it should be part of `IosUtils` so it has some consistency

app/ios/iosimagepicker.cpp
```diff
@@ -36,6 +36,16 @@ void IOSImagePicker::showImagePicker( int sourceType, const QString  &targetDir
...
+{
+#ifdef Q_OS_IOS
+  return readExifDirect( filepath, tag );
+#else
+  Q_UNUSED( filepath )
+  Q_UNUSED( tag )
+#endif
+}
+
 QString IOSImagePicker::targetDir() const
```
PeterPetrik, line 49: why this function is changed?

app/ios/iosimagepicker.mm
```diff
@@ -28,3 +27,11 @@
   IOSInterface *obj = [[IOSInterface alloc]init];
   [obj showImagePicker:sourceType:handler];
 }
+
+QString IOSImagePicker::readExifDirect( const QString &filepath, const QString &tag )
+{
+  // TODO make static call
```
PeterPetrik, line 33: todo?

app/ios/iosinterface.mm
```diff
@@ -80,6 +89,53 @@ static QImage fromUIImage( UIImage *image )
...
+
+static NSObject *readExifAttribute( NSString *imagePath, NSString *tag )
+{
+  NSData *data = [NSData dataWithContentsOfFile:imagePath];
+  CGImageSourceRef source = CGImageSourceCreateWithData( ( __bridge CFDataRef )data, NULL );
+  NSDictionary *metadata = [( NSDictionary * )CGImageSourceCopyPropertiesAtIndex( source, 0, NULL )autorelease];
+
+  // Modify EXIF tag for GPS group
+  // TODO more generic handle of EXIF (tagskCGImageProperty*)
+  NSString *newTag = [tag stringByReplacingOccurrencesOfString: @"GPS" withString:@""];
```
PeterPetrik, line 101: should be done in IosUtils::getExitInfo

app/ios/iosinterface.mm
```diff
@@ -80,6 +89,53 @@ static QImage fromUIImage( UIImage *image )
...
+    return nil;
+  }
+
+  CFRelease( source );
+
+  NSObject *rawValue = [GPSDictionary objectForKey:( NSObject * )newTag];
+  return rawValue;
+}
+
+static bool copyFile( NSString *srcPath, NSString *dstPath )
```
PeterPetrik, line 115: why we cannot use the existing c++ implementation?

app/ios/iosinterface.mm
```diff
@@ -117,11 +173,39 @@ -( void )showImagePicker:( int )sourceType : ( IOSImagePicker * )handler
     static IOSViewDelegate *delegate = nullptr;
     delegate = [[IOSViewDelegate alloc] initWithHandler:handler];
 
+    [[NSNotificationCenter defaultCenter] addObserverForName:@"_UIImagePickerControllerUserDidCaptureItem" object:nil queue:nil usingBlock: ^ ( NSNotification * _Nonnull note )
+                                         {
+                                           // TODO fetch Location data when an image is captured
+                                         }];
+
```
PeterPetrik, line 180: leftover

app/ios/iosinterface.mm
```diff
@@ -117,11 +173,39 @@ -( void )showImagePicker:( int )sourceType : ( IOSImagePicker * )handler
...
+      NSString *imagePath = [targetDir stringByAppendingPathComponent:fileNameWithSuffix];
+
+      bool isCameraPhoto = picker.sourceType == UIImagePickerControllerSourceType::UIImagePickerControllerSourceTypeCamera;
+      if ( isCameraPhoto )
+      {
+        // Camera handling // TODO write location metadata
+        // now done in IOSImagePicker::onImagePickerFinished
+      }
+      else
+      {
```
PeterPetrik, line 200: leftover

app/ios/iosinterface.mm
```diff
@@ -133,6 +217,11 @@ -( void )showImagePicker:( int )sourceType : ( IOSImagePicker * )handler
         QImage image = fromUIImage( chosenImage );
         QVariantMap data;
         data["image"] = image;
+        if ( !isCameraPhoto )
+        {
+          QString imagePathData( [imagePath UTF8String] );
+          data["imagePath"] = imagePathData;
+        }
```
PeterPetrik, line 224: ?

app/ios/iosinterface.mm
```diff
@@ -167,4 +256,34 @@ -( void )showImagePicker:( int )sourceType : ( IOSImagePicker * )handler
...
+    NSString *stringResult = [( NSDecimalNumber * )result stringValue];
+    return stringResult;
+  }
+  else
+  {
+    NSString *stringResult = [NSString stringWithFormat:@"%@", result];
+    return stringResult;
+  }
+}
+
```
PeterPetrik, line 288: if various tags returns always same NS type, I would prefer to have 1 function for each tag type? since now you convert native type -> NSString and than in C++ QString -> nativetype...

app/ios/iosinterface.mm
sklencar: There is now extra handling for gallery image - the image file is copied during this action.
For camera image, its kept as it was - image data are sent as response data and saved in the callback.

app/ios/iosinterface.mm
sklencar: Id rather have separate obj-c utils function and pure cpp functions - `readExifAttribute`
returns `NSString` and its dealing with other `NS` classes. 
 Also `fromUIImage` could go to obj-c utils beside `readExifAttribute`.
What do you think?

app/ios/iosinterface.mm
sklencar: This conversion is not ideal indeed. We can have a map of `tag` -> `type` and handle result according it, but the map would be rather long and maybe not complete at some point soon.

What can be quite a good solution is that for both Android and iOS (or other platforms later on), we can omit conversion to string and directly create QVariant out of native type.

app/ios/iosimagepicker.cpp
sklencar: There is now extra handling for gallery image - the image file is copied during the imagePicker action.
For camera image, its kept as it was - image data are sent as response data and saved in this callback.

PR #1319, opened 2021-04-13
Fix of crash due to hasChanges function

### LINKED ISSUE #1301: Crash on  Samsung Galaxy A20e
There has been several occurrences of crash on SG A20e due to `QgsQuickAttributeModel::hasAnyChanges()`

*** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
```
pid: 0, tid: 0 >>> uk.co.lutraconsulting <<<

backtrace:
  #00  pc 000000000002f840  /data/app/uk.co.lutraconsulting-ddddddd==/lib/arm64/libqgis_quick_arm64-v8a.so (QgsQuickAttributeModel::hasAnyChanges()+108)
  #00  pc 0000000000046740  /data/app/uk.co.lutraconsulting-ddddddd==/lib/arm64/libqgis_quick_arm64-v8a.so
  #00  pc 00000000000469e0  /data/app/uk.co.lutraconsulting-ddddddd==/lib/arm64/libqgis_quick_arm64-v8a.so (QgsQuickAttributeModel::qt_metacall(QMetaObject::Call, int, void**)+144)
  #00  pc 00000000002cc1c4  /data/app/uk.co.lutraconsulting-ddddddd==/lib/arm64/libQt5Qml_arm64-v8a.so
```

sklencar:
Crash was caused by a null pointer of edit buffer. Therefore added some checks to avoid the situation.

Will be good to create PR to QgsQuick afterwards - note that this fix and previous changes were not added there yet at all. 

closes #1301

### DIFF
--- a/qgsquick/from_qgis/attributes/qgsquickattributemodel.cpp
+++ b/qgsquick/from_qgis/attributes/qgsquickattributemodel.cpp
@@ -406,7 +406,14 @@ void QgsQuickAttributeModel::create()
 
 bool QgsQuickAttributeModel::hasAnyChanges()
 {
-  return FID_IS_NULL( mFeatureLayerPair.feature().id() ) || mFeatureLayerPair.layer()->editBuffer()->isFeatureAttributesChanged( mFeatureLayerPair.feature().id() );
+  if ( FID_IS_NULL( mFeatureLayerPair.feature().id() ) ) return true;
+
+  if ( mFeatureLayerPair.layer() &&  mFeatureLayerPair.layer()->editBuffer() )
+  {
+    return mFeatureLayerPair.layer()->editBuffer()->isFeatureAttributesChanged( mFeatureLayerPair.feature().id() );
+  }
+
+  return false;
 }
 
 bool QgsQuickAttributeModel::commit()


qgsquick/from_qgis/attributes/qgsquickattributemodel.cpp
```diff
@@ -406,7 +406,14 @@ void QgsQuickAttributeModel::create()
 
 bool QgsQuickAttributeModel::hasAnyChanges()
 {
-  return FID_IS_NULL( mFeatureLayerPair.feature().id() ) || mFeatureLayerPair.layer()->editBuffer()->isFeatureAttributesChanged( mFeatureLayerPair.feature().id() );
+  if ( FID_IS_NULL( mFeatureLayerPair.feature().id() ) ) return true;
```
tomasMizera, line 409: Can `mFeatureLayerPair.feature()` be also null?

tomasMizera:
Please rebase this branch, it is created from master without refactoring and other stuff :)

qgsquick/from_qgis/attributes/qgsquickattributemodel.cpp
sklencar: Its not returning a pointer, but an const instance. So this should be at least safe to call as well as additional call for its id.

However, it can return "empty" feature, which should have invalid id. Maybe the question can be if we want to also check mFeatureLayerPair for validity (checking layer, feature geometry and feature.isValid) or only check feature validity. However, I think its not needed in this function.

sklencar:
Rebased with current master

tomasMizera (approved):

PR #1336, opened 2021-04-16
Multipolygon highlight crash

### LINKED ISSUE #1322: Feature information - crash when selecting multipolygons or polygons with rings
Project dracic/sumarija-novi-vinodolski
layer: ods

Saber has access to my project

Try polygons on south edge

sklencar:
The crash was caused by:
`"qAbs(m_vertices.at(i).x) < (1 << 21)" in file painting/qtriangulator.cpp`

It seems that triangulation is sensitive to extreme values in path dataset (if a vertex is way away from average coordinates of other vertices).

Extreme values for our case were obtain by wrong parsing multipolygon coordinates into an array - it was handled as single polygon feature which resulted into invalid data. Instead, each polygon data should be rather split to be in valid format.

Fix changes how data are parsed - data are in one array, where each rings starts its part with two elements: index 0 - geometry type and index 1 - no. of vertices. Those values are used as indicators that interior ring data starts from there.
This fixes the crash but rather not data format itself. 

PR has eventually been simplified to use single Shape object for rendering.

Screenshot 2021-04-16 at 18 01 40

How the feature is rendered in QGIS:
Screenshot 2021-04-16 at 17 35 03

closes #1322

### DIFF
--- a/app/qml/Highlight.qml
+++ b/app/qml/Highlight.qml
@@ -150,7 +150,7 @@ Item {
 
     let startIndex = 2
     let endIndex = polyData.length
-    elements.push( componentMoveTo.createObject( objOwner, { "x": data[ startIndex ], "y": data[ startIndex + 1 ] } ) )
+    elements.push( componentMoveTo.createObject( objOwner, { "x": polyData[ startIndex ], "y": polyData[ startIndex + 1 ] } ) )
     for ( let j = startIndex; j < endIndex; j += 2 ) {
       elements.push( componentLineTo.createObject( objOwner, { "x": polyData[ j ], "y": polyData[ j + 1 ] } ) )
     }
@@ -197,7 +197,7 @@ Item {
 
         // True if drawing as Shape with multiple ShapePaths
         // False if using single ShapePath and simple filtering data <-- resolves crash
-        let showAsMultipolygonShape = false
+        let showAsMultipolygonShape = true
 
         if (showAsMultipolygonShape) {
           let testData = prepareData(data)
--- a/app/qml/Highlight.qml
+++ b/app/qml/Highlight.qml
@@ -152,7 +152,12 @@ Item {
         elements.push( componentMoveTo.createObject( objOwner, { "x": data[ dataStartIndex ], "y": data[ dataStartIndex + 1 ] } ) )
         for ( let i = dataStartIndex + 2; i < data.length; i += 2 )
         {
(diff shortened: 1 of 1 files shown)


app/qml/Highlight.qml
```diff
@@ -152,7 +152,13 @@ Item {
         elements.push( componentMoveTo.createObject( objOwner, { "x": data[ dataStartIndex ], "y": data[ dataStartIndex + 1 ] } ) )
         for ( let i = dataStartIndex + 2; i < data.length; i += 2 )
         {
-          elements.push( componentLineTo.createObject( objOwner, { "x": data[ i ], "y": data[ i + 1 ] } ) )
+          if (data[ i ] === geometryType) {
```
wonder-sk, line 155: instead of checking if data[i] == geometryType (which is potentially fragile), we should modify the for-cycle...

For each ring we have the number of points in it (the second entry after geometry type). So let's use that count to have something like this:
```
while ( i < data.length )
{
  geomType = data[i++];
  count = data[i++];
  // ... add moveTo ...
  for ( k = 1; k < count; ++k )
  {
    // ... add lineTo ...
  }
}
```

wonder-sk (commented):
Can you please test this also with multi-linestring geometries and make sure their highlight is working fine?

And can you please check if multi-points work as well? From the look into the code, it seems I have forgotten to implement them, however extractGeometryCoordinates() should be ready for multi-points, we just need to check for more points (geometry type 0) after processing the first one... if it should be an easy fix, let's fix it on this PR, if not, let's create a ticket and we will do it later...

saberraz:
Checked with multipoint and works as expected. The only think is, the highlight marker always snaps to the first point (not a big deal), but we should consider having highlighting for all features in future.

app/qml/Highlight.qml
```diff
@@ -148,11 +148,19 @@ Item {
...
+        // Create (multi) geometry for the highlight
+        let i = 0
+        let k = 0
+        while ( i < data.length )
         {
-          elements.push( componentLineTo.createObject( objOwner, { "x": data[ i ], "y": data[ i + 1 ] } ) )
+          let geomType = data[ i++ ];
+          let pointsCount = data[ i++ ];
+          elements.push( componentMoveTo.createObject( objOwner, { "x": data[ i ], "y": data[ i + 1 ] } ) )
+          for ( k = i; k < i + pointsCount * 2; k += 2 )
```
wonder-sk, line 159: it looks like we are adding the first point with both moveTo and lineTo if we start from `k = i`

wonder-sk (approved):

Issue #1281, opened 2021-03-22
Crash when switching project


saberraz:
- Download documentation\form_cascade & lutraconsulting\test_digitizing
- Open documentation\form_cascade,  add a feature and save
- Open lutraconsulting\test_digitizing
- Add a feature (polygon)

As soon as you press done to finish with geometry and open the form, the app crashes.

Tested with 0.9.2 on Android (arm64)

saberraz:
Duplicate of #1233

saberraz:
The issue has been resurfaced again. Every time I switch projects and record the first feature, it causes a crash.

saberraz:
Note that you can switch to any project. Tested in iOS with the latest master.

Issue #1233, opened 2021-03-08
Input creates tab for each widget in form


tomasMizera:
Probably related to #1035. 

@wonder-sk finally found reproducible scenario for this bug.

To reproduce:
 - open project `lutraconsulting/test_forms`
 - open feature form
 - open different project that does not have multiple tabs

Experienced:
 - you will see tabs created for all widgets

Sometimes application crashes when this bug happens.
I could reproduce this issue on all platforms

![image](https://user-images.githubusercontent.com/22449698/110295412-5da6f200-7ff1-11eb-93ff-f1366dba0961.png)

saberraz:
This has been annoying me so much but I could not consistently reproduce the issue. I am glad you got to the bottom of it.

PR #1338, opened 2021-04-19
Do not crash when switching the projects

PeterPetrik:
related to #1281
related to #1233


saberraz:
looks good. It stopped the crash. But introduced another regression. The fields turn into group tabs:
Screenshot_20210419-135551

Issue #1289, opened 2021-03-25
Support CE


sklencar:
* support capabilities flags (core/subscription/registration)
* wrap merginApi functionality into submodules + support submodules flags (capabilities)
* support handling projects from different mergin servers
* fix
    - `login2` -> `login` + billing info request
    - registration
* update billing related UI

Issue #1347, opened 2021-04-20
Support for projects from different servers 


sklencar:
Provide a flawless handling of projects stored on different Mergin servers to avoid possible data corruption.

- add `sever` flag to project's metadata
- prevent any Mergin action for projects from "different" server
- introduce a new project status for such projects

jozef-budac:
@tomasMizera is this still valid for workspaces?

tomasMizera:
Yes, very! :)

PR #1348, opened 2021-04-20
CE support

sklencar:
Changes due to support for CE Mergin server. The idea is to use primary mergin core endpoints. Request related to subscription are sent extra. Therefore login and userInfo request were changed.

Note that there is a possibility to corrupt data by mixing projects from different servers ATM: https://github.com/lutraconsulting/input/issues/1347

**Squash and Merge** option can be ideally used.

Account page with server supporting subscription:
Screenshot 2021-04-22 at 10 06 06

Account page with server NOT supporting subscription (looks quite empty):
Screenshot 2021-04-22 at 10 08 58

related to #1289

### DIFF
--- a/app/qml/ProjectPanel.qml
+++ b/app/qml/ProjectPanel.qml
@@ -176,6 +176,8 @@ Item {
               onClicked: {
                 if (__merginApi.userAuth.hasAuthData() && __merginApi.apiVersionStatus === MerginApiStatus.OK) {
                   __merginApi.getUserInfo()
+                  if (__merginApi.apiSupportsSubscriptions)
+                    __merginApi.getSubscriptionInfo()
                   stackView.push( accountPanelComp )
                 }
                 else
@@ -434,11 +436,9 @@ Item {
           if (__merginApi.apiVersionStatus === MerginApiStatus.OK && stackView.currentItem.objectName === "authPanel") {
             if (__merginApi.userAuth.hasAuthData()) {
               refreshProjectList()
-            } else if (toolbar.highlighted !== homeBtn.text) {
-              if (stackView.currentItem.objectName !== "authPanel") {
+            } else if (stackView.currentItem.objectName !== "authPanel") {
                 stackView.push(authPanelComp, {state: "login"})
               }
-            }
           }
         }
         onAuthRequested: {
--- a/app/qml/ProjectPanel.qml
+++ b/app/qml/ProjectPanel.qml
@@ -436,9 +436,11 @@ Item {
           if (__merginApi.apiVersionStatus === MerginApiStatus.OK && stackView.currentItem.objectName === "authPanel") {
             if (__merginApi.userAuth.hasAuthData()) {
               refreshProjectList()
-            } else if (stackView.currentItem.objectName !== "authPanel") {
+            } else if (toolbar.highlighted !== homeBtn.text) {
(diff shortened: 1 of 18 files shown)


PeterPetrik:
nice, Viktor, can you please

1. add screenshot WHEN billing is on and when is OFF
2. add a manual-input test against CE mergin server?

app/qml/ProjectPanel.qml
```diff
@@ -434,11 +436,9 @@ Item {
           if (__merginApi.apiVersionStatus === MerginApiStatus.OK && stackView.currentItem.objectName === "authPanel") {
             if (__merginApi.userAuth.hasAuthData()) {
               refreshProjectList()
-            } else if (toolbar.highlighted !== homeBtn.text) {
-              if (stackView.currentItem.objectName !== "authPanel") {
+            } else if (stackView.currentItem.objectName !== "authPanel") {
                 stackView.push(authPanelComp, {state: "login"})
               }
-            }
```
PeterPetrik, line 441: what is this change related to?

core/merginapi.cpp
```diff
@@ -663,6 +664,28 @@ void MerginApi::getUserInfo( )
   connect( reply, &QNetworkReply::finished, this, &MerginApi::getUserInfoFinished );
 }
 
+void MerginApi::getSubscriptionInfo()
+{
+  if ( !apiSupportsSubscriptions() )
+  {
+    CoreUtils::log( "subscription info", QStringLiteral( "Request skipped - server doesn't support subscriptions." ) );
```
PeterPetrik, line 671: this should be just qDebug() not sure why it needs to be in the user logs

core/merginapi.cpp
```diff
@@ -823,7 +846,8 @@ void MerginApi::authorizeFinished()
     {
       QJsonObject docObj = doc.object();
       mUserAuth->setFromJson( docObj );
-      mUserInfo->setFromJson( docObj );
+
+      getUserInfo();
```
PeterPetrik, line 850: not sure about his... user info is not part of /login endpoint?

core/merginapi.cpp
```diff
@@ -2120,6 +2144,34 @@ void MerginApi::getUserInfoFinished()
...
+  Q_ASSERT( r );
+
+  if ( r->error() == QNetworkReply::NoError )
+  {
+    CoreUtils::log( "subscription info", QStringLiteral( "Success" ) );
+    QJsonDocument doc = QJsonDocument::fromJson( r->readAll() );
+    if ( doc.isObject() )
+    {
+      QJsonObject docObj = doc.object();
+      mUserInfo->setSubscriptionInfoFromJson( docObj );
```
PeterPetrik, line 2159: hmm, this is not good. we should have CLEAR separatarion what is get from which endpoint. so what you get from 
login endpoint should be 1 or 2 structs, what you get from subscripion endpoint should be separate struct. ideally named mUserSubscriptionInfo

app/qml/ProjectPanel.qml
sklencar: This is just fixed leftover from refactoring - `toolbar` component doesn't exists. I will it move to a separate PR.

PeterPetrik (approved):

sklencar:
closes #1289

PR #1350, opened 2021-04-21
Updated secret docs

sklencar:
Due to change in the path of merginsecrets.cpp

### DIFF
--- a/docs/developers/secrets.md
+++ b/docs/developers/secrets.md
@@ -16,12 +16,12 @@ manifest files.
 
 encrypt
 ```
-cd app/
+cd core/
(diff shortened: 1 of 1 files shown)


PeterPetrik (approved):

islam500:
Can not build android workflow because of the missing I dont know the password for the encrypted merginsecrets.cpp
I have the MERGINSECRETS_DECRYPT_KEY setup but dont know what to put inside.

PeterPetrik:
you do not need merginsecrets.cpp for building of Input or running it, this is secret file that is required only for registration of new user on the merginmaps.com page.

Issue #581, opened 2020-02-11
Store accuracy information as a metadata to the survey table


saberraz:
Or alternatively within an auxiliary storage?

PR #1363, opened 2021-04-27
Position variables

sklencar:
Implementation of Position variables according docs: https://github.com/lutraconsulting/input-docs/pull/31
Manual tests (C) added to: https://github.com/lutraconsulting/input-manual-tests/issues/6 

Note that direction is fetched from QGeoPositionInfo. It might be maybe better to fetch it from compass sensor(?).
Unfortunately, changes are also in attribute model (forms).

### DIFF
--- a/app/digitizingcontroller.cpp
+++ b/app/digitizingcontroller.cpp
@@ -16,6 +16,7 @@
 
 #include "qgsquickutils.h"
 #include "qgsvectorlayerutils.h"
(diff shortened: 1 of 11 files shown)


PeterPetrik:
direction should be indeed fetched from the compass. you may add a variable for the moving direction from position source

saberraz:
Resolves https://github.com/lutraconsulting/input/issues/581

PR #1368, opened 2021-04-30
Added an extra check if project status is "modified"

### LINKED ISSUE #1320: Wrong project status after first open of project
Experienced:
- Open project for the first time
- Do not change anyting and go back to home
- you will see modified sync icon!

Expected:
 - project has upToDate status (no sync icon)

----
Upon clicking on the sync icon, it tries to sync, nothing happens actually and status updates correctly

---

After little investigation, it seems that some files can have last modified datetime later than our metadata file. In project `saber/text_exif` while I was testing data.gpkg has modified time four seconds later than metadata file and that leads Input to think something has been changed ~> modified state

sklencar:
When a project with a geopackage is opened, the header of gpkg file is modified. Therefore gpkg modified timestamp is updated as well. That causes the issue that project acts like there is something to be uploaded.

closes #1320

### DIFF
--- a/core/project.cpp
+++ b/core/project.cpp
@@ -62,19 +62,22 @@ ProjectStatus::Status ProjectStatus::projectStatus( const std::shared_ptr<Projec
     return ProjectStatus::NoVersion;
   }
 
-  //
-  // TODO: this check for local modifications should be revisited
-  //
-
   // Something has locally changed after last sync with server
   QString metadataFilePath = project->local->projectDir + "/" + MerginApi::sMetadataFile;
   QDateTime lastModified = CoreUtils::getLastModifiedFileDateTime( project->local->projectDir );
-  QDateTime lastSync = QFileInfo( metadataFilePath ).lastModified();
+  QDateTime lastSync = QFileInfo( metadataFilePath ).lastModified().toUTC();
   MerginProjectMetadata meta = MerginProjectMetadata::fromCachedJson( metadataFilePath );
   int filesCount = CoreUtils::getProjectFilesCount( project->local->projectDir );
   if ( lastSync < lastModified || meta.files.count() != filesCount )
   {
-    return ProjectStatus::Modified;
+    // When GPKG is opened, its header is updated and therefore lastModified timestamp is updated as well.
+    // Double check if there is really something to upload
+    QList<MerginFile> localFiles = MerginApi::getLocalProjectFiles( project->local->projectDir + "/" );
+    MerginProjectMetadata serverProject = MerginProjectMetadata::fromCachedJson( metadataFilePath );
+    ProjectDiff diff = MerginApi::compareProjectFiles( serverProject.files, serverProject.files, localFiles, project->local->projectDir );
+
+    if ( !diff.localAdded.isEmpty() || !diff.localDeleted.isEmpty() || !diff.localUpdated.isEmpty() )
+      return ProjectStatus::Modified;
   }
 
   // Version is lower than latest one, last sync also before updated


sklencar:
freshly-downloaded vs opened gpkg diff (same file):
Screenshot 2021-04-30 at 11 43 20

It seems that Geodiff ignores these changes and returns a correct diff only with changes in data.

core/project.cpp
```diff
@@ -62,19 +62,22 @@ ProjectStatus::Status ProjectStatus::projectStatus( const std::shared_ptr<Projec
...
   if ( lastSync < lastModified || meta.files.count() != filesCount )
   {
-    return ProjectStatus::Modified;
+    // When GPKG is opened, its header is updated and therefore lastModified timestamp is updated as well.
+    // Double check if there is really something to upload
+    QList<MerginFile> localFiles = MerginApi::getLocalProjectFiles( project->local->projectDir + "/" );
+    MerginProjectMetadata serverProject = MerginProjectMetadata::fromCachedJson( metadataFilePath );
+    ProjectDiff diff = MerginApi::compareProjectFiles( serverProject.files, serverProject.files, localFiles, project->local->projectDir );
+
+    if ( !diff.localAdded.isEmpty() || !diff.localDeleted.isEmpty() || !diff.localUpdated.isEmpty() )
```
PeterPetrik, line 79: in else branch we do not return anything?

sklencar: Its not needed. The branch from line 71 covers only local changes case - we can quickly compare timestamps and check files diff afterwards only when needed. If there are no local changes in diff, we need to go further and check for server changes.

On the other hand, we need to check diff only when timestamp of gpkg file is the one, which fulfil condition on line 71 (gpkg file seems to be update after last sync, but actually gpkg data are not changed - thats the reported bug).

PeterPetrik: bool function must return something all the time?

sklencar: Indeed, it has to. However, in this case, we have several if conditions (not even if else or else) which are evaluated one of the other - if one condition is not fulfilled, app continues with others. If none of them are fulfilled, function returns "default" `up to date` status.

This fix only extend current condition for `modified` status to be sure that correct status is returned.

core/project.cpp
```diff
@@ -62,19 +62,22 @@ ProjectStatus::Status ProjectStatus::projectStatus( const std::shared_ptr<Projec
...
   MerginProjectMetadata meta = MerginProjectMetadata::fromCachedJson( metadataFilePath );
   int filesCount = CoreUtils::getProjectFilesCount( project->local->projectDir );
   if ( lastSync < lastModified || meta.files.count() != filesCount )
   {
-    return ProjectStatus::Modified;
+    // When GPKG is opened, its header is updated and therefore lastModified timestamp is updated as well.
+    // Double check if there is really something to upload
+    QList<MerginFile> localFiles = MerginApi::getLocalProjectFiles( project->local->projectDir + "/" );
+    MerginProjectMetadata serverProject = MerginProjectMetadata::fromCachedJson( metadataFilePath );
+    ProjectDiff diff = MerginApi::compareProjectFiles( serverProject.files, serverProject.files, localFiles, project->local->projectDir );
```
sklencar, line 77: What if gives you error?

sklencar: Error is handled by GeoDiffUtils.
`MerginApi::compareProjectFiles` function is using `GeodiffUtils::hasPendingChanges` function, which in case of error returns true - in-spite of fact that something went wrong, it assumes the file has changed (this is from code/comment). So in the case of error, diff may contain some local changes and therefore the returned status is `Modified`.
The same function and logic is used in MerginApi sync.

PR #1369, opened 2021-04-30
Fix of opening account panel after sign out

### LINKED ISSUE #1367: Wrong behaviour on Sign out
On sign out, Input should bring a user back to ProjectPanel ,but an empty account panel shows instead.

For more details see a video below:

https://user-images.githubusercontent.com/6735606/116540688-ca9c8100-a8ea-11eb-9653-b9ac3d3592f3.MP4

sklencar:
Fix of really annoying bug that has been introduces after CE support PR.

**Story behind:**
When a user signs out, userInfo, subscriptionInfo and authData are cleared. After CE support PR a new signal has been introduced to update user info, when subscription has changed (storage has to be updated - this extra call was needed mostly for tests). 

**The cause:**
This extra call for userInfo triggers authorization, which opens an extra Login page in the stackview. This extra page is unexpected and stackview pops out Login page instead of Account page (where user clicks on Sign out button). Therefore AccountPage was still present (and empty)

**Solution:**
Since after the sign out, all user data are cleared, there is not need to request userInfo at all. Simple condition was added to prevent it.

Another possible solution would be to use different signals for clearing user data and where data are updated. 

closes #1367

### DIFF
--- a/core/merginapi.h
+++ b/core/merginapi.h
@@ -452,7 +452,7 @@ class MerginApi: public QObject
     void pingMerginReplyFinished();
     /**
      * @brief When plan has been changed, an extra userInfo request is needed to update also storage.
-     * Calls user info only when has authData, otherwise slots catches the signal from clering user data after signing out.
+     * Calls user info only when has authData, otherwise slots catches the signal from clearing user data after signing out.
      */
     void onPlanProductIdChanged();
 
--- a/core/merginapi.h
+++ b/core/merginapi.h
@@ -450,7 +450,10 @@ class MerginApi: public QObject
     void authorizeFinished();
     void registrationFinished( const QString &username = QStringLiteral(), const QString &password = QStringLiteral() );
     void pingMerginReplyFinished();
-    //! When plan has been changed, an extra userInfo request is needed to update also storage.
+    /**
+     * @brief When plan has been changed, an extra userInfo request is needed to update also storage.
+     * Calls user info only when has authData, otherwise slots catches the signal from clering user data after signing out.
+     */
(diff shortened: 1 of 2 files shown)


core/merginapi.h
```diff
@@ -450,7 +450,10 @@ class MerginApi: public QObject
     void authorizeFinished();
     void registrationFinished( const QString &username = QStringLiteral(), const QString &password = QStringLiteral() );
     void pingMerginReplyFinished();
-    //! When plan has been changed, an extra userInfo request is needed to update also storage.
+    /**
+     * @brief When plan has been changed, an extra userInfo request is needed to update also storage.
+     * Calls user info only when has authData, otherwise slots catches the signal from clering user data after signing out.
```
PeterPetrik, line 455: clering -> typo

PeterPetrik (changes requested):
now in the "subscribtion plans" panel I do not see storage and the end date

Screen Shot 2021-05-05 at 8 49 47

sklencar:
Did you test on desktop build? Are actually purchasing plans available for regular (non-testing) desktop build? FetchPurchasingPlans has successful reply, but gets empty data for desktop.

The issue is present at least since release 0.9 for desktop builds. The issue is not present for current branch, master or 0.9.0 release on iOS.
I would create a separate ticket for it since it seems to be unrelated to changes of this PR.

PeterPetrik:
server sends the empty data?

sklencar:
Correct behaviour is that manageSubscirptionPlans triggers Merign page as it does on Android

PR #1370, opened 2021-05-03
Fixed behaviour of TextField in ProjectWizard

### LINKED ISSUE #1334: Upper case switch on automaticaly
During creation of new project and adding new field, upper case switch on keyboard is automatically turn on even after turn it off manually. This problem is not visible with project name

https://user-images.githubusercontent.com/78033756/115005892-9b2c5400-9ea8-11eb-9a80-85390ee127fb.mp4

sklencar:
I couldn't replicate the issue, but I could replicate similar problem with auto uppercase - when a user types into an Attribute field with fixed uppercase option (usually double click on shift to "lock" shift), it always switches back to lower case. It doesn't happen for a project name field.

This PR fixes the issue described above. However, the reported issue could be very related and fixed as well. Best to test.
Now the field value is bound to `text` without an extra handling for displayText.

closes #1334

### DIFF
--- a/app/qml/components/FieldRow.qml
+++ b/app/qml/components/FieldRow.qml
@@ -31,7 +31,7 @@ Item {
       Layout.preferredWidth: row.itemSize
 
       Component.onCompleted: text = AttributeName
-      onDisplayTextChanged: AttributeName = displayText
+      onTextChanged: AttributeName = text
     }
 
     ComboBox {
--- a/app/qml/components/FieldRow.qml
+++ b/app/qml/components/FieldRow.qml
@@ -26,10 +26,12 @@ Item {
     InputTextField {
       id: textField
       color: fieldDelegate.color
-      text: AttributeName
       Layout.fillHeight: true
       Layout.fillWidth: true
       Layout.preferredWidth: row.itemSize
+
+      Component.onCompleted: text = AttributeName
+      onDisplayTextChanged: AttributeName = displayText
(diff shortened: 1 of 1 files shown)


sklencar:
@jozefbudac Could you please test it and possibly confirm if it also fixed for the reported issue?
Thanks!

PeterPetrik (changes requested):
you may have fixed the uppercase bug, but now the user's custom Attribute Name is not used in the created project. since you do not assign the user input to AttributeName

app/qml/components/FieldRow.qml
```diff
@@ -26,11 +26,11 @@ Item {
     InputTextField {
       id: textField
       color: fieldDelegate.color
-      text: AttributeName
       Layout.fillHeight: true
       Layout.fillWidth: true
       Layout.preferredWidth: row.itemSize
 
+      Component.onCompleted: text = AttributeName
       onDisplayTextChanged: AttributeName = displayText
```
PeterPetrik, line 34: http://imaginativethinking.ca/bi-directional-data-binding-qt-quick/

instead of onDisplayTextChanged signal it is better to do binding  AttributeName = textField.text

sklencar: Changed for simple textChanged signal handling

jozef-budac:
Tried it with lastest APK and it works fine

PR #1371, opened 2021-05-03
Changed bg color for an invalid project in Project list

### LINKED ISSUE #1345: Invalid project on the home page appear as selected rows
Maybe we can use grey font color instead of highlighting the whole row:
File (37)

sklencar:
Only font color is greyed out.

Screenshot_20210503-112401

closes #1345

### DIFF
--- a/app/qml/components/ProjectDelegateItem.qml
+++ b/app/qml/components/ProjectDelegateItem.qml
@@ -73,6 +73,12 @@ Rectangle {
     return ""
   }
 
+  function getIconColor() {
+    if (root.highlight) return root.primaryColor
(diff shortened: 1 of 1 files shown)


app/qml/components/ProjectDelegateItem.qml
```diff
@@ -73,6 +73,12 @@ Rectangle {
     return ""
   }
 
+  function getIconColor() {
```
PeterPetrik, line 76: best to not use function, just copy paste it to those 2 lines. function does not work with binding events

PeterPetrik:
can you create a manual test for the invalid project?

PeterPetrik (approved):

PR #1378, opened 2021-05-06
Format time of sync

### LINKED ISSUE #1359: Time of sync is a bit too much
Maybe we should do it similar to Mergin (1 hour ago....etc)

sklencar:
Nice format of last update of a project - `<duration><period_name> ago`, where 

- duration is length (number) of a period
- period name - one of: minute, hour,  week, month, year

Things to consider:
(**A**) The formatting is applied for both server and local projects (items), but the logic is untouched. Currently (or after models refactoring), it could be confusing what is actually "update" time. Current logic is that if a project is (also) local, it takes last modified timestamp of projectDir, otherwise takes serverUpdated timestamp. So if there is an update on server after your local changes, a user will see only time difference after local update. I guess this can be consider as bug or feature, so its up to discussion.
(**B**) A lastModified timestamp of a projectDir is used, which could be enough for description text, but it could be incorrect. Similarly to #1320 issue, where also timestamp was not enough to get correct projectStatus - problem was related to gpkg and modification of its header without modifying data. 

Added also test, where is best to see what is supported. Note that a new test class has been introduced + new test task added to auto-tests.

Screenshot 2021-05-06 at 13 05 57

closes #1359

### DIFF
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -112,18 +112,6 @@ QString InputUtils::formatNumber( const double number, int precision )
   return QString::number( number, 'f', precision );
 }
 
-QString InputUtils::formatDuration( int lengthOfCycle, const QString &cycleName )
-{
-  if ( lengthOfCycle == 1 )
-  {
-    return QStringLiteral( "%1 %2 ago" ).arg( lengthOfCycle ).arg( cycleName );
-  }
-  else
-  {
-    return QStringLiteral( "%1 %2s ago" ).arg( lengthOfCycle ).arg( cycleName );
-  }
-}
-
 QString InputUtils::formatDateTimeDiff( const QDateTime &tMin, const QDateTime &tMax )
 {
   qint64 daysDiff = tMin.daysTo( tMax );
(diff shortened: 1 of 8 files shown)


app/inpututils.cpp
```diff
@@ -111,6 +112,70 @@ QString InputUtils::formatNumber( const double number, int precision )
...
 
+QString InputUtils::formatDuration( int lengthOfCycle, const QString &cycleName )
+{
+  if ( lengthOfCycle == 0 || lengthOfCycle == 1 )
+  {
+    return QStringLiteral( "%1 %2 ago" ).arg( lengthOfCycle ).arg( cycleName );
+  }
+  else
+  {
+    return QStringLiteral( "%1 %2s ago" ).arg( lengthOfCycle ).arg( cycleName );
```
wonder-sk, line 123: This is not very translation friendly. Also, I think the correct english version for zero is "0 hours ago", not "0 hour ago"

Qt has support for plural forms in tr() - https://doc.qt.io/qt-5/i18n-source-translation.html#handling-plurals - some languages have multiple plural forms - but not sure if we support it with transifex + we may need to add english translation (to convert generic  "%1 week(s) ago" to "%1 week ago" / "%1 weeks ago" depending on the number. So maybe as a compromise let's stick to `if (length > 1) tr( "%1 weeks ago") else tr("%1 week ago")`. Let's have multiple texts ("%1 day ago", "%1 days ago", "%1 week ago", ...) - it will give better context to translators.

app/test/testutilsfunctions.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+const int DAY_IN_SECS = 60 * 60 * 24;
+const int MONTH_IN_SECS = 60 * 60 * 24 * 31;
+
+void TestUtilsFunctions::testFormatDuration()
+{
+  QDateTime t0 = QDateTime::currentDateTime();
+
+  testFormatDuration( t0, 0, QStringLiteral( "Invalid datetime" ) );
+  testFormatDuration( t0, 1, QStringLiteral( "0 minute ago" ) );
```
wonder-sk, line 23: How about "just now" instead of "0 minute ago" ?

app/test/testutilsfunctions.cpp
```diff
@@ -0,0 +1,43 @@
...
+  testFormatDuration( t0, 14 * DAY_IN_SECS, QStringLiteral( "2 weeks ago" ) );
+  testFormatDuration( t0, MONTH_IN_SECS, QStringLiteral( "1 month ago" ) );
+  testFormatDuration( t0, 2 * MONTH_IN_SECS, QStringLiteral( "2 months ago" ) );
+  testFormatDuration( t0, 12 * MONTH_IN_SECS, QStringLiteral( "1 year ago" ) );
+  testFormatDuration( t0, 24 * MONTH_IN_SECS, QStringLiteral( "2 years ago" ) );
+}
+
+void TestUtilsFunctions::testFormatDuration( const QDateTime &t0, qint64 diffSecs, const QString &expectedResult )
+{
+  QDateTime t1 = t0.addSecs( ( -1 ) * diffSecs );
```
wonder-sk, line 40: maybe if you switch t1,t0 on the next line, we don't need to do multiply secs by -1 here

app/test/testutilsfunctions.cpp
sklencar: just pushed a fix to "a few seconds ago"

sklencar: updated to `just now`

wonder-sk (approved):

wonder-sk:
is it still wip or ready to merge?

sklencar:
Ready to be merged

Issue #1304, opened 2021-04-08
Can't enter credentials


waiotahi:
Input App 0.9.2 / screen resolution 640 x 480 /landscape
Cant enter /signin  with credentials on Login page.
The signup for free & reset links both work but nothing on SIGNUP?

waiotahi:
This is the same behavior as ticket # 1114

saberraz:
To reproduce, you can use your iOS device in Landscape mode.

waiotahi:
![Screenshot_20210518-151846](https://user-images.githubusercontent.com/53323943/118585289-ae567a80-b7ec-11eb-90e9-8d484cf44dad.png)
I finally got around to updating to 0.9.3 and testing - I still don't seem to be able to login - It will send me to reset the password or register a new account but not signin

PR #1384, opened 2021-05-07
Scrollable Login form

sklencar:
- Added ScrollView to Login form similarly as it is in Registration form.
- Fixed also `root` references to have independent Login form component.
- Set bottom padding of AuthPanel to 0 to save a bit of space for landscape mode and to prevent rendering components on bottom padding white space (kind of clipping).
    - Note that the text field for Mergin link is anchored to bottom and basically it acts like a static footer of AuthPanel - not sure if its better to have it like this or rather not. Anyway, I kept it as it was for now.
- Centered Sing up and Forgot password buttons by removing an extra Column wrapper around each of them.

### DIFF
--- a/app/qml/AuthPanel.qml
+++ b/app/qml/AuthPanel.qml
@@ -86,6 +86,7 @@ Item {
     height: root.height - header.height
     width: root.width
     y: header.height
+    bottomPadding: 0
     background: Rectangle {
       color: root.bgColor
     }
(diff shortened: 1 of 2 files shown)


PeterPetrik (approved):

PeterPetrik:
@jozef-budac can you please test?

saberraz:
Fixes https://github.com/lutraconsulting/input/issues/1304

PR #1394, opened 2021-05-11
Fixed auth failed message

### LINKED ISSUE #1379: Sign in notification are missing
When signing in, there is no notification when wrong or no login/password is used.
This steps in TC01 (input-manual-tests#2) failed:

B7. Application should stay at login page and popup with text: Invalid username or password shall be visible
B8. Wait till notification will disappear (in few seconds)
B10. Popup should inform you with text: Please enter your login details
B11. Repeat B9 and B10 with password (leave username empty)

sklencar:
Reverted code to use showMessage as in previous versions.
closes #1379

### DIFF
--- a/app/qml/main.qml
+++ b/app/qml/main.qml
@@ -706,7 +706,7 @@ ApplicationWindow {
           storageLimitDialog.open()
         }
 
-        onNotify: showDialog(msg)
+        onNotify: showMessage(message)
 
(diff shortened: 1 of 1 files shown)


PeterPetrik:
> Reverted code to use showMessage as in previous versions.

this code was not changed 6 years?

PR #1401, opened 2021-05-13
Range widget improvement

### LINKED ISSUE #996: Minimum value of range widget is not correct
Widget type: Range
Field type: double (Real)
I set minimum to 15,0 in QGIS, but in Input app it is 1,5
for report
fro report 2
for report 3

### LINKED ISSUE #1395: Not possible to add decimal numbers in form
- Try to download https://public.cloudmergin.com/projects/lutraconsulting/decimal_issue/tree in Input.
- Add a point and in the form try to add a number in **DBH** field with decimal.

sklencar:
For following field config:
precision =  _p_, step precision = _sp_, that _p_ < _sp_ (e.g _p_=0, _sp_= 1 for step=0.1)
was range widget not working properly.

Another issue was that if `Minimum` or `Maximum` in config was out of range for QML Integer, the spinbox got broken.

Since QML spinbox works with QML integers for its properties (value, from, to, stepSize), all these properties have to be multiplied. This shrinks actual range that could cause overflow issue. Best to get spinbox working with doubles/reals.
However, not sure if it should be done in scope of this PR (?).

Tested on lutraconsulting/decimal_issue and  lutraconsulting/test_decimals projects.

docs: https://github.com/lutraconsulting/input-docs/pull/35
test (E): https://github.com/lutraconsulting/input-manual-tests/issues/6

closes #1395 
closes #996

### DIFF
--- a/app/qml/editor/inputrange.qml
+++ b/app/qml/editor/inputrange.qml
@@ -64,9 +64,10 @@ Item {
       property int multiplier: {
         var m = 1;
         while (fieldItem.step * m < 1) {
-          m +=1
+          m *=10
         }
-        m
+        var precision = fieldItem.precision === 0 ? 1 : Math.pow(10, spinbox.precision)
+        Math.max(m, precision)
       }
       property int precision: fieldItem.precision
       property int intValue: fieldItem.parent.value * multiplier
@@ -149,7 +150,7 @@ Item {
                   font.pixelSize: spinbox.font.pixelSize * 2
                   font.bold: true
                   fontSizeMode: Text.Fit
-                  color: fieldItem.enabled ? customStyle.fields.fontColor : customStyle.toolbutton.backgroundColorInvalid
+                  color: fieldItem.enabled && spinbox.value > spinbox.from ? customStyle.fields.fontColor : customStyle.toolbutton.backgroundColorInvalid
                   leftPadding: customStyle.fields.sideMargin
                   horizontalAlignment: Text.AlignLeft
                   verticalAlignment: Text.AlignVCenter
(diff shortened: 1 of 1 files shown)


saberraz:
@sklencar do you think this PR also fixes https://github.com/lutraconsulting/input/issues/996?

sklencar:
@saberraz Yes, it fixes fixes #996 (added to PR's comment)

saberraz:
I am not sure how you have fixed handling of 0 precision. But can we ensure 0 is equivalent of no limit. Similar to the text with 0 length.

sklencar:
@saberraz There are two precision option. Unfortunately, I was working only with precision set in Layer properties -> Attribute form which basically sets format of number.

So regarding 0 precision set in Layer properties -> fields, I have to investigate how that precision value works there. Have you managed to change it within QGIS?

saberraz:
@sklencar no, according to sqlite/geopackge spec, there is no definition for precision.

PeterPetrik:
@sklencar can you please fix tests and rabase?

tomasMizera:
Good job, from my point of view it works well. 

There is one thing I found that I think we should address. When user edits value manually (with keyboard), he needs to click on `Done` on the keyboard, otherwise clicking on -/+ signs will erase the written text. I think using `displayText` property would help solving this issue, but I am not sure whether it works on `spinbox` the same as in `textfield`. There is the attribute though https://doc.qt.io/qt-5/qml-qtquick-controls2-spinbox.html#displayText-prop

https://user-images.githubusercontent.com/22449698/119360266-ba3db300-bcaa-11eb-989d-e42f8c9212d9.mp4

PeterPetrik:
created new issue for the interaction: https://github.com/lutraconsulting/input/issues/1422

PR #1407, opened 2021-05-17
Fixes after qgsquick refactoring

sklencar:


app/qml/editor/inputcheckbox.qml
```diff
@@ -68,7 +68,7 @@ Item {
       leftPadding: customStyle.fields.sideMargin
     }
 
-    SwitchComponent {
+    Switch {
```
PeterPetrik, line 71: is this right? switch is qt native and switch component is our???

PeterPetrik: @tomasMizera @sklencar

sklencar: Its our custom widget and should be the only usage of the component. 
It has been renamed in #1405 and later on it may bring a bit of confusion indeed.

PeterPetrik: @sklencar can you please rename it and make PR ?

Issue #1437, opened 2021-05-28
Wrong process to upload/sync project  when storage limit is reached


saberraz:
When user has a local project, which the changes exceed the storage quota, and tries to upload it the warning appears about the subscription plan, then a Mergin project is created  with nothing inside.

The project status becomes corrupted locally and user cannot upload the data even if a new plan is purchased.

saberraz:
We should do the following:
1- when user tries to sync or upload a project > Check the storage availability
1.a if enough storage is available proceed with the upload or sync
2- If enough storage is not available:
2.a. Display the storage warning
2.b. DO NOTHING: no new project creation in Mergin, or attempting to upload partial files. Simply NOTHING should be done apart from the message.

PR #1439, opened 2021-05-31
Autoremove of empty project

tomasMizera:
This is a hotfix of #1437

This is what we did when project should migrate to Mergin:
 1. Create empty Mergin project with the same name as local project
 2. Send start uploading request
 3. If the previous request failed with storage limit, we would show to user that he does not have sufficient storage left 

But the project created in step 1 remained! 

This PR removes the project from step 1. Although it would be better to not create it at all, we need a support for this on Mergin, created ticket: https://gitlab.cloud.lutraconsulting.co.uk/mergin/mergin/-/issues/703

### DIFF
--- a/app/projectsmodel.cpp
+++ b/app/projectsmodel.cpp
@@ -499,6 +499,7 @@ void ProjectsModel::onProjectDetachedFromMergin( const QString &projectFullName 
   if ( proj )
   {
     proj->mergin.reset();
+    proj->local->projectNamespace = "";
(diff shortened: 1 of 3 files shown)


sklencar:
Looks good. However, could you please add an automatic test.

tomasMizera:
After discussions I added a manual test: https://github.com/lutraconsulting/input-manual-tests/issues/37, also created a ticket to follow new Mergin's endpoint: #1440

PR #1445, opened 2021-06-01
Fixed attributePreview type enum in QML

### LINKED ISSUE #1276: HTML Tooltip doesn't work when field names contain uppercase letters
I noticed this today. HTML Tooltip doesn't work when field names contain uppercase letters.

```
Display Name: OPIS

HTML Map Tip:

# image
file:///[%@project_folder%]/[% "Foto" %]
```
Neither Display name nor image is shown.

sklencar:
Wrong enum has been used in preview panel, therefore photo preview was not working at all.
Beside it, display name work as expected and its shown correctly as title in preview Panel - tested with lower, upper case and also when field name contains some special chars (e.g !@#$,..).

Demonstration of image preview + display name as the title using fields with upper case.
Screenshot 2021-06-01 at 14 00 54

closes #1276

### DIFF
--- a/app/qml/PreviewPanel.qml
+++ b/app/qml/PreviewPanel.qml
@@ -131,7 +131,7 @@ Item {
                 }
 
                 Image {
-                    visible: controller.type == AttributePreviewController.Image
+                    visible: controller.type == AttributePreviewController.Photo
                     source: controller.photo
                     sourceSize: Qt.size(width, height)
                     fillMode: Image.PreserveAspectFit


sklencar:
Note that renaming a field in QGIS -> layer properties -> fields -> edit `name` cannot be done, if old and new field name are the same (ignoring upper/lower case). For instance, renaming `opis` -> `OPIS` can be done in fields properties, but its not commited into gpkg afterwards and it ends up with following error instead:

Screenshot 2021-06-01 at 14 07 48

PeterPetrik (approved):

PR #1472, opened 2021-06-10
Fix condition for undefined

### LINKED ISSUE #1426: Range slider wrong minimum value
Tested on project `saber/forms_value_multiple` 

Reproduce:
 - open the project
 - create new feature
 - in feature form navigate to tab `Data`
 - move the `Number of plants` slider 

Experienced on current master:
 - slider goes from -inf to 10 

Experienced on beta version from Google play (before forms and range widget refactoring):
 - slider goes from 0 to 10

tomasMizera:
Changed the check for `undefined` max/min range, there was a problem with coercion, 0 was considered false and thus range widget ignored it and worked with `inf`

Fixes #1426

### DIFF
--- a/app/qml/editor/inputrange.qml
+++ b/app/qml/editor/inputrange.qml
@@ -32,7 +32,7 @@ Item {
   property string suffix: config["Suffix"] ? config["Suffix"] : ""
 
    function getRange(rangeValue, defaultRange) {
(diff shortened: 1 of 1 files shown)


tomasMizera:
I am closing this PR, the change will be introduced in the upcoming range widget PR

PR #1484, opened 2021-06-15
fix mixed tabs & fields in form [regression 0.9.6->0.9.7]

### LINKED ISSUE #1467: Mixture of groups and single field in Drag-n-Drop designer breaks the form
If user has a drag-n-drop form designer similar to the one below (normal font: fields, bold font group), Input does not show any of the fields/groups:

- field_1
- **Group1**
- **Group2**
- field_2

See an example of the issue here:
https://public.cloudmergin.com/projects/lutraconsulting/issue_group-field-dnd/tree

sklencar:
Evaluation of the form is done in advance before creating form items. If field and container elements are mixed, then tabs are ignored and all elements are put into single root container. This should be the same behaviour as it was before (v0.9.6 before form refactoring).

For having tabs, all elements must be put into a tab.

Some use-cases may not be considered and for sure some not covered by tests. Also best to write documentation what is expected behaviour for each use-case. Shall it be added in the scope of this ticket?

Test with `documentation/form_setup` and `documentation/form_cascade` and the project from issue.
Added automatic tests.

NOTE that iOS build is failing due to version clash.

https://user-images.githubusercontent.com/6735606/122062174-73e1fc80-cdef-11eb-9c4e-05307fe131db.MP4

closes #1467

### DIFF
--- a/app/attributes/attributecontroller.cpp
+++ b/app/attributes/attributecontroller.cpp
@@ -99,6 +99,17 @@ QgsAttributeEditorContainer *AttributeController::autoLayoutTabContainer() const
   return root.release();
 }
 
+bool AttributeController::evaluateHasTabs( QgsAttributeEditorContainer *container )
+{
+  bool hasTabs = !container->children().isEmpty();
+  for ( QgsAttributeEditorElement *element : container->children() )
+  {
+    QgsAttributeEditorContainer *elemContainer = static_cast<QgsAttributeEditorContainer *>( element );
+    hasTabs = hasTabs && element->type() == QgsAttributeEditorElement::AeTypeContainer && !elemContainer->isGroupBox();
+  }
+  return hasTabs;
+}
+
 VariablesManager *AttributeController::variablesManager() const
 {
   return mVariablesManager;
@@ -290,15 +301,14 @@ void AttributeController::updateOnLayerChange()
         root->setColumnCount( 1 );
       }
 
+      mHasTabs = evaluateHasTabs( root );
       for ( QgsAttributeEditorElement *element : root->children() )
       {
         if ( element->type() == QgsAttributeEditorElement::AeTypeContainer )
         {
           QgsAttributeEditorContainer *container = static_cast<QgsAttributeEditorContainer *>( element );
-          if ( !container->isGroupBox() )
+          if ( !container->isGroupBox() && mHasTabs )
           {
-            mHasTabs = true;
-
             if ( container->columnCount() > 1 )
             {
               qDebug() << "tab " << container->name() << " in manual config has multiple columns. not supported on mobile devices!";
--- a/app/attributes/attributecontroller.cpp
+++ b/app/attributes/attributecontroller.cpp
@@ -99,15 +99,21 @@ QgsAttributeEditorContainer *AttributeController::autoLayoutTabContainer() const
   return root.release();
 }
 
-bool AttributeController::evaluateHasTabs( QgsAttributeEditorContainer *container )
+bool AttributeController::allowTabs( QgsAttributeEditorContainer *container )
 {
-  bool hasTabs = !container->children().isEmpty();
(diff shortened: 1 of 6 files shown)


app/attributes/attributecontroller.cpp
```diff
@@ -99,6 +99,17 @@ QgsAttributeEditorContainer *AttributeController::autoLayoutTabContainer() const
   return root.release();
 }
 
+bool AttributeController::evaluateHasTabs( QgsAttributeEditorContainer *container )
```
PeterPetrik, line 102: rename to "allowTabs"

app/attributes/attributecontroller.cpp
```diff
@@ -99,6 +99,17 @@ QgsAttributeEditorContainer *AttributeController::autoLayoutTabContainer() const
...
+{
+  bool hasTabs = !container->children().isEmpty();
+  for ( QgsAttributeEditorElement *element : container->children() )
+  {
+    QgsAttributeEditorContainer *elemContainer = static_cast<QgsAttributeEditorContainer *>( element );
```
PeterPetrik, line 107: you need to first check element->type() == QgsAttributeEditorElement::AeTypeContainer before doing static cast

app/attributes/attributecontroller.cpp
```diff
@@ -99,6 +99,17 @@ QgsAttributeEditorContainer *AttributeController::autoLayoutTabContainer() const
...
+    hasTabs = hasTabs && element->type() == QgsAttributeEditorElement::AeTypeContainer && !elemContainer->isGroupBox();
+  }
+  return hasTabs;
```
PeterPetrik, line 110: I would rewrite this completely like
```
for (element) {
  if element-type()== container 
  {
   static_cast
    if (!elemContainer->isGroupBox()) return true;
    
  }
}
return false;

File: app/attributes/attributecontroller.cpp [hunk 4]
```diff
@@ -290,15 +301,14 @@ void AttributeController::updateOnLayerChange()
...
       }
 
+      mHasTabs = evaluateHasTabs( root );
       for ( QgsAttributeEditorElement *element : root->children() )
       {
         if ( element->type() == QgsAttributeEditorElement::AeTypeContainer )
         {
           QgsAttributeEditorContainer *container = static_cast<QgsAttributeEditorContainer *>( element );
-          if ( !container->isGroupBox() )
+          if ( !container->isGroupBox() && mHasTabs )
```
PeterPetrik, line 310: this can be checked directly in the before the loop over elements again.

```
if (allowTabs) {
  same as before
} else 
  createTab( root );
```

app/attributes/attributecontroller.h
```diff
@@ -177,6 +177,14 @@ class  AttributeController : public QObject
     // Generates fake root tab for auto-layout
     QgsAttributeEditorContainer *autoLayoutTabContainer() const;
 
+    /**
+     * Checks if ALL first level elements of given container are type of QgsAttributeEditorElement::AeTypeContainer.
+     * Therefore if AeFields are mixed with AeContainers elements, tabs setting is ignored since embedded tabs are not supported.
+     * @param container Suppose to be the root of attributeEditor container.
+     * @return True if all container's elements are type of AeContainer.
```
PeterPetrik, line 184: this is misleading doc. Group is also AeTypeContainer.

the function should check if we are allowed/should use tab layout (=> there are tabs and all other elements are within tabs) or not (=> there are no tabs or there are elements that are not inside any tab)

PeterPetrik (changes requested):

PeterPetrik (approved):
thanks, code looks good

PeterPetrik:
@tomasMizera or @jozef-budac can you please test ?

jozefbudac:
I just verified the fix for arm64-v8a

PR #1495, opened 2021-06-16
Guess widget type for an empty field config

### LINKED ISSUE #1464: The form doesn't use checkbox date widgets with autogenerated form
When I create new project with auto-generated form, in INPUT, checkbox and Date&Time widgets (and probably also range) shows only simple field without slider, calendar or square for check.

you can check this on project tin1/test4

Screenshot_20210608-110012

sklencar:
If widget setup is empty, Input tries to guess widget automatically - note that only numbers, date/time, bool fields are straightforward - for instance, external (photo) widget cannot be guessed only according field type (any ideas how handle such case?)

Added also QgsQuick import in WizardPage (its using QgsQuick.Utils.dp)

tested with project provided within the issue.

closes #1464

### DIFF
--- a/app/attributes/attributecontroller.cpp
+++ b/app/attributes/attributecontroller.cpp
@@ -27,6 +27,7 @@
 #include "qgsattributeeditorcontainer.h"
 #include "qgsvectorlayerutils.h"
 #include "qgsmessagelog.h"
+#include "inpututils.h"
 
 AttributeController::AttributeController( QObject *parent )
   : QObject( parent )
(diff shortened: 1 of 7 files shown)


tomasMizera:
So this issue was affecting only for projects created in Input?

sklencar:
Actually a project created with the project wizard with any field has proper form. Input creates field configuration with proper widget editor setup when project is created.

The issue is present for fields which have editor widget setup type empty. @jozefbudac How was the project from issue created?
In QGIS, this case is handled by evaluating `fieldScore` for each widget type (see more in `QgsEditorWidgetFactory` and inherited classes).

jozefbudac:
From my point of view, it works well. Maybe even better than it should :) If you try tin1/test7 project, on INPUT it quest (correctly) 64 number and decimal number fields. in QGIS, this two fields look similar to text field. 

QGIS:
Screenshot at 2021-06-17 11-32-27

INPUT:
Screenshot_20210617-113314

jozefbudac:
> The issue is present for fields which have editor widget setup type empty. @jozefbudac How was the project from issue created?
I just created new project in QGIS, add a layer with all the fields and didn't touch forms at all.

tomasMizera (approved):
Looking good.

Tests are failing due to a version (iOS) and known purchasing issue that we should look at (auto test, not caused by this PR)

PeterPetrik:
@sklencar can you please rebase/resolve conflicts and merge? thanks

sklencar:
Rebased

Issue #1183, opened 2021-02-12
Handling NULL values in Range


saberraz:
If a field is set to Range, it is defaulting to 0 instead of NULL when filling the form. 
~The work-around is to use Text widget edit.~ << The workaround did not work out! Setting 0 in Input and it will appear as NULL in QGIS.

Issue #493, opened 2019-10-25
Integers in GPKG displayed as "0"


dracic:
I noticed a change since last update two days ago. Every int field in every layer in my gpkg is displayed as "0" in forms.

dracic:
![Screenshot_2019-10-25-13-36-47-407_uk co lutraconsulting](https://user-images.githubusercontent.com/4533263/67568590-eab71700-f72c-11e9-9b39-f28d53f6c3de.png)
![Screenshot_2019-10-25-13-36-52-877_uk co lutraconsulting](https://user-images.githubusercontent.com/4533263/67568591-eab71700-f72c-11e9-95f5-8954386d47ee.png)

saberraz:
Hi @dracic that looks like an upstream issue in QGIS. Do you have the same issue there too in your forms or Display tooltip?

dracic:
In qgis everything looks fine.

dracic:
Well, perhaps it has to do with the fact that QGIS automatically selects range widget for int type, not so much because of int type?

dracic:
I don't see this behavior any more in 0.6.2, so I suppose some fix got in with the new SDK....

dracic:
So the issue appeared again:

![image](https://user-images.githubusercontent.com/4533263/115039245-911e4b80-9ed0-11eb-861d-6fe1da8dbb85.png)

saberraz:
I can confirm the issue. In QGIS, it stays as NULL but in Input it is saved as 0. But still there is an issue with the Slider. In both QGIS and Input the value stays as 0.

dracic:
Also when value is not NULL Input always shows 0, and QGIS shows exact value:

![image](https://user-images.githubusercontent.com/4533263/115050859-7d78e200-9edc-11eb-980c-956e49a3a833.png)

Issue #1497, opened 2021-06-17
Range slider limit is 29, even when is set to 30


jozefbudac:
INPUT: 0.9.8 releases
OS: both iOS & Android

Steps to reproduce:
1) open project  input-testing/forms_widgets
2) add a feature to POI layer
3) in the form on tab 'Numbers' there is a Range slider, which is set to: min -10, max 30, step 3. But the slider allows do to max 29 and can't be moved further

I am not sure, but I believe this worked in previous versions, so I marked is as regression

PeterPetrik:
still in Input 2.1.1

PR #1500, opened 2021-06-18
Range widget

### LINKED ISSUE #1422: Spinbox manual edit vs +/- button interaction
There is one thing I found that I think we should address. When user edits value manually (with keyboard), he needs to click on Done on the keyboard, otherwise clicking on -/+ signs will erase the written text. I think using displayText property would help solving this issue, but I am not sure whether it works on spinbox the same as in textfield. There is the attribute though https://doc.qt.io/qt-5/qml-qtquick-controls2-spinbox.html#displayText-prop

https://github.com/lutraconsulting/input/pull/1401#issuecomment-847085560

### LINKED ISSUE #1426: Range slider wrong minimum value

### LINKED ISSUE #1435: Limit for range widget is sometimes ignored
In range widget, sometimes the max and min limit is not set, when grater (or lower) number is set manually.

https://user-images.githubusercontent.com/83582181/119709720-cddd4b00-be5d-11eb-8176-7d81db36df14.mp4

tomasMizera:
Complete rework of range widget. 

Important parts:
 - Range editable and range slider widgets were separated into two qml files
 - Range editable is no longer using qml `SpinBox` many bugs were coming from there. It is now just RowLayout with:
    - minus sign (new icon)
    - number input (text field)
    - optional suffix (text)
    - plus sign (new icon)
 - State has been introduced for numeric fields, it is the text underneath field name (the same as constraint description). It informs user when a value he/she typed is either invalid (can not be converted to number with `convertCompatible`) or is outside of specified min/max range 
 - We do not use any qml validator, `AttributeController` serves as validator
 - _Step is set to have a value that is visible!_ It is possible to set e.g. precision:3 and step:0.0001 which resulted in number not being incremented (it actually was incremented, but it was not visible). Now step is set as a maximum of step from qgis and minimum visible increment.

Current workflow:
 - User types in number or hits +/- sign
 - Number is converted to double with `Number.fromLocaleString`
 - Value is sent to AttributeController
 - If the value is not valid or the value is outside of the specified range (min/max), state of the formItem is set accordingly
 - If any of formItems have not valid state, it is not possible to save the form
 - When value is changed, it is converted back to the field with `Number.toLocaleString`

Also, there was this toast when user inserted some invalid value to a field, something like "Value x is not compatible with field y". It was not working in recent versions, I would just remove it, what do you think?

Needs to be properly tested, best in combinations with default values and apply-on-update values. I have used projects: `lutraconsulting/test_decimals`, `tomas/decimals` and `tomas/another-decimals`.

There are still some edge case bugs and I also need to completely change the tests because in the end, we are not using any validator (`AttributeController` is our validator). **Please add all spotted bugs to comments.** Best to test it on mobile device, desktop might handle scope differently (mainly due to a virtual keyboard).

Todo:
 - [x] alter and add tests
 - [x] revert unused code and tests
 - [ ] issues to fix:
     - [x]  when groups (thousands) separator is used as decimal separator it provides some unexpected results 
     - [x]  textfield sometimes does not recognize that number is already valid 
     - [x]  slider with min/max set to +-inf
     - [ ]  it is possible to save when hitting save directly before commiting text (due to a Qt bug https://bugreports.qt.io/browse/QTBUG-94499?jql=status%20in%20(Reported%2C%20Open)%20AND%20text%20~%20%22qt%20inputmethod%20visible%22%20ORDER%20BY%20created%20DESC)
     - [ ]  polish loop
     - [ ] "invalid value text" disappears if you commit the same text twice
     - [x]  field item state remains for other features too
     - [x]  would be better to convert text from locale string before user edits it
     - [x] ignore for slider widget
     - [x] handling zero with +/-
 - [ ] input help documentation 
 - [ ] press and hold on +/- signs

Resolves #1426 
Resolves #1435 
Resolves #1422 
~~Should fix also #1183 and #493 but need to double check~~ Needs to be fixed with #1507, we are currently not ideally deal null values

Let me know what you think

https://user-images.githubusercontent.com/22449698/122519296-9869ee80-d012-11eb-9d86-ef3f51e941ad.mp4


app/attributes/attributecontroller.cpp
```diff
@@ -593,6 +593,31 @@ void AttributeController::recalculateDerivedItems( bool isFormValueChange, bool
...
+      std::shared_ptr<FormItem> item = formItemsIterator.value();
+      QVariant value = featureLayerPair().feature().attribute( item->fieldIndex() );
+
+      if ( item->editorWidgetType() == QStringLiteral( "Range" ) && !value.isNull() )
+      {
+        double min = item->editorWidgetConfig()["Min"].toDouble();
+        double max = item->editorWidgetConfig()["Max"].toDouble();
+        double val = featureLayerPair().feature().attribute( item->fieldIndex() ).toDouble();
+
+        if ( !( min <= val && val <= max ) )
```
tomasMizera, line 610: Here I need to check if the range is `inf` and if so, do not compare

app/attributes/attributecontroller.cpp
```diff
@@ -875,14 +919,26 @@ bool AttributeController::setFormValue( const QUuid &id, QVariant value )
...
+        if ( value.toBool() )
         {
-          emit formDataChangedFailed( userFriendlyMsg );
+          item->setState( FormItem::InvalidInput );
+          emit formDataChanged( id, { AttributeFormModel::FieldState } );
+
+//          I vote for removing this toast message. It did not work with recent versions and nobody seems to care
+//          so I would just remove it
+//          It was added for QR code, we can add similar message like this min/max range instead
+//          emit formDataChangedFailed( userFriendlyMsg );
```
tomasMizera, line 931: This is the toast notification. It was not working in previous versions

app/qml/editor/inputrangeslider.qml
```diff
@@ -0,0 +1,134 @@
+﻿/***************************************************************************
+ range.qml
+  --------------------------------------
+  Date                 : 2019
+  Copyright            : (C) 2019 by Viktor Sklencar
+  Email                : viktor.sklencar@lutraconsulting.co.uk
```
tomasMizera, line 6: I kept the licence header here, because the slider had only few changes

app/test/testformeditors.cpp
```diff
@@ -0,0 +1,342 @@
...
+{
+}
+
+void TestFormEditors::cleanup()
+{
+}
+
+void TestFormEditors::testDoublesValidation()
+{
+//  std::unique_ptr<InputNumberValidator> validator = std::unique_ptr<InputNumberValidator>( new InputNumberValidator() );
```
tomasMizera, line 25: All these tests were here for the validator I created at first. But since we are not using it, I will change these tests for attribute controller.

app/attributes/attributecontroller.cpp
tomasMizera: The whole logic can probably be ignored for slider widget

app/attributes/attributecontroller.cpp
```diff
@@ -634,6 +659,25 @@ bool AttributeController::constraintsSoftValid() const
...
+  bool allFieldsAreValid = true;
+
+  // loop over items and see if there is someone with invalid state
+  QMap<QUuid, std::shared_ptr<FormItem>>::iterator formItemsIterator = mFormItems.begin();
+  while ( formItemsIterator != mFormItems.end() )
+  {
+    std::shared_ptr<FormItem> item = formItemsIterator.value();
+    if ( item->state() != FormItem::Valid )
+    {
+      allFieldsAreValid = false;
```
PeterPetrik, line 673: you can directly return here false?

app/attributes/attributecontroller.cpp
PeterPetrik: +1 better to use text below field similar to constrains

app/attributes/attributecontroller.h
```diff
@@ -80,6 +80,10 @@ class  AttributeController : public QObject
     //! Returns TRUE if all soft constraints defined on fields are satisfied with the current attribute values
     Q_PROPERTY( bool constraintsSoftValid READ constraintsSoftValid NOTIFY constraintsSoftValidChanged )
 
+    //! Returns TRUE if all hard constraints are satisfied and all fields have valid state.
+    //! Validity of a field depends on its type, f.e. numeric field is valid when input is convertible to number and number is in correct min/max range
```
PeterPetrik, line 84: I am not sure if it is better to combine or not combine hard constrains and invalid state into single boolean. we may have bool property to evaluate if all values are valid. and in qml/c++ you can do contrainsHardValid && allFieldsValidValues.

just for consideration

app/attributes/attributedata.cpp
```diff
@@ -93,6 +93,23 @@ void FormItem::setConstraintHardValid( bool constraintHardValid )
...
+void FormItem::setState( FormItemState state )
+{
+  // We are currently using the state only for numeric fields,
+  // thus we here check for field type. We can use this functionality
+  // in future for other fields too, however, now we do not want
+  // to break working features, since this could prohibit user from saving the form
+  if ( mField.isNumeric() || mEditorWidgetSetup.type() == QStringLiteral( "Range" ) )
+  {
+    mState = state;
+  }
```
PeterPetrik, line 110: not sure how to break existing features ? setState should be called with the Valiid for anything other. 

I would not be this paranoid :)

jozefbudac:
In this PR there is still problem described here: https://github.com/lutraconsulting/input/issues/1497

app/attributes/attributeformmodel.h
```diff
@@ -54,6 +54,7 @@ class  AttributeFormModel : public QAbstractListModel
       RememberValue, //!< Remember value (whether to remember the value)
       Field, //!< Field
       FieldIndex, //!< Index
+      FieldState, //!< State of the field, Valid, InvalidInput,..
```
PeterPetrik, line 57: it is more about field value than field itself... hmm.. maybe ValueValidity? CurrentValueValidity? not sure

app/editor/rangewidgethelper.cpp
```diff
@@ -0,0 +1,66 @@
+﻿#include "rangewidgethelper.h"
```
PeterPetrik, line 1: copyright

app/editor/rangewidgethelper.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+  if ( mWidgetConfig.contains( "Suffix" ) )
+  {
+    mSuffix = mWidgetConfig["Suffix"].toString();
+    emit suffixChanged( mSuffix );
+  }
+
+  // don't ever use a step smaller than would be visible in the widget
+  // i.e. if showing 2 decimals, smallest increment will be 0.01
+  // https://github.com/qgis/QGIS/blob/master/src/gui/editorwidgets/qgsdoublespinbox.cpp
```
PeterPetrik, line 60: you need permlink here probably not file from master that can move/change

app/editor/rangewidgethelper.h
```diff
@@ -0,0 +1,54 @@
+﻿#ifndef NUMBERWIDGETHANDLER_H
```
PeterPetrik, line 1: copyright

app/qml/FeatureForm.qml
```diff
@@ -442,15 +447,38 @@ Item {
 
         Label {
           id: constraintDescriptionLabel
+
+          property string constraintText: {
```
PeterPetrik, line 451: maybe rename since you use it for invalid input too?

PeterPetrik (commented):
good work!

One top-level question about "constrains [soft/hard]" and "invalid field" etc. In general from QML point of view it is just "string" to show/not show below the field and if user can/cannot save the features. I am wondering if the whole separation of constrains (qgis code concept) and validity (newly introduced input concept) shouldn't be just internal/private implementation detail of attribute controller ....

tomasMizera:
> good work!
> 
> One top-level question about "constrains [soft/hard]" and "invalid field" etc. In general from QML point of view it is just "string" to show/not show below the field and if user can/cannot save the features. I am wondering if the whole separation of constrains (qgis code concept) and validity (newly introduced input concept) shouldn't be just internal/private implementation detail of attribute controller ....
> 
> @wonder-sk @tomasMizera what do you think?

I am 100% for. We could maybe still have two types of constraints:
 - soft: those that let's you save the form, e.g. qgis soft constraint or scan an invalid QR code but the field can be null
 - hard: those that won't let you save the form, as you said range and qgis hard constraints

@PeterPetrik in this case, would you still keep the formItem's `state` property or how would we differentiate between different errors?

PeterPetrik:
maybe just `msgText` and `msgLevel = [None/Info/Warn/Error]` and then have a boolean property in attribute controller possible to save or not? not sure

tomasMizera:
Hm, so AttributeController would set the `msgText` e.g. _number out of range_ or _invalid value_?

tomasMizera:
@PeterPetrik @wonder-sk I have created a separate ticket for the _invalid values_ topic we discussed

app/attributes/attributecontroller.h
tomasMizera: Changed it to have separate boolean for values validity, #1507 will join them to `errorMessage`

PeterPetrik:
let me know when it its ready for review, I still see "WIP" in the title?

tomasMizera:
@PeterPetrik ready for final review, there are some remarks but it is not blocking review

app/attributes/attributecontroller.cpp
```diff
@@ -608,6 +608,41 @@ void AttributeController::recalculateDerivedItems( bool isFormValueChange, bool
...
+        double min = item->editorWidgetConfig()["Min"].toDouble();
+        double max = item->editorWidgetConfig()["Max"].toDouble();
+        double val = featureLayerPair().feature().attribute( item->fieldIndex() ).toDouble();
+
+        if ( !( min <= val && val <= max ) )
+        {
+          item->setState( FormItem::ValueOutOfRange );
+          changedFormItems << item->id();
+        }
+      }
```
PeterPetrik, line 633: best to setup some minimal framework so another validators can be added easily/testable and not mess up the attribute controller logic. 

for example in the folder attributes/ create class fieldvalidator.[h/cpp] a method `FormItem::Validity validate(const std::shared_ptr<FormItem>)` and this implementation to be hidden in this class


OR could be just a static method in rangewidgethelper.h ?

app/attributes/attributecontroller.cpp
```diff
@@ -649,6 +684,23 @@ bool AttributeController::constraintsSoftValid() const
   return mConstraintsSoftValid;
 }
 
+bool AttributeController::fieldValuesValid()
```
PeterPetrik, line 687: this should be evaluated in the evaluateDerived() and stored in bool

app/attributes/attributecontroller.h
```diff
@@ -80,6 +80,10 @@ class  AttributeController : public QObject
     //! Returns TRUE if all soft constraints defined on fields are satisfied with the current attribute values
     Q_PROPERTY( bool constraintsSoftValid READ constraintsSoftValid NOTIFY constraintsSoftValidChanged )
 
+    //! Returns TRUE if all fields have valid state.
```
PeterPetrik, line 83: use 
/**

**/
for multiline comment

app/attributes/attributedata.cpp
PeterPetrik: not fixed yet?

app/attributes/attributedata.h
```diff
@@ -44,6 +44,14 @@ class  FormItem
     };
     Q_ENUMS( FormItemType )
 
+    enum ValueState
+    {
+      ValidValue = 1,
+      InvalidValue,     // did not pass convertCompatible check
```
PeterPetrik, line 50: //!<

app/attributes/attributedata.h
```diff
@@ -44,6 +44,14 @@ class  FormItem
...
+      ValueOutOfRange  // number is out of min/max range
```
PeterPetrik, line 51: //!<

app/qml/FeatureForm.qml
```diff
@@ -441,16 +446,39 @@ Item {
         }
 
         Label {
-          id: constraintDescriptionLabel
+          id: fieldHelperText
```
PeterPetrik, line 449: where is this used?

app/qml/FeatureForm.qml
```diff
@@ -441,16 +446,39 @@ Item {
...
+
+          property string constraintText: {
+            if ( ValueValidity === FormItem.ValueOutOfRange )
+              return qsTr( 'Number is outside of specified range' )
+            else if ( ValueValidity === FormItem.InvalidValue )
+              return qsTr( 'Value is not valid' )
+
+            if ( ConstraintDescription )
+              return ConstraintDescription
```
PeterPetrik, line 458: logic should be in c++

app/qml/FeatureForm.qml
```diff
@@ -441,16 +446,39 @@ Item {
...
+
+          property bool shouldShowConstraintText: {
+            if ( ( !ConstraintHardValid || !ConstraintSoftValid ) && !!ConstraintDescription )
+              return true
+
+            if ( ValueValidity !== FormItem.ValidValue )
+              return true
+
+            return false
+          }
```
PeterPetrik, line 471: logic should be in c++; also misleading name of property

PeterPetrik (commented):
just 1/2 of the review

app/qml/FeatureForm.qml
tomasMizera: Line 428 https://github.com/lutraconsulting/input/pull/1500/files/dd4b3d3b4c45b52363c6dc486b01d60d4baa08f2#diff-055efeb3c73beb5496dc4f32dd5e5467b2e76e572317083b7d788449598bd5a7R428

sklencar (commented):
* position of value and suffix (iOS + A)
* still buggy for values that are completely out of the range (too big numbers): +- buttons doesn't work (iOS + A): 
* different behaviour as in QGIS: when value is out of range, +- buttons sets value to edge range value (QGIS), Input just in/decreases value by a step. Assuming we want to keep invalid values visible for a user, it is just a note that it works differently.
* not only "invalid value text" disappears, but also (hard) constrain disappears on second click on a field (iOS + A)

sklencar:
Also slider doesn't work properly (test project: _tomas/decimals_):

https://user-images.githubusercontent.com/6735606/123099515-8c2aca80-d432-11eb-9ba4-b4f5b2c224cf.MP4

However, field config also seems to not be set correctly - range is rather none that set as <-inf, inf>
Screenshot 2021-06-23 at 14 50 10

sklencar:
Generally besides known issues and small remarks, I can approve it since it works with most of the user cases.
However, best to resolve code comments before.

tomasMizera:
> Also slider doesn't work properly (test project: _tomas/decimals_):
> 
>  RPReplay_Final1624452537.MP4 
> However, field config also seems to not be set correctly - range is rather none that set as <-inf, inf>
> Screenshot 2021-06-23 at 14 50 10

Yes, that is just project discrepancy between field name and actual values in range.
There is however one ongoing issue (was here in previous versions too): #1516

app/qml/FeatureForm.qml
tomasMizera: Will be in https://github.com/lutraconsulting/input/issues/1507

app/qml/FeatureForm.qml
tomasMizera: Will be in #1507

app/attributes/attributecontroller.cpp
tomasMizera: https://github.com/lutraconsulting/input/issues/1517

PeterPetrik (approved):

PR #1503, opened 2021-06-18
Applied focus fix for search bar

### LINKED ISSUE #1420: Hide keyboard on project open
Steps to reproduce:

 - search for a project
 - open project without closing the keyboard
 - go back to projects list
 
Expected:
 - you are in the home page and search bar is cleaner ✔️ 
 - keyboard is hidden ❎ 

Keyboard is actually visible

sklencar:
Also ProjectPanel needs the same fix as it was applied for BrowsingData panel.
Most likely, it would be needed always for the search bar.

Verified that this is Android issue and cannot be replicated on iOS

closes #1420


tomasMizera (approved):
Good idea moving it to searchbar, working on my device

Issue #1226, opened 2021-03-04
app crashes when selecting features containing Value Relation


p21w:
I've just created a project with loads of polygons with many attribute widgets.
I've sorted the widgets in the drag/drop designer (which I've had issues with in the app before, but thought I'd give it a try), and I've made multiple value relations. 

Now when I open the project on my phone and select the polygon with the new value relations, the app immediatly crashes. 
Any ideas?

saberraz:
Hi @p21w 
Could you make a public project on Mergin (with no sensitive data/information) and post the link here so that we can replicate your issue?

p21w:
Not sure how to link it, but here goes:
https://public.cloudmergin.com/projects/p21w/kbhk_tilstandsregistrering_public/tree 

The name is "kbhk_tilstandsregistrering_public" :) 

I've located the problem to being the Value Relation, since I've now actually made a new layer with Value Map, and it seems to be working whereas the old layer with Value Relation still crashes.

saberraz:
Hmmm...the project not found. Do you want to upload the project again and we can look at the issue?

p21w:
I had to remove it since I don't have enough storage to have two of the projects (o: 

I've made the original project public. It is here: 

https://public.cloudmergin.com/projects/p21w/kbhk_faelledparken1/tree

saberraz:
Duplicate of #1257

PR #1505, opened 2021-06-18
Fix of crash when value relation is missing referenced table/fields

### LINKED ISSUE #1257: Crash when the table for Value Relation is missing
Input crashes when trying to open a form in which a field has Value Relation widget and the referenced table is missing in the project.

It is obviously up to users to check if all their layers are packaged with the project, but in any case, we should not have crashes and silently deal with the missing layers.

sklencar:
Added test if related layer has any field and if the layer has fields needed for creating relation.
Note that both key and value fields have to be present (might be the same field though).

Also simplified code since we already got field names from the config, so we dont need to fetch field itself and get name out of it. Only if it handles also aliases or any other case? From quick look on QgsFields implementation it looks safe though. What do you think?

Tested with project provided in https://github.com/lutraconsulting/input/issues/1226

closes #1257

### DIFF
--- a/app/featureslistmodel.cpp
+++ b/app/featureslistmodel.cpp
@@ -216,18 +216,24 @@ void FeaturesListModel::setupValueRelation( const QVariantMap &config )
 
   QgsVectorLayer *layer = QgsValueRelationFieldFormatter::resolveLayer( config, QgsProject::instance() );
 
-  if ( layer )
+  if ( layer && layer->fields().size() != 0 )
   {
     // save key and value field
     QgsFields fields = layer->fields();
 
-    setKeyField( fields.field( config.value( QStringLiteral( "Key" ) ).toString() ).name() );
-    setFeatureTitleField( fields.field( config.value( QStringLiteral( "Value" ) ).toString() ).name() );
+    QString keyFieldName = config.value( QStringLiteral( "Key" ) ).toString();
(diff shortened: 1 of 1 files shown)


tomasMizera (commented):
It seems ok for me too to not fetch the fields again. However, could you double check it please?

sklencar:
I have also double checked how value relation config and fields.name() work - it uses only field's name and not alias. Note that user sees `display name` on the config form, but field's `name` is stored in the config (see screens below, where description field has an alias).

So simplified code should work as its original.

Screenshot 2021-06-21 at 11 45 08
Screenshot 2021-06-21 at 11 45 18

tomasMizera (commented):
Thanks, it looks good. 

I will try to setup a similar project, test and approve afterwards

sklencar:
I would use only some icon or something less invasive than popup notification - I remember we had notification about invalid type for field which was bit annoying.

tomasMizera:
Sure, I am also against a toast message. I meant the text added in place of constraints:
https://user-images.githubusercontent.com/22449698/122519296-9869ee80-d012-11eb-9d86-ef3f51e941ad.mp4

tomasMizera (approved):
Fix is working ✔️ , it behaves the same as in QGIS - you simply do not see any entries in the relation widget.

PR #1508, opened 2021-06-21
Use onTextChanged for multiline-text edit widget.

### LINKED ISSUE #1502: Multiline widget text dissapears
INPUT: 0.9.8
project: tin1/forms_widgets

In multi-line widget, when scrolling out of the window, the text sometimes disappears. 
Steps to reproduce:
1) open project tin1/forms_widgets
2) record the feature in POI layer
3) go to relation tab, set some type, name and country in the Value relation fields
4) go down, set some text into Multiline field, scroll up and down again and see that the text disappears.

https://user-images.githubusercontent.com/78033756/122540942-6912ac00-d029-11eb-9cdf-45fc5b842c03.mp4

sklencar:
Issue affects not only multiline, but also simple textEdit widget (after range widget PR).
As it was reported, value has not been applied when user scrolls away - so a widget gets out of the feature form's viewport and is destroyed afterwards. There was no other signal emitted (e.g of losing focus or editingFinished), so value has not been committed. 

Such extra handling has been added on destroy event.
The condition of saving a new value is that it has to be different from original value. So its questionable if after changing a value and then reverting it back without committing it in between should or should not trigger `valueChanged` (and updating the form afterwards). Current implementation ignores such case.

test project `tin1/form_widgets`

Note that PurchasingTest failed.

closes #1502

### DIFF
--- a/app/qml/editor/inputtextedit.qml
+++ b/app/qml/editor/inputtextedit.qml
@@ -77,6 +77,15 @@ Item {
     onEditingFinished: {
       valueChanged( text, text === undefined )
     }
+
+    //! Commit value if has changed when widget gets out of the FeatureForm (ListView) viewport
+    Component.onDestruction: {
+      if ( textField.activeFocus ) {
+        if ( value !== textField.text ) {
+          valueChanged( textField.text, textField.text === undefined )
+        }
+      }
+    }
   }
 
   TextArea {
@@ -102,8 +111,17 @@ Item {
         radius: customStyle.fields.cornerRadius
     }
 
-    onTextChanged: {
+    onEditingFinished: {
         valueChanged( text, text === undefined )
+      }
+    }
+
+    //! Commit value if has changed when widget gets out of the FeatureForm (ListView) viewport
+    Component.onDestruction: {
+      if ( textArea.activeFocus ) {
+        if ( value !== textArea.text ) {
+          valueChanged( textArea.text, textArea.text === undefined )
+        }
       }
     }
 
--- a/app/qml/editor/inputtextedit.qml
+++ b/app/qml/editor/inputtextedit.qml
@@ -102,7 +102,7 @@ Item {
         radius: customStyle.fields.cornerRadius
     }
 
-    onEditingFinished: {
+    onTextChanged: {
         valueChanged( text, text === undefined )
       }
     }


tomasMizera:
Not sure about this. It changes the logic a lot. While before the text was committed only when field lost focus, now we commit it at each typed letter. This can bring some other problems, because after each letter we now evaluate the whole `recalculateDerivedItems` with default values, constraints, etc.. For example #1500 changes the handler for _simple textfield_ back to `onEditingFinished` because for numeric fields we can not reevaluate everything after each digit (numerous problems). So if someone would set multiline textfield for numeric fields, this filed will no longer accept comma and will result in weird behavior. However, who would set multiline field for numeric value 🤔

sklencar:
Agree, that `onEditingFinished` suppose to be used. Especially after the changes after range widget, where `onEditingFinished` replaced `onTextChanged` due some issue with numbers.

Unfortunately, closing keyboard in such way doesn't emit any signal which can be used. However, such action changes (active) focus of the text field, which [triggers](https://doc.qt.io/qt-5/qml-qtquick-controls-textfield.html#editingFinished-signal) `onEditingFinished`.

I have tried signals related to focus, visibility, even cursor prop and nothing has been emitted. What seems to be promising is using `Component.onDestroy` - when a field disappears from listview's (form's) viewport, delegate destroy signal is emitted. Then, if a textEdit value is different from feature's value, valueChanged signal can be possibly emitted to changed the value.

This approach has to be tested.

app/qml/editor/inputtextedit.qml
```diff
@@ -77,6 +77,15 @@ Item {
     onEditingFinished: {
       valueChanged( text, text === undefined )
     }
+
+    //! Commit value if has changed when widget gets out of the FeatureForm (ListView) viewport
+    Component.onDestruction: {
+      if ( textField.activeFocus ) {
+        if ( value !== textField.text ) {
+          valueChanged( textField.text, textField.text === undefined )
```
tomasMizera, line 85: Shouldn't we rather use `displayText` property? I think `text` does not need to be committed

```suggestion
        if ( value !== textField.displayText ) {
          valueChanged( textField.displayText, textField.displayText === undefined )
```

app/qml/editor/inputtextedit.qml
```diff
@@ -107,6 +116,15 @@ Item {
       }
     }
 
+    //! Commit value if has changed when widget gets out of the FeatureForm (ListView) viewport
+    Component.onDestruction: {
+      if ( textArea.activeFocus ) {
+        if ( value !== textArea.text ) {
+          valueChanged( textArea.text, textArea.text === undefined )
```
tomasMizera, line 123: I know we are using the comparison to undefined all over the place with `valueChanged` signal, but I am thinking whether the empty string would not make more sense

```suggestion
          valueChanged( textArea.text, textArea.text === "" )
```
Or maybe better:
```suggestion
          valueChanged( textArea.text, textArea.text === undefined || textArea.text === "" )
```

tomasMizera (changes requested):
Good idea with the `onDestruction` signal!

app/qml/editor/inputtextedit.qml
sklencar: It is save to change it to displayText. However there was no such problem with text or using onTextChanged, the issue was related to not committing the change to a feature.

Have you experienced any issue during the testing?

app/qml/editor/inputtextedit.qml
sklencar: The second parameter of `ValueChange` signal is `bool isNull`. Handling of null values is not handled properly as I can recall (issues can be easily found by searching "NULL"). I would like to resolve those issues, but not in this PR.

tomasMizera (approved):

PR #1510, opened 2021-06-23
Show relations in form

tomasMizera:
We are currently able to show relations in new widget called `inputrelation.qml`

How it work:
 - relation widget type (form qgis) is handled as another form item (see `AttributeController::flatten`)
 - relationFeaturesModel (derived from featureslistmodel) receives `QgsRelation` and provides the data to relation widget
 - relation widget has 3 states: 
    - _initial_ = there is only text saying how many features are linked
    - _expanded_ = widget expands to three lines with features listed (their display text is used). If feature contains more linked features than can be displayed (see attached video in 0:10), we only show the first n.
    - _page_ =  (not yet implemented) separate page with all linked features listed ~> will be implemented in **future PR**, because some qml changes need to be done first

docu: https://docs.google.com/document/d/1hEpfXa59NoKz7tyyFSghZhQ1QELQftGWiVuAAT57Szc/edit

https://user-images.githubusercontent.com/22449698/123085765-eb80de80-d422-11eb-9ffd-80cf7f57695a.mp4

I will create tickets for following tasks.

### DIFF
--- a/app/attributes/attributecontroller.cpp
+++ b/app/attributes/attributecontroller.cpp
@@ -24,10 +24,13 @@
 #include <QSet>
 #include "qgsvectorlayer.h"
 #include "qgsattributeeditorfield.h"
+#include "qgsattributeeditorrelation.h"
 #include "qgsattributeeditorcontainer.h"
 #include "qgsvectorlayerutils.h"
+#include "qgsrelation.h"
 #include "qgsmessagelog.h"
 #include "inpututils.h"
+#include "coreutils.h"
 
 AttributeController::AttributeController( QObject *parent )
   : QObject( parent )
@@ -249,8 +252,39 @@ void AttributeController::flatten(
       }
 
       case QgsAttributeEditorElement::AeTypeRelation:
-        // todo
+      {
+        QUuid widgetUuid = QUuid::createUuid();
+
+        QgsAttributeEditorRelation *relationField = static_cast<QgsAttributeEditorRelation *>( element );
+        QgsRelation relation = relationField->relation();
+
+        bool isNmRelation = layer->editFormConfig().widgetConfig( relation.id() )["nm-rel"].toBool();
+        if ( isNmRelation )
+        {
+          CoreUtils::log( "Relations", "Nm relations are not supported" );
+          continue;
+        }
+
+        const QString groupName = container->isGroupBox() ? container->name() : QString();
+        const QString label = relationField->label();
+
+        std::shared_ptr<FormItem> formItemData =
+          std::shared_ptr<FormItem>(
+            new FormItem(
+              widgetUuid,
+              groupName,
+              parentTabRow,
+              FormItem::Relation,
+              label,
+              relation
+            )
+          );
+
+        mFormItems[formItemData->id()] = formItemData;
+        items.append( widgetUuid );
+
         break;
+      }
 
       case QgsAttributeEditorElement::AeTypeInvalid:
         // todo
@@ -570,16 +604,19 @@ void AttributeController::recalculateDerivedItems( bool isFormValueChange, bool 
       while ( formItemsIterator != mFormItems.end() )
       {
         std::shared_ptr<FormItem> item = formItemsIterator.value();
-        QStringList errors;
-        bool hardConstraintSatisfied = QgsVectorLayerUtils::validateAttribute( layer,  featureLayerPair().feature(), item->fieldIndex(), errors, QgsFieldConstraints::ConstraintStrengthHard );
-        if ( hardConstraintSatisfied != item->constraintHardValid() )
+        if ( item->type() == FormItem::Field )
         {
-          item->setConstraintHardValid( hardConstraintSatisfied );
-          changedFormItems << item->id();
-        }
-        if ( !hardConstraintSatisfied )
-        {
-          allConstraintsHardValid = false;
+          QStringList errors;
+          bool hardConstraintSatisfied = QgsVectorLayerUtils::validateAttribute( layer,  featureLayerPair().feature(), item->fieldIndex(), errors, QgsFieldConstraints::ConstraintStrengthHard );
+          if ( hardConstraintSatisfied != item->constraintHardValid() )
+          {
+            item->setConstraintHardValid( hardConstraintSatisfied );
+            changedFormItems << item->id();
+          }
+          if ( !hardConstraintSatisfied )
+          {
+            allConstraintsHardValid = false;
+          }
         }
         ++formItemsIterator;
       }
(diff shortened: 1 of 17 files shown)


app/attributes/attributecontroller.cpp
```diff
@@ -249,8 +252,39 @@ void AttributeController::flatten(
       }
 
       case QgsAttributeEditorElement::AeTypeRelation:
-        // todo
+      {
+        QUuid widgetUuid = QUuid::createUuid();
+
+        QgsAttributeEditorRelation *relationField = static_cast<QgsAttributeEditorRelation *>( element );
+        QgsRelation relation = relationField->relation();
```
PeterPetrik, line 259: GitHub highlight "relation" as c++ keyword (there is std::relation..) so maybe use some other name for local variable?

app/attributes/attributedata.cpp
```diff
@@ -40,9 +40,39 @@ FormItem::FormItem(
...
+  , mEditorWidgetSetup( QgsEditorWidgetSetup() )
+  , mFieldIndex( -1 )
+  , mConstraints( QgsFieldConstraints() )
+  , mVisibilityExpression( QgsExpression() )
+    // ----
+
+  , mRelation( relation )
+{
+  mConstraintHardValid = true;
+  mConstraintSoftValid = true;
```
PeterPetrik, line 73: why is this not in the above section?

app/attributes/attributedata.cpp
```diff
@@ -60,6 +90,9 @@ bool FormItem::isEditable() const
 
 QString FormItem::editorWidgetType() const
 {
+  if ( mType == FormItem::Relation )
+    return "relation";
```
PeterPetrik, line 94: QStringLiteral("relation")

app/attributes/attributedata.h
```diff
@@ -58,6 +60,15 @@ class  FormItem
       const QgsExpression &visibilityExpression
     );
 
+    //! Contructor for relation items
```
PeterPetrik, line 63: I think it is better to have just one constructor with ALL const items and then static factory functions 
```
static FormItem createFieldItem(...)
static FormItem createRelationItem(...)
```
with subset of attributes

app/qml/FeatureForm.qml
```diff
@@ -407,9 +407,12 @@ Item {
 
     Item {
       id: fieldContainer
-      visible: Type === FormItemType.Field
+
+      property bool shouldBeVisible: Type === FormItemType.Field || Type === FormItemType.Relation
```
PeterPetrik, line 411: it is definitely ok to do it this way, but just for consideration to have 2 separate items , one for fields (visible: Type === FormItemType.Field) and one for relations  (visible: Type === FormItemType.Relation), since not sure how much implementation/qml they share?

also it would be good to finally use components and try to detangle/un-spagetty the feature form to some components in separate files

app/qml/editor/inputrelation.qml
```diff
@@ -0,0 +1,271 @@
...
+            target: noOfFeaturesText
+            visible: false
+          }
+        },
+        State {
+          name: "page"
+//          PropertyChanges {
+//            target: object
+
+//          }
```
PeterPetrik, line 140: commented out code

app/relationfeaturesmodel.h
```diff
@@ -0,0 +1,53 @@
...
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+
+#ifndef RELATIONFEATURESMODEL_H
+#define RELATIONFEATURESMODEL_H
+
+#include "featureslistmodel.h"
```
PeterPetrik, line 13: is this include required in header?

app/relationfeaturesmodel.h
```diff
@@ -0,0 +1,53 @@
...
+#include "featurelayerpair.h"
```
PeterPetrik, line 14: in general best to use forward declarations where possible for classes in the input/app or qgsquick since it speeds up development (less compilation on change)

PeterPetrik (commented):
👏

app/qml/FeatureForm.qml
tomasMizera: That is a good point. They share field name and some properties from form model like featurepair or config. However, most of the other properties are not used. Even checkboxes for _Reuse last value_ should not be there. 

Yes, feature form needs to be cleaned. Let's have a discussion after relations. My plan is to at least consolidate the styling within relations.

app/relationfeaturesmodel.h
tomasMizera: Base class, did not work in forward declaration

app/relationfeaturesmodel.h
tomasMizera: I tried that, however, QgsRelation and FeatureLayerPair are private members and it did not work me to do forward declaration in these cases

app/attributes/attributecontroller.cpp
tomasMizera: It will still highlight the `->relation` part

app/qml/editor/inputrelation.qml
```diff
@@ -0,0 +1,268 @@
...
+  }
+
+  Component {
+    id: textDelegate
+
+    Item {
+      id: textDelegateContainer
+
+      property real maximumWidth: flowItemView.width
+      property bool isVisible: {
```
sklencar, line 213: It is necessary to limit elements, however some indicator (e.g three dots) is missing in such case

app/qml/editor/inputrelation.qml
```diff
@@ -0,0 +1,268 @@
...
+
+        height: parent.height
+        width: requestedWidth > parent.maximumWidth ? parent.maximumWidth : requestedWidth
+
+        color: customStyle.fields.backgroundColorDark
+        radius: customStyle.fields.cornerRadius
+        border.color: customStyle.fields.backgroundColorDarker
+        border.width: 2 * QgsQuick.Utils.dp
+
+        Text {
```
sklencar, line 238: Maybe consider elide left for long text to see beginning of the relation name. However, its just my preference.

app/qml/editor/inputrelation.qml
```diff
@@ -0,0 +1,268 @@
...
+
+          font.pointSize: customStyle.fields.fontPointSize
+          color: customStyle.fields.fontColor
+          clip: true
+        }
+      }
+
+      MouseArea {
+        anchors.fill: parent
+        onReleased: console.log("opening", model.FeatureId)
```
sklencar, line 256: leftover?

app/attributes/attributecontroller.cpp
```diff
@@ -249,8 +252,39 @@ void AttributeController::flatten(
...
 
       case QgsAttributeEditorElement::AeTypeRelation:
-        // todo
+      {
+        QUuid widgetUuid = QUuid::createUuid();
+
+        QgsAttributeEditorRelation *relationField = static_cast<QgsAttributeEditorRelation *>( element );
+        QgsRelation associatedRelation = relationField->relation();
+
+        bool isNmRelation = layer->editFormConfig().widgetConfig( associatedRelation.id() )["nm-rel"].toBool();
```
sklencar, line 261: QStringLiteral("nm-rel")

app/qml/editor/inputrelation.qml
tomasMizera: Yes, I plan to add it in next PR. It is indeed in the docs.

app/qml/editor/inputrelation.qml
tomasMizera: That is a good point, however, you can have an elide only when `Text` has width property specified, in this case it grows with `paintedWidth` and it did not work well. I will mark it down and have a look in future

PeterPetrik (approved):

Issue #1514, opened 2021-06-23
Open existing / add new linked features


tomasMizera:
After #1513 We will be able to show multiple feature forms. Next step is to open the features and assign parent feature for child form. 

Requirements:
 - you are able to add new child feature
 - you are able to delete existing child feature
 - relation reference (in child feature form) has already prefilled parent id
 - list of features is automatically refreshed after child form is closed

Issue #1252, opened 2021-03-10
Update Qt lib versions in qml


tomasMizera:
In some QML files we use very old versions of Qt libs. For example QtQuick 2.0 (>5 years old), even though there is 2.14

Versions:
https://en.wikipedia.org/wiki/Qt_Quick
https://doc.qt.io/qt-6/qtquickcontrols-index.html#versions

Would be good to update versions in all files.

PR #1524, opened 2021-06-30
Dedicated page for relations

### LINKED ISSUE #1513: Dedicated page to show child features
Have a dedicated page (just like in value relation or browse data) that contains all linked features from child layer. 

https://docs.google.com/document/d/1hEpfXa59NoKz7tyyFSghZhQ1QELQftGWiVuAAT57Szc/

tomasMizera:
QML only refactoring of feature forms.

This PR introduces possibility to:
 - easily open pages from editors (widgets) like value relation page, relation page, ..
 - **open multiple feature forms**

How it looks:
 - We have a new `FormController.qml` that is responsible for showing correct feature forms. It contains a `StackView` and pushes there `FormWrapper` components for each form.
 - `FormWrapper` wraps `FeatureFormPage` and `PreviewPanel`
 - All form components besides `FormController` were moved to subfolder `forms`

It looks like this:
formcontroller

Moreover, `FeatureFormPage` contains another `StackView` called extraView, it is used to display the form and other extra pages like relations page..

We are now able to browse all linked features in relations. Searchbar in relations page is disabled for now.
Resolves #1513 

I have also fixed inf. loop in value relation here. It was introduced in #1500 because we started to emit _dataChanged_ signal more often.

https://user-images.githubusercontent.com/22449698/123920964-d0642080-d986-11eb-9e41-0a171932cf41.mp4


tomasMizera:
Value relations can be tested here `tomas/VR-Filtering`

app/qml/FormController.qml
```diff
@@ -0,0 +1,143 @@
...
+import "."
+import "./form" as Forms
+import QgsQuick 0.1 as QgsQuick
+import lc 1.0
+
+Item {
+  id: root
+
+  /*
+   * FeaturePanel component is responsible for entire feature form, both preview and fullscreen form
```
PeterPetrik, line 22: FeaturePanel? do you mean FormController? 

the name is a bit consusing to me at the first read thinking what is the difference between form controller and attribute controller? Maybe `FormsStackManager`? 

it would be good to add here descrition, e.g. last form is not destroyed but reused, new forms pages are opened e.g. when using relation in forms, ...

app/qml/main.qml
```diff
@@ -737,42 +741,36 @@ ApplicationWindow {
         }
     }
 
-    FeaturePanel {
-        id: featurePanel
+    FormController {
+        id: formController
+
         height: window.height
```
sklencar, line 747: Is it a controller or page as the same time? (it has size and onClose signal)

app/qml/FormController.qml
```diff
@@ -0,0 +1,143 @@
...
+
+  function reload() {
+    formsStack.clear() // removes all objects due to Qt parent system
+  }
+
+  function isNewFeature() {
+    if ( root.activeFormIndex >= 0 && root.activeFormIndex < formsStack.depth ) {
+      let form = formsStack.get( activeFormIndex )
+      return form.isNewFeature()
+    }
```
PeterPetrik, line 76: shouldn't we return something in else statement?

app/qml/FormController.qml
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  function updateFeatureGeometry() {
+    if ( root.activeFormIndex >= 0 && root.activeFormIndex < formsStack.depth ) {
```
PeterPetrik, line 80: do you expect these can happen normally? or it is just super-safe guard? if it shouldn't happen normally I would put debug message in else statement for all these?

app/qml/FormController.qml
```diff
@@ -0,0 +1,143 @@
...
+
+      project: root.project
+
+      previewHeight: root.previewHeight
+
+      onClosed: formsStack.popOneOrClose()
+      onEditGeometry: root.geometryEditingStarted( StackView.index )
+
+      onCreateFeature: {
+        // TODO
```
PeterPetrik, line 139: TODO

app/qml/editor/inputrelation.qml
```diff
@@ -265,4 +270,30 @@ Item {
...
+
+      pageTitle: qsTr( "Linked features" )
+      allowSearch: false //TODO search
+
+      onBackButtonClicked: {
+        root.parent.formView.pop()
+        textModeContainer.state = "expanded"
+      }
+
+//      onFeatureClicked: //TODO
```
PeterPetrik, line 288: TODO

app/qml/form/FeatureToolbar.qml
```diff
@@ -1,109 +1,109 @@
...
+
+            MainPanelButton {
+                id: openProjectBtn
+                width: toolbar.itemSize
+                text: qsTr("Edit")
+                imageSource: InputStyle.editIcon
+
+                onActivated: {
+                    toolbar.editClicked()
+//                    toolbar.state="Edit"
```
sklencar, line 65: why is commented out?

app/qml/form/FeatureFormPage.qml
```diff
@@ -0,0 +1,241 @@
...
+          id: saveButtonText
+
+          text: qsTr("Save")
+
+          height: header.rowHeight
+          visible: featureForm.state === "Edit" || featureForm.state === "Add"
+
+          enabled: featureForm.controller.fieldValuesValid && featureForm.controller.constraintsHardValid
+
+          color: enabled ? InputStyle.highlightColor : "red"
```
PeterPetrik, line 99: can you put "red" in InputStype too?

app/qml/form/FeatureFormStyling.qml
```diff
@@ -0,0 +1,108 @@
+/***************************************************************************
+ featureformstyling.qml
+  --------------------------------------
+  Date                 : January 2018
+  Copyright            : (C) 2018 by Martin Dobias
+  Email                : wonder dot sk at gmail dot com
+ ***************************************************************************
```
PeterPetrik, line 7: remove

PeterPetrik (commented):
Quite a lot of changes, so it needs a proper testing. but the changes looks good to me!

app/qml/form/FeatureToolbar.qml
```diff
@@ -1,109 +1,109 @@
...
+/***************************************************************************
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+
+import QtQuick 2.14
```
sklencar, line 10: as I can remember, PeterP or Martin was complaining about increasing version which can resulted into some unexpected bug with QT.
On the other hand, I would like to have single max. possible version in all imports.

app/qml/FormController.qml
tomasMizera: It should not happen, I added error message

app/qml/editor/inputrelation.qml
tomasMizera: incoming in #1514

app/qml/FormController.qml
tomasMizera: incoming in #1514

app/qml/form/FeatureToolbar.qml
tomasMizera: it was unnecessary, only breaking a binding. removed now

app/qml/FormController.qml
tomasMizera: It returns undefined by default, should we handle it somehow? I think returning false could be misleading

tomasMizera: but it _should not_ happen

app/qml/form/FeatureToolbar.qml
tomasMizera: Hmm, I did not know about that. @PeterPetrik @wonder-sk Shall we increase version everywhere then ( #1252 ) ? Versions in components now differ, some have 2.7, some 2.0 or 2.12 ..

app/qml/main.qml
tomasMizera: It is invisible `Item` with height and width so it can forward it to forms

app/qml/form/FeatureToolbar.qml
wonder-sk: I don't recall any unexpected bug in qt...

- a selfish concern I typically have is not to use higher version number than what Qt I have on my system :-) on my system I have nowadays Qt 5.12, so I would be happy with 2.12. But as I am currently not doing much work on Input, feel free to ignore me and use the version according to Qt version that we use :-)
- a non-selfish concern is that we should ideally keep the imports consistent with our minimum required Qt version, to make sure that all components have consistent behavior and API. So e.g. if we have Qt 5.15 on all platforms, let's raise all imports to 2.15 ...

PeterPetrik: we have 5.15 everywhere now

PeterPetrik (approved):

PR #1527, opened 2021-06-30
Added gallery widget for relations

### LINKED ISSUE #1511: Photo widget for relations
If child table in relations contains photo, we want a different view in relations widget.

Model is already created, so we only need the view.
`Inputrelation.qml` has two places to implement this, first a view (in `photoModeContainer`) and then a delegate (in `photoDelegate`). 

We need to define the logic when to show such view, e.g. everytime when at least one field is `attachment` / contains some special word in alias / ... ? 

Project `tomas/relations` contains such relation, layer `POI` has reference to child layer `POI_photos`.

See relations documentation: https://docs.google.com/document/d/1hEpfXa59NoKz7tyyFSghZhQ1QELQftGWiVuAAT57Szc/

sklencar:
Added gallery widget for relations with images.

**When is shown:** (can be discussed)
The widget is shown if there is an image field if field widget type is 'ExternalResource'.
Note that it can be also beneficial to use some keyword as part of a field name, to clearly specify which field suppose to be used to show relation image. Input would give it priority before the first 'ExternalResource' field.
If we want to have it, what would be the keyword?
 
**Behaviour:**
Gallery can be browsed from left to right. Opening related feature or showing a preview is not implemented yet.
If image is not available or is missing, an icon is shown instead.

In order to resolve path for images, some functions were moved from inputexternalresource.qml to input utils.
Also test have been added for those functions. 
 
Can be tested with: `viktor/test_image_relations`
 
Note that purchasing tests failed
  
To improve:
* more love for icons if an image is not available
* to center images in the gallery row
* interaction events (click, hold)

IMG_3E785B13480F-1

 
closes #1511

### DIFF
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -635,6 +635,68 @@ bool InputUtils::fileExists( const QString &path )
   return ( check_file.exists() && check_file.isFile() );
 }
 
+QString InputUtils::resolveTargetDir( const QString &homePath, const QVariantMap &config, const FeatureLayerPair &pair, QgsProject *activeProject )
+{
+  QString expression;
+  QMap<QString, QVariant> collection = config.value( QStringLiteral( "PropertyCollection" ) ).toMap();
+  QMap<QString, QVariant> props = collection.value( QStringLiteral( "properties" ) ).toMap();
+
+  if ( !props.isEmpty() )
+  {
+    QMap<QString, QVariant> propertyRootPath = props.value( QStringLiteral( "propertyRootPath" ) ).toMap();
+    expression = propertyRootPath.value( QStringLiteral( "expression" ), QString() ).toString();
+  }
+
+  if ( !expression.isEmpty() )
+  {
+    return evaluateExpression( pair, activeProject, expression );
+  }
+  else
+  {
+    QString defaultRoot = config.value( QStringLiteral( "DefaultRoot" ) ).toString();
+    if ( defaultRoot.isEmpty() )
+    {
+      return homePath;
+    }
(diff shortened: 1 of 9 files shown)


wonder-sk:
Any screenshot? :wink:

tomasMizera (commented):
Looking good!

I tested the widget and found out that it works with the provided project `viktor/test_image_relations` but does not with project `tomas/relations` which seems to be a base project for it.

When I test it on the `tomas/relations` I keep getting `D Input   : qml: error: Unable to copy file file:///storage/emulated/0/Pictures/Viber/IMG-46560b97c9a70921964f215eabaacd6e-V.jpg to the project directory` when adding photo to `poi_photos` layer. Is there some configuration error in the project?

Regarding todos of opening feature form and previewing image, I would keep them as they are. I can merge these changes with #1514 so the form will get opened

We can also discuss what to do when image is not available

sklencar:
I tried to use **tomas/relations**, but change a path of an image of one of the referenced features to some existing file on my drive. I have also get `Unable to copy file` message, even on master.

From a debugger it seems that somehow project has still your local path:
Screenshot 2021-07-01 at 08 53 08

even though widget config seems ok as well as project variables:
Screenshot 2021-07-01 at 08 56 41

tomasMizera:
Hm, that is weird. Maybe that is something project specific.
Could you also please double check if image preview is working on Android? It again did not work for me yesterday

sklencar:
Preview should work now

tomasMizera (approved):
Great!

Issue #1530, opened 2021-07-01
Relations followup


tomasMizera:
Set of tasks that needs to be done for relations:
 - ~~Add "Add feature" (plus) and "More" (three dots) icons for relation text widget ... plus icon also for photo widget~~
 - ~~Remove the initial state with text _n linked features_, show the first n features immediately~~
 - Check if the function `acquiredId` passed the `commit` step and if not, show error message (see https://github.com/lutraconsulting/input/pull/1531#discussion_r662225426)
 - ~~Make invalid image icons smaller in widget and add border to them~~

tomasMizera:
Should be addressed already

PR #1531, opened 2021-07-01
Add, edit and remove features in relation

### LINKED ISSUE #1514: Open existing / add new linked features

tomasMizera:
Added ability to work with child features in relations. 

When creating new child feature, `Relation Reference` field is automatically prefilled with its parent id. This is done by assigning the AttributeController with `parentController` (of type AttributeController) and `linkedRelation` (of type QgsRelation). When both properties are assigned, (child) controller looks for relation reference field and updates its value with parent's id.

However, if parent id is not yet valid (in case the parent feature is also being just created), we first commit the parent feature to get its id and then update the child feature relation reference field. To be able to do this, child controller is listening on `featureIdChanged` signal from parent controller.

When you cancel creation of parent feature that was already committed due to a child feature creation, we detect this and remove the parent feature. **However, child features remain created and also their relation reference fields are assigned**.

Todo:
 - [x] fix conflicts (incorporate photo widget)
 - [x] fix focus management (back button)

Further UI changes together with missing _Add feature_ and _More_ icons will be added in #1530

https://user-images.githubusercontent.com/22449698/124115199-ca953a80-da6d-11eb-91bd-dbbee7eb5bb6.mp4

Closes #1514

### DIFF
--- a/app/attributes/attributecontroller.cpp
+++ b/app/attributes/attributecontroller.cpp
@@ -116,7 +116,7 @@ QgsEditorWidgetSetup AttributeController::getEditorWidgetSetup( QgsVectorLayer *
 
 void AttributeController::prefillRelationReferenceField()
 {
-  if ( !mParentFeatureLayerPair.isValid() || !mLinkedRelation.isValid() )
+  if ( !mParentController || !mLinkedRelation.isValid() )
     return;
 
   const QList<QgsRelation::FieldPair> fieldPairs = mLinkedRelation.fieldPairs();
@@ -128,8 +128,9 @@ void AttributeController::prefillRelationReferenceField()
       std::shared_ptr<FormItem> itemData = formItemsIterator.value();
       if ( itemData->field().name() == fieldPair.referencingField() )
       {
-        setFormValue( itemData->id(), mParentFeatureLayerPair.feature().id() );
+        setFormValue( itemData->id(), mParentController->featureLayerPair().feature().id() );
         emit formDataChanged( itemData->id() );
+        break;
       }
       ++formItemsIterator;
     }
@@ -462,6 +463,26 @@ void AttributeController::updateOnFeatureChange()
 bool AttributeController::isNewFeature() const
 {
   return FID_IS_NULL( mFeatureLayerPair.feature().id() );
+}
+
+void AttributeController::acquireId()
+{
+  if ( !mFeatureLayerPair.layer() )
+    return;
+
+  startEditing();
+  QgsFeature feat = mFeatureLayerPair.feature();
+  if ( !mFeatureLayerPair.layer()->addFeature( feat ) )
+  {
+    QgsMessageLog::logMessage( tr( "acquireId: feature could not be added" ),
+                               QStringLiteral( "Input" ),
+                               Qgis::Critical );
+
+  }
+
+  connect( mFeatureLayerPair.layer(), &QgsVectorLayer::featureAdded, this, &AttributeController::onFeatureAdded );
+  commit();
+  disconnect( mFeatureLayerPair.layer(), &QgsVectorLayer::featureAdded, this, &AttributeController::onFeatureAdded );
 }
 
 bool AttributeController::recalculateDefaultValues(
@@ -833,7 +854,10 @@ bool AttributeController::create()
                                Qgis::Critical );
 
   }
+  connect( mFeatureLayerPair.layer(), &QgsVectorLayer::featureAdded, this, &AttributeController::onFeatureAdded );
   commit();
+  disconnect( mFeatureLayerPair.layer(), &QgsVectorLayer::featureAdded, this, &AttributeController::onFeatureAdded );
+
 
   if ( mRememberAttributesController )
   {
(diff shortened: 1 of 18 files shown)


app/digitizingcontroller.h
```diff
@@ -56,8 +56,8 @@ class DigitizingController : public QObject
     Q_INVOKABLE FeatureLayerPair pointFeatureFromPoint( const QgsPoint &point, bool isGpsPoint );
     //! Creates a new QgsFeature with line/polygon geometry from the points stored since the start of recording
     Q_INVOKABLE FeatureLayerPair lineOrPolygonFeature();
-    //! Creates a new QgsFeature without geometry
-    Q_INVOKABLE FeatureLayerPair featureWithoutGeometry();
+    //! Creates a new QgsFeature without geometry in layer. If layer is null, it creates the feature on currently edited layer
```
sklencar, line 59: Thinking about to move this function to some generic/helper class since its used not only while digitizing, but also while creating linked feature. It has got a bit more custom I would say

app/qml/editor/inputrelation.qml
```diff
@@ -279,13 +281,20 @@ Item {
 
       pageTitle: qsTr( "Linked features" )
       allowSearch: false //TODO search
+      layerHasGeometry: false
+      toolbarVisible: !root.parent.readOnly
+      focus: true
```
sklencar, line 286: Why does it need to be set to true?

app/inpututils.h
```diff
@@ -334,6 +334,11 @@ class InputUtils: public QObject
      */
     Q_INVOKABLE static QModelIndex invalidIndex();
 
+    /**
+     * Returns if provided Id is valid ( >0 )
```
wonder-sk, line 338: by the way, zero is also valid fid

app/qml/FormsStackManager.qml
```diff
@@ -120,6 +121,32 @@ Item {
...
+      formState: "ReadOnly",
+      panelState: "form"
+    }
+
+    let latest = formsStack.push( formComponent, props )
+    root.activeFormIndex = latest.StackView.index
+  }
+
+  function createLinkedFeature( parentController, relation ) {
+    let newFeaturePair = digitizingController.featureWithoutGeometry( relation.referencingLayer )
```
wonder-sk, line 136: it would be nice to have forms decoupled from digitizingController - and the featureWithoutGeometry() function could be e.g. in InputUtils or a similar place... but no strong feelings about that

app/attributes/attributecontroller.cpp
```diff
@@ -442,6 +465,26 @@ bool AttributeController::isNewFeature() const
...
+  if ( !mFeatureLayerPair.layer()->addFeature( feat ) )
+  {
+    QgsMessageLog::logMessage( tr( "acquireId: feature could not be added" ),
+                               QStringLiteral( "Input" ),
+                               Qgis::Critical );
+
+  }
+
+  connect( mFeatureLayerPair.layer(), &QgsVectorLayer::featureAdded, this, &AttributeController::onFeatureAdded );
+  commit();
```
wonder-sk, line 484: maybe it is properly handled already, just checking - what happens if commit() fails? Do we show any error message?

The `mFeatureLayerPair.layer()->addFeature( feat )` call generally should not be failing, only when something is really wrong in the app logic, but commit() may fail because provider could reject something (for example if one of the fields has a constraint (e.g. "unique") unfulfilled. If not handled well, let's defer to a separate ticket, just pointing it out so we don't forget about it, as this may be silently loosing data...

wonder-sk (approved):
:clap: :clap:

app/attributes/attributecontroller.cpp
tomasMizera: Hm, yes, it is **not** being handled yet. How do you think we should notify user? Toast message or the popup Viktor made for GPS accuracy?

wonder-sk: I think the correct approach would be to show a big error popup that saving failed, and ideally also detailed message from `layer->commitErrors()`. Also, we should not close the form on failed save - the user may know how to fix the problem and try saving again...

Visually I am thinking about something like this (just a random image from internet, does not need to look like that)

image

app/qml/editor/inputrelation.qml
tomasMizera: Just fixed the focus in forms (for back button). I am not 100% sure, but this should catch focus when other form is being closed.

app/attributes/attributecontroller.cpp
tomasMizera: Added the check to https://github.com/lutraconsulting/input/issues/1530

app/qml/FormsStackManager.qml
tomasMizera: Hm, I am not sure, it is not a static function. It uses variables manager and evaluates the variables in feature's context. 

See https://github.com/lutraconsulting/input/pull/1531/files#diff-a414e2e410de55a5bc2ba342f687a3a7d92d664079d2f7e977217403c53c46beR220-R227

If it would be simple `QgsVectorUtils::createFeature` then I would go for it, but with variables manager I am not really sure. What do you think? We would need to add the variables manager then to inpututils and somehow change it to be static

app/digitizingcontroller.h
tomasMizera: See comment below: https://github.com/lutraconsulting/input/pull/1531#discussion_r662282041

app/qml/FormsStackManager.qml
wonder-sk: I see -- yeah let's keep it as it is for now... and maybe in the future we can have some class that has the project context in it, including variables manager, decoupled from digitizing controller...

PR #1545, opened 2021-07-07
Fixed value relation widget error

### LINKED ISSUE #1498: Multiple selections does not always work
We have found a small issue with Input App which we cannot resolve. 

Two of the fields in our form have an option for multiple select. Both work fine in QGIS but for some unknown reason one of them doesn't save selections in the Input app. 
You can select multiple options, click Done at the bottom to close the field but when you reopen your selection is not saved. Both fields have the exact same setting in form builder.

Please see the attached video.

I have made the project public on Mergin. It is called Thamesmead_v2, and fields are called Multiple Plant Species (works fine) and Multiple Animal Species (problematic).

https://user-images.githubusercontent.com/13314217/122428939-45e4ef80-cf8a-11eb-8c14-faa40f09f69b.MP4

### LINKED ISSUE #1522: Value input relation generates error
It happens for number fields with textedit widget. Can be replicated with project `tin1/forms_widgets`

TypeError:
`qrc:/editor/inputvaluerelation.qml:125: TypeError: Property 'startsWith' of object ... is not a function`

sklencar:
- The widgets generates the error while calling string function on QVariant (with number value): closes #1522 
- Removed white spaces from value comparison: closes #1498

### DIFF
--- a/app/featureslistmodel.cpp
+++ b/app/featureslistmodel.cpp
@@ -356,7 +356,7 @@ QVariant FeaturesListModel::attributeFromValue( const int role, const QVariant &
   for ( int i = 0; i < mFeatures.count(); ++i )
   {
     QVariant d = data( index( i, 0 ), role );
-    if ( d == value )
+    if ( d.toString().trimmed() == value.toString().trimmed() )
     {
       QVariant key = data( index( i, 0 ), requestedRole );
       return key;
--- a/app/qml/editor/inputvaluerelation.qml
+++ b/app/qml/editor/inputvaluerelation.qml
@@ -100,7 +100,7 @@ Item {
     * in order to get values as
     */
   function getCurrentValueAsFeatureId() {
-    if ( allowMultipleValues && widgetValue != null && widgetValue.startsWith('{') )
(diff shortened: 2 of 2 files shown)


app/featureslistmodel.cpp
```diff
@@ -356,7 +356,7 @@ QVariant FeaturesListModel::attributeFromValue( const int role, const QVariant &
   for ( int i = 0; i < mFeatures.count(); ++i )
   {
     QVariant d = data( index( i, 0 ), role );
-    if ( d == value )
+    if ( d.toString().trimmed() == value.toString().trimmed() )
```
sklencar, line 359: Add comment why `trimmed` is used

jozef-budac:
It looks like it fixes https://github.com/lutraconsulting/input/issues/1498, but I don't have original project Thamesmead_v2 to verify. Thamesmead_v3 project have changed configuration and use just TextEdit widget for animals. After changing the configuration to value relations, it looks it work fine.

In input-testing/forms_widget, the multiple choice is still not working, but it is a different problem (probably miss-configuration)

tomasMizera (approved):
It is working for me. Please consider the comparators and I will merge

Issue #1536, opened 2021-07-05
Issues in 0.9.9 - 2.15.210702185053 testflight


jozef-budac:
1) terms and conditions on SignIn page are not working (404 error)

2) "learn more" on subscription page still doesn't work (404 error)

3) subscription doesn't work for individial same as professional

4) there is sporadic "undo - undo" on home screen
![IMG_0051](https://user-images.githubusercontent.com/83582181/124434545-49db8480-dd74-11eb-91ed-bc5681e0795b.PNG)
I don't have way to simulate it. It just randomly appears. I suspect, it can be something with subscription, because I tried to subscribe multiple times and now this errors appears randomly when I am working with INPUT

5)  "Reuse last value" feature doesn't work. After first record, the check boxes disappears.

6) the infinite range slider doesn't work

7) "Number is outside of specific range" - is not set for translation

jozef-budac:
8) for relations, I can't add photo for referencing feature. Steps:
 - open input-testing/tc05_relations project
 - in 'pois' layer open a feature and edit it
 - add to attached photos
 - try to take a picture with camera or choose from a gallery
 - the photo widget is not updated with a picture

sklencar:
Points 1, 5, 8 addressed in https://github.com/lutraconsulting/input/pull/1549

Some notes to reported issues:
2. What link should be used instead?
3. Please, provide more info about the issue. Is it a case of https://github.com/lutraconsulting/input/issues/1475 ? 	
4. Regarding "undo - undo" on home screen- It can be replicated by shaking an iOS device - you can read more and change the setting in Settings -> accessibility -> Touch -> Shake to undo
6. The infinite range slider doesn't work - QML Slider has limitation and does not work properly with wide range and small step size. Such case suppose to be rather rare and as Martin suggested, this bug can be ignored for a time-being. It would require custom implementation of slider widget.
7. I could not find that specific string, however found similar "Number is outside of specific range" which seems to have translation and is wrapped in translation function. Can you please be more specific
8. Image default path is set incorrectly(home/tomasmizera/lutra/projects/qgis-projects/transportation) - photo field config should be fixed for a project input-testing/tc05_relations. However, notification has been added if an error occurs during photo/gallery image handling.

9. and 10. kept for another round.

wonder-sk:
For 9+10 there's now a ticket #1550

jozef-budac:
3. Please, provide more info about the issue. Is it a case of #1475 ?
-> yes it is

[...]

PR #1549, opened 2021-07-09
Some bugfixes

sklencar:
- changed TOS link
- fixed remember last value checkboxes
- notification about incorrect photo path

### DIFF
--- a/app/androidutils.cpp
+++ b/app/androidutils.cpp
@@ -214,11 +214,8 @@ void AndroidUtils::handleActivityResult( int receiverRequestCode, int resultCode
 
     QAndroidJniObject errorJNI = data.callObjectMethod( "getStringExtra", "(Ljava/lang/String;)Ljava/lang/String;", RESULT_STRING.object<jstring>() );
     // Internal cancelation due to an error
-    if ( !errorJNI.isValid() )
-    {
-      QString errorMsg = errorJNI.toString();
-      showToast( errorMsg );
-    }
+    QString errorMsg = errorJNI.toString();
+    showToast( errorMsg );
     return;
   }
 
--- a/app/qml/RegistrationForm.qml
+++ b/app/qml/RegistrationForm.qml
@@ -212,7 +212,7 @@ Rectangle {
         height: fieldHeight
         width: registerForm.width
         text: qsTr("I accept the Mergin %1Terms and Conditions%3 and %2Privacy Policy%3")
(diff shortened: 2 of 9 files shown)


saberraz:
Related to https://github.com/lutraconsulting/input/issues/1536

PeterPetrik (approved):

jozef-budac:
Fixes were verified

app/qml/RegistrationForm.qml
```diff
@@ -212,7 +212,7 @@ Rectangle {
         height: fieldHeight
         width: registerForm.width
         text: qsTr("I accept the Mergin %1Terms and Conditions%3 and %2Privacy Policy%3")
-              .arg("<a href='"+__merginApi.apiRoot + "assets/mergin_terms_of_service.pdf'>")
+              .arg("<a href='"+__merginApi.apiRoot + "assets/tos.html'>")
```
tomasMizera, line 215: There is already `__inputHelp.merginTermsLink`, let's use that one instead

app/qml/form/FeatureForm.qml
```diff
@@ -197,7 +197,6 @@ Item {
     else
     {
       controller.create()
-      state = "Edit"
```
tomasMizera, line 200: Why is this removed?

app/androidutils.cpp
```diff
@@ -214,11 +214,8 @@ void AndroidUtils::handleActivityResult( int receiverRequestCode, int resultCode
...
     QAndroidJniObject errorJNI = data.callObjectMethod( "getStringExtra", "(Ljava/lang/String;)Ljava/lang/String;", RESULT_STRING.object<jstring>() );
     // Internal cancelation due to an error
-    if ( !errorJNI.isValid() )
-    {
-      QString errorMsg = errorJNI.toString();
-      showToast( errorMsg );
-    }
+    QString errorMsg = errorJNI.toString();
+    showToast( errorMsg );
     return;
```
tomasMizera, line 219: Can you give some info why is this check removed now?

app/qml/RegistrationForm.qml
sklencar: Good point! However, `__inputHelp.merginTermsLink` uses a hardcoded Mergin api root pointing at public Mergin server. If we want to support CE Mergin server, its probably better to use current client's mergin api root (`__merginApi.apiRoot`).

Issue #1533, opened 2021-07-02
Issues in master branch 8a32b03


jozef-budac:
1) opening complex forms are sometimes laggy
2) 'Reuse last value' feature is not working
3) After restarting the app, layer which is not present in the them, can be set as default for recording. This can be very confusing.

https://user-images.githubusercontent.com/83582181/124245443-49e34680-db20-11eb-845e-af27516d42f4.mp4

jozef-budac:
4) if there is a validation problem on a feature and I delete a feature, the validation error remains on the new feature.
Steps to reproduce:
 - open project input-testing/decimals
 - go to default value tab, set 11 into the 2nd field 
 - not delete the feature
 - open new feature and you are not able to save it, even when there is not constrain violated

https://user-images.githubusercontent.com/83582181/124250550-456d5c80-db25-11eb-88c6-43fb6b34e478.mp4

jozef-budac:
5) there is a problem with value relations, that after user searches for value and click on it, it is not set into the field
 Steps to reproduce:
 - open project tomas/VR-Filtering
 - create new feature with all data
 - edit the feature and and in Car-Manufactured search for Audi, and click on the result. 
 - No value is set into the field


https://user-images.githubusercontent.com/83582181/124252974-d5aca100-db27-11eb-96f5-3633325751bf.mp4

jozef-budac:
6) slider widgets for infinite values are working strangely. 
open project input-testing/decimals and go to 2nd tab

https://user-images.githubusercontent.com/83582181/124254675-98e1a980-db29-11eb-9d46-6fa04c441a45.mp4

jozef-budac:
7) issue described in ticket https://github.com/lutraconsulting/input/issues/1502 is there again: 

https://user-images.githubusercontent.com/83582181/124262329-2f19cd80-db32-11eb-8021-fe1ccb97fe72.mp4

tomasMizera:
> 1. slider widgets for infinite values are working strangely.
>    open project input-testing/decimals and go to 2nd tab
> 
>  20210702_113350.mp4

There is a ticket for this one: https://github.com/lutraconsulting/input/issues/1516

jozef-budac:
Closing this tiket:
1) performance problem on older phones
2) fixed in latest master
3) fixed in latest master
4) separate ticket https://github.com/lutraconsulting/input/issues/1560
5) separate ticket https://github.com/lutraconsulting/input/issues/1559
[...]

Issue #1355, opened 2021-04-24
Menu with overlapping texts


saberraz:
For German users, there is overlapping text (Zoom to project extent) on  moto g8:
![Screenshot_20210423-225226](https://user-images.githubusercontent.com/2649047/115949161-e4912a80-a4ca-11eb-9145-bddbeca897ef.png)

jozef-budac:
There is one more overlapping error with Diagnostic log. 
In German there is a "An die Entwickler senden" button and it goes out the borders

PeterPetrik:
we have new Toolbar in 2024.1.0, reopen if it is still an issue

PR #1552, opened 2021-07-15
Bugfixing #1355

### LINKED ISSUE #1543: Expression for default value evalues in opposite way
In the project input-testing/forms_derived there is field "Name" on tab "Data" with default value set by expression "if("Danger" IS 'false' ,'Bear safe','Bear dangerous')". 
Please note that to see Data tab you need to insert some value into "Information" tab into field "Value"

So it should set name "Bear safe" when Danger = FALSE and "Bear dangerous" then Danger = TRUE, but it do just the opposite.

sklencar:
From #1533 :
- [x] 2. 'Reuse last value' feature is not working - fixed in https://github.com/lutraconsulting/input/pull/1549
- [x] 3. After restarting the app, layer which is not present in the them, can be set as default for recording. This can be very confusing.
- [x] 7. issue described in ticket Multiline widget text disappears
	-> used onTextChanged for text widget
- [x] 8. closes #1543

### DIFF
--- a/app/attributes/attributecontroller.cpp
+++ b/app/attributes/attributecontroller.cpp
@@ -729,6 +729,12 @@ void AttributeController::recalculateDerivedItems( bool isFormValueChange, bool 
         }
       }
 
+      if ( !isFormValueChange ) // reset state for new feature layer pair
+      {
+        item->setState( FormItem::ValidValue );
+        changedFormItems << item->id();
+      }
+
       ++formItemsIterator;
     }
     updateFieldValuesValidity();
--- a/app/attributes/attributecontroller.cpp
+++ b/app/attributes/attributecontroller.cpp
@@ -729,12 +729,6 @@ void AttributeController::recalculateDerivedItems( bool isFormValueChange, bool 
         }
       }
 
-      if ( !isFormValueChange ) // reset state for new feature layer pair
-      {
-        item->setState( FormItem::ValidValue );
-        changedFormItems << item->id();
-      }
-
       ++formItemsIterator;
     }
     updateFieldValuesValidity();
--- a/app/loader.cpp
+++ b/app/loader.cpp
@@ -31,16 +31,17 @@
 
 const QString Loader::LOADING_FLAG_FILE_PATH = QString( "%1/.input_loading_project" ).arg( QStandardPaths::standardLocations( QStandardPaths::TempLocation ).first() );
 
-Loader::Loader(
-  MapThemesModel &mapThemeModel
-  , AppSettings &appSettings
-  , ActiveLayer &activeLayer
-  , QObject *parent ) :
+Loader::Loader( MapThemesModel &mapThemeModel
+                , AppSettings &appSettings
+                , ActiveLayer &activeLayer
+                , LayersProxyModel &recordingLayerPM
+                , QObject *parent ) :
 
   QObject( parent )
   , mMapThemeModel( mapThemeModel )
   , mAppSettings( appSettings )
   , mActiveLayer( activeLayer )
+  , mRecordingLayerPM( recordingLayerPM )
 {
   // we used to have our own QgsProject instance, but unfortunately few pieces of qgis_core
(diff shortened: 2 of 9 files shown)


jozef-budac:
There is still problem with `Reuse last value` feature. When this feature is turn on, in 2nd recorded feature, stored values are used in new form, but checkbox are missing.

https://user-images.githubusercontent.com/83582181/126116009-ffd90345-6aa1-4d4d-ae97-fb60046f00ec.mp4

jozef-budac:
There is still problem with 5. point
After I choose another Car manufacturer with search (or without), `Car model` should be cleared, but is isn't. 
When I store the feature, Car model is not stored (which is correct).

https://user-images.githubusercontent.com/83582181/126117306-23d333d0-397a-496c-949d-5a8a41830ffe.mp4

jozef-budac:
7. I have still the same problem with this PR


https://user-images.githubusercontent.com/83582181/126117965-98b9d69c-6641-4fa9-9915-09b0c8e0b301.mp4

tomasMizera:
> There is still problem with 5. point
> After I choose another Car manufacturer with search (or without), `Car model` should be cleared, but is isn't.
> When I store the feature, Car model is not stored (which is correct).
> 
>  20210719_090438.mp4

I think this one is fixed in #1545

tomasMizera:
Can you please rebase master? There are several fixes merged, thanks

sklencar:
PR rebased with master (some other bugfixes have been applied).

Issue 5. - It has been moved to separate issue https://github.com/lutraconsulting/input/issues/1559
Issue 7. - I cant replicate it. Can anyone confirm is still present?

app/loader.cpp
```diff
@@ -107,7 +108,13 @@ bool Loader::forceLoad( const QString &filePath, bool force )
...
     mActiveLayer.resetActiveLayer();
     mMapThemeModel.reloadMapThemes( mProject );
-    setActiveLayerByName( mAppSettings.defaultLayer() );
+
+    QgsVectorLayer *defaultLayer = mRecordingLayerPM.layerFromLayerName( mAppSettings.defaultLayer() );
+    if ( defaultLayer )
+      setActiveLayer( defaultLayer );
+    else
+      setActiveLayer( mRecordingLayerPM.firstUsableLayer() );
+
```
tomasMizera, line 117: Similar thing is happening here: https://github.com/lutraconsulting/input/blob/c8c8e23e584763ae36093f0ab33961a1ae628fbf/app/qml/main.qml#L211

Can we maybe erase one of these two?

app/attributes/attributecontroller.cpp
```diff
@@ -729,12 +729,6 @@ void AttributeController::recalculateDerivedItems( bool isFormValueChange, bool
         }
       }
 
-      if ( !isFormValueChange ) // reset state for new feature layer pair
```
tomasMizera, line 732: Have you checked if the state is reset and new features do not inherit it?

app/loader.cpp
sklencar: Similar, but I think a bit different. 
The former is called when a project is loaded and handles setting theme and layer accordingly.
The latter is used in `updateRecordToolbar` function which is called in several cases. Most relevant reason to keep it there is that it handles the case when:

1. user changes map theme which doesn't contain current active layer
2. starts to digitizing

Another option would be to replace/fix `Loader::setActiveLayerByName` used in `Loader::forceLoad`.
Essentially, the fix would be similar (checking if default layer is present in digitizing proxy layer model).

sklencar:
I am afraid that issue related with calculating validity and evaluating constrains is broken. Seems to be complex issue and thus rough estimate to revise all the process is rather counted in days.

Known issues:
- invalid state of field can be set to valid easily - value out of range can be reset just by clicking on another field and back to the field. Most likely, editingFinished is triggered which triggers following actions. It can be replicated as Jozef described.
- invalid default value is ignored for new features - if default value breaks the range from config (config range 0-10, 11 is set as default value), the feature can be saved anyway

PR #1553, opened 2021-07-16
Fixed of cascaded expression evaluation

### LINKED ISSUE #1543: Expression for default value evalues in opposite way

sklencar:
The issue seems to be related with evaluation of cascade expressions, when expression value depends on another expression value that has been evaluated just before.
Seems like updating expression context has fixed it.

closes #1543

### DIFF
--- a/app/attributes/attributecontroller.cpp
+++ b/app/attributes/attributecontroller.cpp
@@ -539,6 +539,8 @@ bool AttributeController::recalculateDefaultValues(
           if ( val != oldVal )
           {
             mFeatureLayerPair.featureRef().setAttribute( item->fieldIndex(), val );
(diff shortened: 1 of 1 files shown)


jozef-budac:
Verified that this PR fix https://github.com/lutraconsulting/input/issues/1543

tomasMizera (approved):

PR #1555, opened 2021-07-19
Temporary fix qgsquick check by comparing Input-SDK QGIS version

tomasMizera:
We are currently not using the most recent qgsquick version from QGIS and so qgsquick check is always failing. 

I changed it to be compared to the version from Input-SDK https://github.com/lutraconsulting/input-sdk/blob/2d4d7623320df72c1c779196d62ccd47305caba5/android/recipes/qgis/recipe.sh#L10

### DIFF
--- a/.github/workflows/qgsquick.yml
+++ b/.github/workflows/qgsquick.yml
@@ -10,6 +10,10 @@ jobs:
       - name: download qgis
         run: |
           git clone https://github.com/qgis/QGIS.git --depth 1
+          cd QGIS
+          git fetch --depth 1 origin 1d17bf5bd35d7872f53c8e1c8b0a1e371616bf07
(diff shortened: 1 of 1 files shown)


.github/workflows/qgsquick.yml
```diff
@@ -10,6 +10,10 @@ jobs:
       - name: download qgis
         run: |
           git clone https://github.com/qgis/QGIS.git --depth 1
+          cd QGIS
+          git fetch --depth 1 origin 1d17bf5bd35d7872f53c8e1c8b0a1e371616bf07
```
PeterPetrik, line 14: can we have this `1d17bf5bd35d7872f53c8e1c8b0a1e371616bf07` as env variable on top of the file?

PeterPetrik (approved):

PR #1561, opened 2021-07-21
Add relation reference editor

### LINKED ISSUE #1550: Add "Reference relation" widget
For linked features that have "reference relation" widget assigned to an attribute in form configuration, Input currently only shows a simple text edit with the linked feature ID. It is desirable to have the "reference relation" widget implemented, so that:
- user can see the textual representation of the linked feature, not just its ID
- user can click on the linked feature to open its attribute form
- [in edit mode] user can open the list of features and change the link to some other feature
- [in edit mode] user can unlink existing feature (if "Allow NULL value" option is checked in the widget config)

It would be similar to Value Relation widget, with few extra features...

tomasMizera:
PR adds new relation reference editor. Child features in relations use this editor to reference parent feature.

Parent feature now shows feature title (if set) instead of its ID. Upon clicking on parent name, Input opens parent's feature form.
Field also contains icon that manages the linked parent, supports changing different parent and completely unlinking the child feature from any parent (if `Allow Null` is allowed for relation reference widget in QGIS form setup). 

https://user-images.githubusercontent.com/22449698/126474059-2d4b311c-09f3-492a-8cbf-e76d48df8890.mp4

Relation reference is handled as just another editor, instantiated from `FeatureForm`. Each Relation reference editor instantiates helper cpp class called `RelationReferenceFeaturesModel`. It parses editor's config and loads features from parent layer.
For future ease of changing editor, `AbstractEditor` is introduced that serves as a base class for new editor (and overtime will be used for all other editors). It automatically sets colors, sizes and other general options that different editors have in common.

Resolves #1550


sklencar (approved):
Tested with sample project (tc05-relations), works and looks good.

PR #1572, opened 2021-07-28
Selective sync

### LINKED ISSUE #911: Selective sync
Similar to dropbox, it will be good to allow users to select which files/folders can be synchronised. For example when surveyors collecting several (1000s) of photos, due to the large number of files, synchronisation takes a very long time.

sklencar:
### Description
To prevent download images files to Input client from the server.

Selective sync is supported only on Input (client) and has to be configured manually ATM by defining a config file. The idea is to have a tool for setting a configuration up in the mergin plugin.

Note that only images files (not case sensitive extensions: *.jpg, *.jpeg, *.png) are restricted from the sync which are located at `input-selective-sync-dir` or its subdirectories - a client can upload images, but doesn't download images from the server according the configuration. However, all images files are listed in `.mergin.json`. 

The config file:
- named as `mergin-config.json`
- is located in the project directory
- has JSON format
- is rather generic so can be used not only for selective-sync
- contains following variables:
    - `input-selective-sync`: `true/false`
    - `input-selective-sync-dir`: `<relative-path-to-project-dir>`; if empty, project directory is considered for selective sync.

Related automatic tests have been added.

### To improve:
A) **`mergin-config.json` location**
Since Input ignores hidden files ATM, mergin-config.json is located in the project directory. The best would be to hide it either to .mergin folder or just make it as a hidden file, but with ability to sync. Note that it can be modified over time although Input suppose to only read it.

B) **The initial download use case**
When an Input client downloads a project for the first time, the mergin-config is suppose to be downloaded as well, thus selective sync is not configured at that state. The best would be request mergin config from a server in advance and then continue with download/sync having mergin-config context already set up. Or, request a mergin config as the very first file for the download if its listed in remotely added files.

C) **New status for images in the photo widget**
When some photos are excluded from the sync and therefore not present on client's drive, but actually are used as value for an attachment field, the external resource widget shows it as missing image. The better would be to notify user that the image is stored on server and it only hasn't been synced on the user's device due to selective sync.

close #911

### DIFF
--- a/app/test/testmerginapi.cpp
+++ b/app/test/testmerginapi.cpp
@@ -1543,7 +1543,7 @@ void TestMerginApi::testSelectiveSync()
 
   // Create photo files
   QDir dir;
-  QString photoPath( projectDir + "/do_not_upload" );
+  QString photoPath( projectDir + "/subdir" );
   if ( !dir.exists( photoPath ) )
     dir.mkpath( photoPath );
 
@@ -1555,10 +1555,10 @@ void TestMerginApi::testSelectiveSync()
 
   // Download the project and copy mergin config file containing selective sync properties
   downloadRemoteProject( mApiExtra, mUsername, projectName );
-  QString configFilePathExtra( projectDirExtra + "/.mergin-config.json" );
+  QString configFilePathExtra( projectDirExtra + "/mergin-config.json" );
   QVERIFY( QFile::copy( mTestDataPath + "/mergin-config-project-dir.json", configFilePathExtra ) );
 
-  // Sync event 1
+  // Sync event 1:
   // Client 1 uploads images
   uploadRemoteProject( mApi, mUsername, projectName );
   // Download project and check
@@ -1567,21 +1567,21 @@ void TestMerginApi::testSelectiveSync()
   QFile fileExtra( projectDirExtra + "/photo.jpg" );
   QVERIFY( !fileExtra.exists() );
 
-  QFile fileExtra1( projectDirExtra + "/do_not_upload/photo.jpg" );
+  QFile fileExtra1( projectDirExtra + "/subdir/photo.jpg" );
   QVERIFY( !fileExtra1.exists() );
 
   // Sync event 2:
-  // Client 2: upload an image
-  QString configFilePath( projectDir + "/.mergin-config.json" );
+  // Client 2 uploads an image
+  QString configFilePath( projectDir + "/mergin-config.json" );
   QVERIFY( QFile::copy( mTestDataPath + "/mergin-config-project-dir.json", configFilePath ) );
 
   QFile fileExtra2( projectDirExtra + "/" + "photoExtra.png" );
   fileExtra2.open( QIODevice::WriteOnly );
 
-  // Client  2 uploads a new image
+  // Client 2 uploads a new image
   uploadRemoteProject( mApiExtra, mUsername, projectName );
 
-  // Client 1: sync without Client 2 new image and without removing own images.
+  // Client 1 syncs without Client 2's new image and without removing own images.
   downloadRemoteProject( mApi, mUsername, projectName );
 
   QVERIFY( file.exists() );
@@ -1639,7 +1639,10 @@ void TestMerginApi::testExludeFromSync()
 
   QVERIFY( mApi->excludeFromSync( selectiveSyncDir + "/not-existing.jpg", config ) );
   QVERIFY( mApi->excludeFromSync( selectiveSyncDir + "/image.png", config ) );
+  QVERIFY( mApi->excludeFromSync( selectiveSyncDir + "/image.PNG", config ) );
   QVERIFY( mApi->excludeFromSync( selectiveSyncDir + "/image.jpg", config ) );
+  QVERIFY( mApi->excludeFromSync( selectiveSyncDir + "/image.JPG", config ) );
+  QVERIFY( mApi->excludeFromSync( selectiveSyncDir + "/image.jpeg", config ) );
   QVERIFY( mApi->excludeFromSync( selectiveSyncDir + "/image.JPEG", config ) );
   QVERIFY( mApi->excludeFromSync( selectiveSyncDir + "/subdir/image.jpg", config ) );
 
--- a/app/test/testmerginapi.cpp
+++ b/app/test/testmerginapi.cpp
@@ -108,7 +108,7 @@ void TestMerginApi::initTestCase()
   deleteRemoteProject( mApiExtra, mUsername, "testMigrateDetachProject" );
   deleteRemoteProject( mApiExtra, mUsername, "testSelectiveSync" );
 
-  deleteLocalDir( mApi->projectsPath() + "/testExludeFromSync" );
+  deleteLocalDir( mApi, "testExcludeFromSync" );
 }
 
 void TestMerginApi::cleanupTestCase()
@@ -1609,10 +1609,10 @@ void TestMerginApi::testRegister()
   QVERIFY( spy.wait( TestUtils::LONG_REPLY ) );
 }
 
-void TestMerginApi::testExludeFromSync()
+void TestMerginApi::testExcludeFromSync()
 {
   // Set selective sync directory
-  QString selectiveSyncDir( mApi->projectsPath() + "/testExludeFromSync" );
+  QString selectiveSyncDir( mApi->projectsPath() + "/testExcludeFromSync" );
 
   QList<QString> testFiles =
   {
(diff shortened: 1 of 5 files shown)


core/merginapi.cpp
```diff
@@ -2592,6 +2650,24 @@ bool MerginApi::isInIgnore( const QFileInfo &info )
   return sIgnoreExtensions.contains( info.suffix() ) || sIgnoreFiles.contains( info.fileName() );
 }
 
+bool MerginApi::excludeFromSync( const QString &filePath, const MerginConfig &config )
+{
+  if ( config.selectiveSyncEnabled )
+  {
+    QFileInfo info( filePath );
+    bool isImage = sIgnoreImageExtensions.contains( info.suffix() );
```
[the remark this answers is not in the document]
tomasMizera: Maybe better to rename to something more generic?

```suggestion
    bool isExcluded = sIgnoreExtensions.contains( info.suffix() );
```

core/merginapi.cpp
```diff
@@ -2592,6 +2650,24 @@ bool MerginApi::isInIgnore( const QFileInfo &info )
...
+    if ( config.selectiveSyncDir.isEmpty() )
+    {
+      return isImage;
```
tomasMizera, line 2661: In this case `isImage` is always false, right? Can we rather return `false` explicitly?

tomasMizera: Maybe that condition can go to the top

core/merginapi.cpp
```diff
@@ -1948,6 +1998,10 @@ void MerginApi::uploadInfoReplyFinished()
     QList<MerginFile> diffFiles;
     for ( QString filePath : transaction.diff.localAdded )
     {
+      // filter out files from selective sync
+      if ( excludeFromSync( filePath, merginConfig ) )
```
tomasMizera, line 2002: Hm, so we are not sending photos on the server?

core/merginapi.cpp
```diff
@@ -29,8 +29,10 @@
 #include <geodiff.h>
 
 const QString MerginApi::sMetadataFile = QStringLiteral( "/.mergin/mergin.json" );
+const QString MerginApi::sMerginConfigFile = QStringLiteral( "/.mergin-config.json" );
```
tomasMizera, line 32: Maybe we should also place it to `.mergin` subfolder?

core/merginapi.cpp
```diff
@@ -29,8 +29,10 @@
...
 const QString MerginApi::sDefaultApiRoot = QStringLiteral( "https://public.cloudmergin.com/" );
 const QSet<QString> MerginApi::sIgnoreExtensions = QSet<QString>() << "gpkg-shm" << "gpkg-wal" << "qgs~" << "qgz~" << "pyc" << "swap";
+const QSet<QString> MerginApi::sIgnoreImageExtensions = QSet<QString>() << "jpg" << "JPEG" << "png";
```
tomasMizera, line 35: From what I understood, Martin was talking about having extensions in the config file too, is that right @wonder-sk? So that in plugin you can ignore any particular file extension.

core/merginapi.cpp
```diff
@@ -258,6 +260,43 @@ QString MerginApi::getApiKey( const QString &serverName )
...
+    }
+    else
+    {
+      qDebug() << "MerginConfig: invalid content!";
+    }
+
+  }
+  else
+  {
+    qDebug() << "MerginConfig: cannot read a config file!";
```
tomasMizera, line 294: This will be printed for all projects that do not have config file, right? I was thinking about something less disturbing in that case like

```suggestion
    qDebug() << "MerginConfig: Project do not contain config file";
```
Or possibly no log at all

tomasMizera (changes requested):
Looks really good, good job!

core/merginapi.cpp
sklencar: If `selectiveSyncDir` is empty thus is not set, then the whole project dir is taken by default. In that case is it enough to check only if extension is from the exclude extension list.
If its defined, beside extension condition there is also an extra check for a location

core/merginapi.cpp
sklencar: I was thinking about it, but the config file cannot be excluded from the sync.

core/merginapi.cpp
sklencar: ATM its hardcoded according the discussion we had - most common images extensions are supported (not case sensitive), but the list can be easily extended.

core/merginapi.cpp
tomasMizera: Ok cool, I see now. Can you please add comment there stating how it behaves?

tomasMizera (approved):
Worked on my project. I think we are ready for merge / final review from @PeterPetrik ?

Can you please add demo config somewhere to the project so that we now in the future what should be there and not need to go through the code to find out? Thanks!

app/test/testmerginapi.cpp
```diff
@@ -106,6 +106,9 @@ void TestMerginApi::initTestCase()
   deleteRemoteProject( mApiExtra, mUsername, "testMigrateProject" );
   deleteRemoteProject( mApiExtra, mUsername, "testMigrateProjectAndSync" );
   deleteRemoteProject( mApiExtra, mUsername, "testMigrateDetachProject" );
+  deleteRemoteProject( mApiExtra, mUsername, "testSelectiveSync" );
+
+  deleteLocalDir( mApi->projectsPath() + "/testExludeFromSync" );
```
PeterPetrik, line 111: typo Exlude->Exclude

app/test/testmerginapi.cpp
```diff
@@ -1543,6 +1609,51 @@ void TestMerginApi::testRegister()
   QVERIFY( spy.wait( TestUtils::LONG_REPLY ) );
 }
 
+void TestMerginApi::testExludeFromSync()
```
PeterPetrik, line 1612: typo Exlude->Exclude

app/test/testmerginapi.cpp
```diff
@@ -1590,12 +1701,16 @@ void TestMerginApi::deleteLocalProject( MerginApi *api, const QString &projectNa
   LocalProject project = api->getLocalProject( MerginApi::getFullProjectName( projectNamespace, projectName ) );
   QVERIFY( project.isValid() );
   QVERIFY( project.projectDir.startsWith( api->projectsPath() ) );  // just to make sure we don't delete something wrong (-:
-//  QDir projectDir( project.projectDir );
```
PeterPetrik, line 1593: remove commented out code

app/test/testmerginapi.cpp
```diff
@@ -1590,12 +1701,16 @@ void TestMerginApi::deleteLocalProject( MerginApi *api, const QString &projectNa
...
-//  projectDir.removeRecursively();
 
   api->localProjectsManager().removeLocalProject( project.id() );
 }
 
+void TestMerginApi::deleteLocalDir( const QString &dirPath )
+{
+  QDir dir( dirPath );
+  QVERIFY( dir.removeRecursively() );
```
PeterPetrik, line 1711: ou, hoping noone will do error during development and try to delete `/` here ... what about deleting the local dir relative to the (hardcoded) temp path ? since we know the temp folder location where we store the test results?

app/test/testmerginapi.h
```diff
@@ -66,9 +66,13 @@ class TestMerginApi: public QObject
     void testMigrateProject();
     void testMigrateProjectAndSync();
     void testMigrateDetachProject();
+    void testSelectiveSync();
 
     void testRegister();
 
+    // mergin functions
+    void testExludeFromSync();
```
PeterPetrik, line 74: typo Exlude->Exclude

core/merginapi.cpp
```diff
@@ -258,6 +260,43 @@ QString MerginApi::getApiKey( const QString &serverName )
...
+
+      if ( !selectiveSyncDir.isEmpty() )
+      {
+        QDir rootDir( projectDir );
+        config.selectiveSyncDir = rootDir.canonicalPath() + "/" + selectiveSyncDir;
+      }
+    }
+    else
+    {
+      qDebug() << "MerginConfig: Invalid content of the config file!";
```
PeterPetrik, line 288: this should go to the log at least

core/merginapi.cpp
```diff
@@ -1781,8 +1820,14 @@ void MerginApi::startProjectUpdate( const QString &projectFullName, const QByteA
   transaction.diff = compareProjectFiles( oldServerProject.files, serverProject.files, localFiles, transaction.projectDir );
   CoreUtils::log( "pull " + projectFullName, transaction.diff.dump() );
 
+  MerginConfig merginConfig = parseMerginConfig( transaction.projectDir );
```
PeterPetrik, line 1823: const MerginConfig merginConfig

PeterPetrik:
created https://github.com/lutraconsulting/input/issues/1573

app/test/testmerginapi.cpp
sklencar: That is what has been done

app/test/testmerginapi.cpp
sklencar: Good point!

app/test/testmerginapi.cpp
PeterPetrik: sure, sorry

PR #1585, opened 2021-08-05
Add features with geometry from relations widget

### LINKED ISSUE #1551: Add geometry for feature with null geometry
Right now, if Input creates a new linked feature, it is created with null geometry. For spatial tables, it would be very useful to have a button "add geometry" in the attribute form, which would show the map and start digitizing of the feature's geometry.

tomasMizera:
This PR mainly adds possibility to create linked features in relation widget with geometry. 

Workflow: 
 1. Open form with relations widget 
 2. Click on `edit` (if it is form of existing feature)
 3. Click on `+ add` in relations widget
 4. If the linked layer has geometry, forms will hide and you will see map in record mode. After geometry is recorded, form will show up (including previously opened forms)
 5. If the linked layer is no geometry layer, just new form will be opened 

Technical details:
 - we have one map with different states (`view` / `edit geometry` / `record features` / `record features in specific layer only` <- used to create linked features )
 - we hide opened forms (close them without destroying, in state `hidden` ) when geometry is being edited or new feature is being recorded 

Note: PR also fixes previously not working geometry editing.

I will add video asap

Closes #1551

### DIFF
--- a/app/qml/FormsStackManager.qml
+++ b/app/qml/FormsStackManager.qml
@@ -27,11 +27,11 @@ Item {
   property var project
   property real previewHeight
 
-  property int activeFormIndex: 0
+  property int activeFormIndex: formsStack.depth - 1
 
   signal closed()
   signal editGeometryRequested( var pair )
-  signal createLinkedFeatureRequested( var parentController, var relation )
+  signal createLinkedFeatureRequested( var targetLayer, var parentPair )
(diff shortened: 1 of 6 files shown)


PeterPetrik (approved):
Hi, maybe add manual test with cooperation with @jozef-budac ?

PR #1606, opened 2021-08-24
Selective sync fixes

### LINKED ISSUE #1608: Selective sync deletes projects directories on the mergin
Using selective synchronization can lead to deleting project directories in mergin portal. 

I have a project with 'mergin-config.json' configuration file, 'photos_a' directory will not synchronize to the INPUT app.

`{
  "input-selective-sync": true,
  "input-selective-sync-dir": "photos_a" 
}`

This works correctly

Then I change configuration not to use selective synchronization in mergin portal (or remove the configuration file completely):

`{
  "input-selective-sync": false,
  "input-selective-sync-dir": "photos_a" 
}`

After this, INPUT firstly synchronize new configuration file.
Then want to synchronize again, and it detects, that it doesn't have `photos_a` directory and it removes it from the mergin server. 
The expectation is, that the directory will be downloaded from the mergin server.

Steps to reproduce:
1) download project input-testing/tc10_selective_sync to the INPUT
2) in the mergin project set input-selective-sync: false in 'mergin-config.json'  OR remove the config file 
3) synchronize the project on the INPUT (twice)
4) you can check that "photos_a" directory is deleted in project

### LINKED ISSUE #1609: Selective sync - local changes shows files from disabled directory
When selective synchronization is enabled, `Local changes` shows photos/pictures from a directory, which were not synchronized.

My expectation is, that this pictures would not be part of local changes waiting to synchronize to the server.

Steps to reproduce:
1) download project 
2) input-testing/tc10_selective_sync
3) in home page on INPUT, hit 3. dots and then 'Local changes'

screenshot_20210825-105400

tomasMizera:
Several issues were found for Selective Sync, this PR fixes:
 - [x] using relative paths instead of absolute paths
 - [x] fix #1609 
 - [x] fix #1608 
 - [x] add tests covering different `mergin-config.json` scenarios 

----
### Description for relative paths:

There was an issue in selective sync. We were using absolute paths when deciding if file should or should not be synced. This is a problem mainly when downloading project for the first time -> files are not on the device yet, so we can not compare their absolute paths.

So previously we compared absolute paths like:
`/home/tomas/..../assets/projects/projectA/dcim` (config path) to
`/home/tomas/..../assets/projects/projectA/dcim/photoA.jpg` (photo path) <- this was an issue, when file is not locally available, Qt FileInfo() returned some default path.

And now we just compare paths relative to project:
`dcim` (config path) to `dcim/photoA.jpg` (photo path)

---
### Other fixes

The logic behind selective sync was changed significantly. We no longer exclude files from calculated diff, but now we do not include those files to diff at all (comparison of project files was changed).

**Logic to check for correct configuration:** 

selective-sync

Selective sync can now also be disabled in code by setting `apiSupportsSelectiveSync` to `false` on `MerginApi` instance.

---
This PR must go to release.


tomasMizera:
@wonder-sk tests added

core/merginapi.cpp
```diff
@@ -2359,7 +2477,7 @@ ProjectDiff MerginApi::compareProjectFiles( const QList<MerginFile> &oldServerFi
       if ( chkOld == chkLocal )
       {
         // R-D
-        diff.remoteDeleted << filePath;
+        diff.remoteDeleted << filePath; // here (ak si to nikdy nemal, ani to nemaz) -> skusim to bez toho
```
wonder-sk, line 2480: let's translate the comment to English :-)

core/merginapi.cpp
```diff
@@ -2451,7 +2569,27 @@ ProjectDiff MerginApi::compareProjectFiles( const QList<MerginFile> &oldServerFi
...
+              lastSyncConfig.isValid &&
+              MerginApi::excludeFromSync( file.path, lastSyncConfig );
+
+          if ( previouslyIgnoredButShouldDownload )
+          {
+            diff.remoteAdded << file.path;
+            continue;
+          }
+        }
+        diff.localDeleted << file.path; // here (nechceme oznacit fotky, ktore nemame lokalne, za vymazane, su na serveri)
```
wonder-sk, line 2592: and here

core/merginapi.cpp
```diff
@@ -2462,7 +2600,14 @@ ProjectDiff MerginApi::compareProjectFiles( const QList<MerginFile> &oldServerFi
...
       // R-A
-      diff.remoteAdded << file.path;
+      if ( allowConfig )
+      {
+        if ( MerginApi::excludeFromSync( file.path, config ) )
+        {
+          continue;
+        }
+      }
+      diff.remoteAdded << file.path; // here (nechceme stiahnut fotky, ktore ignorujeme a pridal ich niekto iny)
```
wonder-sk, line 2610: and here

core/merginapi.h
```diff
@@ -191,6 +182,8 @@ class MerginApi: public QObject
     Q_PROPERTY( MerginSubscriptionInfo *subscriptionInfo READ subscriptionInfo NOTIFY subscriptionInfoChanged )
     Q_PROPERTY( QString apiRoot READ apiRoot WRITE setApiRoot NOTIFY apiRootChanged )
     Q_PROPERTY( bool apiSupportsSubscriptions READ apiSupportsSubscriptions NOTIFY apiSupportsSubscriptionsChanged )
+    // apiSupportsSelectiveSync if true, fetches mergin-config.json in project and changes sync behavior based on its content (selective sync)
+    Q_PROPERTY( bool apiSupportsSelectiveSync READ apiSupportsSelectiveSync NOTIFY apiSupportsSelectiveSyncChanged )
```
wonder-sk, line 186: may I suggest a different name for this? because the above "apiSupportsSubscriptions" is really about whether Mergin server has API for dealing with subscriptions, while selective sync does not deal with remote API. Maybe just drop "api" and have "supportsSelectiveSync" ?

wonder-sk (commented):
:clap:  Nice!

core/merginapi.cpp
tomasMizera: my bad 🤦🏻‍♂️

PeterPetrik:
@tomasMizera would it be possible to add "master diagram" of the flow tot the devel documenation in the repo? (ideally with the modifiable source for future editions).

tomasMizera:
@PeterPetrik it should be there https://github.com/lutraconsulting/input/tree/master/docs/developers/code/media

PR #1617, opened 2021-08-31
Discover relations in autogenerated config

### LINKED ISSUE #1568: Auto-generated form does not include relation widget
If I use the default option to have form auto-generated (not the "drag&drop designer" option), in QGIS relation widget is automatically included in the form, but in Input it is not.

tomasMizera:
Autogenerated form layouts by default do not include relations that are in a project. 

This PR adds logic, that each time layer is changed in the `AttributeController` (and the new layer has autogenerated layout) Input uses `RelationManager` to discover all relations for such layer. We want both parent and child relations of this layer because `relation reference` widget was also wrongly being assigned as `range` (because config contained no information about the field, so we considered it as numeric). 

Resolves #1568


app/test/testformeditors.cpp
```diff
@@ -156,3 +157,67 @@ void TestFormEditors::testNumericFields()
...
+
+  const TabItem *tabItem = controller.tabItem( 0 );
+  QVector<QUuid> formItems = tabItem->formItems();
+
+  for ( const auto &itemId : formItems )
+  {
+    const FormItem *item = controller.formItem( itemId );
+    if ( item->editorWidgetType() == QStringLiteral( "relation" ) )
+      containsRelation = true;
+  }
```
wonder-sk, line 201: would be good to check there's exactly one relation + there's no relation ref widget
(and the other way around for the other layer)

wonder-sk (approved):

PR #1630, opened 2021-09-10
Form validation

### LINKED ISSUE #493: Integers in GPKG displayed as "0"

### LINKED ISSUE #1183: Handling NULL values in Range

### LINKED ISSUE #1196: Allow rich text with hyperlink
For preview panel and feature form.

Use case: creating navigation to a point using URL builder and point geometry.

### LINKED ISSUE #1383: No notification for QR scan for incompatible type
According TC 04 :
F7. Try to scan a code with text value for a field with numeric type (num_qrcode). A notification message of incompatible type conversion should pop up.

No notification appears.

### LINKED ISSUE #1455: Text field limit is ignored for alphabetical characters
Then there is Text Edit widget type on QString field with max length 10 characters, during recording of the new feature INPUT limits the length for numbers but not for alphabetical characters. 
After saving the feature, the string is trimmed to 10 characters

https://user-images.githubusercontent.com/83582181/121004659-3e099c00-c78f-11eb-9006-543120ea7f6c.mp4

### LINKED ISSUE #1456: Decimal number is possible to set into integer field
Decimal number is possible to set into integer field

https://user-images.githubusercontent.com/83582181/121004749-57124d00-c78f-11eb-850c-3361d0419f02.mp4

### LINKED ISSUE #1461: Hard constrain applies only every second feature
INPUT version: 0.9.7.
project: input-testing/form_derived

Hard constrain applies only on every 2nd feature.

https://user-images.githubusercontent.com/83582181/121015476-5384c300-c79b-11eb-9e20-29dce21e4dfe.mp4

### LINKED ISSUE #1502: Multiline widget text dissapears

### LINKED ISSUE #1507: Handle invalid feature form values (unify constraints)
It would be great to change a little bit how we handle user input in forms. It is not the greatest experience to just remove data from field when they are invalid, e.g. when someone scans QR code with text into numeric field or types in `34,51hg` to numeric field. 

**Right now when user saves the form with invalid value, last valid value is saved.** This can be misleading, user thinks value A is saved, but actually some other value is saved. 

Pr #1500 introduced new way how to handle invalid values. There is a text under field name (previously used only for constraint description) that can show texts like _"Not a valid number"_ or _"Number out of range"_ for numeric fields. We can easily use this for all other fields.

Proposal:
 - Even when value is invalid, save it to `mFeatureLayerPair` in `AttributeController`
 - Let's have a boolean array (flags) with entry for each field to know if value in a field is valid
 - QGIS Constraints and invalid values could be unified:
    - Controller no longer serving hard and soft constraints, rather `errorMessage` for each field and `errorLevel (info, warning, error)`
    - Invalid value in field and hard constraint would be `error`
    - Soft constraints as `warnings`
 - Do not let user save form if there is any field with `error` state
 - `errorMessage` is set in `AttributeController` and shown under field name with color according to `errorLevel`

**TLDR** Let user type in any value (even invalid) to field and tell him it is an invalid value (and why). Do not let them save the form when there is an invalid value.

See more in #1500 comments

### LINKED ISSUE #1517: Validators for fields
best to setup some minimal framework so another validators can be added easily/testable and not mess up the attribute controller logic.

for example in the folder attributes/ create class fieldvalidator.[h/cpp] a method FormItem::Validity validate(const std::shared_ptr<FormItem>) and this implementation to be hidden in this class

OR could be just a static method in rangewidgethelper.h ?

See comment in https://github.com/lutraconsulting/input/pull/1500#discussion_r655933272

### LINKED ISSUE #1518: Range editable widget followup
Known issues from #1500 

- [x] it is possible to save when hitting save directly before commiting text (due to a Qt bug https://bugreports.qt.io/browse/QTBUG-94499?jql=status%20in%20(Reported%2C%20Open)%20AND%20text%20~%20%22qt%20inputmethod%20visible%22%20ORDER%20BY%20created%20DESC)
- [x] "invalid value text" disappears if you commit the same text twice
- [x] polish loop (when field has suffix, we need to center it)

~CRASH!: _also sometimes the app crashes when editing existing feature. An it crashed in a way, that even screen recorder crashes. So maybe there is a bigger problem (in android?). The project used for testing is input-testing/test_decimals_~

Double check these issues: https://github.com/lutraconsulting/input/pull/1500#pullrequestreview-690441941

### LINKED ISSUE #1541: Forms not saved when a field exceeds the character length
In QGIS, fields length (e.g. a text field with 120 character length) is not required when user fills the form, i.e. user can add a 200 characters length inside the field and QGIS happily saves the Geopackage. But in Input, forms do not get saved.

To replicate the issue:
1- Create a project with a layer that has a field with 2 characters length
2- Set the widget type to value relation
3- Within the value relation, add some values which exceed 2 characters length
4- In QGIS, you can select the values and save the form
5- In Input, when you edit the existing features and save the form, when you go back, your values have disappeared

### LINKED ISSUE #1542: Hard constrain are not applied after clearing the field value
When I set value in field with hard constrain and then clear the value, hard constrain is not applied to the field again.
This issue is not present in published version of the app. 

Steps to reproduce:
1) download input-testing/forms_derived
2) set all field in "Hard constrain" to 123 and then clear it
3) you are able to save the feature even when it volatiles all the constrains

https://user-images.githubusercontent.com/83582181/124556974-3fcf8980-de39-11eb-9a16-7d6d589b91fa.mp4

### LINKED ISSUE #1560: Constrain validation problem
Version: 57994275d071fe6efe7e4a2f830e9364c8cdabc2
OS: Android arm64-v8a

if there is a validation problem on a feature and I delete a feature, the validation error remains on the new feature.
Steps to reproduce:
 - open project input-testing/decimals
 - go to default value tab, set 11 into the 2nd field 
 - now delete the feature
 - open new feature and you are not able to save it, even when there is not any violated constrain 
 
https://user-images.githubusercontent.com/83582181/124250550-456d5c80-db25-11eb-88c6-43fb6b34e478.mp4

### LINKED ISSUE #1562: 'Reuse last value' checkbox clears the number field
Version: 0.9.9. Beta

Steps to reproduce:

1) download project input-testing/tc04_recording
2) set "Reuse last value" option to ON
3) record feature in LineString layer
4) set something into 'Lenght' field and then hit "Reuse last value" checkbox for the field
5) the field is cleared

https://user-images.githubusercontent.com/83582181/126756410-0e54dddf-ddd2-457a-914a-9635181c346b.mp4

### LINKED ISSUE #1563: Not null constrain is not applied on empty fields
Version: 0.9.9 Beta

Steps to reproduce:
1) download project input-testing/forms_derived
2) record the feature
3) on the first tab, set 12 to every field. 
4) now remove the first field .. you are able to set the field, even when it is empty (null) 
5) problem is also, that value is stored in this field, even when I remove it

https://user-images.githubusercontent.com/83582181/126760157-67ca1e5c-e23f-47b3-9eb4-8a4a263d438e.mp4

### LINKED ISSUE #1564: Unique constrain is triggered when feature is edited
version: 0.9.9 

When I have a feature with Unique constrains on the fields and I edit it, constrains are applied when I open the form. This should not happen as the only feature with field values is the one which is open. 
Probably the constrain compares this feature to itself.

Steps to reproduce:
1) download project input-testing/forms_derived
2) Set all values in the first tab "Hard constrains"
3) save the feature
4) edit the feature and see, that Unique constrains on 2nd and 5th field are triggered

### LINKED ISSUE #1592: Field is not saved when scrolling too far
Hi,
I'll start by saying that Input is awesome.
I am experiencing an issue when filling a long form. When I fill information in a field, the information is lost if I scroll too far, as you can see in this clip:

https://user-images.githubusercontent.com/88735158/128910627-658a7949-f157-4fc9-9983-283518248b26.mp4

tomasMizera:
PR introduces new way of form validation. We now save whatever value comes from user input and validate it based on several rules.

Validation step replaces previous hard/soft QGIS constraints checks and so called _value validity_ (renamed to validation/validationMessage). Current implementation merges all these three steps into one inside `FieldValidator` and enhances value validations.

**How it works:**
 - User types in / scans / selects from calendar some data
 - Data are sent via new signal (`editorValueChanged`) to controller
 - Emit signal about form value changed
 - Controller saves the value without any checks to responding field
 - Controller then runs `recalculateDerivedItems`, in which it:
   - evaluates default values
   - evaluates visibility for both tabs and fields
   - runs validation of fields, in which:
       - provided value is checked via `convertCompatible` check of QgsField
       - moreover, numerical fields are checked for additional: a) min/max limits b) double values stored in integer fields
       - hard QGIS constraints
       - soft QGIS constraints
       - if any of the 4 previous check fails, `validationMessage` is set accordingly from `appresources.h` and saving of the form is prevented (except for when soft constraints are violated) 
   - emits signals for fields

Todo:
- [x] fix tests
- [x] fix autoupdate fields
- [x] fix save button sometimes not responding correctly

---

**Related tickets**

Resolves #1507 
Resolves #1517 
Resolves #1592 
Resolves #1542 
Resolves #1563 
Resolves #1461 
Resolves #1502
Resolves #1518 
Resolves #1456
Resolves #1560 
Resolves #1564  
Resolves #1196
Resolves #1562 

Resolves #493 (slider will always have value, it can not be `NULL`)
Resolves #1183 

Resolves #1455
Resolves #1383


PeterPetrik:
wow, nice list of fixed bugs!

saberraz:
@tomasMizera this PR completely breaks the cascade forms for me: https://public.cloudmergin.com/projects/documentation/form_cascade/tree

app/attributes/fieldvalidator.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+  bool hardConstraintSatisfied = QgsVectorLayerUtils::validateAttribute( pair.layer(),  pair.feature(), item.fieldIndex(), errors, QgsFieldConstraints::ConstraintStrengthHard );
+  if ( !hardConstraintSatisfied )
+  {
+    validationMessage = constructConstraintValidationMessage( item, errors );
+    level = Error;
+    return false;
+  }
+
+  errors.clear();
```
wonder-sk, line 70: do we need to clear? isn't it supposed to be empty?

app/attributes/fieldvalidator.cpp
```diff
@@ -0,0 +1,259 @@
...
+}
+
+QString FieldValidator::constructConstraintValidationMessage( const FormItem &item, const QStringList &unmetConstraints )
+{
+  const QgsField field = item.field();
+  QgsFieldConstraints fldCons = field.constraints();
+  QStringList validationMessages;
+
+  bool hasNotNullConstraint = fldCons.constraints() & QgsFieldConstraints::ConstraintNotNull;
+  bool notNullViolated = unmetConstraints.contains( QStringLiteral( "value is NULL" ) );
```
wonder-sk, line 195: these messages from qgis are translatable unfortunately, but as long as we don't ship translations for qgis code, it should work :-) maybe at some point we should just make an internal copy of QgsVectorLayerUtils::validateAttribute() to provide "our" texts (or fix the function in qgis to provide better messages)

app/attributes/fieldvalidator.cpp
```diff
@@ -0,0 +1,259 @@
...
+  return true;
+}
+
+bool FieldValidator::validateTextField( const FormItem &item, QVariant &value, QString &validationMessage, ValidationMessageLevel &level )
+{
+  const QgsField field = item.field();
+
+  if ( !field.convertCompatible( value ) )
+  {
+    validationMessage = Resources::Texts::Validation::genericValidationFailed;
```
wonder-sk, line 89: My preference would be to have the error messages inline in the code, as it is easier to read the code + they are only used once in the code anyway (in tests it is best if the expected error messages are intentionally duplicated)

app/attributes/fieldvalidator.h
```diff
@@ -0,0 +1,56 @@
...
+     * \brief validate function validates value saved in feature inside pair for specific FormItem (field), validates value validity
+     * checks, hard and soft constraints from QGIS.
+     * \param pair feature within layer to be validated
+     * \param item form item, it describes which field are we validating
+     * \param validationMessage [out] message will be set if there is a problem with field, empty if field has a valid value
+     * \param level [out] importance level of the message, either error (feature can not be saved), warning (can be
+     * saved, but have problems) or info either for no error or for helping user understand what field accepts.
+     * \return true if have valid value - no error nor warning, false otherwise
+     */
+    static bool validate( const FeatureLayerPair &pair, const FormItem &item, QString &validationMessage, ValidationMessageLevel &level );
```
wonder-sk, line 46: is it useful to have the "bool" return value? the client code needs to check "level" anyway to see if the issue was a warning or an error... maybe simpler if the return value is void or to have the "level" as return value?

app/qml/editor/inputvaluerelation.qml
```diff
@@ -88,13 +88,13 @@ Item {
...
-      combobox.popup.close()
-    }
-  }
+//  function dataUpdated( feature ) {
+//    if ( !isReadOnly ) {
+//      vrModel.currentFeature = feature
+//      updateField()
+//      combobox.popup.close()
+//    }
+//  }
```
wonder-sk, line 97: remove?

wonder-sk (approved):

tomasMizera:
@wonder-sk All comments are resolved and tests added/fixed

PR #1632, opened 2021-09-16
Show correct subscription price

tomasMizera:
Reordered parsing of `v1/user/service`, we now first parse product data and after that subscription data.

Fixes issue when **new price**(37.99) for individual plan was being shown to users who subscribed prior to price change(should be 15.99). It was due to an order of parsing - we first parsed correct price from subscription data and after that parsed new product price which rewrote the correct price (15.99) to the new one (37.99).

### DIFF
--- a/core/merginsubscriptioninfo.cpp
+++ b/core/merginsubscriptioninfo.cpp
@@ -44,6 +44,24 @@ void MerginSubscriptionInfo::clear()
 
 void MerginSubscriptionInfo::setFromJson( QJsonObject docObj )
 {
+  // parse service.plan data
+  QJsonObject planObj = docObj.value( QStringLiteral( "plan" ) ).toObject();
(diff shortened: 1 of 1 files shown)


core/merginsubscriptioninfo.cpp
```diff
@@ -44,6 +44,24 @@ void MerginSubscriptionInfo::clear()
 
 void MerginSubscriptionInfo::setFromJson( QJsonObject docObj )
 {
+  // parse service.plan data
```
PeterPetrik, line 47: can you add a small comment here that why this needs to be parsed before the other block

PeterPetrik (approved):

tomasMizera:
iOS build failed due to a version, it is fixed in #1630

PeterPetrik:
this approach shows correct price but in EUR, not localized.

PR #1648, opened 2021-09-23
Update translation strings after validation

tomasMizera:
#1630 added around 10 new strings about validation state, this PR adds them to translations


tomasMizera:
It seems that it is already done

PR #1652, opened 2021-09-23
Fix nested relations - geometry check

### LINKED ISSUE #1624: Issue with the 1-to-many relations within 1-to-many relations
I have set up a survey as follows:
1- A survey point: showing the location of tree survey
2- Visit (non-spatial table): to be filled in by the surveyor for each visit to a tree: e.g. visit 2019, 2020, etc with some measurements.
3- Photo (non-spatial table): for each visit, the surveyor might want to take multiple photos.

The project works as expected in QGIS, but in Input only relations between 1 and 2 works. The relation between 2 and 3 does not show up at all.

See the attached project:
[tree-survey.zip](https://github.com/lutraconsulting/input/files/7104771/tree-survey.zip)

tomasMizera:
There was an issue that when you set up project with no geo layer having relation to other no geo layer, relations in forms did not work (you could not see any features).

This issue was actually in a way we check if feature is valid. For no geometry features, we **were always returning False!** 

Resolves #1624 

@wonder-sk could this potentially break something else?

### DIFF
--- a/app/featurelayerpair.cpp
+++ b/app/featurelayerpair.cpp
@@ -60,10 +60,10 @@ bool FeatureLayerPair::hasValidGeometry() const
 {
   Q_ASSERT( mLayer );
 
-  if ( !mFeature.hasGeometry() )
-    return false;
+  if ( !mFeature.hasGeometry() && !mLayer->isSpatial() )
+    return true;
 
-  if ( mFeature.geometry().type() != mLayer->geometryType() )
+  if ( !mFeature.hasGeometry() || mFeature.geometry().type() != mLayer->geometryType() )
     return false;
 
   return true;


app/featurelayerpair.cpp
```diff
@@ -60,10 +60,12 @@ bool FeatureLayerPair::hasValidGeometry() const
 {
   Q_ASSERT( mLayer );
 
-  if ( !mFeature.hasGeometry() )
-    return false;
+  bool isNoGeometryLayer = mLayer->geometryType() == QgsWkbTypes::NullGeometry;
```
wonder-sk, line 63: I would suggest to use `mLayer->isSpatial()` - its body allows both NullGeometry and UnknownGeometry.

saberraz:
Nice one. Works well with my test project.

wonder-sk (approved):

PR #1664, opened 2021-09-29
Extract value relations

### LINKED ISSUE #1640: Value relations after forms refactoring
Value relations had _hot reloading_ (update on data changed) feature before forms refactoring, however, after refactoring this led to several binding loops and many "hacky" fixes. 

Currently cascaded value relations are not working. 
We should have a dedicated `ValueRelationsFeaturesModel` like we have one for relations and relation references, to clean current `FeatureListModel`. 

+ _Sometimes_ search in value relation page is not working

Closing other related tickets in favour of this one. Some issues are in production, some coming from dev.

tomasMizera:
This PR extracts `ValueRelationFeaturesModel` from previous `FeaturesListModel` that contained overcomplicated logic mixing logic for `BrowseData` and `ValueRelations` in one class. `FeaturesListModel` is also a base class for `RelationFeaturesModel` and `RelationReferenceFeaturesModel` as it can be seen on image below.

`ValueRelation-`, `Relation-` and `RelationReference-` `FeaturesModels` are used in respective form editors to list features from desired layers. They override several methods from base class like methods that set up feature request, title field or general setup method where configs are parsed.

featuresModels

Resolves #1640 

- [x] add test for `ValueRelationFeaturesModel`
- [x] add test for `FeaturesModel`


app/inpututils.cpp
```diff
@@ -786,6 +786,26 @@ const QUrl InputUtils::getEditorComponentSource( const QString &widgetName, cons
     return QUrl( path.arg( QLatin1String( "textedit" ) ) );
   }
 
+  if ( widgetName == QStringLiteral( "valuerelation" ) )
+  {
+    QgsMapLayer *referencedLayer = QgsProject::instance()->mapLayer( config.value( "Layer" ).toString() );
+    QgsVectorLayer *layer = qobject_cast<QgsVectorLayer *>( referencedLayer );
```
PeterPetrik, line 792: const?

app/valuerelationfeaturesmodel.h
```diff
@@ -0,0 +1,59 @@
+/***************************************************************************
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+
+#pragma once
```
PeterPetrik, line 10: https://en.wikipedia.org/wiki/Include_guard ?

PeterPetrik (commented):
just a brief read. 👏

app/inpututils.cpp
```diff
@@ -786,6 +786,26 @@ const QUrl InputUtils::getEditorComponentSource( const QString &widgetName, cons
...
+
+    if ( layer )
+    {
+      int featuresCount = layer->allFeatureIds().count();
```
wonder-sk, line 796: This may be inefficient for layers with many features as allFeatureIds() iterates over the whole layer to extract feature IDs. A more efficient approach should be to do layer->dataProvider()->featureCount() - normally the count is already cached within the provider (and typically calculated faster)

app/featuresmodel.h
```diff
@@ -0,0 +1,126 @@
+/***************************************************************************
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+
+#pragma once
```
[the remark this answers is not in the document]
tomasMizera: I find it easier to use, especially when class is being renamed, but sure I can change it.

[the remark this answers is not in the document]
tomasMizera: Hmm, I found that we are using `pragma` in few other classes too, but in a weird way, e.g.:

`MerginSubscriptionType`:

```c++
#ifndef MERGINSUBSCRIPTIONTYPE_H
#define MERGINSUBSCRIPTIONTYPE_H
#pragma once
```

Is it good for something to use both variants? :O

app/valuerelationfeaturesmodel.h
tomasMizera: same

tomasMizera:
Please do not merge, I found one issue that needs to be fixed!

app/featuresmodel.h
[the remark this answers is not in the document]
PeterPetrik: I think it is either that or other one @wonder-sk which one we should stick to ? (QGIS uses guards, so that would be my preference )

PR #1681, opened 2021-10-04
add form validation diagram

tomasMizera:
Added workflow (diagram) of form validation into developer docs.

form-validation


PeterPetrik:
can you add reference to the makrdown file too?

PeterPetrik (approved):

wonder-sk:
... and ideally also add the source file for the diagram in case we will need to modify it in the future :wink: 

Nice diagram!

tomasMizera:
Added both!

PR #1697, opened 2021-10-08
Fix value relations scientific notations bug

tomasMizera:
There was an issue that when `key` in value relation was numeric and the value was high enough to be rounded in scientific notation (e.g. > 10000), `QVariant::toString` automatically used scientific notation. Comparisons inside value relations model were then not working properly.

It seems that it only happens when number was sent from QML/JS, so I fixed it by first converting `fid` value to string in JS and then call C++ model function to convert it to `key` column.

### DIFF
--- a/app/qml/editor/inputvaluerelationcombobox.qml
+++ b/app/qml/editor/inputvaluerelationcombobox.qml
@@ -78,7 +78,10 @@ AbstractEditor {
     model: vrModel
 
     onItemClicked: {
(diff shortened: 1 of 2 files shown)


tomasMizera:
Good point. `.toString` converts number as is, so if the original Number would have scientific notation, the resulting string would have it too. However, JS is using scientific notation for numbers with over 20 digits, which I think is way more than sufficient for us. What do you think @wonder-sk ?

Issue #1698, opened 2021-10-08
Inverted GPS lat, lng saved in pictures taken on iOS


gilles-morain:
I have an ongoing field survey where enumerators collect polygons but are required to attach 2 pictures to the form. to check where the pictures were taken, my form grabs the GPS EXIF data on form save with read_exif_longitude() and read_exif_latitude(). Photos collected by Android phones get the coords right, while iOS devices report switched lat and long.

exiftool shows the wrong value also when checking the synced image, while a regular picture taken with the iPhone camera has the coordinates in the correct order. I do not get how thats's possible if iOS is in charge of saving the metadata with the picture.

gilles-morain:
Thanks! Meanwhile, for those affected, here is a procedure to swap coords in images affected:

1/ isolate all affected pictures in a dedicated folder (you do not want to swap coords from Android devices!)
2/ cd to that directory
3/ use exiftool like this (one liner for efficiency):

```
$ exiftool "-comment<GPSLatitude" -overwrite_original . ; exiftool "-GPSLatitude<GPSLongitude" -overwrite_original . ; exiftool "-GPSLongitude<Comment" -overwrite_original .
```

Notes: 
- it will overwrite the EXIF `Comment` field, so use it carefully
- it does not clean the `Comment` field afterwards so you end up with one longitude inside the comment field

PeterPetrik:
will be released in input 1.1.0

tomasMizera:
@gilles-morain you are indeed right, thanks for being helpful and patient, PR https://github.com/lutraconsulting/input/pull/1706 fixes the issue you mentioned.

PR #1706, opened 2021-10-14
Replace coordinate refs

tomasMizera:
We had mixed coordinates in iOS Exif. This is followup PR after #1699

See discussion in #1698, also see #1699


gilles-morain:
This fix apparently made it into Input iOS version 1.2 but not listed in the release note FYI

saberraz:
Thanks @gilles-morain for pointing this out. Fixed now!

Issue #1703, opened 2021-10-11
Issue with OnePlus and sync


saberraz:
It appears there is a bug specifically related to the OnePlus devices and sync.
The permissions are set correctly on the device (1+) and the project works perfectly on other phones. 

The only thing I can suspect of is the issue related to SQLite and Android version of OnePlus:
https://forums.oneplus.com/threads/unable-to-access-the-sqlite-database-from-our-app.429079/

PeterPetrik:
Still an issue with v2.4.0?

PR #1711, opened 2021-10-18
Scoped storage + support for active user + data migration

### LINKED ISSUE #116: Internal app files created in user folder

### LINKED ISSUE #289: Input project folder & Android cache cleaner apps
I noticed that Android cleaner apps consider Input Project folder as app cache location, so users unaware of that can loose their data. I also think that it could be good idea to allow users to configure project folder on some other location - SD cards for example.

### LINKED ISSUE #453: Sign in, Sign out
There appears to be no way to fully sign out of Input. For security reasons (loss of device/privacy of data) I would like to sign out of input after use. Even after signing out I can still get to home and add data to projects. It never really shuts down completely. Maybe this is only on testflight?

### LINKED ISSUE #1007: support Andoid 11+ scoped storage
https://developer.android.com/training/data-storage/use-cases#opt-out-scoped-storage

we use workaround for android 10 with `android:requestLegacyExternalStorage="true"`. 

we need to 
1. migrate data from /sdcard/INPUT to  Context#getExternalCacheDir()
2. use  Context#getExternalCacheDir() for the future

### LINKED ISSUE #1100: Input hangs when granting storage permission from home screen
To reproduce:
1- Make sure you clear storage and cache (so that it will prompt you with the permissions).
2- Deny permission to Storage
3- When taken to the Home screen, select the option to grant storage setting.
4- The screen below appears and I have to force close Input:
Screenshot_20210118-112758

tomasMizera:
This PR makes Input compatible with scoped storage model. 

Summary of changes:
 - Android now considers EXIF data as sensitive (since they contain user's location), so we need to opt for new permission called _ACCESS\_MEDIA\_LOCATION_. Even though it is a runtime permission (one need to opt for it in code), it has no GUI element (is granted automatically without user's input / user do not see anything).
 - We no longer use _INPUT_ folder in user's internal storage, new location is provided by Android, most often: _Android/data/0\<or other number based on active user\>/uk.co.lutraconsulting_

As a side effect, we now support active user (When someone changes active user on Android device) - each user has his/her own projects, they do not collide anymore.

@wonder-sk @PeterPetrik do you have any idea for some autotests?

Resolves #1100 
Resolves #453 (if I understand the ticket description properly, support for active user will deal with this - at least on Android)
Resolves #1007
Resolves #116 
Resolves #289

Bunch of things to **consider** adding to manual tests (if not already exists): (cc @jozef-budac)
 - Test that captured photos are copied in the full resolution to the project folder
 - Test that photos from gallery are copied in full resolution
 - Test different Android versions - try devices with Android 9/10/11, see if photos are stored and synced
 - Test EXIF - see if data are read from captured photo / from gallery image (if it contains such data!)
 - Test Active user - test if two users can work on the same device 
 - Test data migration - simulate app update to version with scoped storage and see if your data were copied correctly

---

**Legacy Folder Migration logic**

We run data migration only for Android. In each application run we try to determine if app should do the migration based on rules described in following diagram. 

Copy and rename of legacy folder is done on different thread to not block UI.

---

**Screenshots**

<table>
  <tr>
    <td>Update window</td>
    <td></td>
  </tr>
  <tr>
    <td>General error</td>
    <td></td>
  </tr>
  <tr>
    <td>Insufficient space</td>
    <td></td>
  </tr>
 </table>

### DIFF
--- a/app/i18n/input_en.ts
+++ b/app/i18n/input_en.ts
@@ -83,27 +83,27 @@
 <context>
     <name>AndroidUtils</name>
     <message>
-        <location filename="../androidutils.cpp" line="229"/>
+        <location filename="../androidutils.cpp" line="233"/>
         <source>Without storage permission you will not be able to access previous projects</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <location filename="../androidutils.cpp" line="280"/>
+        <location filename="../androidutils.cpp" line="287"/>
         <source>Storage permission is permanently denied, please allow it in settings in order to load pictures from gallery</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <location filename="../androidutils.cpp" line="284"/>
+        <location filename="../androidutils.cpp" line="291"/>
         <source>Input needs a storage permission in order to load pictures from gallery</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <location filename="../androidutils.cpp" line="302"/>
+        <location filename="../androidutils.cpp" line="309"/>
         <source>Camera permission is permanently denied, please allow it in settings</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
-        <location filename="../androidutils.cpp" line="306"/>
+        <location filename="../androidutils.cpp" line="313"/>
         <source>We need a camera permission in order to take a photo</source>
         <translation type="unfinished"></translation>
     </message>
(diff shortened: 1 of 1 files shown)


tomasMizera:
Might be worth checking if this fixes #1703

PR #1752, opened 2021-11-11
Add attention banner

### LINKED ISSUE #1731: Action required banner
Add banner to application based on design doc: https://gitlab.cloud.lutraconsulting.co.uk/mergin/mergin/-/issues/799

varmar05:
PR adds 'Attention required' banner in Project Panel. Its purpose is to draw attention of user that there is some issue which needs to be resolved in Mergin dashboard.

Appearance of the banner depends on flag coming from server response (`/v1/user/service`). Status is refreshed on panel open action to ensure current state is reflected before sync.

Resolves #1731 

input_banner

### DIFF
--- a/core/merginsubscriptioninfo.cpp
+++ b/core/merginsubscriptioninfo.cpp
@@ -111,6 +111,8 @@ void MerginSubscriptionInfo::setFromJson( QJsonObject docObj )
     emit planProductIdChanged();
   }
 
+  // check if some user action is required
+  mActionRequired = docObj.value( QStringLiteral( "action_required" ) ).toBool();
   emit subscriptionInfoChanged();
 }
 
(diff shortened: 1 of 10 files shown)


PeterPetrik:
does it work with Mergin CE ?

core/merginsubscriptioninfo.cpp
```diff
@@ -111,6 +111,8 @@ void MerginSubscriptionInfo::setFromJson( QJsonObject docObj )
     emit planProductIdChanged();
   }
 
+  // check if some user action is required
+  mActionRequired = docObj.value( QStringLiteral( "action_required" ) ).toBool();
```
tomasMizera, line 115: As a rule of thumb, we only set property value when we know it has changed (and emit its specific signal):

```suggestion
  bool isActionRequired = docObj.value( QStringLiteral( "action_required" ) ).toBool();
  if ( isActionRequired != mActionRequired )
  {
    mActionRequired = isActionRequired;
  } 
```

tomasMizera (changes requested):
Great job!

varmar05:
> does it work with Mergin CE ?

yes, tested with CE server (banner never appears as it is not supported)

tomasMizera (approved):

tomasMizera:
I will test this on Monday on iOS and close it then

Issue #1581, opened 2021-08-03
GPS accuracy popup link is not working


tomasMizera:
Banner used for gps accuracy contains clickable link. Currently clicking on link _only sometimes_ works.

I saw that its internal state is changing each time position is changed (in streaming mode). MouseArea is enabled based on this state so I think this is the problem.

<img width=300 src="https://user-images.githubusercontent.com/22449698/128027505-b695f3e7-aa0a-4ce8-a266-4e0e182f84ec.jpg">

wonder-sk:
Ah yes, I wanted to report this a couple of times already :-) 

While at it, I think we should also change the color of the "Learn more" link, so it is more obvious it is a clickable link...

tomasMizera:
Also: you can open the link when the banner is hidden, when you hit the exact link place

tomasMizera:
Today this bug has occured again - I can click on a space where normally should be the link and even though banner is not visible, I am still redirected to the webpage

tomasMizera:
No longer there

tomasMizera:
Occured again

jozef-budac:
@tomasMizera  I can't reproduce this on my devices

PR #1758, opened 2021-11-17
Fix GPS accuracy link [#1581]

NEDJIMAbelgacem:
This PR fixes the bug reported in:
https://github.com/lutraconsulting/input/issues/1581
The problem was that the height of some components was not set properly.
I couldn't reproduce the link click when the banner is hidden. If the issue still happens, I will try to find a fix.


PeterPetrik:
@NEDJIMAbelgacem if you can recreate directly without using fork please

PR #1760, opened 2021-11-18
Fix gps accuracy link [#1581]

NEDJIMAbelgacem:
This PR fixes the bug reported in: https://github.com/lutraconsulting/input/issues/1581
The problem was that the height of some components was not set properly.
I couldn't reproduce the link click when the banner is hidden. If the issue still happens, I will try to find a fix.

### DIFF
--- a/app/projectwizard.cpp
+++ b/app/projectwizard.cpp
@@ -17,6 +17,8 @@
 #include "qgsvectorfilewriter.h"
 #include "qgsdatetimefieldformatter.h"
 #include <qgsmarkersymbollayer.h>
+#include <qgssymbol.h>
+#include <qgsmarkersymbol.h>
 #include "inpututils.h"
 
 ProjectWizard::ProjectWizard( const QString &dataDir, QObject *parent )
--- a/app/projectwizard.cpp
+++ b/app/projectwizard.cpp
@@ -16,7 +16,7 @@
 #include "qgsvectorlayer.h"
 #include "qgsvectorfilewriter.h"
 #include "qgsdatetimefieldformatter.h"
-#include <qgsmarkersymbol.h>
+#include <qgsmarkersymbollayer.h>
 #include "inpututils.h"
 
 ProjectWizard::ProjectWizard( const QString &dataDir, QObject *parent )
--- a/app/projectwizard.cpp
+++ b/app/projectwizard.cpp
@@ -16,8 +16,6 @@
 #include "qgsvectorlayer.h"
 #include "qgsvectorfilewriter.h"
 #include "qgsdatetimefieldformatter.h"
-#include <qgsmarkersymbollayer.h>
-#include <qgssymbol.h>
 #include <qgsmarkersymbol.h>
 #include "inpututils.h"
 
--- a/app/qml/Banner.qml
+++ b/app/qml/Banner.qml
@@ -72,7 +72,7 @@ Rectangle {
       textItem.rightPadding: InputStyle.innerFieldMargin
       textItem.text: "<style>a:link { color: " + banner.linkColor
             + "; text-decoration: underline; }</style>" +
-            qsTr("%1<br><a href='%2'>Learn more</a>").arg(banner.text).arg(banner.link)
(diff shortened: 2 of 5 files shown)


app/projectwizard.cpp
```diff
@@ -17,6 +17,8 @@
 #include "qgsvectorfilewriter.h"
 #include "qgsdatetimefieldformatter.h"
 #include <qgsmarkersymbollayer.h>
+#include <qgssymbol.h>
+#include <qgsmarkersymbol.h>
```
tomasMizera, line 21: Please remove the `qgsmarkersymbol.h` include, it is due to a mismatch of QGIS versions.

tomasMizera: Android build is using previous QGIS version where this file does not exist. That is why CI does not pass

app/qml/Banner.qml
```diff
@@ -71,7 +72,7 @@ Rectangle {
       textItem.rightPadding: InputStyle.innerFieldMargin
       textItem.text: "<style>a:link { color: " + banner.linkColor
             + "; text-decoration: underline; }</style>" +
-            qsTr("%1<br><a href='%2'>Learn more</a>").arg(banner.text).arg(banner.link)
+            qsTr("%1<br><a href='%2' style=\"color: darkblue;\">Learn more</a>").arg(banner.text).arg(banner.link)
```
tomasMizera, line 75: What is the reason for changing the color? If so, can you please attach a screenshot and put this color into `InputStyle.qml`?

tomasMizera (commented):
So right now the link is working? I once tried to debug this and I think I found out that states inside the Banner were not being set properly. Could you please check that? 

just like `onStateChanged` console log to see if they are properly set

app/qml/Banner.qml
NEDJIMAbelgacem: MartinD mentioned in a comment on the issue that we should change the color to look more like a link than a normal text.
Here is how it looks like with the new color:
Screenshot from 2021-11-18 22-35-08

NEDJIMAbelgacem:
The link now works with this PR. I think the states issue is unrelated to this.
What I did to debug this is I added a rectangle over the textItem in TextWithIcon.qml, by doing this I noticed that the used height fieldHeight was not right as you can see here:
Screenshot from 2021-11-18 22-48-20

app/projectwizard.cpp
NEDJIMAbelgacem: Oh I see. Can you please tell me which version of QGIS is used in the CI so that I can add #ifdef to support the later version I use ?

app/qml/map/MapWrapper.qml
```diff
@@ -450,7 +450,7 @@ Item {
     .arg( __inputUtils.formatNumber( _positionKit.accuracy ) )
     link: __inputHelp.gpsAccuracyHelpLink
 
-    showWarning: shouldShowAccuracyWarning
+    showWarning: true//shouldShowAccuracyWarning
```
tomasMizera, line 453: I guess you did not want to commit this change?

app/projectwizard.cpp
tomasMizera: Let's not add ifdefs for now. If you could just remove this code (I understand that for local desktop you will need to add this), upcoming aab PR is going to sort this thing out :)

tomasMizera:
@NEDJIMAbelgacem I tried to run the apk and it is still possible to click on an empty space (place that would be occupied by the banner) to open the link 🤨 

https://user-images.githubusercontent.com/22449698/142593714-e94e8ea6-e509-45ba-98ff-47380262c7c8.mp4

app/qml/map/MapWrapper.qml
NEDJIMAbelgacem: Yes - sorry about it

tomasMizera:
Hi @NEDJIMAbelgacem thanks, it looks good. However, there is one more thing I noticed: we have lost a fade out animation the banner.

https://user-images.githubusercontent.com/22449698/142819880-c9bb6b9c-5b86-4e38-889c-af55a98b83ae.mp4

tomasMizera (approved):

PR #1761, opened 2021-11-18
Add qml formatting

tomasMizera:
This PR adds action that checks formatting of QML files.

On macos, install qmlfmt via homebrew, on Linux you need to build it: https://github.com/jesperhh/qmlfmt

Let's first merge this PR and format everything in the next one.

There is a new script that we can run to format QML files: format_qml.bash. or we can use format_all.bash to format both c++ and qml files.


.gitignore
```diff
@@ -10,6 +10,7 @@ test/temp_extra_projects/
 input.pro.user*
 app/android/assets/qgis-data/proj_custom/
 app/android/assets/qgis-data/projects/
+scripts/format_all.bash
```
PeterPetrik, line 13: why this is in ignore?

scripts/format_all.bash.default
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
PeterPetrik, line 1: i would just copy it without "default"

PeterPetrik (approved):

PeterPetrik (approved):

Issue #1046, opened 2020-12-02
Use the project extent instead of WMS setting


saberraz:
In the past, we used WMS extent for setting the project extent in Input. We should now use this:
https://github.com/qgis/QGIS/pull/40310

saberraz:
See https://github.com/lutraconsulting/input/pull/732

PR #1762, opened 2021-11-22
use project extent instead of WMS extent [1046]

NEDJIMAbelgacem:
This PR addresses this issue:
https://github.com/lutraconsulting/input/issues/1046
I'm not sure if it's very correct to use project extent instead of WMS extent all the time and I'm just assuming it will just work...

### DIFF
--- a/app/loader.cpp
+++ b/app/loader.cpp
@@ -29,6 +29,8 @@
 #endif
 #include <QDebug>
 
+#include <qgsprojectviewsettings.h>
+
 const QString Loader::LOADING_FLAG_FILE_PATH = QString( "%1/.input_loading_project" ).arg( QStandardPaths::standardLocations( QStandardPaths::TempLocation ).first() );
 
 Loader::Loader( MapThemesModel &mapThemeModel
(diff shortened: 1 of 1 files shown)


saberraz:
The project extent should work (at least for the time being) both with the new extent the WMS extent. Eventually, we can remove the WMS extent.

wonder-sk:
The logic should be:
1. if project extent is set - use that.
2. else if WMS extent is set - use that (for backward compatibility)
3. else use full extent calculated from map layers.

It looks like you have removed the options 2+3 :-)

tomasMizera (approved):
looking good!

jozef-budac:
In the latest APK, WMS extent is not used as project extent and after hitting `Zoom to project` in input, maximum extent is used.

NEDJIMAbelgacem:
> In the latest APK, WMS extent is not used as project extent and after hitting `Zoom to project` in input, maximum extent is used.
> 
> Steps to reproduce:
> 
> 1. download tc03_map_canvas project
> 2. in INPUT press Zoom to project. It should show Bratislava. This is OK
> 3. In QGIS, remove project extent and synchronize
> 4. in INPUT hit Zoom to project. Not whole map is present, even when WMS extent is set to Prague city

Should be fixed now, I was using the wrong functions :'/

tomasMizera:
Let's wait for @jozef-budac to test out his project once again and if that is fine, let's merge this one :)

jozef-budac:
Latest update looks ok!

Issue #1263, opened 2021-03-15
Detect missing layers and show warnings


saberraz:
Users get confused when their layers do not appear on Input but shows up in their QGIS sessions. This at least one bug related to missing layer (#1257). In other occasions (I think if the missing layer is an online service) it takes a long time for any map to appear.

It would be good to show warnings and point the user to the packaging help documents, as this is very likely caused by incorrect packaging of layers/services.

jozef-budac:
Related to https://github.com/lutraconsulting/input/issues/580

PR #1763, opened 2021-11-22
Display notification when some layers are invalid [1263]

NEDJIMAbelgacem:
Addresses https://github.com/lutraconsulting/input/issues/1263

When some of the project's layers are invalid, a notification is displayed saying that some layers are invalid:
warning-on-missing-layers2

Maybe I should find a way to display a notification the same way the GPS accuracy warning is displayed?

### DIFF
--- a/app/loader.cpp
+++ b/app/loader.cpp
@@ -22,6 +22,8 @@
 #include "qgslayertreegroup.h"
 #include "qgsmapthemecollection.h"
 #include "qgsquickmapsettings.h"
+#include <qgsapplication.h>
+#include <qgslogger.h>
 
 #if VERSION_INT >= 30500
 // this header only exists in QGIS >= 3.6
@@ -94,6 +96,14 @@ bool Loader::forceLoad( const QString &filePath, bool force )
   flagFile.open( QIODevice::WriteOnly );
   flagFile.close();
 
+  QString logFilePath = CoreUtils::logFilename();
+  qint64 alreadyAppendedCharsCount = 0;
+
+  {
+    QFile file( logFilePath );
+    alreadyAppendedCharsCount = file.size();
+  }
+
   // Give some time to other (GUI) processes before loading a project in the main thread
   QEventLoop loop;
   QTimer t;
@@ -120,23 +130,39 @@ bool Loader::forceLoad( const QString &filePath, bool force )
     emit projectReloaded( mProject );
   }
 
-  QMap<QString, QgsMapLayer *> layersMap = mProject->mapLayers();
+  QString qgisLog;
+
+  QFile file( logFilePath );
+  if ( file.open( QIODevice::ReadOnly ) )
+  {
+    QByteArray logFileData = file.readAll();
+    QByteArray neededLogFileData( logFileData.data() + alreadyAppendedCharsCount );
+    qgisLog = QString::fromStdString( neededLogFileData.toStdString() );
+    file.close();
+  }
+
+  bool foundInvalidLayer = false;
   QStringList invalidLayers;
-  for ( QgsMapLayer *layer : layersMap.values() )
+  for ( QgsMapLayer *layer : mProject->mapLayers().values() )
   {
     if ( !layer->isValid() )
-      invalidLayers.push_back( layer->name() );
+    {
+      invalidLayers.append( layer->name() );
+      foundInvalidLayer = true;
+      emit reportIssue( layer->name(), "layer is invalid" );
+    }
   }
 
   flagFile.remove();
   if ( !force )
   {
     emit loadingFinished();
-    if ( !invalidLayers.isEmpty() )
+    if ( foundInvalidLayer )
     {
       QString message = QStringLiteral( "WARNING: The following layers are invalid: %1" ).arg( invalidLayers.join( ", " ) );
       CoreUtils::log( "project loading", message );
-      emit loadedInvalidLayer( invalidLayers );
+      emit submitQgisLog( qgisLog );
+      emit loadingErrorFound();
     }
   }
   return res;
--- a/app/loader.cpp
+++ b/app/loader.cpp
@@ -44,6 +44,7 @@ Loader::Loader( MapThemesModel &mapThemeModel
   , mAppSettings( appSettings )
   , mActiveLayer( activeLayer )
   , mRecordingLayerPM( recordingLayerPM )
+  , mQgisLog( "" )
 {
   // we used to have our own QgsProject instance, but unfortunately few pieces of qgis_core
   // still work with QgsProject::instance() singleton hardcoded (e.g. vector layer's feature
@@ -96,6 +97,8 @@ bool Loader::forceLoad( const QString &filePath, bool force )
   flagFile.open( QIODevice::WriteOnly );
   flagFile.close();
 
+  mQgisLog.clear();
+
   QString logFilePath = CoreUtils::logFilename();
   qint64 alreadyAppendedCharsCount = 0;
 
@@ -130,17 +133,6 @@ bool Loader::forceLoad( const QString &filePath, bool force )
     emit projectReloaded( mProject );
   }
 
-  QString qgisLog;
-
-  QFile file( logFilePath );
-  if ( file.open( QIODevice::ReadOnly ) )
-  {
-    QByteArray logFileData = file.readAll();
-    QByteArray neededLogFileData( logFileData.data() + alreadyAppendedCharsCount );
-    qgisLog = QString::fromStdString( neededLogFileData.toStdString() );
-    file.close();
-  }
-
   bool foundInvalidLayer = false;
   QStringList invalidLayers;
   for ( QgsMapLayer *layer : mProject->mapLayers().values() )
@@ -149,7 +141,7 @@ bool Loader::forceLoad( const QString &filePath, bool force )
     {
       invalidLayers.append( layer->name() );
       foundInvalidLayer = true;
-      emit reportIssue( layer->name(), "layer is invalid" );
+      emit reportIssue( layer->name(), layer->publicSource() );
     }
   }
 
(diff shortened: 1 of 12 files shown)


app/loader.cpp
```diff
@@ -120,9 +120,17 @@ bool Loader::forceLoad( const QString &filePath, bool force )
...
+  for ( QgsMapLayer *layer : layersMap.values() )
+  {
+    if ( !layer->isValid() )
+      invalidLayers.push_back( layer->name() );
+  }
+
   flagFile.remove();
   if ( !force )
-    emit loadingFinished();
+    emit loadingFinished( invalidLayers );
```
tomasMizera, line 133: Let's not couple `loadingFinished` signal with invalid layers. I suggest to introduce a new signal that will be emitted after loading is finished.

app/qml/main.qml
```diff
@@ -402,7 +402,23 @@ ApplicationWindow {
...
+            projectLoadingScreen.visible = false;
+            if ( invalidLayers.length > 0 )
+            {
+                let warningMessage = qsTr( "WARNING: the following layers are invalid: " );
+                for ( let i = 0; i < invalidLayers.length - 1; ++i )
+                {
+                    warningMessage += qsTr( "'%1' " ).arg( invalidLayers[i] );
+                }
+                if ( invalidLayers.length > 1 )
+                    warningMessage += qsTr( "and '%1' " ).arg( invalidLayers[invalidLayers.length - 1] );
```
tomasMizera, line 416: Similar to previous point. I am afraid when translators should add special characters

```suggestion
                    warningMessage += qsTr( "and" ) + " %1".arg( invalidLayers[invalidLayers.length - 1] );
```

jozef-budac:
I tried latest APK and in works strangely in one scenario:
1) create project with one additional layer in its own gpkg file
2) synchronize it with mergin and input
3) remove the gpkg with additional layer in mergin
4) synchronize the changes to INPUT. 
5) layer is still available for recording, but after pressing save, nothing happens (no feature is added). Also there isn't any notification of missing or wrong layer.

app/qml/main.qml
```diff
@@ -402,7 +402,23 @@ ApplicationWindow {
     Connections {
         target: __loader
         onLoadingStarted: projectLoadingScreen.visible = true
-        onLoadingFinished: projectLoadingScreen.visible = false
+        onLoadingFinished:
+        {
+            projectLoadingScreen.visible = false;
+            if ( invalidLayers.length > 0 )
+            {
+                let warningMessage = qsTr( "WARNING: the following layers are invalid: " );
```
tomasMizera, line 410: How about some less prominent message like "We looked everywhere but could not find some of your layers: .... <layers> .... see [documentation to learn how to package them properly](https://help.cloudmergin.com/plugin-new-project?_ga=2.122664654.477250255.1637567361-1897708595.1630658758)" with link to documentation.
@saberraz @wonder-sk do you have any suggestions for the message?

tomasMizera (changes requested):

app/loader.cpp
```diff
@@ -120,9 +120,17 @@ bool Loader::forceLoad( const QString &filePath, bool force )
     emit projectReloaded( mProject );
   }
 
+  QMap<QString, QgsMapLayer *> layersMap = mProject->mapLayers();
+  QStringList invalidLayers;
+  for ( QgsMapLayer *layer : layersMap.values() )
+  {
+    if ( !layer->isValid() )
+      invalidLayers.push_back( layer->name() );
```
PeterPetrik, line 128: maybe it would be worth to add it to the logger for debugging @wonder-sk ? `CoreUtils::warn( "Input", msg );`

app/qml/main.qml
```diff
@@ -402,7 +402,23 @@ ApplicationWindow {
...
+                for ( let i = 0; i < invalidLayers.length - 1; ++i )
+                {
+                    warningMessage += qsTr( "'%1' " ).arg( invalidLayers[i] );
+                }
+                if ( invalidLayers.length > 1 )
+                    warningMessage += qsTr( "and '%1' " ).arg( invalidLayers[invalidLayers.length - 1] );
+                else
+                    warningMessage += qsTr( "'%1' " ).arg( invalidLayers[0] );
```
PeterPetrik, line 418: - can you post a screenshot when there is more layers (so how it behaves when overflowing).
- Also I think we had some "warning message panel", doesn't map has other function than "notify"?

wonder-sk:
We should have something more sophisticated - showing that some layers are missing is a good start, but ideally we should also tell _why_ they are missing (which is something that normally gets reported in QgsMessageLog - we can tap to its output). Also, missing layers is not the only way how project can have some loading problems, so let's make it more flexible, with a full page screen for further details, something like this:

image

saberraz:
+1 for @wonder-sk suggestions. There could be problems with network (for online maps or layers within private network) not loading.

For grid datum shift, best to move the warning to this PR and get rid of the not very pleasant dialog window.

wonder-sk:
> For grid datum shift, best to move the warning to this PR and get rid of the not very pleasant dialog window.

I would suggest to handle datum shift error in a separate follow up PR to keep the scope limited, and schedule fixing datum shift message sometime soon...

app/qml/main.qml
NEDJIMAbelgacem: The overflow is handled correctly:
warning-not-overflowing resized

NEDJIMAbelgacem:
> I would suggest to handle datum shift error in a separate follow up PR to keep the scope limited, and schedule fixing datum shift message sometime soon...

Maybe we can make the GPS accuracy banner into a notification pop up ?
I think that way we can have specific warning messages with links to explanation material without crowding Input with these resources. One thing that might be good also is to have a panel containing all the pushed notifications.

This can be implemented in another PR as notification banner I think.

NEDJIMAbelgacem:
Here is what the new notification looks like and what it reports:

https://user-images.githubusercontent.com/29183781/143433974-9747baed-cc75-4016-829b-b1c0c40883bd.mp4

I added a signal to Loader to report issues that will be added to the project issues panel.
If an issue was found the notification banner will automatically show up.

app/qml/map/MapWrapper.qml
```diff
@@ -577,6 +584,15 @@ Item {
     }
   }
 
+  NotificationBanner {
```
tomasMizera, line 587: I would say let's leave map out from this one. The `NotificationBanner` can be instantiated in `main.qml`, it does not use anything from map canvas / map elements and in future can be used to display custom messages.

tomasMizera: So I vote for having the `notificationBanner` in `main.qml`. In that case `function pushNotification` can be inside `NotificationBanner`.

app/qml/main.qml
```diff
@@ -430,6 +448,12 @@ ApplicationWindow {
         target: __loader
         onLoadingStarted: projectLoadingScreen.visible = true
         onLoadingFinished: projectLoadingScreen.visible = false
+        onLoadingErrorFound: map.pushNotification( "There were issues loading the project." )
+
+        onReportIssue: projectIssuesPanel.reportIssue( layerName, message )
+        onSetProjectIssuesHeader: projectIssuesPanel.headerText = text
+        onQgisLogChanged: projectIssuesPanel.qgisLog = __loader.qgisLog()
```
tomasMizera, line 455: Good, however, this is still going to copy the log on project load. My idea was that only when user clicks on `details` in the notification, qml will ask for the log that is saved inside the loader.

tomasMizera (changes requested):

tomasMizera:
Overall this looks good, thanks for implementing the notes. There are two more points that should be addressed. However, these do NOT block testing so @jozef-budac feel free to check this one out

jozef-budac:
There are some problems.
1) Notification appears for every project, even without problems in layers. 
2) Notification appears even for welcome projects, but with empty description

https://user-images.githubusercontent.com/83582181/144014353-1267119b-8fa0-4174-8ee7-bfefaaee5698.mp4

app/qml/NotificationBanner.qml
```diff
@@ -0,0 +1,129 @@
...
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+import QtQuick 2.7
+import QtQuick.Layouts 1.3
+import QtGraphicalEffects 1.0
+import QtQuick.Controls 2.7
```
tomasMizera, line 12: One more note, please use most recent versions like in `ProjectIssuePanel.qml`

NEDJIMAbelgacem:
> 

Sorry about this - I missed to fix this. I addressed with the latest commit.

app/loader.cpp
```diff
@@ -230,6 +274,11 @@ bool Loader::layerVisible( QgsMapLayer *layer )
   return false;
 }
 
+QString Loader::qgisLog()
+{
+  return mQgisLog;
```
tomasMizera, line 279: Can we rename this `mQgisLog` to something like `mProjectLoadingLog` so that we know in future what it contains?

tomasMizera (approved):
Good job!

@wonder-sk if you are ok with the current design, we could merge this.
@jozef-budac would it be possible to retest this one? Thanks!

jozef-budac:
The issue is fixed, but I found another problem on signip page:
screenshot_20211202-083837

jozef-budac:
Also I have a question about how often and how long the notification should appear. 
1) the notification remains over all the pages in the app (home, shared projects etc.)
2) notification popups for every time project opens. I am not sure it this is not annoying for users. 


https://user-images.githubusercontent.com/83582181/144382291-a8c18345-3027-464a-bec0-28867d1aaf26.mp4

tomasMizera:
> The issue is fixed, but I found another problem on myaccount page: screenshot_20211202-083837

This is already reported here https://github.com/lutraconsulting/input/issues/1773

app/qml/main.qml
```diff
@@ -211,6 +211,20 @@ ApplicationWindow {
         }
     }
 
+    NotificationBanner {
+      id: notificationBanner
+
+      shouldHide: projectPanel.visible && settingsPanel.visible && stateManager.state !== "view"
```
tomasMizera, line 217: I can not see where this property is used or declared.

You should also probably use `||` instead of `&&`

wonder-sk:
Indeed good points @jozef-budac
- the notifications should only be visible on the main page with map
- the "Details" button could have maybe slightly larger font size
- the window with details will need a bit of graphical tuning to make it look nicer, e.g.
  - better split layer name from URI
  - better use of spacing
  - probably slightly larger font size
  image

app/qml/main.qml
NEDJIMAbelgacem: Sorry I forgot to remove this one.

NEDJIMAbelgacem:
The notification only appears when the map is viewed now and the layout and font sizes are fixed.
Here is what the project issues panel looks like:

missing_lsyers

wonder-sk:
Let's merge as is and we can eventually tweak the design later...

PR #1776, opened 2021-11-30
Compare qgsquick to latest QGIS again

tomasMizera:
Removed comparison of qgsquick lib from a specific commit to the latest qgis master.
Also changed that qgsquick check now runs on each change in app/* too.

### DIFF
--- a/.github/workflows/qgsquick.yml
+++ b/.github/workflows/qgsquick.yml
@@ -4,6 +4,7 @@ on:
     paths:
     - 'qgsquick/**'
     - 'scripts/**'
+    - 'app/**'
 
   release:
     types:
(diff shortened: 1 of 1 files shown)


PeterPetrik (commented):
I think we should keep the QGIS_COMMIT_HASH and for example add there "master" if you really want master. 

But my preference is to keep it in sync with  https://github.com/lutraconsulting/input-sdk/blob/e3779fcaa1253f9b0714a8826366634538cb3af5/versions.conf#L53

so when you bump SDK, you bump the hash too .... sooner or later again the qgis master can be different to our version and we may still want green tests

tomasMizera:
Good point! Let me revert that

PeterPetrik (approved):

PR #1779, opened 2021-11-30
[CI] QgsQuick check - compare to existing files

tomasMizera:
Script was checking existence of files we just removed (quick utils). Replaced it with map canvas.


tomasMizera:
Now the check correctly fails with our changes that are not yet in QGIS

PeterPetrik (approved):

Issue #1772, opened 2021-11-29
Sync may break with geodiff 0.x


wonder-sk:
Steps to reproduce:
1. Input user A: create a project + upload it (v1)
2. Input user A: add a point + sync (v2)
3. share the project with user B
4. Input user B: download + add a point + sync (v3) + add a point + sync (v4)
5. Input user A: add two points + sync

At the last step, sync will fail (server will not accept push of v5).

Problem 1: user does not get an error message that the sync failed

Problem 2: user is not offered "sync" button for that project anymore, yet there are unsynced changes that can't be pushed.

Further analysis:
- step 1 creates a project where data.gpkg has gpkg_contents with some last modified time value (t0)
- in step 2 the last modified time is updated (t1) and uploaded, but server (with geodiff >=1.0) does not apply the modified time bit (keeps it at t0 but Input user A thinks it is t1)
- in step 4 the download gets latest project, with time t0, then sends a diff t0->t2 (v3) and then sends a diff t2->t3 (v4)
- in step 5 when doing pull, user A tries to apply v3's changes (t0->t2) to the basefile but there's conflict because basefile has t1. This causes "server file assembly failed" error in log when applying v3 and applying v4 to basefile is completely skipped (!). Due to that, rebase will not run correctly and the new points get fid increased by one (but should be increased by two if v4 would be applied to the basefile). So when doing push, the diff is rejected because there is conflict with the existing fids.

This needs a couple of things in order to break:
- clients with geodiff < 1.0 (i.e. not the recent Mergin qgis plugin versions >= 2021.4)
- one client downloads project after the other one did some changes (so that the last modified time gets out of sync)
- client A pulls multiple versions at once, so that some versions are not applied to its basefile
- clients add new features - when updating/deleting features, fids are not renumbered, so less likely that server will reject diff

With upgrade of Input to geodiff 1.0 this issue should go away, but we should still make sure that we report error properly (problem 1) and if in some bad state already, do something about it (problem 2).

jozef-budac:
It looks like this issue is fixed in latest INPUT master (v1.2.0)

wonder-sk:
Thanks for checking - let's close this ticket. The remaining issues moved to #1849

PR #1781, opened 2021-12-01
Update sdk versions to use geodiff to 1.0.5

tomasMizera:
There are new SDK versions that use geodiff 1.0.5. This PR adds them to CI builds.


tomasMizera:
@wonder-sk should this fix the #1772 issue?

PeterPetrik (approved):

wonder-sk:
@tomasMizera yes it should to large degree. #1772 also mentions buggy behavior of the GUI when a sync error happens, but I am happy if we extract those to a separate ticket and address at some point later.

PR #1782, opened 2021-12-01
Android CI build with AAB

tomasMizera:
PR fixes Android CI build.

CI now generates armeabi and arm64 apks for each commit. AABs are generated for each commit on master branch.

Versioning of builds also changed. For android_version_code we now use just a sequence of numbers (previously we used major-minor-fix-abi) starting at current build max from google play console (101004). We increment the number by number of commits from beginning of the repo.

> **Note:** version name did not change! We still use major.minor.fix schema


scripts/android-apk.bash
```diff
@@ -52,14 +52,14 @@ echo "API: $ANDROIDAPI"
...
-ln -s ${BUILD_DIR}/.gradle /root/.gradle
 
 pushd ${BUILD_DIR}
-cp ${SOURCE_DIR}/scripts/ci/config.pri ${SOURCE_DIR}/app/config.pri
+
 ${QT_BASE}/bin/qmake -spec android-clang ANDROID_ABIS="${ARCH}" ${SOURCE_DIR}/app/input.pro
+ls ${ANDROID_NDK_ROOT}/prebuilt/
 ${ANDROID_NDK_ROOT}/prebuilt/${ANDROID_NDK_HOST}/bin/make qmake_all
 make -j ${CORES}
+
```
PeterPetrik, line 62: wondering if it is not worth to have android-apk.bash and android-aab.bash  as one script, or even do not have bash script and have it as part of yml file... because you build aab and both apks same way., no?

PeterPetrik (approved):

PR #1848, opened 2021-12-13
Replace pixel size of fonts with point sizes

### LINKED ISSUE #1833: Large fonts and GUI elements
In the version to be released, everything in the GUI is larger on my phone.

Before / after:

tomasMizera:
PR replaces use of pixels in font sizes with points.

Video from my Android:
(DPR: 2.625)

https://user-images.githubusercontent.com/22449698/145811032-57f023ab-23ac-485b-8584-c87d66e72171.mp4

@wonder-sk I think we will need to higher the font size in the entire application, however, could you please try to run on your device first and see if it helps with #1833 ?

Resolves #1833


wonder-sk:
This is how it looks like now:

new2

Thoughts:
- fonts
  - "normal" font size is now roughly the same as it used to be (~34px old/new)
  - "small" font size is now a bit small - I suggest increasing 12 -> 12.5 (it should not be a problem to have point sizes as real numbers)
  - "title" / "big" font size looks fine
  - we can probably finally get rid of "rado2lutra" scale factor :-)
- all the element sizes are still increased - e.g. those based on InputStyle.rowHeight
  - height of an item in project list: old ~200 px vs new ~228 px
  - height of page title or search bar: old ~169 px vs new ~192 px
  - looks like if we divide all the constant sizes by ~1.14, we could get to the original sizing

tomasMizera:
@wonder-sk thanks for such tests! I have applied the scale

wonder-sk:
Great, thanks for the updates! Now it looks roughly the same as it used to in terms of fonts and element sizing...

To be checked on other devices...

On my phone:

new3

wonder-sk:
One more thing that caught my eye is that the record button's "donut" icon has very small hole in the donut now :-)
(not sure why the icon is "dynamic" like that - we should be using SVGs for icons)

Screenshot_20211213_210816

app/qml/components/ProjectDelegateItem.qml
```diff
@@ -187,10 +187,10 @@ Rectangle {
       ProgressBar {
         id: progressBar
 
-        property real itemHeight: InputStyle.fontPixelSizeSmall
+        property real itemHeight: 20
```
PeterPetrik, line 190: best to use the numbers from InputStyle?

app/qml/components/ProjectDelegateItem.qml
```diff
@@ -187,10 +187,10 @@ Rectangle {
...
 
         anchors.top: mainText.bottom
-        height: InputStyle.fontPixelSizeSmall
+        height: 20
```
PeterPetrik, line 193: best to use the numbers from InputStyle?

PeterPetrik (approved):

tomasMizera:
@wonder-sk tested it on two additional devices, it seems to be small on them (I can now see 5 icons in the main toolbar on map, previously there were only 4). We need to find a way to scale it appropriately based on device's resolution 

<table>
  <tr>
    <th>iPhone 11</th>
    <th>Samsung Galaxy A32</th>
  </tr>
  <tr>
    <td>

https://user-images.githubusercontent.com/22449698/145958710-2311953d-278e-4c8c-9a0c-055525a82fc4.mov

</td>
    <td>

https://user-images.githubusercontent.com/22449698/145958790-0ca472ed-9ef6-4fb2-85dd-8032d3bb1929.mp4

</td>
  </tr>
 </table>

tomasMizera:
From my tests, feature form editor labels are bigger now, however previously I think they were too small on Android

app/inpututils.cpp
```diff
@@ -928,6 +928,38 @@ QString InputUtils::geometryFromLayer( QgsVectorLayer *layer )
...
 
+qreal InputUtils::calculateScreenDpr()
+{
+  const QWindowList windows = QGuiApplication::topLevelWindows();
+
+  if ( !windows.isEmpty() )
+  {
+    QScreen *screen = windows.at( 0 )->screen();
+    double dpiX = screen->physicalDotsPerInchX();
+    double dpiY = screen->physicalDotsPerInchY();
```
wonder-sk, line 939: Hmm actually this is already giving the scaled DPI (e.g. 140.67 on my phone instead of the actual 422)

I guess we need to multiply it by `screen->devicePixelRatio()`

app/main.cpp
```diff
@@ -511,12 +511,11 @@ int main( int argc, char *argv[] )
...
-  // Enabling HighDPI scaling attribute (at the beggining of the main function) removes the
-  // need for manually calculating dp factor and multiplying it with each pixel value in qml.
-  // However, we keep the multiplications in place (right now we multiply only with 1),
-  // in case we would encounter a screen that cannot be scaled automatically.
+  // Even though enabling QT's HighDPI scaling removes the need to multiply pixel values with dp,
+  // there are screens that need a "little help", because system DPR has different value than the
+  // one we calculated. In these scenarios we use a ratio between real (our) DPR and DPR reported by QT.
   // Use `value * __dp` for each pixel value in QML
-  engine.rootContext()->setContextProperty( "__dp", 1 );
+  engine.rootContext()->setContextProperty( "__dp", InputUtils::calculateDpRatio() );
```
wonder-sk, line 518: I am wondering if this sets __dp correctly - probably there are no top-level windows yet, so it just uses 1 ? because on my phone it has still the same increased scale of the whole GUI

app/inpututils.cpp
tomasMizera: That is correct

app/main.cpp
tomasMizera: Correct again

wonder-sk (approved):
Looks good now! :tada:

PR #1854, opened 2021-12-15
Set up concurrency between workflow runs

### LINKED ISSUE #1843: Stop concurrent CI jobs on the same branch
There is an option in github actions called `concurrency` which allows us to stop all running jobs with the same id. 

We can use this to stop workflows that run on the same branch at the same time (e.g. when you commit and push twice in a short time). You will no longer need to stop other workflows manually

tomasMizera:
PR adds `concurrency` to workflow files for ios/mac/linux/android. When you push new changes to a branch, already running workflows **on the same branch** are going to be stopped automatically.

Moreover, to make autotests a little bit more stable, I added concurrency also for autotests file. It does not have the `cancel-in-progress` option set, so already runnning tests **on different branch** are not going to be canceled. However, if there would be some workflows waiting to start autotests, this queue will be cleared. See more here: https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idconcurrency

Resolves #1843

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -23,7 +23,7 @@ env:
   CCACHE_DIR: /Users/runner/work/ccache
 
 concurrency:
-  group: ci-${{github.ref}}-linux
+  group: ci-${{github.ref}}-macos
   cancel-in-progress: true
 
 jobs:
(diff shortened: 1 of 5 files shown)


.github/workflows/macos.yml
```diff
@@ -22,10 +22,23 @@ env:
   INPUT_SDK_VERSION: mac-20211201-46
   CCACHE_DIR: /Users/runner/work/ccache
 
+concurrency:
+  group: ci-${{github.ref}}-linux
```
PeterPetrik, line 26: copy-paste error: should be "macos"

PeterPetrik (approved):

PeterPetrik (approved):

PR #1873, opened 2021-12-22
Implementation of staking out points long range navigation

NEDJIMAbelgacem:
Accidentally closed https://github.com/lutraconsulting/input/pull/1845 by a bad push operation 😞


app/qml/main.qml
```diff
@@ -126,8 +126,18 @@ ApplicationWindow {
...
+      onFeatureIdentified: {
+        let wasOpen = navigationPanel.isOpen;
+        navigationPanel.endNavigation();
+        if ( !wasOpen )
+          formsStackManager.openForm( pair, "readOnly", "preview" );
+      }
+
+      onNothingIdentified: {
+        if ( navigationPanel.isOpen )
+          navigationPanel.endNavigation();
```
tomasMizera, line 138: Identification should be disabled during navigation
```suggestion

```

app/qml/map/MapWrapper.qml
```diff
@@ -223,10 +231,22 @@ Item {
...
       }
+      case "navigation": {
+        if ( _digitizingHighlight.visible )
+          _digitizingHighlight.visible = false
+
+        if ( _highlightIdentified.visible )
+          _highlightIdentified.visible = false
+
+        if ( _digitizingController.recording )
+          _digitizingController.stopRecording()
```
tomasMizera, line 242: there is never a transition between recording and navigation
```suggestion

```

app/qml/map/MapWrapper.qml
```diff
@@ -256,7 +276,7 @@ Item {
     onIsRenderingChanged: _loadingIndicator.visible = isRendering
 
     onClicked: {
-      if ( !root.isInRecordState )
+      if ( !root.isInRecordState && root.state != "navigation" )
```
tomasMizera, line 279:
```suggestion
      if ( root.state === "view" )
```

tomasMizera (changes requested):
We are almost there ❗ 

I have two questions:

1. what is the reason for having `NavigationHighlight.qml` component that is so similar to `Highlight.qml`? I thought it is going to be a minimal component like this: 

```qml
Item {

  property gpsPoint
  property targetPoint
  
  // other properties for ui of the navigation line (thickness, color,..)

  function constructLine() {
    // convert both points to dest crs
    ...

    // create line from gpsPoint and targetPoint
    lineShape.pathElements = [gpsPoint, targetPoint]
  }

  onGpsPointChanged: constructLine()

  Shape {...}
}
```
However, in `NavigationHighlight.qml` I see markerItems, crosshair points, componentMarker which effectively makes it the original `Highlight.qml` 🤔 

Also, I see that you construct linegeometry in utils and that is passed to navigationhighligt via `featureLayerPair`. Is it necessary? Navigation will always be a simple line from point A to point B. 

2. the second one might be more of a point of view. Should `NavigationPanel` alter states of `MapWrapper`? I wonder if functions `startNavigation`, `endNavigation`, `updateNavigation` shouldn't be rather placed somewhere else (maybe `MapWrapper`?). They basically do nothing with navigationPanel itself, only alter states and extent of the map

app/qml/form/PreviewPanel.qml
```diff
@@ -7,7 +7,7 @@
  *                                                                         *
  ***************************************************************************/
 
-import QtQuick 2.0
+import QtQuick 2.3
 import QtQuick.Layouts 1.3
 import QtQuick.Controls 2.2
 import QtGraphicalEffects 1.0
```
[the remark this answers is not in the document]
NEDJIMAbelgacem: > We are almost there exclamation rocket
> 
> I have two questions:
> 
> 1. what is the reason for having `NavigationHighlight.qml` component that is so similar to `Highlight.qml`? I thought it is going to be a minimal component like this:
> 
> ```qml
> Item {
> 
>   property gpsPoint
>   property targetPoint
>   
>   // other properties for ui of the navigation line (thickness, color,..)
> 
>   function constructLine() {
>     // convert both points to dest crs
>     ...
> 
>     // create line from gpsPoint and targetPoint
>     lineShape.pathElements = [gpsPoint, targetPoint]
>   }
> 
>   onGpsPointChanged: constructLine()
> 
>   Shape {...}
> }
> ```
> 
> However, in `NavigationHighlight.qml` I see markerItems, crosshair points, componentMarker which effectively makes it the original `Highlight.qml` thinking
> 
> Also, I see that you construct linegeometry in utils and that is passed to navigationhighligt via `featureLayerPair`. Is it necessary? Navigation will always be a simple line from point A to point B.
> 
> 2. the second one might be more of a point of view. Should `NavigationPanel` alter states of `MapWrapper`? I wonder if functions `startNavigation`, `endNavigation`, `updateNavigation` shouldn't be rather placed somewhere else (maybe `MapWrapper`?). They basically do nothing with navigationPanel itself, only alter states and extent of the map

Thanks for the reviews @tomasMizera 
- I turned the navigation highlight into more minimal implementation,
- For your second point, my main reasoning is to have the inner code of the map wrapper handle the exclusivity between the modes represented by the states, so if some another menu item places the map wrapper in a different state the map wrapper handles that any mode that is enabled is ended. The `startNavigation` and `endNavigation` are parts of the navigations logic and in my opinion better to be contained in close to their presentation's UI (NavigationPanel.qml) than the presentation of some other UI. Ideally this logic (code to move things around) and the presentation (actual UI: buttons, text, panels...) are to be seperated into different areas but QML is not good at that :'/ .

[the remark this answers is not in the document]
tomasMizera: Yes, I agree with that. I was also thinking about some logical component. Maybe in future we can find a way how to interact with cpp in these scenarios.

app/qml/map/MapWrapper.qml
```diff
@@ -274,6 +294,8 @@ Item {
           root.nothingIdentified()
         }
       }
+      if ( root.state === "navigation" )
+        exitNavigationClicked();
```
tomasMizera, line 298: I wonder whether we want this at all. What do you think @NEDJIMAbelgacem? I am afraid that users might accidentally tap on their screen and thus lose the navigation.. When I look at other apps, they do not seem to have such "feature". I would leave the **X** button an only way how to exit navigation :)

tomasMizera (commented):
Looks great! Good job and thanks for addressing my comments.

app/qml/map/MapWrapper.qml
NEDJIMAbelgacem: I was thinking that it is intuitive for me to minimize panels when the map is clicked. But after thinking about it for a bit I think navigation shouldn't be stopped because you clicked on the map (I remembered Google maps navigation).

app/inpututils.cpp
```diff
@@ -621,6 +621,7 @@ QgsPointXY InputUtils::transformPoint( const QgsCoordinateReferenceSystem &srcCr
                                        const QgsCoordinateTransformContext &context,
                                        const QgsPointXY &srcPoint )
 {
+  qDebug() << srcCrs.geographicCrsAuthId() << " " << destCrs.geographicCrsAuthId();
```
tomasMizera, line 624: I do not think we need this

app/qml/map/NavigationHighlight.qml
```diff
@@ -66,7 +66,9 @@ Item {
   {
     if ( !destinationPair || !gpsPosition || !mapSettings ) return
 
+    print( "gpsMapCRS" )
```
tomasMizera, line 69: remove

app/qml/map/NavigationHighlight.qml
```diff
@@ -66,7 +66,9 @@ Item {
...
     let gpsMapCRS = __inputUtils.transformPoint( __inputUtils.coordinateReferenceSystemFromEpsgId( 4326 ), mapSettings.destinationCrs, mapSettings.transformContext(), __inputUtils.pointXY( gpsPosition.x, gpsPosition.y ) );
+    print( "gpsMapCRS ", destinationPair.layer.crs )
```
tomasMizera, line 71: remove

app/inpututils.h
```diff
@@ -402,12 +402,23 @@ class InputUtils: public QObject
...
+    Q_INVOKABLE static QgsPointXY extractPointFromFeature( const FeatureLayerPair &feature );
+
+    // Returns the extent of the navigation feature representing the source and the destination of the navigation
+    Q_INVOKABLE QgsRectangle navigationFeatureExtent( const FeatureLayerPair &pair, QgsPoint gpsPosition, QgsQuickMapSettings *mapSettings, double panelOffsetRatio );
+
+    // Returns the distance from \a gpsPos to the feature \a pair
+    Q_INVOKABLE QString distanceToFeature( QgsPoint gpsPosition, const FeatureLayerPair &targetFeature, QgsQuickMapSettings *mapSettings );
+
+    // Returns the title of the feature
+    Q_INVOKABLE static QString featureTitle( const FeatureLayerPair &pair, QgsProject *project );
```
tomasMizera, line 415: Can you please cover these functions in auto tests?

tomasMizera:
One more thing, GPS icon with accuracy is hidden under navigation panel, it should be visible and move above the panel when user is in navigation mode

tomasMizera (approved):
After the remaning comments are resolved (gps icon, debug logs removal and adding auto tests for utils functions) I am +1 to merge this one.

I have tested the navigation on both platforms and seems to work fine.

@wonder-sk we agreed with Belgacem that UI is going to be focused on in future PR - this one is already massive.

Passing to you @wonder-sk / @PeterPetrik for final review and to @jozef-budac / @PatrikMizera for testing. Once again, good job @NEDJIMAbelgacem 👍🏻

jozef-budac:
I have this comments:

1) GPS panel should probably not be visible when in navigation mode
image0

2) navigation doesn't work well with follow GPS feature, where the map is centered around the location of the user. Follow GPS should probably be switched off.
image1

3) feature should bot be available for lines and polygons
image2

4) INPUT crashes with MultuPoint geometry line
https://user-images.githubusercontent.com/83582181/147910846-7fb052e0-df32-4d91-92f0-776f7d93f3b2.mov

jozef-budac:
As display of GPS panel is desired functionality, probably it would be better to hide preview panel when GPS panel is visible
image0(1)

jozef-budac:
Also there should be probably some shadow over navigation line, so on the background with same color, it should be visible

image0(2)

jozef-budac:
7) this should work: "when stake out mode is closed, we show the preview panel, and if the point is outside of the current map view, we will re-center"
The preview panel is show, when stake mode is closed, but when point is outside the current map view, we didn't recenter

8) also this should work but doesn't  "if location is not available, show an error notification when trying to enter stake out mode" 
https://user-images.githubusercontent.com/83582181/147915939-a4788759-6769-4318-a4f7-861f44a9cb7d.mov

jozef-budac:
I have two comments for latest release:
9) in horizontal position, GPS button is partially outside the screen

https://files.slack.com/files-pri/TBXE5PUU9-F02SN5WAHDM/screenshot_2022-01-07-09-49-34-369_uk.co.lutraconsulting.jpg

10) then in navigation mode and I open GPS panel
      - when I close the GPS panel with arrow in top right corner, navigation mode is closed. I think, it should be still there.
      - when I close the GPS panel with back button on android, INPUT app is closed

https://user-images.githubusercontent.com/83582181/148520177-7d66cf34-0f31-4299-96aa-24a98cd1b1ea.mp4

scripts/run_all_tests.bash
```diff
@@ -9,44 +9,44 @@ fi
...
 
-$INPUT_EXECUTABLE --testFormEditors
-NFAILURES=$(($NFAILURES+$?))
+#$INPUT_EXECUTABLE --testFormEditors
+#NFAILURES=$(($NFAILURES+$?))
 
-$INPUT_EXECUTABLE --testModels
-NFAILURES=$(($NFAILURES+$?))
+#$INPUT_EXECUTABLE --testModels
+#NFAILURES=$(($NFAILURES+$?))
```
tomasMizera, line 49: Uncomment back :)

scripts/run_all_tests.bash
```diff
@@ -9,44 +9,44 @@ fi
...
 
 echo "Total $NFAILURES failures found in testing"
```
tomasMizera, line 52: Please do not commit changes in these test files: `test/test_data/planes/constraint-layers.gpkg` `test/test_data/planes/geo-layers.gpkg` `test/test_data/project_relations/db.gpkg`. They were probably edited when you run tests locally

tomasMizera (commented):
Hi @NEDJIMAbelgacem can you please have a look at these few comments?

tomasMizera (commented):
Few more comments I found during testing..

Note for future (we will fix it in next PR): navigation highlight disappears when one zooms in (about <10m)! This needs to be addressed in upcoming PR. The color of navigation line is also changed to black soon before disappearing.

PR #1878, opened 2022-01-03
Add rebase files to sync

### LINKED ISSUE #1738: Rebase conflict files get stored outside of the project directory
Currently Input stores any rebase conflict files in the `.mergin` subdirectory:
https://github.com/lutraconsulting/input/blob/master/core/merginapi.cpp#L1451

This is not correct, because the conflict files are then never seen by users.

We also need to figure out naming convention, ideally consistent with Mergin plugin.

Best to fix this after upgrading to geodiff version that fixes https://github.com/lutraconsulting/geodiff/issues/152 (causes creation of conflict files even when not needed)

tomasMizera:
PR saves EDIT conflict files to a project directory so that they are synced and available for users in Mergin.
Edit conflict files are generated when geodiff could not rebase local changes upon server changes.
There is also a new naming convention for conflict files. We differentiate between edit conflict (conflict in diffable file, output is json file from geodiff) and copy conflict (conflict in non diffable file or gpkg that is not possible to rebase).

@wonder-sk I could not find a test that would check existence of an edit conflict file. Shall we add one or are geodiff tests sufficient here?

Closes #1738


wonder-sk:
> I could not find a test that would check existence of an edit conflict file. Shall we add one or are geodiff tests sufficient here?

Yes please, that would be very useful to have an explicit test that mergin client will create edit conflict file...

Few notes about naming of files:
- edit conflicts are JSON files, so we should use `.json` suffix instead of `.gpkg`
- it would be great if we improved findUniquePath() a bit to use nicer alternatives - right now for `data.gpkg` it would just use `data.gpkg1`, `data.gpkg2` etc which break the file associations based on suffix. A great alternative would be to use this naming scheme - if `data.gpkg` already exists, we would use `data (1).gpkg`, `data (2).gpkg` etc
- I had a chat about naming the two different types of conflicts, and how to make things more human readable, and we'd like to settle with the following scheme - when starting from `data.gpkg` file:
  - `data (conflicted copy, martin v5).gpkg` for conflicted copy (the "copy conflict" term does not work that well as it implies that it's a conflict due to some copying - the term "conflicted copy" is used elsewhere e.g. in [Dropbox](https://help.dropbox.com/files-folders/share/conflicted-copy))
  - `data (edit conflict, martin v5).json` for edit conflict

tomasMizera:
@wonder-sk all 4 comments are now addressed

app/test/testmerginapi.cpp
```diff
@@ -1009,12 +1010,67 @@ void TestMerginApi::testConflictRemoteAddLocalAdd()
...
+  //
+  // now client A syncs, resulting in edit conflict
+  //
+
+  uploadRemoteProject( mApi, mUsername, projectName );
+
+  QDir projDir( projectDir );
+
+  // check the edit conflict file presence
+  QVERIFY( InputUtils::fileExists( projectDir + "/" + QString( "data (edit conflict, %1 v2).gpkg" ).arg( mUsername ) ) );
```
wonder-sk, line 1067: I think edit conflicts should always have `.json` suffix even though they are related to a `.gpkg` file - otherwise people may get confused and try to load edit conflict files as geopackages...

wonder-sk (commented):
Looking great, just one detail...

wonder-sk (approved):
Thanks!

PR #1886, opened 2022-01-18
Selection of position providers

tomasMizera:
PR adds UI for position providers page and adding new GPS receiver page based on design doc. To change position provider navigate to settings -> `Select GPS receiver`.

Internal GPS provider is the default device GPS receiver. Desktop platforms have there also additional simulated provider.

Showcase video:

https://user-images.githubusercontent.com/22449698/150008161-c6e2c8c4-296a-4007-ba29-f86bdb91482d.mp4

**Note: connection to BT device still needs a little love, we need to show somehow that mobile is trying to connect to a device.**

Note2: that `eeee` device is just test, it is thus duplicated on video. With normal devices it does not happen.

@PeterPetrik I have not removed all TODOs from previous PR. There will be another one :)

### DIFF
--- a/app/position/positionprovidersmodel.h
+++ b/app/position/positionprovidersmodel.h
@@ -20,7 +20,8 @@ struct PositionProvider
 {
   QString name;
   QString description;
-  QString providerId; // simulated/internal/BT address
+  QString providerType; // either internal or external
+  QString providerId; // holds BT address for external provider and simulated/internal for internal provider
 
   bool operator==( const PositionProvider &other ) const
   {
@@ -32,16 +33,11 @@ struct PositionProvider
     return !( *this == other );
   }
 
-  PositionProvider( const QString &name, const QString &desc, const QString &id )
-    : name( name ), description( desc ), providerId( id )
+  PositionProvider( const QString &name, const QString &desc, const QString &type, const QString &id )
+    : name( name ), description( desc ), providerType( type ), providerId( id )
   {}
 
   PositionProvider() {}
-
-  bool isPermanent() const
-  {
-    return ( providerId == QStringLiteral( "internal" ) || providerId == QStringLiteral( "simulated" ) );
-  }
 };
 
 class PositionProvidersModel : public QAbstractListModel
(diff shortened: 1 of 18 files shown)


app/position/positionprovidersmodel.h
```diff
@@ -0,0 +1,84 @@
...
+#ifndef POSITIONPROVIDERSMODEL_H
+#define POSITIONPROVIDERSMODEL_H
+
+#include <QObject>
+#include <qglobal.h>
+#include <QAbstractListModel>
+
+class AppSettings;
+
+struct PositionProvider
```
PeterPetrik, line 19: I personally do not like structs :). Alternative to this is c++ class where all members are not modifiable/final (?) and you have only gettters. Anyways, this is good too, matter of taste

app/position/positionprovidersmodel.h
```diff
@@ -0,0 +1,84 @@
...
+class PositionProvidersModel : public QAbstractListModel
+{
+    Q_OBJECT
+
+    Q_PROPERTY( AppSettings *appSettings READ appSettings WRITE setAppSettings NOTIFY appSettingsChanged )
+
+  public:
+
+    explicit PositionProvidersModel( QObject *parent = nullptr );
+    virtual ~PositionProvidersModel() = default;
```
PeterPetrik, line 52: I think = default should be in c++ , some compilers does not like it in headers

app/qml/misc/PositionProviderPage.qml
```diff
@@ -28,84 +30,266 @@ Page {
...
 
-          text: model.DeviceAddress + ", signal strength: " + model.SignalStrength
+          Text {
+            id: deviceSecondaryText
+
+            width: parent.width
+            height: parent.height * 0.5
+
+            // add BT address to text if it is external receiver
+            text: model.ProviderDescription + ( model.ProviderType === "external" ? " (" + model.ProviderId + ")": "" )
```
PeterPetrik, line 159: I see this multiple times in the qml, maybe have another prop in model like "title" ? with this sting?

PeterPetrik (approved):

PR #1888, opened 2022-01-19
PositionKit as QML singleton

tomasMizera:
Changed position kit to be a singleton in QML. Previously we needed to pass `PositionKit` qml instance all over the place. It is handy to be able to access device position from any part of the app.

### DIFF
--- a/app/loader.cpp
+++ b/app/loader.cpp
@@ -62,12 +62,6 @@ QgsProject *Loader::project()
   return mProject;
 }
 
-void Loader::setPositionKit( PositionKit *kit )
-{
-  mPositionKit = kit;
-  emit positionKitChanged();
-}
-
 void Loader::setRecording( bool isRecordingOn )
 {
   if ( mRecording != isRecordingOn )
@@ -308,18 +302,6 @@ void Loader::appStateChanged( Qt::ApplicationState state )
 
   logHelper << "Application changed state to: " << state;
   CoreUtils::log( "Input", msg );
-
-  if ( !mRecording && mPositionKit )
-  {
-    if ( state == Qt::ApplicationActive )
-    {
-      mPositionKit->startUpdates();
-    }
-    else
-    {
-      mPositionKit->stopUpdates();
-    }
-  }
 }
 
 void Loader::appAboutToQuit()
--- a/app/qml/misc/GpsDataPage.qml
+++ b/app/qml/misc/GpsDataPage.qml
@@ -18,9 +18,8 @@ import lc 1.0
 Page {
   id: root
 
-  property var positionKit
-  property var mapPositioning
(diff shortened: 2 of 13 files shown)


app/qml/misc/GpsDataPage.qml
```diff
@@ -35,6 +34,13 @@ Page {
     }
   }
 
+  MapPosition {
+    id: mapPositioning
+
+    positionKit: __positionKit
```
PeterPetrik, line 40: maybe we can use singleton and not pass it around?

app/loader.cpp
```diff
@@ -308,18 +302,6 @@ void Loader::appStateChanged( Qt::ApplicationState state )
 
   logHelper << "Application changed state to: " << state;
   CoreUtils::log( "Input", msg );
-
```
PeterPetrik, line 311: should here be called PositionKit::appStateChanged?

tomasMizera: Well, I connected it directly from main.cpp [here](https://github.com/lutraconsulting/input/pull/1888/files#diff-c1b709b8a90bef4573ddad8a9b2fbc5d5dbc85e4dc759f95b6d9905931fec24fR450) in order to remove the coupling of loader (which from my point of view should serve only to load a project) and position.

app/qml/misc/GpsDataPage.qml
tomasMizera: You mean to convert `MapPosition` to singleton too?

tomasMizera: `MapPosition` is a cpp class that needs mapSettings to work. MapSettings is not a singleton class so we would need to pass new map settings to it each time project is loaded

PeterPetrik: No I mean why you need to have `positionKit: __positionKit` here and not use `__positionKit` directly in the element?

app/loader.cpp
PeterPetrik: ok fair enough. is the original "Loader: appStateChanged" needed at all now?

PeterPetrik (approved):
good to go

app/qml/misc/GpsDataPage.qml
tomasMizera: Because it is a C++ class.

tomasMizera:
@PeterPetrik there was actually one place where I was still passing the positionKit :D `PositionMarker.qml`

app/qml/misc/GpsDataPage.qml
PeterPetrik: make sense , sorry :)

tomasMizera:
@PeterPetrik Autotests (purchasing tests) are failing due to a known old issue with test db.

PR #1890, opened 2022-01-19
Make satellite source optional (fixes Apple)

tomasMizera:
`InternalPositionProvider` previously had to have both geo sources (position and satellite) valid in order to start receiving updates. This is wrong on few devices including Apple, where `QGeoSatelliteInfoSource` does not work.

Provider now works in a separates validity of each of the two sources and acts based on that (turns them on/off).

Position can now be read also on mac! :)


saberraz:
@tomasMizera brilliant work. It works as expected now.

PeterPetrik (approved):

PR #1892, opened 2022-01-20
Opt to enable bluetooth when it is disabled

tomasMizera:
Added check if bluetooth is turned on before starting discovery process. 
If:
 - bluetooth is turned on, discovery process starts automatically
 - bluetooth is turned off, show system dialog to enable bluetooth, if:
    - user enables bluetooth in the dialog, discovery process starts
    - user do not enable it, app goes back to list of existing providers and toast message pops up with text "You need to enable Bluetooth in order to connect new GPS receiver"

### DIFF
--- a/app/androidutils.cpp
+++ b/app/androidutils.cpp
@@ -275,6 +275,40 @@ void AndroidUtils::handleLegacyFolderMigration( AppSettings *appsettings, bool d
 #endif
 }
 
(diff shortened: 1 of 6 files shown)


tomasMizera:
This implementation is only for Android, iOS is working out of the box

PeterPetrik (approved):

Issue #1887, opened 2022-01-19
Horizontal and vertical accuracy appears NAN when connected to the Bluetooth GPS device


saberraz:
I am testing https://github.com/lutraconsulting/input/pull/1886 and when I am connected to the external GPS, the horizontal and vertical accuracy appears as nan

jozef-budac:
It behaves the same using android device as external GPS provider

![screenshot_20220119-100522](https://user-images.githubusercontent.com/83582181/150099624-8a97ee77-3164-44e4-aee7-0d1f9c3e9183.png)

PeterPetrik:
@tomasMizera this is fixed?

tomasMizera:
@PeterPetrik no, not yet

PR #1901, opened 2022-01-26
Add position provider to GPS panel

tomasMizera:
Adds GPS receiver name, status, fix and HDOP to GPS panel.
Clicking on the GPS receiver takes user to position providers list. 

Currently it looks like this:
image

@wonder-sk I am not sure if user would understand that he/she can click on the GPS receiver and it will take him to the list of providers .. what do you think? Do you have any idea how to visually indicate that it does something?


saberraz:
what about change the entry for the GPS receiver to a button?

wonder-sk:
I would suggest:
- move the "Connected" text to the right column with e.g. "GPS receiver status" label on top of it (where is the empty space now)
- to change the GPS receiver, I had in mind a small orange button with cog icon that would take people to the provider setup - but not sure if we can align it nicely with the rest of the content - or we can skip adding the button...

tomasMizera:
@saberraz @wonder-sk how about this?

<table>
  <tr>
    <th>External receiver</th>
    <th>Internal receiver</th>
  </tr>
  <tr>
    <td></td>
    <td></td>
  </tr>
 </table>

wonder-sk:
That looks good, thanks Tomas!

Just wondering, for the internal GPS provider it looks a bit strange to have lat/lon and X/Y split diagonally... would it be possible to keep the top-right empty, so that lat/lon and x/y are always on the same line?

One more question - about the external GPS provider - how is it possible that we are connected, we have valid position, but accuracy, hdop, last fix are not available? :)

For the button at the bottom, I would probably prefer the cog icon used also for settings, I found it a bit difficult to decode the icon as a satellite :)

tomasMizera:
@wonder-sk true, that icon looks like a trumpet 😄 

<table>
  <tr>
    <th>External receiver</th>
    <th>Internal receiver</th>
  </tr>
  <tr>
    <td></td>
    <td></td>
  </tr>
 </table>

tomasMizera:
@wonder-sk , re your question:

>One more question - about the external GPS provider - how is it possible that we are connected, we have valid position, but accuracy, hdop, last fix are not available? :)


it is the same issue as in #1887 . I used mobile GPS output and it simply does not provide GPRMC (for datetime), GPGST (for accuracy) and does not fill HDOP field in GPGSA sentence.

tomasMizera:
Example NMEA from that app: https://github.com/lutraconsulting/input/blob/master/test/test_data/position/test_nmea_petrzalka.txt

uclaros:
X and Y without a hint on the crs used drew my attention as a topographer!

wonder-sk:
Thanks for the update, it looks nice. My further minor suggestions:
- having "No data from GPS" repeated multiple times does not look so nice - probably just having "-" would be cleaner (or "N/A" as before)
- in the first line, the "receiver" and "GPS" text is repeated in both label and value ("GPS receiver" + "Internal GPS receiver" and "GPS receiver status" + "Connecting to receiver") - maybe we could avoid that, for example use:
  - "Source" label instead of "GPS receiver" label
  - "Internal GPS receiver" could be just "Internal"
  - "GPS receiver status" label could be just "Status"
  - "Connecting the receiver" could be replaced just by more brief "Connecting"

I see regarding input NMEA... on that note, for me the "date not available" text for "last fix" is quite confusing, and we have GGA sentence with time, so I would suggest to simply always assume the date is today if we don't have that information explicitly from other sentences...

app/position/abstractpositionprovider.h
```diff
@@ -26,10 +26,14 @@ class GeoPosition : public QgsGpsInformation
 
     double verticalSpeed = -1;
 
+    QString fixStatusString;
```
PeterPetrik, line 29: If it is class, why we do not use private members and setters/getters? I am not big fan of public members in general :)

app/position/positionkit.h
```diff
@@ -60,6 +60,9 @@ class PositionKit : public QObject
     // GPS satellites advanced data (Horizontal dilution of precision)
     Q_PROPERTY( double hdop READ hdop NOTIFY hdopChanged )
 
+    // GPS fix status
+    Q_PROPERTY( QString fix READ fix NOTIFY fixChanged )
```
PeterPetrik, line 64: is "fix" common name for this? for me it sounds more like you want to fix your GPS connection or so?

app/qml/misc/GpsDataPage.qml
```diff
@@ -27,199 +27,318 @@ Page {
...
-      TextRowWithTitle {
-        id: verticalAccuracy
+            titleText: qsTr( "X" )
+            text: {
+              if ( !__positionKit.hasPosition ) {
+                return qsTr( "No data from GPS" ) // if you do not have projected position yet
+              }
+              __inputUtils.formatNumber( mapPositioning.mapPosition.x, 2 )
+            }
+          }
```
PeterPetrik, line 169: does this stuff work "live" or you got it refreshed only when you close/open the panel?

app/position/positionkit.h
wonder-sk: "fix" or "position fix" is a common term - see e.g. https://en.wikipedia.org/wiki/Geopositioning#Background or https://en.wikipedia.org/wiki/Time_to_first_fix

app/position/abstractpositionprovider.h
tomasMizera: Well yeah, it's used like this in `QgsGpsInformation` (base class of this), look: https://github.com/qgis/QGIS/blob/master/src/core/gps/qgsgpsconnection.h#L98 I only added few other members

app/qml/misc/GpsDataPage.qml
tomasMizera: Yes, it works live, `mapPosition` is a property with `changed` signal

app/position/positionkit.h
tomasMizera: I think so, @wonder-sk ?

PeterPetrik (approved):

tomasMizera:
> "Connecting the receiver" could be replaced just by more brief "Connecting"

Already renamed here: https://github.com/lutraconsulting/input/pull/1903

wonder-sk (approved):

PR #1906, opened 2022-01-27
Do not use bluetooth qt API on iOS platform

tomasMizera:
As iOS does not allow us to read serial from bluetooth device, this PR removes all connections to bluetooth API for ios build.

There is a new define "HAVE_BLUETOOTH" that is used on max/android/linux.

### DIFF
--- a/app/android.pri
+++ b/app/android.pri
@@ -77,6 +77,8 @@ android {
     QT += multimedia
     QT += printsupport
     QT += androidextras
+    QT += bluetooth
+    DEFINES += "HAVE_BLUETOOTH"
 
     QMAKE_CXXFLAGS += -std=c++11
 
--- a/app/bluetoothdiscoverymodel.cpp
+++ b/app/bluetoothdiscoverymodel.cpp
@@ -10,10 +10,14 @@
 #include "bluetoothdiscoverymodel.h"
 #include "coreutils.h"
 
-#include "qdebug.h"
+#ifdef HAVE_BLUETOOTH
+#include <QBluetoothUuid>
+#endif
+
 
 BluetoothDiscoveryModel::BluetoothDiscoveryModel( QObject *parent ) : QAbstractListModel( parent )
 {
+#ifdef HAVE_BLUETOOTH
   mDiscoveryAgent = std::unique_ptr<QBluetoothDeviceDiscoveryAgent>( new QBluetoothDeviceDiscoveryAgent() );
 
   connect( mDiscoveryAgent.get(), &QBluetoothDeviceDiscoveryAgent::deviceDiscovered, this, &BluetoothDiscoveryModel::deviceDiscovered );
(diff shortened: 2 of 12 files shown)


tomasMizera:
Currently the GPS receivers page looks like this:


I suggest to remove the possibility to navigate to this page from settings. What do you think @saberraz @wonder-sk @PeterPetrik ?

wonder-sk:
I would probably keep that page, just with this kind of note (using normal colors, not red as in my mockup :slightly_smiling_face:  ):

image

tomasMizera:
Hmm, isn't Apple review process picky when something is not available on their platform? I see that it is blocked by them but still..

wonder-sk:
Well, if they don't provide support for that on the OS level, we should be fine IMHO - I guess that kind of rule is meant if some functionality is explicitly disabled in app on their platform even if it was supported...

DunktheLunk:
Sorry to jump in here-I find it interesting you say that iOS doesn’t support Bluetooth GPS. Yet Leica have this via the Zeno Connect App (https://apps.apple.com/gb/app/zeno-connect/id1310344749). Or am I missing something?

wonder-sk:
@DunktheLunk good point, the thing is that only Bluetooth 4.0 Low Energy is supported by iOS, bat many GNSS receivers on market today do not support this protocol - so in fact those can't be used with iOS...

And Bluetooth LE is quite different from what I understand, so we are unable to support it with the code we have now, and currently we don't have a Bluetooth LE capable receiver around for development...

There's also alternative interface offered by some GNSS receivers - HTTP connection through the local network between the phone and receiver.

So maybe the note just should be more made more accurate :-)

tomasMizera:
Hi @DunktheLunk thanks for jumping in! 
I am going to add to @wonder-sk's reply. Yes, there are GPS devices that can be connected via its manufacturer app which then provides position to your phone (e.g. Leica as you mention[1], EOS https://eos-gnss.com/products/software/eos-tools-pro or possibly Trimble's https://www.frontierprecision.com/wp-content/uploads/quickstartguide.pdf). 

Any other app then (including Input) is able to show this high-precision position. The thing is that the communication goes via _operating system_. Manufacturer's app (e.g. Zeno) provides position to OS and then any other app can read it with the same method as if there would not be any receiver and only mobile's GPS would be used. It has some advantages - you do not need to do anything special in Input (or similar apps) and you have the position right away. On the other hand, not that many GPS receivers supports this. Our intention was to be able to connect Input directly with a GPS receiver without going through operating system. This works on Android, but is unfortunately banned on iOS. 

Long story short, you can use Leica, but in Input settings you will see it as Internal GPS receiver (mobile's receiver).

[1] https://www.youtube.com/watch?v=l1rlXlODU6Y

PeterPetrik:
For iOS I think that the text in red should be for now that you should connect the device directly and the use "internal"? and the link to the official docs?

PeterPetrik (approved):

tomasMizera:
Link leads here: https://merginmaps.com/docs/howto/external_gps/?utm_source=input-help&utm_medium=help&utm_campaign=input I have updated ticket for that docs to include this case

DunktheLunk:
Thanks @wonder-sk and @tomasMizera for the responses! That all makes sense, and I think the new message reads well and would make sense to a user such as me. Cheers!

PR #1910, opened 2022-01-31
Autotests for position

tomasMizera:
PR adds auto tests for classes related to new positioning system. It also includes few fixes found during testing and minor changes to code so that it can be tested properly.


PeterPetrik (approved):

app/appsettings.h
```diff
@@ -75,6 +75,9 @@ class AppSettings: public QObject
     const QString &activePositionProviderId() const;
     void setActivePositionProviderId( const QString &id );
 
+    static const QString GROUP_NAME;
```
PeterPetrik, line 78: I liked it a bit more before, string defines make me a bit dizzy, but this is matter of preference :)

PR #1916, opened 2022-02-04
Report server errors

### LINKED ISSUE #1849: No error message when sync fails
If sync fails for some reason when doing the "push finish" request, no error message is reported and the sync button is not shown anymore, giving user the impression that sync went fine. (Follow up of #1772)

It seems that if sync fails before the "finish" request, error message does get reported - e.g. when doing push start (typically if quota is exceeded or if other user is uploading a new version). It would be good to double check all code paths to make sure the errors are reported.

Also worth checking the second part of the issue, that after failed "push finish" request, the sync button is not offered anymore.

To simulate push finish request, one could replace basefile of a geopackage with an incorrect one. For example, if a table has already fids 1,2,3 (also on mergin server), then basefile could get turned into invalid state by deleting fid=3... this will make sync try to push fid=3 as a newly added feature, but it will fail in push finish.

uclaros:
Added missing error reporting for some cases of failed sync.
Also remove the diff files created on such failures.
Fix #1849

### DIFF
--- a/core/merginapi.cpp
+++ b/core/merginapi.cpp
@@ -1788,8 +1788,10 @@ void MerginApi::updateInfoReplyFinished()
   }
   else
   {
+    QString serverMsg = extractServerErrorMsg( r->readAll() );
     QString message = QStringLiteral( "Network API error: %1(): %2" ).arg( QStringLiteral( "projectInfo" ), r->errorString() );
     CoreUtils::log( "pull " + projectFullName, QStringLiteral( "FAILED - %1" ).arg( message ) );
+    emit networkErrorOccurred( serverMsg, QStringLiteral( "Mergin API error: updateInfo" ) );
 
     transaction.replyProjectInfo->deleteLater();
     transaction.replyProjectInfo = nullptr;
@@ -2275,8 +2277,10 @@ void MerginApi::uploadInfoReplyFinished()
   }
   else
   {
+    QString serverMsg = extractServerErrorMsg( r->readAll() );
     QString message = QStringLiteral( "Network API error: %1(): %2" ).arg( QStringLiteral( "projectInfo" ), r->errorString() );
     CoreUtils::log( "push " + projectFullName, QStringLiteral( "FAILED - %1" ).arg( message ) );
+    emit networkErrorOccurred( serverMsg, QStringLiteral( "Mergin API error: uploadInfo" ) );
 
     transaction.replyUploadProjectInfo->deleteLater();
     transaction.replyUploadProjectInfo = nullptr;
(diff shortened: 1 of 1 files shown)


uclaros:
The failed test was a hickup due to the server unexpectedly returning a 502:
```
2022-02-04T08:54:06.8350920Z QDEBUG : TestMerginApi::testPushModifiedFile() "2022-02-04T08:54:06.834Z push test_inputapp/testPushModifiedFile: Requesting transaction finish: 97264daf-602b-47b9-bd58-f76e86790c55\n"
2022-02-04T08:54:07.9829290Z QDEBUG : TestMerginApi::testPushModifiedFile() "2022-02-04T08:54:07.982Z push test_inputapp/testPushModifiedFile: FAILED - Network API error: uploadFinish(): Error transferring https://test.dev.cloudmergin.com/v1/project/push/finish/97264daf-602b-47b9-bd58-f76e86790c55 - server replied: Bad Gateway. <html>\r\n<head><title>502 Bad Gateway</title></head>\r\n<body>\r\n<center><h1>502 Bad Gateway</h1></center>\r\n<hr><center>nginx/1.15.9</center>\r\n</body>\r\n</html>\r\n\n"
2022-02-04T08:54:07.9831860Z QDEBUG : TestMerginApi::testPushModifiedFile() "2022-02-04T08:54:07.982Z sync test_inputapp/testPushModifiedFile: ### FAILED ###\n\n"
2022-02-04T08:54:07.9902460Z FAIL!  : TestMerginApi::testPushModifiedFile() 'arguments.at( 2 ).toBool()' returned FALSE. ()
2022-02-04T08:54:07.9902930Z    Loc: [../input/app/test/testmerginapi.cpp(273)]
```

uclaros:
Hey @wonder-sk , could you review this?
I insisted on not using a simple, translatable string for those errors for homogeneity. Then we can change to simpler strings for all cases of `emit networkErrorOccurred()` or directly at the qml on a separate PR.

wonder-sk (approved):
Looking good!

wonder-sk:
@uclaros do you have a screenshot of how such error upon "upload finish" looks like?

uclaros:
It depends upon how the server responds. By replicating #1849 on a local mergin in docker, the server handles the request and fails gracefully with this:
image

The public server returns an internal server error so the user gets this:
image

Admittedly, both look ugly and do not offer much help to the average user (the user needs to know 2 things: whether sync failed and whether he should try again or call support ;P), but this is also how errors are reported if the sync fails on a previous step.

On a separate pr we can change what gets reported to the user for all cases of failed network replies.

Issue #822, opened 2020-07-29
Support for staking out points


saberraz:
https://en.wikipedia.org/wiki/Construction_surveying#:~:text=Construction%20surveying%20or%20building%20surveying,such%20as%20roads%20or%20buildings.

An example of how it works in other apps:
https://docs.emlid.com/reachrs2/common/reachview/survey/#staking-out-points

jancelin:
The objectives would be:
 * View the quality of GNSS reception (fix, float, single)
 * To select a point to reach.  Have the distance, the line to the point and the direction.
* To have an acquisition of geographic data averaged over X seconds with the acquisition
quality.

An example : 
![Screenshot_20200729-111400](https://user-images.githubusercontent.com/6421175/88781404-a4e02100-d18c-11ea-8d0f-835906d4ee49.png)

pchirouze:
Good idea !

jancelin:
Hi @saberraz,
 I am relaunching the discussion on this subject.
 how is it possible to imagine the development of this feature?

saberraz:
Hi @jancelin 

This is indeed a nice feature. We usually ask for these types of features to be funded. Would you be interested in sponsoring the work? Feel free to drop us an email.

saberraz:
It would be good to display the coordinates in QGIS CRS as well as lat/long
![Screenshot_20211007-121700_UTM_Map](https://user-images.githubusercontent.com/2649047/136356981-9b193b34-d7e4-4357-b798-8ac9ba889570.png)

saberraz:
Here is my proposed workflow:

1- User selects an existing point on the map (we will only support Point layer)
2- The preview panel comes with an extra option: Navigate to: (DISPLAY NAME of)  the feature
![Untitled-2021-11-12-0848(1)](https://user-images.githubusercontent.com/2649047/141431648-926170db-670c-4934-95f3-726cc514f3bf.png)

wonder-sk:
Few notes from brainstorming:
- auto-zooming / auto-panning can be disabled if user manually moves/zooms map - in such situation we would show a button to re-center which will enable auto-zooming / auto-panning again
- when stake out mode is closed, we show the preview panel, and if the point is outside of the current map view, we will re-center
- it may be useful to show the feature title when in stake out mode
- icon to enter stake out to be picked
- accuracy will be shown in the "bubble" as it is now (not inside the panel)

tomasMizera:
[...]

PR #1920, opened 2022-02-08
Stake out - short mode

tomasMizera:
PR adds short mode for stakeout. 
Short mode is activated when your position is below one meter from the destination point. It allows users with high-precision GNSS receivers to better orientate in the field when going close to a feature.

When distance between your position and target point is under 10 cm, navigation goes green.

Showcase:

https://user-images.githubusercontent.com/22449698/152962912-d8c143b0-14b6-45f8-adbe-ebeb154b8883.mp4

**Visual UPDATE:**

https://user-images.githubusercontent.com/22449698/153176776-96313f01-4d20-46e1-86fb-fae9598699a7.mp4

For test purposes I also added another feature - long tap on a map changes your position to the place you tapped. _This is working only when you use simulated provider_

There are few issues to address in next PR:
 - make the navigation scale properly on different screens - currently I use few hard coded values
 - show GPS panel button (+ handle back button properly)
 - ~~show "recenter button"~~
 - consider size of gps position marker when changing extent (so that your position is visible on the new extent)

Another part of #822 (basically almost all of that, but let's close it after those minor issues are fixed)

### DIFF
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -1390,8 +1390,9 @@ qreal InputUtils::distanceToFeature( QgsPoint gpsPosition, const FeatureLayerPai
 
   QgsDistanceArea distanceArea;
   distanceArea.setSourceCrs( mapSettings->destinationCrs(), mapSettings->transformContext() );
-
-  return distanceArea.measureLine( gpsPosition, targetPoint );
+  qreal distance = distanceArea.measureLine( gpsPosition, targetPoint );
+  distance = distanceArea.convertLengthMeasurement( distance, QgsUnitTypes::DistanceMeters );
+  return distance;
 }
 
 qreal InputUtils::bearingToFeature( QgsPoint gpsPosition, const FeatureLayerPair &targetFeature, QgsQuickMapSettings *mapSettings )
@@ -1440,7 +1441,6 @@ qreal InputUtils::bearingToFeature( QgsPoint gpsPosition, const FeatureLayerPair
   QgsDistanceArea distanceArea;
   distanceArea.setSourceCrs( mapSettings->destinationCrs(), mapSettings->transformContext() );
 
-//  return distanceArea.measureLine( gpsPosition, targetPoint );
   return distanceArea.bearing( gpsPosition, targetPoint );
 }
 
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -1345,16 +1345,16 @@ QgsRectangle InputUtils::navigationFeatureExtent( const FeatureLayerPair &target
   return bbox;
 }
 
-QString InputUtils::distanceToFeature( QgsPoint gpsPosition, const FeatureLayerPair &targetFeature, QgsQuickMapSettings *mapSettings )
+qreal InputUtils::distanceToFeature( QgsPoint gpsPosition, const FeatureLayerPair &targetFeature, QgsQuickMapSettings *mapSettings )
 {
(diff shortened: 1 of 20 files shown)


app/inpututils.cpp
```diff
@@ -1354,15 +1381,65 @@ QString InputUtils::distanceToFeature( QgsPoint gpsPosition, const FeatureLayerP
...
+
+  QgsVectorLayer *layer = targetFeature.layer();
+  QgsFeature f = targetFeature.feature();
+
+  // Only points are supported
+  if ( layer->geometryType() != QgsWkbTypes::GeometryType::PointGeometry )
+    return -1;
+
+  // Transform gps position to map CRS
+  QgsCoordinateReferenceSystem wgs84( "EPSG:4326" );
```
varmar05, line 1408: why this differs from the line (686) in previous function?
QgsCoordinateReferenceSystem crsGPS = coordinateReferenceSystemFromEpsgId( 4326 )

app/inpututils.cpp
```diff
@@ -1354,15 +1381,65 @@ QString InputUtils::distanceToFeature( QgsPoint gpsPosition, const FeatureLayerP
...
+    catch ( QgsCsException &e )
+    {
+      Q_UNUSED( e )
+      return -1;
+    }
+  }
+
+  // Transform target point to map CRS
+  QgsPoint target( extractPointFromFeature( targetFeature ) );
+  QgsCoordinateTransform ct2( targetFeature.layer()->crs(), mapSettings->destinationCrs(), mapSettings->transformContext() );
```
varmar05, line 1425: maybe wrap in some util function? (the same as above)

app/position/simulatedpositionprovider.h
```diff
@@ -33,9 +33,9 @@ class SimulatedPositionProvider : public AbstractPositionProvider
      *  Set flightRadius to 0 in order to get constant position (no movement)
      */
     explicit SimulatedPositionProvider(
-      double longitude = 17.130,
-      double latitude = 48.130,
-      double flightRadius = 0.1,
+      double longitude = 17.107137342092614,
```
varmar05, line 36: maybe add some comment on this, why is set with such precision?

app/qml/NavigationPanel.qml
```diff
@@ -11,222 +11,390 @@ import QtQuick 2.14
...
+  property real featureToGpsDistance: -1
+  property var navigationTargetFeature
+
+  property var mapCanvas
+
+  property bool autoFollow: true
+  property var calculatedNavigationExtent
+  property string mapStateBeforeNavigation
+
+  property real distanceThresholdToShortMode: 1 // in metres
```
varmar05, line 36: just wondering if this should be hardcoded

wonder-sk: Let's keep it hard-coded for the time being... it will be easy to add configuration to the settings later if needed :-)

app/inpututils.cpp
tomasMizera: Unified now

app/inpututils.cpp
tomasMizera: Can you please check again if it is fine now? I changed it in this commit: https://github.com/lutraconsulting/input/pull/1920/commits/76d987e781be54c2c4cb81873b74b980f2a0ac00

varmar05: it is ok

varmar05 (approved):

PR #1921, opened 2022-02-08
Fix highlight numerical precision issues

wonder-sk:
Instead of using map coordinates for line/polygon shapes and then `Transform` for map to screen conversion, we now use screen coordinates at the time the highlight is constructed ("reference") and the `Transform` just adjusts to the change of scale/offset when the map is moved/zoomed. While both of these approaches seem equivalent, the difference is that in the former case, the `Transform` would deal with big numbers (for offsets), causing numerical issues due to single precision floating point math. Now the calculation with big numbers is done with doubles before creating shapes, and the transform deals with much smaller transforms. This does not make the issue go completely away, but it should be much less apparent under normal circumstances.

Also fixes typos "highligh" -> "highlight" and increases default polygon outline width from 1dp to 2dp as 1dp is a bit too thin.


tomasMizera (approved):
Great job!

PR #1939, opened 2022-02-15
Stakeout extent movements

tomasMizera:
Previous movement of map extent during stakeout was "very shaky". Extent have been unlimitedly zoomed in resulting in scales like 5mm or even smaller -> in that case even smallest change of gps position looked like movement of a 1 meter.

We are now moving extent based on conditions mentioned here: https://github.com/lutraconsulting/input/compare/stakeout-extent?expand=1#diff-7855d101d45d951847ae00d48eb0839f4cae76de50783fd4a4db4682b781e522R1334-R1341 Generally, we use distance between GPS and target point to determine the correct scale factor. We also do not zoom in more than 1:25 scale.

There were also several other issues fixed / enhancements:
 - in `PreviewPanel.qml` I removed redundant `Connections` component and  used property binding
 - in `PreviewPanel.qml` edit icon was shifted to the left when stakeout icon was invisible (for non-point layers)
 - in `MapPosition.cpp` I replaced use of QPointF with QgsPointXY. QPointF uses floats internally which led to other problems with floating point precision, mainly unterminated loops
 - Computation of map extent is moved to `MapWrapper`, it should not be anywhere else
 - Added signal to `qgsquickmapcanvas.qml` to be able to determine whether user iteracted with map

<table>
  <tr>
    <th>Previous</th>
    <th>New</th>
  </tr>
  <tr>
    <td></td>
    <td></td>
  </tr>
 </table>

There is a change to `qgsquickmapcanvas.qml` component (from qgs quick), created PR to QGIS: https://github.com/qgis/QGIS/pull/47364

#{23djxek}[Review]

### DIFF
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -1028,6 +1028,13 @@ bool InputUtils::equals( const QPointF &a, const QPointF &b, double epsilon )
 
 bool InputUtils::equals( const QgsPointXY &a, const QgsPointXY &b, double epsilon )
 {
+  if ( a.isEmpty() && b.isEmpty() )
+    return true;
+  if ( a.isEmpty() && !b.isEmpty() )
+    return false;
+  if ( !a.isEmpty() && b.isEmpty() )
+    return false;
+
   return qgsDoubleNear( a.x(), b.x(), epsilon ) && qgsDoubleNear( a.y(), b.y(), epsilon );
 }
 
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -1,4 +1,4 @@
-/***************************************************************************
+﻿/***************************************************************************
  *                                                                         *
  *   This program is free software; you can redistribute it and/or modify  *
  *   it under the terms of the GNU General Public License as published by  *
@@ -1019,6 +1019,16 @@ qreal InputUtils::calculateDpRatio()
   }
 
   return 1;
+}
+
+bool InputUtils::equals( const QPointF &a, const QPointF &b, double epsilon )
+{
+  return qgsDoubleNear( a.x(), b.x(), epsilon ) && qgsDoubleNear( a.y(), b.y(), epsilon );
+}
+
+bool InputUtils::equals( const QgsPointXY &a, const QgsPointXY &b, double epsilon )
+{
+  return qgsDoubleNear( a.x(), b.x(), epsilon ) && qgsDoubleNear( a.y(), b.y(), epsilon );
 }
 
 QString InputUtils::formatPoint(
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -1333,32 +1333,15 @@ QgsRectangle InputUtils::stakeoutPathExtent(
 
   //
   // In order to compute stakeout extent, we first compute distance to target feature and
-  // based on that we update the extent like this:
-  //   - if distance is > 10m, use 5m scale and center to GPS point
-  //   - if distance is 3-10m, use 2m scale and center to GPS point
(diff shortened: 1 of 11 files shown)


app/inpututils.cpp
```diff
@@ -1302,52 +1312,111 @@ QgsQuickMapSettings *InputUtils::setupMapSettings( QgsProject *project, QgsQuick
...
+  else if ( distance <= 10 && distance > 3 )
+  {
+    scale = 105; // ~ 2m scale
+
+    QgsPointXY centerInCanvasXY( gpsPointInCanvasXY.x(), gpsPointInCanvasXY.y() + panelOffset );
+    QgsPointXY center = mapSettings->screenToCoordinate( centerInCanvasXY.toQPointF() );
+
+    extent = mapSettings->mapSettings().computeExtentForScale( center, scale );
+  }
+  else if ( distance <= 3 && distance > 1 )
```
varmar05, line 1385: up to here it seems like action is the same only scale differs, maybe better to have distance2scale mapping and then just do the action (less repetitive)?

tomasMizera: Addressed

app/inpututils.cpp
```diff
@@ -1021,6 +1021,23 @@ qreal InputUtils::calculateDpRatio()
...
 }
 
+bool InputUtils::equals( const QPointF &a, const QPointF &b, double epsilon )
+{
+  return qgsDoubleNear( a.x(), b.x(), epsilon ) && qgsDoubleNear( a.y(), b.y(), epsilon );
+}
+
+bool InputUtils::equals( const QgsPointXY &a, const QgsPointXY &b, double epsilon )
+{
+  if ( a.isEmpty() && b.isEmpty() )
```
varmar05, line 1031: why these checks are only in PointXY version?

varmar05 (approved):

app/inpututils.cpp
```diff
@@ -1021,6 +1021,23 @@ qreal InputUtils::calculateDpRatio()
...
+    return true;
+  if ( a.isEmpty() && !b.isEmpty() )
+    return false;
+  if ( !a.isEmpty() && b.isEmpty() )
+    return false;
```
PeterPetrik, line 1036: if ( a.isEmpty() != b.isEmpty() )

app/inpututils.cpp
```diff
@@ -1302,52 +1319,103 @@ QgsQuickMapSettings *InputUtils::setupMapSettings( QgsProject *project, QgsQuick
   return settings;
 }
 
-QgsRectangle InputUtils::stakeoutFeatureExtent( const FeatureLayerPair &targetFeature, QgsPoint gpsPosition, QgsQuickMapSettings *mapSettings, double panelOffsetRatio )
+QgsRectangle InputUtils::stakeoutPathExtent(
+  MapPosition *mapPositioner,
```
PeterPetrik, line 1323: MapPosition *mapPosition

app/inpututils.cpp
```diff
@@ -1302,52 +1319,103 @@ QgsQuickMapSettings *InputUtils::setupMapSettings( QgsProject *project, QgsQuick
...
-    try
-    {
-      gpsPosition.transform( ct1 );
-    }
-    catch ( QgsCsException &e )
-    {
-      Q_UNUSED( e )
-      return QgsRectangle();
-    }
+    panelOffset = mapExtentOffset / 2;
```
PeterPetrik, line 1356: / 2.0 ?

PeterPetrik (approved):
looks good

tomasMizera:
Merging this one, auto tests are failing due to an inconsistent state on a test server. I will address that soon.

PR #1944, opened 2022-02-16
Fix camera issue on Android 12

### LINKED ISSUE #1904: Android 12 and blank white screen for taking photos
When I try to take a photo the camera appears with a blank white screen.

tomasMizera:
Android introduced property `queries` in manifest that should contain information about our use of external applications. It is called `Package visibility`. As we did not declare it, they disabled our access to device's camera.

In order to use the new `queries` keyword **we needed to update Gradle version** to new point release that Google released for [this use case](https://android-developers.googleblog.com/2020/07/preparing-your-build-for-package-visibility-in-android-11.html).

In the end the issue had nothing to do with `HIGH_SAMPLING_RATE_SENSORS` permission. However, on Android 12, sensor `TYPE_MAGNETIC_FIELD` (we use it in `OrientationSensor.java` to determine orientation when taking a photo) does not allow us to read data like we read it now (ASAP). 

_One more thing to note: Updating Gradle creates a mismatch between our and Qt 5.14.2's Gradle version. It is a difference only in point release_ 

Read more:
https://developer.android.com/training/package-visibility
https://developer.android.com/training/package-visibility/declaring
https://android-developers.googleblog.com/2020/07/preparing-your-build-for-package-visibility-in-android-11.html
https://stackoverflow.com/questions/67455534/how-to-add-queries-declaration-in-the-manifest-for-using-camera

Fixes #1904

### DIFF
--- a/app/android/AndroidManifest.xml
+++ b/app/android/AndroidManifest.xml
@@ -110,7 +110,7 @@
     <!-- Camera permissions -->
     <uses-permission android:name="android.permission.CAMERA"/>
 
-    <!-- In order to target Android 12 (API LVL 31), uncomment this permission, it is used for orientation sensor! -->
+    <!-- Add the following permission when app should target Android 12 (API LVL 31), it is used for orientation sensor! -->
     <!-- <uses-permission android:name="android.permission.HIGH_SAMPLING_RATE_SENSORS"/> -->
 
     <!-- The following comment will be replaced upon deployment with default features based on the dependencies of the application.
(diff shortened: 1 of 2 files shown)


saberraz:
Excellent work. It behaves as expected on my pixel.

app/android/AndroidManifest.xml
```diff
@@ -110,11 +110,22 @@
     <!-- Camera permissions -->
     <uses-permission android:name="android.permission.CAMERA"/>
 
+    <!-- In order to target Android 12 (API LVL 31), uncomment this permission, it is used for orientation sensor! -->
```
PeterPetrik, line 113: is this sentence correct?

PeterPetrik (approved):
good work!

app/android/AndroidManifest.xml
tomasMizera: Changed that

Issue #1821, opened 2021-12-08
Highlight of invalid polygon may cause crash


wonder-sk:
In some cases, our highlight of polygons (using Qt Quick Shapes) that have self-intersections may end up with a crash. It has been confirmed this can affect recording of new polygon features and in theory it may also cause crash when an existing polygon is picked and highlighted.

https://user-images.githubusercontent.com/193367/145278961-85bfe810-ad24-46fa-aca8-5b8e5a1a3bac.mp4

The underlying Qt problem seems to be https://bugreports.qt.io/browse/QTBUG-94233

A possible fix may be to always ensure that polygons passed to `ShapePath` are already "simple" (i.e. without self-intersecting ring) by processing them with GEOS.

Backtrace:
```
#0  QTriangulator<unsigned int>::ComplexToSimple::decompose() (this=this@entry=0x7fffffffc700) at painting/qtriangulator.cpp:939
#1  0x00007ffff3fa109a in QTriangulator<unsigned int>::triangulate() (this=this@entry=0x7fffffffc7b0) at painting/qtriangulator.cpp:785
#2  0x00007ffff3f9515d in qTriangulate(QVectorPath const&, QTransform const&, double, bool)
    (path=..., matrix=..., lod=lod@entry=1, allowUintIndices=allowUintIndices@entry=true) at painting/qtriangulator.cpp:2283
#3  0x00007fffc51a9ad4 in QQuickShapeGenericRenderer::triangulateFill(QPainterPath const&, QQuickShapeGenericRenderer::Color4ub const&, QVector<QSGGeometry::ColoredPoint2D>*, QVector<unsigned int>*, QSGGeometry::Type*, bool)                                                                                                            
    (path=..., fillColor=..., fillVertices=fillVertices@entry=0x555559a22fe8, fillIndices=fillIndices@entry=0x555559a22ff0, indexType=indexType@entry=0x555559a22ff8, supportsElementIndexUint=<optimized out>) at qquickshapegenericrenderer.cpp:436
#4  0x00007fffc51aacc4 in QQuickShapeGenericRenderer::endSync(bool) (this=0x555558107b10, async=<optimized out>) at qquickshapegenericrenderer.cpp:370
#5  0x00007fffc51a0fd4 in QQuickShapePrivate::sync() (this=this@entry=0x55555831cee0) at qquickshape.cpp:1131
[...]

PR #1950, opened 2022-02-18
Fix guideline elements after float precision fix in highlight

tomasMizera:
As we have changed how coordinates for `Highlight.qml` are calculated, recording guideline was not working correctly. It used map CRS coordinates for its geometry while other parts used screen's XY in pixels. This resulted in guideline being out of the map or on some weird places.

<table>
  <tr>
    <th>Previous</th>
    <th>Fixed</th>
  </tr>
  <tr>
    <td></td>
    <td></td>
  </tr>
 </table>

It also seems that with 939606b the issue https://github.com/lutraconsulting/input/issues/1821 is gone.

#{2695p3z}[Review]

### DIFF
--- a/app/qml/map/Highlight.qml
+++ b/app/qml/map/Highlight.qml
@@ -89,34 +89,16 @@ Item {
     return crosshairCoord
   }
 
-  function updateGuideLine() {
+  // Transforms X coordinate from map CRS to screen XY with regards to scale and HighDPI
+  function transformX( xcoord )
+  {
+    return ( xcoord + highlight.mapTransformOffsetX ) * highlight.mapTransformScale / displayDevicePixelRatio
+  }
 
-    if ( !guideLineAllowed ) { // remove any elements
-      let resetElems = []
-      resetElems.push ( componentMoveTo.createObject( guideLine ) )
-      guideLine.pathElements = resetElems
-    }
-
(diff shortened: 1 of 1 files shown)


PeterPetrik:
I assume this is for the current release, right? 
Also test failure is expected?

PeterPetrik:
Task linked: [CU-2695p3z Recording Glytch - release blocker](https://app.clickup.com/t/2695p3z)

tomasMizera:
Fixes #1952

tomasMizera:
@PeterPetrik Yes, it is for this sprint.

app/qml/map/Highlight.qml
```diff
@@ -216,19 +205,9 @@ Item {
...
   }
 
-  onGuideLineAllowedChanged: {
-    if ( guideLineAllowed )
-      constructHighlights()
-    else updateGuideLine()
-  }
-
-  onPositionChanged: {
-    if ( !recordingInProgress )
```
PeterPetrik, line 226: it looks like here are 2 changes: 1) simple fix of CRS 2) some cleaning (construct + update ===> constuct only). 

1) looks OK
2) I am not able to asses if the previous and new flow is the same ... (looks like now we do always some extra checks on l 122)

jozef-budac:
Happy as Larry :)

PR #1953, opened 2022-02-18
Unique user for autotests

### LINKED ISSUE #902: Unique user for testing
This is a suggestion and topic to discuss rather than reporting a bug.

While running automatic tests against deployed test.dev server with the same user, it might happen (with small but existing probability) that sets of tests running in parallel cause inconsistent state in DB for the user which can result into failing tests.
To avoid such coincidence and confusion it would be good to have a unique user for each set of tests.

tomasMizera:
PR makes auto tests fine to run in parallel by adding a unique user that is used during auto tests.

Autotests can now be used in two ways:
  1. We either pass env variables as we used before - in this case Input will use the provided credentials, or
  2. Do not pass credentials via env variables - in this case Input will register new unique user and fills its credentials to env var

I also made all test projects public, so that we have easier access to them if something goes wrong

Resolves #902 

#23dnz99[Review]

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -29,15 +29,6 @@ jobs:
   macos_build:
     if: ( github.repository == 'lutraconsulting/input' ) && (!contains(github.event.head_commit.message, 'Translate '))
     runs-on: macos-10.15
-    # Set concurrency to this job so that there will be only one workflow instance
-    # running throughout all branches. This limitation can be removed once we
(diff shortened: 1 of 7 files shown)


PeterPetrik:
Task linked: [CU-23dnz99 Unique user for testing](https://app.clickup.com/t/23dnz99)

tomasMizera:
Closes #902

PeterPetrik:
👏

tomasMizera:
I will merge this one once we tag and release master

PR #1954, opened 2022-02-18
Handle situation when position is lost during stakeout

tomasMizera:
When connection to external GPS was lost (e.g. when user minimizes the app) during stakout, stakeout panel had undefined behaviour (gets closed).


tomasMizera:
This one is going to release

PR #1961, opened 2022-02-22
Fix replaced lat/long in GPS data panel

### LINKED ISSUE #1960: Wrong cardinal direction in GPS panel
There is small problem in GPS panel for longitude and latitude.
Longitude should be East (E) or West(W) from London, but our GPS panel displays **N**
Latitude: should be North(N) or South(S) from equator but our GPS panel displays **E** 

a
b

tomasMizera:
`QGeoCoordinate::toString` function takes input in order `(lat, long)` which is different than in QGIS (and Input too) - `(long,lat)`.

Screenshot 2022-02-22 at 14 15 31

Good catch @jozef-budac!!

#26mg000[review]
Closes #1960

### DIFF
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -246,10 +246,12 @@ QString InputUtils::degreesString( const QgsPoint &point )
 {
   if ( point.isEmpty() )
   {
-    return QStringLiteral();
+    return QLatin1String();
   }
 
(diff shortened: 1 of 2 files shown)


PeterPetrik:
Task linked: [CU-26mg000 Wrong cardinal direction](https://app.clickup.com/t/26mg000)

PR #1962, opened 2022-02-22
Trigger repaint for empty geometry of guideline path

### LINKED ISSUE #1958: IOS - shadow of previous polygon in stream mode while recording another polygon feature
Input: 1.3.0 
IOS 15.1 

1. record feature in polygon layer or Multipolygon layer. 
2. record polygon (or Multipolygon) feature in stream mode (holding down GPS button) and you will see shadow glitch of 1. polygon you just did.

https://user-images.githubusercontent.com/94351263/154962334-b0d27f8d-5801-4614-b642-b7cf57df07a6.MP4

tomasMizera:
In #1950 I removed the entire `updateGuideline` function as we moved to a different handling of transformations in `Highlight.qml`.

However, I forgot to add a piece of code that triggers repaint of QML shape when the geometry is empty. Here: https://github.com/lutraconsulting/input/commit/59ed5f89ee45d5031fbae19725073a6eab7ab6a2#diff-c1350d1e69584b4f380f7721e2721f360b8f1417f9b94629abfd240dfd3df883L94-L98

Thus, the issue was that QML shape responsible for painting guideline did not repaint itself when it received empty geometry (newGuideLineElements was empty).

I have also added a little optimization - we do not need to reconstruct highlight when position changed in _NOT_ recording state. It is used only in guideline and that is visible only when we are in recording state.

Closes #1958

### DIFF
--- a/app/qml/map/Highlight.qml
+++ b/app/qml/map/Highlight.qml
@@ -1,4 +1,4 @@
-﻿/***************************************************************************
+/***************************************************************************
  *                                                                         *
  *   This program is free software; you can redistribute it and/or modify  *
  *   it under the terms of the GNU General Public License as published by  *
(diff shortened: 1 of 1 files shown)


PeterPetrik:
Task linked: [CU-26mfzy7 Recording bug](https://app.clickup.com/t/26mfzy7)

PeterPetrik (approved):

PR #1970, opened 2022-03-04
Handle expired token in listProjectsByName API call

### LINKED ISSUE #1499: Not possible to sync without opening my projects
Prerequisite:
 - you need to be logged in and have a project with modified state that can be synced
 - app is closed

To reproduce:
 - open application 
 - in home page try to sync the project
 
Experienced:
 - app says you are either not logged in or does not have sufficient rights

Expected:
 - project is synced

If one clicks on `My projects` and then back to `Home`, project can be synced then.
We are probably just not calling some API on app start because previously it was not possible to sync project from `Home`.

https://user-images.githubusercontent.com/22449698/122517675-9e5ed000-d010-11eb-8132-9ac9b8fb0f48.mp4

tomasMizera:
Calling `listProjectsByName` API was missing a check for auth token expiration. 
When token was expired, server returned for private projects 401, which in Input was understood as user without sufficient rights and did not allow user to sync the project.

I also added check for API status - it is in all other calls too.

Resolves #1499

### DIFF
--- a/app/test/testmerginapi.cpp
+++ b/app/test/testmerginapi.cpp
@@ -149,6 +149,38 @@ void TestMerginApi::testListProject()
   QVERIFY( !mApi->localProjectsManager().projectFromMerginName( mUsername, projectName ).isValid() );
 }
 
+void TestMerginApi::testListProjectsByName()
+{
+  QString projectName = "testListProjectByName";
+
+  // create the project on the server with other client
+  createRemoteProject( mApiExtra, mUsername, projectName, mTestDataPath + "/" + TEST_PROJECT_NAME + "/" );
+
+  // let's invalidate main client's auth token and see if the listProjectsByName gets new one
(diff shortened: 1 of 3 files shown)


app/test/testmerginapi.cpp
```diff
@@ -149,6 +149,38 @@ void TestMerginApi::testListProject()
...
+void TestMerginApi::testListProjectsByName()
+{
+  QString projectName = "testListProjectByName";
+
+  // create the project on the server with other client
+  createRemoteProject( mApiExtra, mUsername, projectName, mTestDataPath + "/" + TEST_PROJECT_NAME + "/" );
+
+  // let's invalidate main client's auth token and see if the listProjectsByName gets new one
+  QDateTime now = QDateTime::currentDateTimeUtc();
+  mApi->userAuth()->setTokenExpiration( now );
```
varmar05, line 161: maybe just set expiration to (now - 3 s) and you do not need to wait?

PR #1984, opened 2022-03-10
Loader cleanup

tomasMizera:
As part of autosync I started with Loader. It contained some parts that were not used anymore (like `mRecording`) and a lot of code that should not be there (rather be in InputUtils or main) like getting correct icon from layer based on its geometry or listening to change of application state.

I've also added new property - `project` - that bears information about local project that is currently active. Previously there was `project` property but it was a QgsProject instance, so I renamed it to `qgsProject`.

Some tests are added for functions that were moved from `Loader` to `Utils`. 

As of `SynchronizationController`, I only created the class and let it listen to signals when active project is changed.

Merging only to integration branch.

### DIFF
--- a/app/qml/ProjectIssuesPanel.qml
+++ b/app/qml/ProjectIssuesPanel.qml
@@ -21,7 +21,7 @@ Item {
   property real rowHeight: InputStyle.rowHeight
   property var projectIssuesModel: ListModel {}
   property string projectLoadingLog: ""
-  property string headerText: ""
+  property string headerText: qsTr( "The following layers failed loading" ) + ":"
 
   function reportIssue( layerName, message ) {
     projectIssuesModel.append( { name: layerName, message: message } );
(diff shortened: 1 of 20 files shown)


app/qml/ProjectIssuesPanel.qml
```diff
@@ -21,7 +21,7 @@ Item {
   property real rowHeight: InputStyle.rowHeight
   property var projectIssuesModel: ListModel {}
   property string projectLoadingLog: ""
-  property string headerText: ""
+  property string headerText: qsTr( "The following layers failed loading" ) + ":"
```
PeterPetrik, line 24: relevant change?

PeterPetrik (approved):
👏

app/qml/ProjectIssuesPanel.qml
tomasMizera: Yes, it was previously sent from the Loader

PR #1986, opened 2022-03-11
Add pre-commit hooks

tomasMizera:
I wanted to add our `format_cpp.sh` script as pre-commit hook, but I came across this cool software called [pre-commit](https://pre-commit.com/#installation). It utilizes different pre-commit hooks and let's you easily set them up via yaml configs.

There are tons of already created hooks (https://pre-commit.com/hooks.html) we can try to add (in future). I tried the `check-yaml` and it seems to work fine.

I added docs so that anyone else can install it, but it is trivial actually - just `brew`/`pip` install and then one command: `pre-commit install` in the repo root.

image

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -10,7 +10,7 @@ on:
     - '.github/workflows/linux.yml'
 
   release:
-    types
+    types:
       - published
 
 env:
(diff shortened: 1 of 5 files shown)


tomasMizera:
We could also use it in CI https://pre-commit.ci/

PeterPetrik (approved):
I assume it somewhat works on linux too?

tomasMizera:
Should work on all major platforms

PR #1990, opened 2022-03-15
fix issue when listprojectsbyname asks for login

tomasMizera:
In last sprint we fixed an issue with expired token in `ListProjectsByName` API.
This fix, however, did not only refresh the expired token, but also opted user to log in when his/her credentials were not set (not logged in).
Now it really just refreshes the token is auth data are set.

Before fix:

https://user-images.githubusercontent.com/22449698/158348713-940270cb-dd7b-4111-8a00-1613c95efb79.mp4

Now with fix it does not happen :)

### DIFF
--- a/core/merginapi.cpp
+++ b/core/merginapi.cpp
@@ -1,4 +1,4 @@
-﻿/***************************************************************************
+/***************************************************************************
  *                                                                         *
  *   This program is free software; you can redistribute it and/or modify  *
  *   it under the terms of the GNU General Public License as published by  *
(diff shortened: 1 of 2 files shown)


core/merginapi.cpp
```diff
@@ -127,8 +127,10 @@ QString MerginApi::listProjectsByName( const QStringList &projectNames )
     return QLatin1String();
   }
 
-  // Authentification is optional in this case, as there might be public projects without the need to be logged in
-  validateAuthAndContinute();
+  // Authentification is optional in this case, as there might be public projects without the need to be logged in.
+  // We only want to include auth token when user is logged in.
+  // User's token, however, might have already expired, so let's just refresh it.
+  refreshAuthToken();
```
varmar05, line 133: so what would happen if refresh is not successful? login window will appear?

varmar05:
I suggest to cherry pick this and apply to master, since it is not related to autosync?

core/merginapi.cpp
tomasMizera: Yes and no, this is public API, client sends a list of projects and server responds with their description. If request is sent without authorization header, private projects will bear only 401/we.

PR #1992, opened 2022-03-15
WIP: Autosync controller logic

tomasMizera:


app/activeprojectmanager.cpp
```diff
@@ -108,9 +125,9 @@ bool ActiveProjectManager::forceLoad( const QString &filePath, bool force )
     res = mQgsProject->read( filePath );
     mActiveLayer.resetActiveLayer();
     mMapThemeModel.reloadMapThemes( mQgsProject );
-    mProject = mLocalProjectsManager.projectFromProjectFilePath( filePath );
+    mProject = mLocalProjectsManager.projectPtrFromProjectFilePath( filePath );
```
PeterPetrik, line 128: I would keep original name of function "Ptr" is something you will see in your code-editor and doesn't bring any value to the reader of the code

app/autosynccontroller.cpp
```diff
@@ -8,33 +8,107 @@
...
 {
-  // raise NotImplemented :)
+  if ( !mActiveQgsProject || !mActiveProject )
+  {
+    CoreUtils::log( QStringLiteral( "Autosync" ), QStringLiteral( "Received an invalid active project data" ) );
+    return;
+  }
+
+  // Register for data change of project's vector layers
+  for ( const QgsMapLayer *layer : mActiveQgsProject->mapLayers( true ) )
```
PeterPetrik, line 32: not sure, but isn't better to call `mActiveQgsProject->mapLayers( true )` outside the for loop?

app/autosynccontroller.cpp
```diff
@@ -8,33 +8,107 @@
...
+  {
+    const QgsVectorLayer *vecLayer = qobject_cast<const QgsVectorLayer *>( layer );
+    if ( vecLayer )
+    {
+      QObject::connect( vecLayer, &QgsVectorLayer::afterCommitChanges, this, &AutosyncController::handleLocalChange );
+    }
+  }
+
+  if ( mActiveProject->localVersion < 0 )
```
PeterPetrik, line 41: can we create function in ActiveProject "isMerginProject" that is called here? If you need to write comment in the code it is usually better to have function with the comment content...

app/autosynccontroller.cpp
```diff
@@ -8,33 +8,107 @@
...
-{
-  // raise NotImplemented :)
-}
+#include "qgsproject.h"
+#include "qgsvectorlayer.h"
 
-AutosyncController::~AutosyncController()
+AutosyncController::AutosyncController(
+  LocalProject *openedProject,
+  QgsProject *openedQgsProject,
```
PeterPetrik, line 18: I am not completely sure without looking deeper, but do we need QgsProject as separate argument everywhere? why it is not part of LocalProject ?

app/autosynccontroller.cpp
```diff
@@ -8,33 +8,107 @@
...
 
+AutosyncController::~AutosyncController() = default;
+
 AutosyncController::SyncStatus AutosyncController::syncStatus()
 {
   return mSyncStatus;
 }
 
-void AutosyncController::setActiveProject( Project )
+void AutosyncController::synchronizationProgressed( const QString &projectFullName, qreal progress )
```
PeterPetrik, line 55: synchronization progressed -> what does it mean? it is running and the progress is updated?

app/autosynccontroller.cpp
```diff
@@ -8,33 +8,107 @@
...
+  }
+}
+
+void AutosyncController::synchronizationFinished( const QString &, const QString &projectFullName, bool successfully, int )
 {
-  // raise NotImplemented :)
+  if ( !mActiveProject )
+    return;
+
+  QString activeProjectName = MerginApi::getFullProjectName( mActiveProject->projectNamespace, mActiveProject->projectName );
```
PeterPetrik, line 76: I do not understand these checks all over the code. How it can happen that projectFullName != activeProjectName if AutoSync controller is destroyed on project change?

app/autosynccontroller.h
```diff
@@ -29,40 +30,45 @@ class AutosyncController : public QObject
...
+    void setSyncStatus( SyncStatus status );
 
     SyncStatus mSyncStatus = Synced;
 
-    Project *mActiveProject = nullptr;
-
-    QgsProject *mActiveQgsProject = nullptr;
+    LocalProject *mActiveProject = nullptr; // not owned
+    QgsProject *mActiveQgsProject = nullptr; // not owned
+    MerginApi *mBackend = nullptr; // not owned
```
PeterPetrik, line 66: I usually prefer to call instance m+name_of_class. better readable in the cpp file usually

app/synchronizationmanager.cpp
```diff
@@ -8,21 +8,36 @@
...
+  {
+    QObject::connect( mBackend, &MerginApi::syncProjectFinished, this, &SynchronizationManager::syncProjectFinished );
+    QObject::connect( mBackend, &MerginApi::syncProjectStatusChanged, this, &SynchronizationManager::syncProjectStatusChanged );
+  }
 
+  QObject::connect( this, &SynchronizationManager::autosyncAllowedChanged, this, [this]( bool allowed )
+  {
+    if ( allowed )
+    {
+      this->setupAutosync();
```
PeterPetrik, line 30: whatabout create `this->enableAutosync(boolean)` function?

app/synchronizationmanager.h
```diff
@@ -58,10 +62,20 @@ class SynchronizationManager : public QObject
...
+    void setupAutosyncController();
+
+    void clearAutosyncController();
+
     bool mAutosyncAllowed = false;
-    std::unique_ptr<AutosyncController> mAutosyncController;
+    std::unique_ptr<AutosyncController> mAutosyncController; // owned
 
-    MerginApi *mBackend = nullptr;
+    MerginApi *mBackend = nullptr; // not owned
```
PeterPetrik, line 77: mMerginApi ?

PeterPetrik (commented):
nice nice...

app/activeprojectmanager.cpp
tomasMizera: Yeah, but there are actually two functions, one (old one) is `projectFromProjectFilePath` which returns a copy and this new one `projectPtrFromProjectFilePath` that returns pointer to original data. So I could not keep the same name

app/autosynccontroller.cpp
tomasMizera: Yeah, ideally we should consider this. Currently `LocalProject` does not include `QgsProject` as you can have LocalProject instance even without it being loaded. We could have something similar to `ActiveLayer` class (maybe `ActiveProject`) that would group these and possibly other properties that are needed for opened (active) project.

app/autosynccontroller.cpp
tomasMizera: Correct, it is a slot

app/autosynccontroller.cpp
PeterPetrik: best to extract to separate call and store to const var, for readability

app/autosynccontroller.cpp
PeterPetrik: synchronizationProgressed -> onSynchronizationProgressUpdated?

PeterPetrik:
@tomasMizera this iis WIP, right?

tomasMizera:
@PeterPetrik Yes, I will add the refactored code here.

app/activeprojectmanager.cpp
tomasMizera: Addressed in refactoring based on review

app/autosynccontroller.cpp
tomasMizera: No longer valid

app/synchronizationmanager.cpp
tomasMizera: Outdated now due to refactoring

tomasMizera:
@PeterPetrik this one is back in review :))

PeterPetrik:
ok, lets do this :)

app/autosynccontroller.cpp
```diff
@@ -8,33 +8,34 @@
...
+  }
 
-void AutosyncController::setActiveQgsProject( QgsProject * )
-{
-  // raise NotImplemented :)
+  // Register for data change of project's vector layers
+  const QMap<QString, QgsMapLayer *> layers = mQgsProject->mapLayers( true );
+  for ( const QgsMapLayer *layer : layers )
+  {
+    const QgsVectorLayer *vecLayer = qobject_cast<const QgsVectorLayer *>( layer );
```
PeterPetrik, line 33: do we need all layers or only editable? wondering if you have some OSM vector layer bg map with 100x layers if we need to register all here ?

app/projectsmodel.h
```diff
@@ -125,77 +125,84 @@ class ProjectsModel : public QAbstractListModel
     Q_INVOKABLE void fetchAnotherPage( const QString &searchExpression );
 
     //! Merges local and remote projects based on the model type
-    void mergeProjects( const MerginProjectsList &merginProjects, Transactions pendingProjects, bool keepPrevious = false );
+    void mergeProjects( const MerginProjectsList &merginProjects, bool keepPrevious = false );
```
PeterPetrik, line 128: (optional) usually it is more readable instead of bool to use small enum with 2 options. so instead of keepPrevious = true/false have MergeProjectStrategy{KeepPrevious, DiscardPrevious(?)} so in callers you see what is done and not just `false`

app/synchronizationmanager.cpp
```diff
@@ -9,81 +9,165 @@
...
+    QObject::connect( mMerginApi, &MerginApi::pushCanceled, this, &SynchronizationManager::onProjectSyncCanceled );
+    QObject::connect( mMerginApi, &MerginApi::syncProjectFinished, this, &SynchronizationManager::onProjectSyncFinished );
+    QObject::connect( mMerginApi, &MerginApi::syncProjectStatusChanged, this, &SynchronizationManager::onProjectSyncProgressChanged );
+  }
 }
 
-SynchronizationManager::~SynchronizationManager()
+SynchronizationManager::~SynchronizationManager() = default;
+
+void SynchronizationManager::syncProject( const Project &project, bool withAuth )
```
PeterPetrik, line 29: so the SynchronizationTransaction is not returned? or it will be added later? (as it is in the schema )

PeterPetrik (commented):
👏 nice!

app/synchronizationmanager.cpp
tomasMizera: No, it is not returned, synchronization status is read directly from `SynchronizationManager`. SynchronizationManager only informs ProjectsModel that some data has changed. SynchronizationManager holds all the sync data.

PeterPetrik: So the schema is not updated?

PeterPetrik (approved):

app/projectsmodel.h
tomasMizera: That is very cool, thanks, I added that

app/autosynccontroller.cpp
tomasMizera: Good idea! Like this? https://github.com/lutraconsulting/input/pull/1992/commits/e6d292b28afd67d4bdfc32fa53b1e2765051be05

PR #2003, opened 2022-03-25
elide project name from the left side to keep project name more readable (fix #1996)

### LINKED ISSUE #1996: Shorten the namespace display
Some users/organisations namespace is long. In Input, it shortens the name from right, so it often results in project names not visible in small screens. We should probably shorten the namespace and display more of the project name.

For example for a project with the following name:
`mylongorganisationnamespace\myprojectwithlongname`

Current case:
`mylongorganisationnamespace\pr...`

Proposed:
`mylo..\myprojectwithlongname`

alexbruy:
Some users/organisations have long names. This results in non-readable project names on small screens, as the names are eldided from the right side. To make project names more readable better to elide their names from the left.

Fixes #1996.

Before
image
After
image

### DIFF
--- a/app/qml/components/ProjectDelegateItem.qml
+++ b/app/qml/components/ProjectDelegateItem.qml
@@ -164,7 +164,7 @@ Rectangle {
                else root.secondaryColor
         horizontalAlignment: Text.AlignLeft
         verticalAlignment: Text.AlignBottom
-        elide: Text.ElideRight
+        elide: Text.ElideLeft
       }
 
       Text {


app/qml/components/ProjectDelegateItem.qml
```diff
@@ -164,7 +164,7 @@ Rectangle {
                else root.secondaryColor
         horizontalAlignment: Text.AlignLeft
         verticalAlignment: Text.AlignBottom
-        elide: Text.ElideRight
+        elide: Text.ElideLeft
```
PeterPetrik, line 167: That's one small step for man, one giant leap for "organizations-with-long-names"

PeterPetrik (approved):

tomasMizera (approved):

PR #2047, opened 2022-04-11
reload project if conflict copies found after sync (fix #1798)

### LINKED ISSUE #1798: Some functionality broken after sync that created conflict file of a gpkg
Steps to reproduce:
1. create a project with a geopackage that is going to break geodiff's rebase (e.g. add a custom trigger)
2. open the project in Input, modify the geopackage (e.g. add a feature)
3. open the project elsewhere (e.g. in qgis plugin or other input) and also update the geopackage (but make sure not to update the QGIS project if using qgis plugin)
4. sync the project in Input - this should cause creation of a conflict file with local changes and the server version will replace the local changes

After these steps, some things will not work:
- adding/editing/removing features in that geopackage will be broken
- map refresh may sometimes show and sometimes not show the features from the geopackage (broken access to spatial index)

If user restarts Input, these issues are gone (switching to another project and back should work too).

It seems that OGR provider still keeps connection to the original DB file that got renamed to a conflict copy, at the same time there's a new file with that name but different content... and therefore things break. Probably reloading QGIS project after a conflict file got created will fix the issue.

alexbruy:
Check if there are conflicted copies after sync and if there are any we will reload active project in order to reset OGR provider connections to layers.

Fixes #1798.

### DIFF
--- a/app/activeproject.cpp
+++ b/app/activeproject.cpp
@@ -125,10 +125,6 @@ bool ActiveProject::forceLoad( const QString &filePath, bool force )
   if ( !force )
     emit loadingStarted();
 
-  // store current project extent
-  if ( mMapSettings )
-    QgsRectangle extent = mMapSettings->extent();
-
   QFile flagFile( LOADING_FLAG_FILE_PATH );
   flagFile.open( QIODevice::WriteOnly );
   flagFile.close();
@@ -215,9 +211,6 @@ bool ActiveProject::forceLoad( const QString &filePath, bool force )
   {
     setAutosyncEnabled( true );
   }
-
-  if ( mMapSettings )
-    mMapSettings->setExtent( extent );
 
   return res;
 }
@@ -370,4 +363,19 @@ void ActiveProject::setActiveLayer( QgsMapLayer *layer ) const
     mActiveLayer.setActiveLayer( layer );
     mAppSettings.setDefaultLayer( mActiveLayer.layerName() );
   }
+}
+
+bool ActiveProject::reloadPreservingExtent()
+{
+  // store current project extent
+  QgsRectangle extent;
+  if ( mMapSettings )
+    extent = mMapSettings->extent();
+
+  bool res = forceLoad( mQgsProject->fileName(), true );
+
+  if ( mMapSettings && !extent.isNull() )
+    mMapSettings->setExtent( extent );
+
+  return res;
 }
--- a/app/activeproject.cpp
+++ b/app/activeproject.cpp
@@ -125,6 +125,9 @@ bool ActiveProject::forceLoad( const QString &filePath, bool force )
   if ( !force )
     emit loadingStarted();
 
+  // store current project extent
+  QgsRectangle extent = mMapSettings->extent();
+
   QFile flagFile( LOADING_FLAG_FILE_PATH );
   flagFile.open( QIODevice::WriteOnly );
   flagFile.close();
@@ -211,6 +214,8 @@ bool ActiveProject::forceLoad( const QString &filePath, bool force )
   {
     setAutosyncEnabled( true );
   }
+
+  mMapSettings->setExtent( extent );
 
   return res;
 }
--- a/app/activeproject.cpp
+++ b/app/activeproject.cpp
@@ -369,12 +369,3 @@ void ActiveProject::setActiveLayer( QgsMapLayer *layer ) const
     mAppSettings.setDefaultLayer( mActiveLayer.layerName() );
   }
 }
-
-bool ActiveProject::hasConflictedCopies()
-{
-  LocalProject project = localProject();
-  if ( project.isValid() )
-    return project.hasConflictedCopies();
-
-  return false;
-}
--- a/app/activeproject.cpp
+++ b/app/activeproject.cpp
@@ -125,6 +125,11 @@ bool ActiveProject::forceLoad( const QString &filePath, bool force )
   if ( !force )
     emit loadingStarted();
 
+  // store current project extent
+  QgsRectangle extent;
+  if ( mMapSettings )
+    extent = mMapSettings->extent();
+
   QFile flagFile( LOADING_FLAG_FILE_PATH );
   flagFile.open( QIODevice::WriteOnly );
   flagFile.close();
@@ -211,6 +216,10 @@ bool ActiveProject::forceLoad( const QString &filePath, bool force )
   {
     setAutosyncEnabled( true );
   }
+
+  // restore extent
+  if ( mMapSettings && !extent.isNull() )
+    mMapSettings->setExtent( extent );
 
   return res;
 }
@@ -363,19 +372,4 @@ void ActiveProject::setActiveLayer( QgsMapLayer *layer ) const
     mActiveLayer.setActiveLayer( layer );
     mAppSettings.setDefaultLayer( mActiveLayer.layerName() );
   }
-}
-
-bool ActiveProject::reloadPreservingExtent()
-{
-  // store current project extent
-  QgsRectangle extent;
-  if ( mMapSettings )
-    extent = mMapSettings->extent();
-
-  bool res = forceLoad( mQgsProject->fileName(), true );
-
-  if ( mMapSettings && !extent.isNull() )
-    mMapSettings->setExtent( extent );
-
-  return res;
 }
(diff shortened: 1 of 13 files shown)


app/qml/map/MapWrapper.qml
```diff
@@ -710,6 +710,13 @@ Item {
           {
             // just banner
             syncSuccessfulBanner.show()
+
+            // check for confliected copies and reload project if any to
+            // reset connections to the GPKG files (see #1798)
+            if ( __activeProject.hasConflictedCopies() )
+            {
+              __activeProject.reloadProject( __activeProject.qgsProject.homePath )
+            }
```
wonder-sk, line 719: It would be good to have this kind of logic in c++ rather than in qml - maybe inside ActiveProject class?

core/project.cpp
```diff
@@ -25,6 +25,22 @@ QString LocalProject::fullName() const
...
+    QDirIterator it( projectDir, QStringList() << "*conflicted copy*", QDir::Files | QDir::NoDotAndDotDot, QDirIterator::Subdirectories );
+    while ( it.hasNext() )
+    {
+      it.next();
+      count++;
+    }
+    return count > 0 ? true : false;
+  }
+  return false;
+}
```
wonder-sk, line 42: shall we ignore .mergin directory, just in case?

wonder-sk (commented):
I guess there's unintended modification of proj.db in this PR?

Looking at the implementation, I am wondering about one more case: 1. user A modifies gpkg schema in qgis plugin (e.g. remove a column) and syncs, 2. user B syncs in Input (with no local changes). In this scenario, there should be no conflict file created, but still we will probably have to reload project, because OGR may not realize the schema has changed, and adding/updating features in the modified table would probably fail... If that's true, I guess we will need a flag from MerginApi when pull is finished, signalling whether any gpkg file was updated by copying (without geodiff), and reload project in such case

core/project.cpp
alexbruy: At least on unix-based systems `QDirIterator` by default ignores hidden files and directories. But to be safe additional check was added.

app/qml/map/MapWrapper.qml
alexbruy: Moved to C++.

app/activeproject.cpp
```diff
@@ -364,3 +369,12 @@ void ActiveProject::setActiveLayer( QgsMapLayer *layer ) const
     mAppSettings.setDefaultLayer( mActiveLayer.layerName() );
   }
 }
+
+bool ActiveProject::hasConflictedCopies()
```
tomasMizera, line 373: Leftover, please remove

core/project.h
```diff
@@ -74,6 +74,9 @@ struct LocalProject
     //!       Project->isMergin() is recommended to use over this one
     bool hasMerginMetadata() const { return localVersion > -1; }
 
+    //! Returns true if project has conflicted copies
+    bool hasConflictedCopies();
```
tomasMizera, line 78: Leftover, please remove

core/merginapi.h
```diff
@@ -610,6 +610,7 @@ class MerginApi: public QObject
     QString mApiRoot;
     LocalProjectsManager &mLocalProjects;
     QString mDataDir; // dir with all projects
+    bool mReloadedNeeded = false; // indicate whether project reload is needed after sync
```
tomasMizera, line 613: Let's move this flag to `TransactionStatus` as there might be several syncs running at the same time. 
We can also call it something like "GpkgSchemaChanged" or w/e related to what actually happened.

app/main.cpp
```diff
@@ -485,6 +485,13 @@ int main( int argc, char *argv[] )
   QObject::connect( &activeProject, &ActiveProject::projectWillBeReloaded, &inputProjUtils, &InputProjUtils::resetHandlers );
   QObject::connect( &pw, &ProjectWizard::notify, &iu, &InputUtils::showNotificationRequested );
   QObject::connect( &iosUtils, &IosUtils::showToast, &iu, &InputUtils::showNotificationRequested );
+  QObject::connect( &syncManager, &SynchronizationManager::syncFinished, &activeProject, [&activeProject]( const QString & projectFullName, bool successfully, int version, bool reloadNeeded )
+  {
+    if ( reloadNeeded )
+    {
+      activeProject.reloadProject( activeProject.qgsProject()->homePath() );
```
tomasMizera, line 492: As we discussed, let's have a dedicated function for reloading of project (it can actually be this one when you remove the duplicated reload connection here https://github.com/lutraconsulting/input/pull/2047/files#diff-c1b709b8a90bef4573ddad8a9b2fbc5d5dbc85e4dc759f95b6d9905931fec24fR482). It does not need to receive any arguments as `activeProject` already has the correct `qgsProject` set.

app/activeproject.cpp
```diff
@@ -212,6 +215,8 @@ bool ActiveProject::forceLoad( const QString &filePath, bool force )
     setAutosyncEnabled( true );
   }
 
+  mMapSettings->setExtent( extent );
```
tomasMizera, line 218: This can probably lead to crash

app/activeproject.cpp
```diff
@@ -125,6 +125,9 @@ bool ActiveProject::forceLoad( const QString &filePath, bool force )
   if ( !force )
     emit loadingStarted();
 
+  // store current project extent
+  QgsRectangle extent = mMapSettings->extent();
```
tomasMizera, line 129: This can probably lead to crash

core/merginapi.cpp
```diff
@@ -2887,7 +2891,7 @@ void MerginApi::finishProjectSync( const QString &projectFullName, bool syncSucc
   }
   else
   {
-    emit syncProjectFinished( projectFullName, syncSuccessful, newVersion );
+    emit syncProjectFinished( projectFullName, syncSuccessful, newVersion, mReloadedNeeded );
```
tomasMizera, line 2894: As agreed over call, there is a duplicate signal called `reloadProject` just few lines below this one.
Let's merge them together in the `reloadNeeded` argument of `syncProjectFinished`.

MerginApi will once more be less coupled with the rest of the app 🎉

tomasMizera (changes requested):
As discussed over call

alexbruy:
Test failure seems unrelated

app/activeproject.cpp
```diff
@@ -364,3 +364,18 @@ void ActiveProject::setActiveLayer( QgsMapLayer *layer ) const
     mAppSettings.setDefaultLayer( mActiveLayer.layerName() );
   }
 }
+
+bool ActiveProject::reloadPreservingExtent()
```
PeterPetrik, line 368: is there any situation when you do not want to preserve extent? if yes, maybe add `preserveExtent=True` to `reload()` function?

alexbruy: From my point of view we should preserve extent all the time, but I can be not aware about all possible use cases.

tomasMizera: A function called `reloadProject` is currently not being used anywhere (as we removed the duplicated signal in MerginApi). I'd say let's have only the `reloadProject` function that is going to preserve the extent (as it is only used on one place)

tomasMizera:
Hi @alexbruy, can you also remove the `reloadProject` signal definition from Merginapi.h:455?

core/merginapi.cpp
```diff
@@ -2884,21 +2895,19 @@ void MerginApi::finishProjectSync( const QString &projectFullName, bool syncSucc
     QString projectNamespace, projectName;
     extractProjectName( projectFullName, projectNamespace, projectName );
     pushProject( projectNamespace, projectName );
+    // preserve information about GPKG schema changes detected at pull
+    mTransactionalStatus[projectFullName].gpkgSchemaChanged = hasSchemaChange;
```
tomasMizera, line 2899: This can not be there. `[]` operator of QHash inserts a new element to the hashmap if there is none already.

We need to preserve this information differently - I suggest storing it in sync manager.

app/activeproject.cpp
```diff
@@ -125,6 +125,11 @@ bool ActiveProject::forceLoad( const QString &filePath, bool force )
   if ( !force )
     emit loadingStarted();
 
+  // store current project extent
+  QgsRectangle extent;
+  if ( mMapSettings )
+    extent = mMapSettings->extent();
```
tomasMizera, line 131: I suggest to have this in the `reloadProject` function :)

tomasMizera (changes requested):

tomasMizera (approved):
Looks good to me. Thanks for addressing the comments.

@wonder-sk / @PeterPetrik would you like to have a look or can we merge this?

PR #2051, opened 2022-04-13
Show error message when adding/editing/removing feature fails (fix #1797) 

### LINKED ISSUE #1797: No error shown when adding/editing/removing feature fails
If user got into some trouble and changes are not getting committed for some reason, this is only logged to the diagnostic log, but otherwise user gets no warning at all. This is not a very common situation, but when it happens, it's really bad...

We should show a big warning that changes failed to save, it would also make sense to not close the form where the add/edit/remove action was triggered.

Failed addition of a new feature:
```
2021-12-07T09:13:13.205Z QGIS : Warning: Commit errors:
ERROR: 1 feature(s) not added.

Provider errors:
OGR error creating feature -3: failed to execute insert : attempt to write a readonly database
2021-12-07T09:13:13.206Z QGIS Input: Error: Could not save changes. Rolling back.
2021-12-07T09:13:13.207Z QGIS OGR: Warning: Unbalanced call to leaveUpdateMode() w.r.t. enterUpdateMode()
```

Failed removal of a feature:
```
2021-12-07T09:38:28.191Z QGIS : Warning: Commit errors:
ERROR: 1 feature(s) not deleted.

Provider errors:
OGR error deleting feature 47: sqlite3_exec(DELETE FROM "field_notes" WHERE "fid" = 47) failed: attempt to write a readonly database
2021-12-07T09:38:28.191Z QGIS Input: Error: Could not save changes. Rolling back.
2021-12-07T09:38:28.192Z QGIS OGR: Warning: Unbalanced call to leaveUpdateMode() w.r.t. enterUpdateMode()
```

Failed edit of a geometry:
```
2021-12-07T09:40:27.627Z QGIS : Warning: Commit errors:
ERROR: 1 geometries not changed.
ERROR: 1 attribute value change(s) not applied.

Provider errors:
OGR error setting feature 47: failed to execute update : attempt to write a readonly database
OGR error setting feature 47: failed to execute update : attempt to write a readonly database
2021-12-07T09:40:27.628Z QGIS Input: Error: Could not save changes. Rolling back.
2021-12-07T09:40:27.658Z QGIS OGR: Warning: Unbalanced call to leaveUpdateMode() w.r.t. enterUpdateMode()
```

alexbruy:
Show small warning banner when committing changes to vector layer fails.

зображення

Fixes #1797.

### DIFF
--- a/app/qml/form/FeatureFormPage.qml
+++ b/app/qml/form/FeatureFormPage.qml
@@ -14,6 +14,7 @@ import QtQuick.Dialogs 1.3
 import lc 1.0
 import ".."
 import "../components"
+import "../banners"
 
 Item {
   id: root
@@ -166,6 +167,12 @@ Item {
           onFormStateChanged: featureForm.state = root.formState
         }
 
+        Connections {
+          target: featureForm.controller
+          onChangesCommited: featureForm.saved()
+          onChangesRolledback: editingFailedBanner.show()
+        }
+
         Component.onCompleted: {
           if ( root.parentController && root.linkedRelation ) {
             featureForm.controller.parentController = root.parentController
@@ -232,6 +239,19 @@ Item {
           }
           visible = false
         }
+      }
+
+      AutoHideBanner {
+        id: editingFailedBanner
+
+        width: featureForm.width - 16
+        height: InputStyle.rowHeight
+
+        bgColor: InputStyle.warningBannerColor
+
+        source: InputStyle.noIcon
+
+        text: qsTr( "Failed to save changes." )
       }
 
       ExternalResourceBundle {
--- a/app/qml/form/FeatureFormPage.qml
+++ b/app/qml/form/FeatureFormPage.qml
@@ -14,7 +14,6 @@ import QtQuick.Dialogs 1.3
 import lc 1.0
 import ".."
 import "../components"
-import "../banners"
 
 Item {
   id: root
@@ -157,7 +156,7 @@ Item {
 
         onSaved: root.close()
         onCanceled: root.close()
-        onEditingFailed: editingFailedBanner.show()
+        onEditingFailed: editingFailedDialog.open()
         onOpenLinkedFeature: root.openLinkedFeature( linkedFeature )
         onCreateLinkedFeature: root.createLinkedFeature( parentController, relation )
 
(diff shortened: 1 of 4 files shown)


app/qml/form/FeatureFormPage.qml
```diff
@@ -234,6 +241,19 @@ Item {
         }
       }
 
+      AutoHideBanner {
+        id: editingFailedBanner
+
+        width: featureForm.width - 16
```
tomasMizera, line 247: Let's not use random constants in code (Look at file `InputStyle.qml`,  all constants are defined there). 
Moreover, pixel values needs to be multiplied by `* __dp`.

```suggestion
        width: featureForm.width - InputStyle.innerFieldMargin
```
or
```suggestion
        width: featureForm.width - InputStyle.smallGap
```

app/qml/form/FeatureFormPage.qml
```diff
@@ -166,6 +167,12 @@ Item {
           onFormStateChanged: featureForm.state = root.formState
         }
 
+        Connections {
```
tomasMizera, line 170: Let's move the entire `Connections` to `FeatureForm.qml` and emit feature form's `saved` signal on success. 

When there was an error in commit let's have a new signal in `FeatureForm`

tomasMizera (changes requested):
@wonder-sk I wonder what to do when feature can not be saved. 
The way it is implemented now is that user will see a banner for few seconds only with text "Failed to save changes." and form will not be closed.

Shouldn't we at least let user know to try to restart the app?

wonder-sk:
> The way it is implemented now is that user will see a banner for few seconds only with text "Failed to save changes." and form will not be closed.
>
> Shouldn't we at least let user know to try to restart the app?

Yeah I think it should be a modal dialog that needs to be closed explicitly by user, because it is quite a serious issue.

Good idea to give user some hint after the main message, I would suggest something like e.g. "This should not happen normally. Please restart the app and try again - if that does not help, please contact support."

alexbruy:
Banner replaced with modal dialog (build on Linux, as it is faster)

зображення

app/qml/form/FeatureFormPage.qml
```diff
@@ -236,17 +235,19 @@ Item {
...
+        visible: false
+        title: qsTr( "Saving failed" )
+        text: qsTr( "Failed to save changes. This should not happen normally. Please restart the app and try again — if that does not help, please contact support." )
+        icon: StandardIcon.Warning
+        standardButtons: StandardButton.Close
 
-        text: qsTr( "Failed to save changes. Try to restart app." )
+        //! Using onButtonClicked instead of onAccepted,onRejected which have been called twice
+        onButtonClicked: {
+          visible = false
```
tomasMizera, line 249: Let's keep it like this, but generally it is better to use `.open()` and `.close()` (instead of `visible = true/false`) because it does not break property binding.

alexbruy: I have followed the approach used in the existing dialogs. Do you want me to update all occurrences in this file?

tomasMizera: Let's keep it like this :)

Issue #2001, opened 2022-03-22
Build with iOS 15 SDK


PeterPetrik:
requirement until 25 April 2022

- also probably input-sdk?

Issue #1983, opened 2022-03-10
iOS crash on 12.x and 13.x


saberraz:
Some users have been reporting (see #1972 ) issues with earlier versions of iOS. In Android, you can detect the supported versions and do not display it for those on earlier versions.

I am not sure how it is done in iOS but we should not have the option for users to update to the latest version or install it from the App store if it is not supported.

PeterPetrik:
Not working on: iOS iPhone SE with 13.3

PeterPetrik:
@jozef-budac on which iOS OS versions we did tests?

tomasMizera:
We might be missing this one https://developer.apple.com/documentation/bundleresources/information_property_list/minimumosversion

saberraz:
@PeterPetrik we usually test it on the most recent version(s) of iOS. Our phones/tablets are still supported by the latest iOS version and we keep them up to date with Apple iOS upgrades.

PeterPetrik:
Not working: 12.5.5

lollisoft:
When you plan to deploy on older iOS versions like my version, you need testing for that. At least some test devices in pre flight. If you only test on current or latest versions, you risk unpredictable errors. I have XCode 10.3 and if known to be supported with your project, I'll give it a try...

PeterPetrik:
input-sdk change between 1.2.0 and 1.3.0 
https://github.com/lutraconsulting/input-sdk/compare/ios-20211201-22...ios-20220212-36

input change between 1.2.0 and 1.3.0
https://github.com/lutraconsulting/input/compare/1.2.0...1.3.0

jozef-budac:
We tested the latest release on iOS 15.1

lollisoft:
Hi,

I can report that the currently released app version works on my iOS 13.3 on iPhone SE. Now, I may be able to use the app for further testing within the app.

Thank you,

Lothar

![2022-04-22 10 51 41](https://user-images.githubusercontent.com/3633520/164670531-5a28f4e4-fff8-4c6e-ba54-247f1efd82fa.jpg)
![2022-04-22 10 51 51](https://user-images.githubusercontent.com/3633520/164670577-f9f13de3-9827-48ae-8fd7-833caf684b55.jpg)

PeterPetrik:
https://iosref.com/ios

[...]

PR #2057, opened 2022-04-22
SDK 15 on iOS

PeterPetrik:
following https://github.com/lutraconsulting/input-sdk/commit/300d1ae56c173daa492f52a8d2074fa8389c0a7d

fix https://github.com/lutraconsulting/input/issues/2001
fix https://github.com/lutraconsulting/input/issues/1983

also update of:
- exiv2=0.27.5
- expat=2.4.8 
- geodiff=1.0.6
- qgis=3.22.6


.github/workflows/ios.yml
```diff
@@ -226,7 +226,11 @@ jobs:
             QMAKE_MAC_XCODE_SETTINGS+=qstyle \
             qstyle.name=CODE_SIGN_STYLE \
             qstyle.value=Manual
-
+            
+          # Use old deprecated build system in XCode 13
+          /usr/libexec/PlistBuddy -c "Add :DisableBuildSystemDeprecationDiagnostic bool" ./Input.xcodeproj/project.xcworkspace/xcshareddata/WorkspaceSettings.xcsettings
+          /usr/libexec/PlistBuddy -c "Set :DisableBuildSystemDeprecationDiagnostic true" ./Input.xcodeproj/project.xcworkspace/xcshareddata/WorkspaceSettings.xcsettings
+
```
tomasMizera, line 233: Do we need to use the old build system?
Locally I applied this patch and it now generates the new build system https://codereview.qt-project.org/c/qt/qtbase/+/353119

PeterPetrik: if it works without it now, I would keep the old one  for this commit. we should focus on qt6 path rather.

tomasMizera: Ok, I agree, though it seems that it is neither fixed in qt6 https://bugreports.qt.io/browse/QTCREATORBUG-25368

tomasMizera (approved):
Good job!

PR #2063, opened 2022-05-03
Rebranding

tomasMizera:
Rebranding of MM in Input.

Changes:
 - new logos ~ replaced Mergin and Input logos in the splash screen, login, project loading and about page
 - added a  banner to tell users about rebranding (we can remove it  when rebranding is finished)

Sneak peek 🙂  :
**Homescreen**

**Login page**

**Splashscreen**


PeterPetrik (approved):

tomasMizera:
Linux build will be fixed in different PR and the same for iOS build (it _should_ work on master)

PR #2065, opened 2022-05-04
WIP: Fix linux build

tomasMizera:


PeterPetrik:
looks like it was only temporarly unreachable ubuntu repos, so I do not think we need to handle it

PR #2075, opened 2022-05-05
add special handling for the names of the conflicted files for QGIS project files

alexbruy:
As discussed in https://github.com/lutraconsulting/qgis-mergin-plugin/issues/382 add `~` at the end of the conflicted copy if this is a QGIS project file.

### DIFF
--- a/core/coreutils.cpp
+++ b/core/coreutils.cpp
@@ -214,7 +214,12 @@ QString CoreUtils::generateConflictedCopyFileName( const QString &file, const QS
 
   QFileInfo f( file );
 
-  return QString( "%1/%2 (conflicted copy, %3 v%4).%5" ).arg( f.path(), f.baseName(), username, QString::number( version ), f.completeSuffix() );
+  QString suffix = f.completeSuffix();
+  if ( !suffix.compare( QStringLiteral( "qgs" ), Qt::CaseInsensitive ) || !suffix.compare( QStringLiteral( "qgz" ), Qt::CaseInsensitive ) )
+  {
+    suffix += "~";
+  }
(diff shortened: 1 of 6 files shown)


core/coreutils.cpp
```diff
@@ -214,7 +214,12 @@ QString CoreUtils::generateConflictedCopyFileName( const QString &file, const QS
 
   QFileInfo f( file );
 
-  return QString( "%1/%2 (conflicted copy, %3 v%4).%5" ).arg( f.path(), f.baseName(), username, QString::number( version ), f.completeSuffix() );
+  QString suffix = f.completeSuffix();
+  if ( !suffix.compare( QStringLiteral( "qgs" ), Qt::CaseInsensitive ) || !suffix.compare( QStringLiteral( "qgz" ), Qt::CaseInsensitive ) )
```
tomasMizera, line 218: We have a function in Mergin API called `hasProjecFileExtension`, can you please move it here to coreutils, fix typo in its name and use it instead of checking suffixes here again? 🙂  Thanks! 

It would also be good to have a test for that function.

alexbruy: Done

tomasMizera (approved):
Thanks, lgtm 👏

core/coreutils.cpp
```diff
@@ -225,3 +230,8 @@ QString CoreUtils::generateEditConflictFileName( const QString &file, const QStr
   QFileInfo f( file );
   return QString( "%1/%2 (edit conflict, %3 v%4).json" ).arg( f.path(), f.baseName(), username, QString::number( version ) );
 }
+
+bool CoreUtils::hasProjecFileExtension( const QString filePath )
```
tomasMizera, line 234: `hasProjecFileExtension` the typo is still there 😄

PR #2083, opened 2022-05-11
Add delete account functionality (fix #1896)

### LINKED ISSUE #1896: Add option to remove account from within the app - APPLE REQUIREMENT - JUNE 30
Just found this in my inbox - I am not sure if deletion via web is sufficient, probably not:

**Account deletion within apps**. 
App Store Review Guideline 5.1.1 provides people with greater control over their personal data by stating that **all apps that allow for account creation must also allow users to initiate deletion of their account from within the app.** This requirement was set to apply to all app submissions by January 31, 2022. Due to the complexity of implementing this requirement, **we’ve extended the deadline to June 30, 2022 to give you more time.**

Please keep these requirements in mind when updating your app:

The account deletion option should be easy to find in your app.
**It’s insufficient to only provide the ability to temporarily disable or deactivate an account. People should be able to delete the account along with their personal data.**
Apps in highly-regulated industries may need to provide additional support flows to confirm and facilitate the account deletion process.
Follow applicable legal requirements for storing and retaining user account information. This includes complying with local laws in different countries or regions. As always, check with your legal counsel.

Learn more about account deletion in the Human Interface Guidelines:
[iOS and iPadOS apps](https://developer.apple.com/design/human-interface-guidelines/ios/user-interaction/accounts/)
[macOS apps](https://developer.apple.com/design/human-interface-guidelines/macos/user-interaction/accounts/)

alexbruy:
Add button to the Account page allowing user to remove Mergin Maps account.
зображення
When account removal requested user is informed that all local projects will be removed from the device and that he should manually cancel Apple subscription if there is one. Also user should type his username in the text field before proceeding.
зображення
In case if account removal failed an error message is shown.

Fixes #1896.

### DIFF
--- a/app/qml/AccountPage.qml
+++ b/app/qml/AccountPage.qml
@@ -326,14 +326,15 @@ Page {
       visible: false
       title: qsTr( "Delete account?" )
 
-      ColumnLayout {
+      contentItem: ColumnLayout {
         id: column
         width: parent.width
         Label {
           id: label
-          text: qsTr("This will remove all local projects from the device and your Mergin Maps account." +
-                     "If you have an Apple subscription you should cancel it manually.\n\n" +
-                     "To continue enter your username in the field below and click Yes.")
+          text: qsTr("This action will delete your Mergin Maps account with all your projects, " +
+                     "both on the device and on the server. This action cannot be undone." +
+                     "If you have an Apple subscription you need to cancel it manually.\n\n" +
+                     "In order to delete your account, enter your username in the field below and click Yes.")
           Layout.fillWidth: true
           wrapMode: Text.WordWrap
         }
--- a/app/qml/AccountPage.qml
+++ b/app/qml/AccountPage.qml
@@ -10,6 +10,8 @@
 import QtQuick 2.14
 import QtQuick.Controls 2.14
 import QtGraphicalEffects 1.14
+import QtQuick.Layouts 1.12
+import QtQuick.Dialogs 1.3 as Dialogs
 import lc 1.0
 import "."  // import InputStyle singleton
 import "./components"
@@ -19,6 +21,7 @@ Page {
   signal managePlansClicked
   signal signOutClicked
   signal restorePurchasesClicked
+  signal accountDeleted
   property color bgColor: "white"
   property real fieldHeight: InputStyle.rowHeight
 
@@ -303,6 +306,8 @@ Page {
             color: root.bgColor
           }
 
+          onClicked: accountDeleteDialog.open()
+
           contentItem: Text {
             text: deleteAccountButton.text
             font: deleteAccountButton.font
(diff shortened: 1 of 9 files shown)


app/qml/AccountPage.qml
```diff
@@ -259,35 +262,149 @@ Page {
...
-          color: InputStyle.highlightColor
-          horizontalAlignment: Text.AlignHCenter
-          verticalAlignment: Text.AlignVCenter
-          elide: Text.ElideRight
+    Dialog {
+      id: accountDeleteDialog
+      visible: false
+      title: qsTr( "Delete account?" )
+
+      ColumnLayout {
```
tomasMizera, line 329: This should most likely be binded to `contentItem:`

```suggestion
      contentItem: ColumnLayout {
```

tomasMizera (changes requested):
Thanks Alex 🚀 , here are my notes:

1. From my point of view, we should first try to delete an account on Mergin and only after this step succeeds continue with the removal of local projects and local auth. The thing is that Mergin server that a user is using might not even support `DELETE` which would result in simple removal of all downloaded projects.
2. When user hits `Yes` in account deletion dialog, there should be a _busy indicator_ running showing that we are doing something on background. It should keep running while wait for server response, delete all local projects and remove auth.
3. I know that styling of QML dialog might be depressing 😬, especially with buttons. However, could we make it look more like the one below? I suggest to keep the standard buttons and not worry about styling them, so only style the background and text of the popup (e.g. give it some margins from screen edges, center popup on a screen, blur/dim everything around the popup - this should be possible out of the box with Dialog coming from `qtquick controls2`)



 4. Lastly, I know we had a discussion about which library to use for dialogs, it seems that QT recommends to use QtQuick controls2 variant. Mind the comment here: https://doc.qt.io/qt-5/qml-qtquick-dialogs-dialog.html#details 
image

alexbruy:
@tomasMizera hope you are happy now

tomasMizera:
@alexbruy regarding failing tests, it might be that you need to wait LONG_REPLY instead:

```
FAIL!  : TestMerginApi::testAccountDelete() 'spy.wait( TestUtils::SHORT_REPLY )' returned FALSE. ()
   Loc: [../input/app/test/testmerginapi.cpp(2491)]
```

jozef-budac:
I am not sure it we fulfill the requirement with this change. After deletion of the user, the account is not deleted, it is just deactivated. According apple quidelines:
 "It’s insufficient to only provide the ability to temporarily disable or deactivate an account. People should be able to delete the account along with their personal data."

Right now, when I delete account in input, I am not able to register again, as the username exists and when I try to log in with deleted account in input od web app, it says, **Account is not activated**

PeterPetrik:
Lets see if they accept it or not,. we can change it after Apple review if needed. I think it is safer to wait 5 days as it is now, so in case of the unwanted removal they can still reactivate the account via support

PR #2102, opened 2022-05-13
trigger layer repaint after successful commit (fix #2100)

### LINKED ISSUE #2100: Dot for new feature is vissible only after moving of the map
In 1.5.1 RC, when new feature is recorded, dot appears only when map moves (rerenders). In current public version, it is present on the map.

https://user-images.githubusercontent.com/83582181/168240981-670d242f-91ae-482d-90cb-93eb0006f80f.mp4

alexbruy:
Request layer repaint request on successful commit, so any changes made in it will be immediately visible.

Fixes #2100.

### DIFF
--- a/app/attributes/attributecontroller.cpp
+++ b/app/attributes/attributecontroller.cpp
@@ -912,6 +912,7 @@ bool AttributeController::commit()
   }
   else
   {
+    mFeatureLayerPair.layer()->triggerRepaint();
     return true;
   }
 }


tomasMizera:
How come this has worked before and does not after recent changes?

alexbruy:
Seems this caused by the upstream change in the QgsQuick library in https://github.com/qgis/QGIS/commit/6143a52f799abd02407a876dc6590f1fb35cbcf0. If I revert back to this code
```
mLayerConnections << connect( layer, &QgsMapLayer::repaintRequested, this, &QgsQuickMapCanvasMap::refresh );
```
it works again.

Issue #1611, opened 2021-08-26
Custom folder for storing photos doesn't work for empty folders


jozef-budac:
When I have project with custom folder for storing photos in QGIS, and this folder is empty, it is not synchronized to mergin, and also to INPUT. 
When creating a photo feature in INPUT, it shows and error "Cannot copy a photo to working directory"

We should synchronize the empty directory, or INPUT should be able to create a directory in case it is needed.
For testing you can use project input-testing/tc10_custom_photo_directory

Here is configuration of the photo field in the feature

![Screenshot at 2021-08-26 09-51-37](https://user-images.githubusercontent.com/83582181/130923605-1f76e67c-c4fe-426c-823d-ee2fb60c2a15.png)

acgtdev:
This doesn't work out for me. Have been trying to create a empty directory 'photos'. It will be created at local but not in mergin.

alexbruy:
In this particular case the problem is in incorrect configuration of the attachment widget. Expression should be set using data-defined override, not entered directly in the field.

PeterPetrik:
is mergin plugin in QGIS reporting the warning?

jozef-budac:
@tomasMizera With @iiLubos patch, this issue is solved

PR #2107, opened 2022-05-25
create directories used in the attachment widget configuration if they are not exists.

alexbruy:
This is needed as geodiff works only with files and empty directories are not synced. As a result if attachment widget is configured to use subfolder and this path does not exists capturing photos will fail.

Also show a warning dialog if copying image to the destination directory fails.

Refs #1611 and https://github.com/lutraconsulting/qgis-mergin-plugin/issues/278

### DIFF
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -78,11 +78,27 @@ bool InputUtils::copyFile( const QString &srcPath, const QString &dstPath )
     modSrcPath = modSrcPath.replace( "file://", "" );
   }
 
+  QFileInfo fi( dstPath );
+  QDir dir = fi.absoluteDir();
+  if ( !dir.exists() )
+  {
+    if ( !dir.mkpath( dir.path() ) )
+    {
+      return false;
+    }
+  }
+
   // https://github.com/lutraconsulting/input/issues/418
   // does not work for iOS files with format
   // file:assets-library://asset/asset.PNG%3Fid=A53AB989-6354-433A-9CB9-958179B7C14D&ext=PNG
 
   return QFile::copy( modSrcPath, dstPath );
+}
+
+bool InputUtils::createDirectory( const QString &path )
+{
+  QDir dir;
+  return dir.mkpath( path );
 }
 
 QString InputUtils::getFileName( const QString &filePath )
@@ -748,8 +764,6 @@ QString InputUtils::resolveTargetDir( const QString &homePath, const QVariantMap
       return defaultRoot;
     }
   }
-
-
 }
 
 QString InputUtils::resolvePrefixForRelativePath( int relativeStorageMode, const QString &homePath, const QString &targetDir )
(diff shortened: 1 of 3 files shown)


app/inpututils.cpp
```diff
@@ -78,13 +78,29 @@ bool InputUtils::copyFile( const QString &srcPath, const QString &dstPath )
...
 
+  QFileInfo fi( dstPath );
+  QDir dir = fi.absoluteDir();
+  if ( !dir.exists() )
+  {
+    if ( !dir.mkpath( dir.path() ) )
+    {
+      return false;
+    }
+  }
```
PeterPetrik, line 89: use `InputUtils::createDirectory()` ?

app/inpututils.cpp
```diff
@@ -78,13 +78,29 @@ bool InputUtils::copyFile( const QString &srcPath, const QString &dstPath )
...
   // does not work for iOS files with format
   // file:assets-library://asset/asset.PNG%3Fid=A53AB989-6354-433A-9CB9-958179B7C14D&ext=PNG
 
   return QFile::copy( modSrcPath, dstPath );
 }
 
+bool InputUtils::createDirectory( const QString &path )
+{
+  QDir dir;
+  return dir.mkpath( path );
```
PeterPetrik, line 101: add `if ( !dir.exists() )`

app/qml/ExternalResourceBundle.qml
```diff
@@ -29,6 +29,12 @@ Item {
          */
         property var capturePhoto: function capturePhoto(itemWidget) {
           externalResourceHandler.itemWidget = itemWidget
+          if ( !__inputUtils.createDirectory( itemWidget.targetDir ) )
+          {
+            errorDialog.errorText = qsTr( "Could not create directory %1. Please ensure that your project configured correctly" ).arg( itemWidget.targetDir )
```
PeterPetrik, line 34: I would remove "Please ensure that your project configured correctly". Also make sure it is logged in console log too!

app/qml/ExternalResourceBundle.qml
```diff
@@ -117,7 +123,8 @@ Item {
             var success = __inputUtils.copyFile(imagePath, absolutePath)
             if (!success)
             {
-                console.log("error: Unable to copy file " + imagePath + " to the project directory")
+              errorDialog.errorText = qsTr( "Failed to copy image file to %1. Please ensure that your project configured correctly" ).arg( absolutePath )
```
PeterPetrik, line 126: I would remove "Please ensure that your project configured correctly". Also make sure it is logged in console log too!

app/inpututils.cpp
alexbruy: Not sure this is really needed, as `mkpath` will return `true` if path already exists.

PeterPetrik: ok

PeterPetrik (approved):

PR #2110, opened 2022-06-01
update public server URL

alexbruy:


tomasMizera:
Mind that we also need to update `merginsecrets.cpp` file. It is probably why other builds are failing

PR #2126, opened 2022-06-21
Snapping support

alexbruy:
Add snapping support when digitizing/editing geometries. There is no GUI part for this change, Input uses snapping mode defined at the project level.

Snapping mode used by the project is configured in Mergin plugin, see https://github.com/MerginMaps/qgis-mergin-plugin/pull/430 for more information.


app/snaputils.cpp
```diff
@@ -0,0 +1,203 @@
...
+{
+  mUseSnapping = useSnapping;
+}
+
+void SnapUtils::setupSnapping()
+{
+  int mode = mQgsProject->readNumEntry( QStringLiteral( "Mergin" ), QStringLiteral( "Snapping" ), 0 );
+  switch( mode )
+  {
+    case 0: break;
```
tomasMizera, line 182: Shouldn't this set `mUseSnapping` to false?

app/qml/map/RecordCrosshair.qml
```diff
@@ -8,23 +8,233 @@
...
+
+      height: root.innerDotSize * 2
+      width: height
+      sourceSize.width: width
+      sourceSize.height: height
+
+      source: InputStyle.crosshairCenterCircleIcon
+    }
+
+    Connections {
```
tomasMizera, line 235: This connection to `_digitizingController` works just because the `RecordCrosshair` is instantiated within `MapWrapper`, however, if we would move the `RecordCrosshair` somewhere else, this is logic is going to break.

I suggest to create a new property in `RecordCrosshair` something like `shouldUseSnapping` and inside `MapWrapper` bind it to:

```qml
RecordCrosshair {
  id: recordCrosshair
  ...
  shouldUseSnapping: _digitizingController.manualRecording && !_digitizingController.useGpsPoint
}
```
instead.

That way if we would move the `RecordCrosshair` component somewhere else we would just need to fix the binding 🙂

app/snaputils.cpp
```diff
@@ -0,0 +1,203 @@
...
+    return;
+
+  if ( mMapSettings )
+  {
+    disconnect( mMapSettings, &QgsQuickMapSettings::extentChanged, this, &SnapUtils::setup );
+    disconnect( mMapSettings, &QgsQuickMapSettings::destinationCrsChanged, this, &SnapUtils::setup );
+    disconnect( mMapSettings, &QgsQuickMapSettings::mapUnitsPerPixelChanged, this, &SnapUtils::setup );
+    disconnect( mMapSettings, &QgsQuickMapSettings::visibleExtentChanged, this, &SnapUtils::setup );
+    disconnect( mMapSettings, &QgsQuickMapSettings::outputSizeChanged, this, &SnapUtils::setup );
+    disconnect( mMapSettings, &QgsQuickMapSettings::outputDpiChanged, this, &SnapUtils::setup );
```
tomasMizera, line 50: We can simplify the code thanks to the handy [QT disconnect syntax ](https://doc.qt.io/qt-5/qobject.html#disconnect): 

```suggestion
    disconnect( mMapSettings, nullptr, this, nullptr );
```

It should disconnect all connections between `mMapSettings` and `this` receiver.

app/qml/map/Highlight.qml
```diff
@@ -72,6 +74,16 @@ Item {
   property real refTransformOffsetX: 0
   property real refTransformOffsetY: 0
 
+  SnapUtils {
```
tomasMizera, line 77: We might go with a different approach here. The thing is we want to reuse `Highlight` in different places and not all instances need to have the ability to snap.

I was thinking about having a `center` property here in `Highlight.qml` that would be binded by default to `Qt.point( highlight.width / 2, highlight.height / 2 )` and inside `MapWrapper` we can override this for `_highlightIdentified` with the logic you added to `crosshairPoint()` function.

That way we (a) do not need to have 2 `SnapUtils` objects, (b) can remove the `crosshair()` function altogether - just replace its one occurrence with the new property - and (c) `Highlight` can be reused 🙂

app/snaputils.cpp
alexbruy: Probably yes. And we should also handle this case (snapping in QGIS enabled, but disabled in Input) in the plugin and notify user. I will add another validation for it.

app/qml/map/Highlight.qml
alexbruy: Hmm, if I understand correctly, in that case we should move snapUtils to map wrapper and pass it as a property to both crosshair and highlight objects, right?

tomasMizera (changes requested):
Great, we are now even closer to merge this 🎉 . I left a few minor comments, but there are two things that blocks us from being able to start the testing:
 1. `Highlight`, especially guide line, is not working properly. When you start to digitize you will see a lot of `qrc:/map/Highlight.qml:162: ReferenceError: centerPoint is not defined` errors. It is actually just because the local variable in function `constructHighlights()` is named the same as the property `centerPoint`
 2. This one is worse, I tested it a little bit, tried to create some features in Input by snapping and synced to QGIS. Then I compared if the snapped points have exactly the same coordinates as points "that they snapped to" and found out that there are slight differences.

On left & right side you can see some examples of coordinates of point captured (by snapping) in MM Input and the point it snapped to. You can see that most of the time, coordinate values do not match, they differ a little. I suspect that it is due to the use of `QPointF` as `snappedPosition`, it internally uses floats - we probably face the float precision issue. Coordinates I compared have 14 decimal places, e.g. 48,10457350410248.
We might want to try to use `QgsPointXY` instead, it uses doubles.

image

image

image

It would be great to have an auto test to cover this scenario.

PR #2128, opened 2022-06-22
Add maptools - Refactoring of digitizing controller & splitting map tool

tomasMizera:
This PR completely rewrites `DigitizingController`, there is a separate class for it now called `RecordingMapTool`. 
I have not yet removed the `DigitizingController`, it is still used inside QML.

Our current _"digitizing"_ process is a bit of spaghetti 🍝 ,  this is a first improvement out of few (next one `Highlight.qml` and `MapWrapper.qml`)

There are two TODOs in code I need to figure out, maybe @wonder-sk could help me there? 🙏🏻 (: 

The basic idea how the digitizing is going to work:
 - `RecordingMapTools` will be instantiated from QML only for the time of recording a new feature (previously 🆙 all the time)
 - `RecordingMapTools` holds recorded points, can add and remove them; on each change generates new `QgsGeometry ~ recordedGeometry` that will be used by `Highlight.qml` to show the feature on a map
 - At the end of recording (when user hits _Done_ ), QML will take the geometry and pass it together with _layer_ and `VariablesManager` to `InputUtils::createFetureLayerPair`
 - `RecordingMapTools` is then destroyed 💣 

Moreover, there is `AbstractMapTool` that will be reused as a base class for other map tools like the tool for splitting, editing or measuring.

Update
----

This PR now also includes `SplittingMapTool`, it works similar way to `RecordingMapTool`, but in a simplified terms. It only captures geometry (add/remove point) and in the end calls `QgsVectorLayer::splitFeatures` with the captured geometry in order to split specified feature.

Behavior of copying the data is the same as in QGIS.

### DIFF
--- a/app/maptools/recordingmaptool.cpp
+++ b/app/maptools/recordingmaptool.cpp
@@ -1,0 +1,253 @@
+/***************************************************************************
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+
+#include "recordingmaptool.h"
+
+#include <QDebug>
+
+#include "qgsvectorlayer.h"
+#include "qgspolygon.h"
+#include "qgsvectorlayerutils.h"
+
+#include "position/positionkit.h"
+#include "variablesmanager.h"
+#include "inpututils.h"
+
+RecordingMapTool::RecordingMapTool( QObject *parent )
+  : AbstractMapTool{parent}
+{
+}
+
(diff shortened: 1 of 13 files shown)


tomasMizera:
Merging to integration branch: `CU-2fdz4a3_map-tools-splitting-refactoring`

app/maptools/recordingmaptool.cpp
```diff
@@ -0,0 +1,253 @@
...
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+
+#include "recordingmaptool.h"
+
+#include <QDebug>
```
alexbruy, line 12: Leftover?

alexbruy (approved):

Issue #2130, opened 2022-06-23
Snapping - does not record the same coordinates


tomasMizera:
I created a feature in Input with snapping, synced to QGIS and then compared if the snapped vertex is the same as the one it snapped to. Turns out that for some (not all) vertices coordinate values differ a little.

<img width="1135" alt="image (1)" src="https://user-images.githubusercontent.com/22449698/175327207-cc4ce593-1bd8-4412-a4e7-205d4711c6d4.png">

alexbruy:
I can't reproduce this using my phone and with Linux build. Might be related to hidpi screens.

tomasMizera:
@jozef-budac

jozef-budac:
Can be related to this:
https://lutraconsulting.slack.com/files/U01T1L12335/F03MDHEN6AV/image_from_ios.mp4

jozef-budac:
I tried this scenario, in input on Android I created a polygon with 5 points.
Then I created 2nd same polygon, but with snapping tool with exact same 5 points as first one.

This are the coordinates of first polygon:

Polygon ((17.04835226784347668 48.1988806214264045, 17.0474370366437995 48.19901134511145813, 17.04755144054376359 48.19964317155354649, 17.04836861125776082 48.19974121288925062, 17.04890794392900233 48.19931636574605704, 17.04835226784347668 48.1988806214264045))

This are the coordinates of second polygon

Polygon ((17.04835226784347668 48.19888062142638319, 17.0474370366437995 48.19901134511145813, 17.04755144054376359 48.19964317155352518, 17.04836861125776082 48.19974121288925062, 17.04890794392900233 48.19931636574604994, 17.04835226784347668 48.19888062142638319))

There are small differences there

Issue #2132, opened 2022-06-24
Snapping - in maximum zoom, cursor doesn't snap


jozef-budac:
When zooming out a lot, snapping is not working. Cursor doesn't snap to the line

https://user-images.githubusercontent.com/83582181/175507407-6cda2bc3-e1de-42f8-940a-884cb5926f7f.MP4

tomasMizera:
Not a bug, we should just not allow zooming to 0mm :)

Issue #2129, opened 2022-06-23
Snapping glytch - center of crosshair not aligned to vertex


tomasMizera:
How to reproduce:
 - start recording in layer that has snapping enabled
 - go near vertex
   - Input should snap to the vertex ✔️ 
- pan map a little bit from vertex
  -  Input should still show snap on the vertex ❌ 

Snap indicator is moved with map until it snaps again to another place.

<img width="109" alt="image" src="https://user-images.githubusercontent.com/22449698/175326484-95354850-22b8-4e16-bc20-1e61813e1f16.png">

PR #2134, opened 2022-06-24
Fixes for snapping issues

alexbruy:
Better comparison of old and new snapped positions.

### DIFF
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -1453,7 +1453,7 @@ QgsGeometry InputUtils::stakeoutGeometry( const QgsPoint &mapPosition, const Fea
     return QgsGeometry();
 
   QgsPointXY targetInLayerCoordinates = target.feature().geometry().asPoint();
-  QgsPointXY t = transformPoint( target.layer()->crs(), mapSettings->destinationCrs(), mapSettings->transformContext(), targetInLayerCoordinates );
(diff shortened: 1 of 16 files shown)


app/snaputils.cpp
```diff
@@ -200,7 +199,9 @@ void SnapUtils::setupSnapping()
     }
     case 2:
     {
-      mSnappingUtils.setConfig( mQgsProject->snappingConfig() );
+      QgsSnappingConfig config = mQgsProject->snappingConfig();
+      config.setTolerance( config.tolerance() * InputUtils::calculateDpRatio() );
```
tomasMizera, line 203: We need to first check if the tolerance is set in pixels :)

app/inpututils.cpp
```diff
@@ -681,6 +681,35 @@ QgsPointXY InputUtils::transformPoint( const QgsCoordinateReferenceSystem &srcCr
...
+  {
+    return QgsPoint();
+  }
+
+  try
+  {
+    QgsCoordinateTransform ct( srcCrs, destCrs, context );
+    if ( ct.isValid() )
+    {
+      const QgsPoint pt( ct.transform( srcPoint.x(), srcPoint.y() ) );
```
tomasMizera, line 701: Don't we drop `Z` and `M` coordinates this way @alexbruy ?

alexbruy: You are right. Fixed.

tomasMizera:
Hi @alexbruy, can you please link all tickets this should close?

alexbruy:
Should fix #2130 and #2129. Probably also #2132.

PR #2135, opened 2022-06-28
Map tools & highlight refactoring

tomasMizera:
Map tools are now implemented in QML too.

What happened:
 - added `splittingMapTool` ~ ability to split geometries
 - added `recordingMapTool` ~ refactored digitizingController
 - refactored `stakeoutTool` ~ not using custom highlight anymore

The main idea is that now we instantiate components only when we need them - when user hits `Record`, we instantiate `RecordingTools.qml` and similarly for `Splitting` and `Stakeout`.


app/inpututils.cpp
```diff
@@ -1372,6 +1384,19 @@ QgsRectangle InputUtils::stakeoutPathExtent(
   return extent;
 }
 
+QgsGeometry InputUtils::stakeoutGeometry( const QgsPoint &mapPosition, const FeatureLayerPair &target, QgsQuickMapSettings *mapSettings )
+{
+  if ( !mapSettings || !target.isValid() )
+    return QgsGeometry();
+
+  QgsPointXY targetInLayerCoordinates = target.feature().geometry().asPoint();
+  QgsPointXY t = transformPoint( target.layer()->crs(), mapSettings->destinationCrs(), mapSettings->transformContext(), targetInLayerCoordinates );
```
alexbruy, line 1393: In #2134 this was renamed to `transformPointXY`

app/qml/map/SplittingTools.qml
```diff
@@ -0,0 +1,106 @@
...
+import QtQuick.Layouts 1.14
+import QtQuick.Shapes 1.14
+
+import QgsQuick 0.1
+import lc 1.0
+
+import "../"
+
+/**
+  * RecordingTools is a set of tools that are used during recording/editing of a geometry.
```
alexbruy, line 20: Shouldn't it be `SplittingTools`? The rest of the comment also from another tool.

app/qml/map/SplittingTools.qml
```diff
@@ -0,0 +1,106 @@
...
+    crosshairPosition: crosshair.recordPoint
+  }
+
+  Highlight {
+    id: guideline
+
+    height: root.map.height
+    width: root.map.width
+
+    lineColor: Qt.rgba( 0.67, 0.7, 0.74, 0.5 )
```
alexbruy, line 52: Maybe put this into `InputStyle` so we can re-use it elsewhere

alexbruy (approved):

PR #2163, opened 2022-07-18
Geometry editing support

alexbruy:
Add support for line/polygon geometry editing.


tomasMizera:
Let's use this one as an integration branch :)

PR #2164, opened 2022-07-19
Fix crashes in snaputils

tomasMizera:
Added bunch of sanity checks for snapping and cleared all its pointers when project is about to be reloaded.

I have also change loading of a project a little bit. We now first destroy `ActiveLayer` and only then load the project, previously we started with project load and only after that removed the old active layer (which was a dangling pointer already)


tomasMizera:
@PatrikMizera can you please verify this fixes the crash while snapping with schema change?

PR #2165, opened 2022-07-19
Helper functions for geometry editing

alexbruy:
Add several helper methods to RecordingMapTool needed to support geometry editing:

- `extractGeometryVertices` extracts geometry vertices in order to visualize them. Supports single- and multipart geometries, extracted vertices are returned as MultiPoint geometry.
- `extractMidSegmentVertices` extracts midpoints for every segment of the geometry. Supports single- and multipart geometries, extracted vertices are returned as MultiPoint geometry. If input geometry is Point/MultiPoint then a null geometry is returned.
- `createHandles` create "handles" at the beginning and end of line geometry (will be used to extend lines). Supports single- and multipart geometries, handles are returned as MultiLineString geometry, if input is not a LineString/MultiLineString then a null geometry is returned.


app/inpututils.cpp
```diff
@@ -33,6 +33,8 @@
 #include "qgslayertree.h"
 #include "qgsprojectviewsettings.h"
 #include "qgsvectorlayerutils.h"
+#include "qgsmultipoint.h"
+#include "qgsmultilinestring.h"
```
tomasMizera, line 37: unused import

app/maptools/recordingmaptool.cpp
```diff
@@ -275,15 +277,87 @@ void RecordingMapTool::setInitialGeometry( const QgsGeometry &newInitialGeometry
...
+      multiPoint->addGeometry( p.clone() );
+    }
+
+    // if input geometry is a line we need to add virtual nodes and the beginning
+    // and at the end of the line. They will be used to extend line
+    if ( geometry.type() == QgsWkbTypes::LineGeometry )
+    {
+      p = QgsGeometryUtils::interpolatePointOnLine( line->pointN( 0 ), line->pointN( 1 ), -0.1 );
+      multiPoint->insertGeometry( p.clone(), 0 );
+      p = QgsGeometryUtils::interpolatePointOnLine( line->pointN( line->numPoints() - 2 ), line->pointN( line->numPoints() - 1 ), 1.1 );
```
tomasMizera, line 327: if `line->numPoints()` is 1, what would `line->pointN( line->numPoints() - 2 )` do?

app/maptools/recordingmaptool.cpp
```diff
@@ -275,15 +277,87 @@ void RecordingMapTool::setInitialGeometry( const QgsGeometry &newInitialGeometry
...
+      multiPoint->addGeometry( p.clone() );
+    }
```
tomasMizera, line 329: What is the difference between this code and `RecordingMapTool::createHandles`?

app/qml/map/RecordingTools.qml
```diff
@@ -69,6 +69,7 @@ Item {
 
     positionKit: __positionKit
     layer: __activeLayer.vectorLayer
+    initialGeometry: root.initialGeometry.feature.geometry
```
tomasMizera, line 72: `root.initialGeometry` can be null

tomasMizera:
Can you please change the base to the integration branch, not master?

app/maptools/recordingmaptool.cpp
alexbruy: Here we add points at the beginning and end of the line, while `createHandles` creates lines which follow first or last segment direction

app/maptools/recordingmaptool.cpp
alexbruy: Hmm, so we can have lines with just one point? I can't digitize such lines in Input and QGIS.

alexbruy: Than it should not be relevant, no? I mean we use these only when editing existing geometries, not when digitizing them.

PR #2173, opened 2022-07-21
Look up vertex id from the clicked position

alexbruy:
Add method to retrieve vertex id from screen coordinates. Takes into account both real and "virtual" (mid points and start/end points) vertices, supports search threshold.

If there no veritices found within search threshol returns invalid vertex id.

### DIFF
--- a/app/maptools/recordingmaptool.cpp
+++ b/app/maptools/recordingmaptool.cpp
@@ -285,8 +285,8 @@ void RecordingMapTool::setInitialGeometry( const QgsGeometry &newInitialGeometry
     mPoints.push_back( QgsPoint( *pointIt ) );
   }
 
-  createNodesAndHandles();
   rebuildGeometry();
+  createNodesAndHandles();
 
   emit initialGeometryChanged( mInitialGeometry );
 }
--- a/app/maptools/recordingmaptool.cpp
+++ b/app/maptools/recordingmaptool.cpp
@@ -406,3 +406,16 @@ void RecordingMapTool::setHandles( const QgsGeometry &newHandles )
   mHandles = newHandles;
   emit handlesChanged( mHandles );
 }
+
+const QString &RecordingMapTool::state() const
+{
+  return mState;
+}
+
+void RecordingMapTool::setState( const QString &newState )
+{
+  if ( mState == newState )
+    return;
+  mState = newState;
+  emit stateChanged( mState );
+}
--- a/app/maptools/recordingmaptool.cpp
+++ b/app/maptools/recordingmaptool.cpp
@@ -285,87 +285,82 @@ void RecordingMapTool::setInitialGeometry( const QgsGeometry &newInitialGeometry
     mPoints.push_back( QgsPoint( *pointIt ) );
   }
 
-  setExistingVertices( extractGeometryVertices( mInitialGeometry ) );
-  setMidPoints( extractMidSegmentVertices( mInitialGeometry ) );
-  setHandles( createHandles( mInitialGeometry ) );
+  createNodesAndHandles();
 
   emit initialGeometryChanged( mInitialGeometry );
 }
 
-
-QgsGeometry RecordingMapTool::extractGeometryVertices( const QgsGeometry &geometry )
+void RecordingMapTool::createNodesAndHandles()
 {
-  QgsMultiPoint *multiPoint = new QgsMultiPoint();
-  QgsGeometry outputGeom( multiPoint );
-  for ( auto pointIt = geometry.vertices_begin(); pointIt != geometry.vertices_end(); ++pointIt )
-    multiPoint->addGeometry( ( *pointIt ).clone() );
-  return outputGeom;
-}
+  mVertexIds.clear();
 
-QgsGeometry RecordingMapTool::extractMidSegmentVertices( const QgsGeometry &geometry )
-{
-  if ( geometry.type() == QgsWkbTypes::PointGeometry )
+  QgsPoint vertex;
+  QgsVertexId vertexId;
+  const QgsAbstractGeometry *geom = mInitialGeometry.constGet();
+
+  QgsMultiPoint *existingVertices = new QgsMultiPoint();
+  mExistingVertices.set( existingVertices );
+
+  QgsMultiPoint *midPoints = new QgsMultiPoint();
+  mMidPoints.set( midPoints );
+
+  QgsMultiLineString *handles = new QgsMultiLineString();
+  mHandles.set( handles );
+
+  int currentPart = -1;
+  int currentRing = -1;
+
+  while ( geom->nextVertex( vertexId, vertex ) )
   {
-    return QgsGeometry();
-  }
+    existingVertices->addGeometry( vertex.clone() );
+    mVertexIds.push_back( qMakePair( vertexId, vertex ) );
 
-  QgsMultiPoint *multiPoint = new QgsMultiPoint();
-  QgsGeometry outputGeom( multiPoint );
-  QgsPoint p;
-
-  const QVector< QgsLineString * > lines = QgsGeometryUtils::extractLineStrings( geometry.constGet() );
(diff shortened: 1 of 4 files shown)


app/maptools/recordingmaptool.cpp
```diff
@@ -406,3 +401,94 @@ void RecordingMapTool::setHandles( const QgsGeometry &newHandles )
...
+void RecordingMapTool::lookForVertex( const QPointF &clickedPoint, double searchRadius )
+{
+  double minDistance = std::numeric_limits<double>::max();
+  double currentDistance = 0;
+  QgsPoint point;
+  QgsVertexId vertexId;
+
+  QgsPoint pnt = mapSettings()->screenToCoordinate( clickedPoint );
+
+  if ( mInitialGeometry.isEmpty() )
```
tomasMizera, line 442: Don't we want to look into mRecordedGeometry instead?

app/maptools/recordingmaptool.cpp
```diff
@@ -406,3 +401,94 @@ void RecordingMapTool::setHandles( const QgsGeometry &newHandles )
...
+  {
+    setClickedVertexId( vertexId );
+    return;
+  }
+
+  for ( QPair<QgsVertexId, QgsPoint> pair : mVertexIds )
+  {
+    currentDistance = QgsGeometryUtils::sqrDistance2D( pnt, pair.second );
+    if ( currentDistance <= minDistance && currentDistance <= searchRadius )
```
tomasMizera, line 451: Why do we need the `minDistance`?

app/maptools/recordingmaptool.cpp
```diff
@@ -406,3 +401,94 @@ void RecordingMapTool::setHandles( const QgsGeometry &newHandles )
...
+  if ( geom->deleteVertex( id ) )
+    geometry.set( geom );
+
+  setRecordedGeometry( geometry );
+}
+
+void RecordingMapTool::insertVertex( QgsVertexId id, const QgsPoint &point )
+{
+  QgsGeometry geometry;
+  QgsAbstractGeometry *geom = mInitialGeometry.get()->clone();
```
tomasMizera, line 477: Again, we probably want to take `mRecordedGeometry` instead of `mInitialGeometry`

app/maptools/recordingmaptool.cpp
```diff
@@ -406,3 +401,94 @@ void RecordingMapTool::setHandles( const QgsGeometry &newHandles )
...
+  if ( geom->insertVertex( id, point ) )
+    geometry.set( geom );
+
+  setRecordedGeometry( geometry );
+}
+
+void RecordingMapTool::updateVertex( QgsVertexId id, const QgsPoint &point )
+{
+  QgsGeometry geometry;
+  QgsAbstractGeometry *geom = mInitialGeometry.get()->clone();
```
tomasMizera, line 488: Again, we probably want to take `mRecordedGeometry` instead of `mInitialGeometry`

app/maptools/recordingmaptool.cpp
```diff
@@ -406,3 +401,94 @@ void RecordingMapTool::setHandles( const QgsGeometry &newHandles )
...
+    }
+  }
+
+  setClickedVertexId( vertexId );
+}
+
+void RecordingMapTool::removeVertex( QgsVertexId id )
+{
+  QgsGeometry geometry;
+  QgsAbstractGeometry *geom = mInitialGeometry.get()->clone();
```
tomasMizera, line 466: `mRecordedGeometry` again here

app/maptools/recordingmaptool.cpp
```diff
@@ -285,87 +285,82 @@ void RecordingMapTool::setInitialGeometry( const QgsGeometry &newInitialGeometry
...
 
-QgsGeometry RecordingMapTool::extractMidSegmentVertices( const QgsGeometry &geometry )
-{
-  if ( geometry.type() == QgsWkbTypes::PointGeometry )
-  {
-    return QgsGeometry();
-  }
+  QgsPoint vertex;
+  QgsVertexId vertexId;
+  const QgsAbstractGeometry *geom = mInitialGeometry.constGet();
```
tomasMizera, line 299:
```suggestion
  const QgsAbstractGeometry *geom = mRecordedGeometry.constGet();
```

app/maptools/recordingmaptool.cpp
```diff
@@ -285,87 +285,82 @@ void RecordingMapTool::setInitialGeometry( const QgsGeometry &newInitialGeometry
...
-  QgsGeometry outputGeom( multiLine );
-  QgsPoint p;
-  QgsLineString handle;
-
-  const QVector< QgsLineString * > lines = QgsGeometryUtils::extractLineStrings( geometry.constGet() );
-  for ( QgsLineString *line : lines )
-  {
-    if ( line->numPoints() >= 2 )
+    // for lines and polygons create midpoints
+    if ( mInitialGeometry.type() != QgsWkbTypes::PointGeometry && vertexId.vertex < geom->vertexCount( vertexId.part, vertexId.ring ) - 1 )
```
tomasMizera, line 319:
```suggestion
    if ( mRecordedGeometry.type() != QgsWkbTypes::PointGeometry && vertexId.vertex < geom->vertexCount( vertexId.part, vertexId.ring ) - 1 )
```

app/maptools/recordingmaptool.cpp
```diff
@@ -285,87 +285,82 @@ void RecordingMapTool::setInitialGeometry( const QgsGeometry &newInitialGeometry
...
+      QgsPoint midPoint = QgsGeometryUtils::midpoint( geom->vertexAt( vertexId ), geom->vertexAt( id ) );
+      midPoints->addGeometry( midPoint.clone() );
+      mVertexIds.push_back( qMakePair( id, midPoint ) );
+    }
 
-      p = QgsGeometryUtils::interpolatePointOnLine( line->pointN( line->numPoints() - 2 ), line->pointN( line->numPoints() - 1 ), 1.1 );
-      handle = QgsLineString( line->pointN( line->numPoints() - 1 ), p );
-      multiLine->addGeometry( handle.clone() );
+    // for lines also create start/end points and handles
+    if ( mInitialGeometry.type() == QgsWkbTypes::LineGeometry && ( vertexId.part != currentPart && vertexId.ring != currentRing ) )
```
tomasMizera, line 328:
```suggestion
    if ( mRecordedGeometry.type() == QgsWkbTypes::LineGeometry && ( vertexId.part != currentPart && vertexId.ring != currentRing ) )
```

app/maptools/recordingmaptool.cpp
```diff
@@ -285,87 +285,82 @@ void RecordingMapTool::setInitialGeometry( const QgsGeometry &newInitialGeometry
     mPoints.push_back( QgsPoint( *pointIt ) );
   }
 
-  setExistingVertices( extractGeometryVertices( mInitialGeometry ) );
-  setMidPoints( extractMidSegmentVertices( mInitialGeometry ) );
-  setHandles( createHandles( mInitialGeometry ) );
+  createNodesAndHandles();
```
tomasMizera, line 288: We miss a call to build recorded geometry from the initial one here.

```suggestion
  createNodesAndHandles();
  rebuildGeometry();
```

tomasMizera (commented):
Really good job @alexbruy !
I like the idea with interpolation. 

There is basically only one comment that I left, we should use `mRecordedGeometry` instead of the initial one. `mInitialGeometry` is here just to be set once in the beginning of editing - we will copy it to `mRecordedGeometry` and use the recorded geometry from that point on.

app/maptools/recordingmaptool.cpp
alexbruy: We discussed it a few days ago. If user taps somewhere too far from the any vertex we still will be able to find closest vertex based on the distance to the clicked position. So we want to search only within some radius around each vertex.

tomasMizera (approved):

PR #2176, opened 2022-07-22
QDirIterator on Windows - Fix Windows sync

tomasMizera:
We had a broken sync on Windows because `QDirIterator` on Unix systems ignores the hidden folders, but lists them on  Windows (as the folder can start with a dot). It resulted in `.mergin` folder files being scheduled for synchronisation. They were assigned a chunk id, but Mergin correctly ignored these files. This led to 404s due to chunk IDs mismatch.

I fixed this by ignoring all files in `.mergin` folder in `MerginApi::isInIgnore`. It should not change anything on UNIX systems as this folder was never listed (always ignored).

Note, our Windows release is still just experimental, so it includes other bugs.


jozef-budac:
It is verified that synchronization now works with this changes

PR #2182, opened 2022-07-28
Editing fixes

alexbruy:
More fixes and updates to the geometry editing:

- better threshold for selecting active vertex
- hide marker for active vertex
- hide handle and start/end marker if start/end point of the line is selected
- fix toolbar buttons availability depending on the current state
- show crosshair when new feature is recorded
- fix broken test
- remove unused code

### DIFF
--- a/app/maptools/recordingmaptool.cpp
+++ b/app/maptools/recordingmaptool.cpp
@@ -25,6 +25,7 @@ RecordingMapTool::RecordingMapTool( QObject *parent )
 {
   connect( this, &RecordingMapTool::initialGeometryChanged, this, &RecordingMapTool::prepareEditing );
   connect( this, &RecordingMapTool::recordedGeometryChanged, this, &RecordingMapTool::createNodesAndHandles );
+  connect( this, &RecordingMapTool::activeVertexChanged, this, &RecordingMapTool::createNodesAndHandles );
 }
 
 RecordingMapTool::~RecordingMapTool() = default;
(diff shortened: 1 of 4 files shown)


app/maptools/recordingmaptool.cpp
```diff
@@ -412,9 +420,43 @@ void RecordingMapTool::cancelGrab()
...
+      } );
+    }
+    emit existingVerticesChanged( mExistingVertices );
+  }
+}
+
+double RecordingMapTool::pixelsToMapUnits( double numPixels )
+{
+  QgsRenderContext context = QgsRenderContext::fromMapSettings( mapSettings()->mapSettings() );
+  return numPixels * context.scaleFactor() * context.mapToPixel().mapUnitsPerPixel();
```
tomasMizera, line 457: I believe we will need to incorporate DPI here as well

tomasMizera (approved):
Good stuff! 🚀

app/maptools/recordingmaptool.cpp
tomasMizera: Can you please multiply the number of pixels by `InputUtils::calculateScreenDpr`?

app/maptools/recordingmaptool.cpp
```diff
@@ -322,9 +326,10 @@ void RecordingMapTool::lookForVertex( const QPointF &clickedPoint, double search
 {
   double minDistance = std::numeric_limits<double>::max();
   double currentDistance = 0;
+  QgsRenderContext context = QgsRenderContext::fromMapSettings( mapSettings()->mapSettings() );
```
tomasMizera, line 329: This `context` is not used anywhere, right?

app/maptools/recordingmaptool.cpp
```diff
@@ -412,9 +420,83 @@ void RecordingMapTool::cancelGrab()
...
+
+    if ( part != -1 )
+    {
+      mmHiddenHandle.set( geom->lineStringN( part ).clone() );
+      mHandles.deletePart( part );
+    }
+  }
+
+  emit handlesChanged( mHandles );
+}
```
tomasMizera, line 492: I have mixed feelings about these functions. 
Wouldn't it be enough to just call the `createNodesAndHandles` function when a vertex is grabbed / released and inside the function just filter the `activeVertex` based on id? Do I miss something? 

Also `toggleSelectedVertexVisibility` misses the `emit activeVertexChanged()` signal

tomasMizera (approved):

Issue #1417, opened 2021-05-21
Input defaults to another language


saberraz:
...if there is one in the system. In the example below, Input defaults to Japanese despite the fact that the first language setting is English. The work-around it to remove the other language(s).
![image](https://user-images.githubusercontent.com/2649047/119096962-1d5dea00-ba0c-11eb-9ce5-a6d4f24a10c3.png)

saberraz:
I think the issue is the type of English, in this case United States. The app should automatically pick any English instead of defaulting to the 2nd choice.

saberraz:
The work-around is to remove the 2nd language.

sklencar:
Strings from cpp classes using `tr` function seem to be using correct language (locale).
Translation function in QML `qsTr` is not working properly though. Essentially, when a device has several languages, it takes first and tries to load related translation file. If it fails, takes the next language from the list. 

Note that A) locale (and related translation file) is named according country code and sometimes with region postfix (e.g. `en-GB`); B) the translation fallback looks at the whole (country + region) code.  

As an example, lets have language setup from screenshot below. 
Even though Input has some Portuguese, English or Croatian translations, those will be ignored (not matching region code) and eventually Japanese is applied.

Generating and including translation files with region code  **resolves** the issue. However, its not ideal.
Would be better to change fallback in a way that if there is no marching `country+region.ts` file, it tries to load country.ts file first and only after this would fail, it will take the next language from the list?

![Screenshot_20210617-170437](https://user-images.githubusercontent.com/6735606/122434620-8f880700-cf97-11eb-8592-d1d87a752f4f.png)

sklencar:
Probably better solution is to upgrade QT - it seems that they have changed how languages are ordered.

From QT 5.15 they have changed behaviour of the fallback that it tries to truncate country code before trying to load the next language from the list.
It seems that such fix has been done here:
[https://github.com/qt/qtbase/commit/8df2bf04917e742470bbd1ff69e003811c5975ff](https://github.com/qt/qtbase/commit/8df2bf04917e742470bbd1ff69e003811c5975ff) 

Here are [docs](https://doc.qt.io/qt-5.14/qtranslator.html#load-1) describing behaviour of the `QTranslator::load` function (used to install translation in the app) in QT 5.14 which we use.
Either from commit reference above or [QT 5.15 docs](https://doc.qt.io/qt-5/qtranslator.html#load-1) you can see difference in order to load translation files.

However, I haven't tried it with bumped version of QT.

[...]

Issue #743, opened 2020-06-05
Allow manual override of language preferences 


saberraz:
Add an option to override the system language with one of the supported language in the app.

PeterPetrik:
related to #1417

tomasMizera:
This is working now on iOS and should work in the new Android versions too

PeterPetrik:
iOS 13+ and android 14+ supports it natively in system settings , closing

PR #2186, opened 2022-07-28
Fix language selection for several languages

tomasMizera:
This issue is well described here: https://github.com/MerginMaps/input/issues/1417#issuecomment-863382773 

QT tries to exactly match systems' locale with a translation file we provide. This is a problem for codes like `en_US`, `fr_CA`, where we only provide file with name `en` or `fr`. These translation files would never be picked, despite being translated. To fight this issue QT has a following system: https://doc.qt.io/archives/qt-5.14/qtranslator.html#load-1 (TL;DR - if it can not exactly match any system language, it tries to cut out the country code and match only language codes). 

---

Imagine situation like this:
  - Your system languages
    1. `en_US`
    2. `kr_KR`

 - Input provides these translation files:
   - `en`
   - `kr_KR`

 - QT tries to match it in two rounds:
   1. exact match:
       - `en_US` exactly compared with `en` (FAIL)
       - `kr_KR` exactly compared with `kr_KR` (SUCCESS, STOP)

`kr_KR` would be thus picked as app language, despite we also provided translation for english.

---

This issue is _fixed_ by removing country codes from locales for some languages (english, german, french, spanish,..). The PR, therefore, fixes these languages:
 - English
 - French
 - Spanish
 - German
 - Slovak
 - Polish
 - Croatian
 - Turkish
 - Italian

Upgrade to new QT version would fix this properly; we can remove the added code then.

Fixes (partially, for some languages) #1417 
Related #743

CU-2a6zn2p

@PeterPetrik do we also want to have the language selection page?

### DIFF
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -500,6 +500,48 @@ void InputUtils::quitApp()
   QCoreApplication::quit();
 }
 
+QLocale InputUtils::fixLocaleCountry( QLocale applocale )
+{
+  QLocale out = applocale;
+
+  if ( applocale.language() == QLocale::English )
+  {
+    out = QLocale( QLocale::English, QLocale::AnyCountry );
+  }
+  else if ( applocale.language() == QLocale::German || applocale.language() == QLocale::LowGerman || applocale.language() == QLocale::SwissGerman )
+  {
+    out = QLocale( QLocale::German, QLocale::AnyCountry );
+  }
+  else if ( applocale.language() == QLocale::French )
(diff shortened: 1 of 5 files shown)


alexbruy (approved):

PeterPetrik:
@jozef-budac I think this will need an update how we test language support.

PeterPetrik (approved):
🎉

jozef-budac:
I can add testcase for this fix to our regression runs

Issue #1060, opened 2020-12-10
Crash in thread::~thread() on Huawei


wonder-sk:
Seems to happen when user tries to exit with the android back button. Only Huawei seems affected.

Not specific to Input - other apps seem to have this problem as well:
https://bugreports.qt.io/browse/QTBUG-82617
https://bugreports.qt.io/browse/QTBUG-85449

PeterPetrik:
this is probably critical discussion: https://codereview.qt-project.org/c/qt/qtbase/+/318099

tomasMizera:
https://developernote.com/2022/03/crash-at-std-thread-and-std-mutex-destructors-on-android/#comment-694101

tomasMizera:
Can't see it in logs anymore :)

PR #2187, opened 2022-07-29
Reduce Huawei crash rate on back button

tomasMizera:
This is an effort to reduce the number of crashes happening on Huawei devices when exiting the app with back button.
There are two fixes:
 - First: we do not close the app immediately after the back button is tapped; instead, we show toast saying "hit the back button again to close the app". This way we reduce the number of accidental hits of the back button.
 - Second: instead of letting QT quit the app, we now call Java to exit the app. This is due to some blogs/stackoverflow's that tried to find a workaround to this issue. See: https://developernote.com/2022/03/crash-at-std-thread-and-std-mutex-destructors-on-android/#comment-694101 and https://stackoverflow.com/questions/61321845/qt-app-crashes-at-the-destructor-of-stdthread-on-android-10-devices

So far the second fix is applied only to Huawei devices. If it will behave well, we can apply it to all manufacturers (Recently there are crashes on other devices too).

QT has not yet provided a fix to this issue (see https://bugreports.qt.io/browse/QTBUG-82617)

Fixes ( partially ) #1060

### DIFF
--- a/app/android/src/uk/co/lutraconsulting/InputActivity.java
+++ b/app/android/src/uk/co/lutraconsulting/InputActivity.java
@@ -2,9 +2,14 @@ package uk.co.lutraconsulting;
 import android.os.Bundle;
 import android.view.WindowManager;
 import org.qtproject.qt5.android.bindings.QtActivity;
+import android.util.Log;
+import android.os.Build;
+import java.lang.Exception;
 
 public class InputActivity extends QtActivity
 {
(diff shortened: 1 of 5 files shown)


PeterPetrik (commented):
looks good

tomasMizera:
So keen to see if this will help 🍀 🍀 🙏🏻 👍🏻

PR #2188, opened 2022-07-29
Fix vertex editing and handling of handles

alexbruy:
More work on the geometry editing:

- fixes recording and editing of points
- fix guideline when handle point is selected
- attempt to fix removal of non-existent vertex

### DIFF
--- a/app/maptools/recordingmaptool.cpp
+++ b/app/maptools/recordingmaptool.cpp
@@ -145,9 +145,20 @@ void RecordingMapTool::removePoint()
   }
   else if ( mState == MapToolState::Record )
   {
-    // TODO: select first/last existing vertex as active
-    // TODO: change state to GRAB
+    // select first/last existing vertex as active and change state to GRAB
+    if ( mNewVertexOrder = NewVertexOrder::End )
+    {
+      mActiveVertex.setVertexId( QgsVertexId( 0, 0,  mRecordedGeometry.constGet()->vertexCount() - 1 ) );
+    }
+    else if ( mNewVertexOrder = NewVertexOrder::Start )
+    {
+      mActiveVertex.setVertexId( QgsVertexId( 0, 0, 1 ) );
+    }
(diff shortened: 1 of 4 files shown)


app/maptools/recordingmaptool.cpp
```diff
@@ -135,55 +145,67 @@ void RecordingMapTool::removePoint()
   }
   else if ( mState == MapToolState::Record )
   {
-    // TODO: select first/last existing vertex as active
-    // TODO: change state to GRAB
-
-    if ( mRecordedGeometry.constGet()->vertexCount() > 0 )
+    // select first/last existing vertex as active and change state to GRAB
+    if ( mNewVertexOrder = NewVertexOrder::End )
```
tomasMizera, line 149: Assignment instead of comparison

app/maptools/recordingmaptool.cpp
```diff
@@ -135,55 +145,67 @@ void RecordingMapTool::removePoint()
...
-      if ( mRecordedGeometry.type() == QgsWkbTypes::PolygonGeometry )
-      {
-        QgsLineString *r = qgsgeometry_cast<QgsLineString *>( qgsgeometry_cast<const QgsPolygon *>( mRecordedGeometry.constGet() )->exteriorRing() );
-        if ( !r )
-        {
-          return;
-        }
+      mActiveVertex.setVertexId( QgsVertexId( 0, 0,  mRecordedGeometry.constGet()->vertexCount() - 1 ) );
+    }
+    else if ( mNewVertexOrder = NewVertexOrder::Start )
```
tomasMizera, line 153:
```suggestion
    else if ( mNewVertexOrder == NewVertexOrder::Start )
```

app/maptools/recordingmaptool.cpp
tomasMizera:
```suggestion
    if ( mNewVertexOrder == NewVertexOrder::End )
```

tomasMizera (approved):
Just the assignments, otherwise looks good!

tomasMizera (approved):

PR #2191, opened 2022-08-02
if one vertex left — release returns to the record state

alexbruy:


tomasMizera:
Well, this fix works only partially. 
- For polygons, we also want such logic when there are 2 coordinates. 
- It should also take into account part and ring.

I believe it could be tackled in a more generic way:
 - if you release point that is first/last in linestring, just continue in `MapToolState::Record` no matter how many points are in it - this is better UX
 - for polygons use such logic but with fix to the two mentioned points

PR #2192, opened 2022-08-02
Handle more edge cases

alexbruy:
Handle cases:

- releasing point if polygon contains only 2 or 1 points
- releasing first/last line vertex after removing handle

Also change "Remove" button icon.

### DIFF
--- a/app/maptools/recordingmaptool.cpp
+++ b/app/maptools/recordingmaptool.cpp
@@ -554,7 +554,6 @@ void RecordingMapTool::lookForVertex( const QPointF &clickedPoint, double search
   else
   {
     // nothing found
-
     setActiveVertex( Vertex() );
     setState( MapToolState::View );
   }
@@ -610,14 +609,14 @@ void RecordingMapTool::releaseVertex( const QgsPoint &point )
   // if it is a first or last vertex of the line we go to the recording mode
   if ( mRecordedGeometry.type() == QgsWkbTypes::LineGeometry )
   {
-    if ( mActiveVertex.vertexId().vertex == 0 )
+    if ( mActiveVertex.type() != Vertex::Existing && mActiveVertex.vertexId().vertex == 0 )
     {
       setState( MapToolState::Record );
       mActiveVertex.setCoordinates( QgsPoint() );
       emit activeVertexChanged( mActiveVertex );
       return;
     }
-    else if ( mActiveVertex.vertexId().vertex == vertexCount - 1 )
+    else if ( mActiveVertex.type() != Vertex::Existing && mActiveVertex.vertexId().vertex == vertexCount - 1 )
     {
       QgsVertexId id = mActiveVertex.vertexId();
       id.vertex += 1;
--- a/app/maptools/recordingmaptool.cpp
+++ b/app/maptools/recordingmaptool.cpp
@@ -527,7 +527,6 @@ void RecordingMapTool::lookForVertex( const QPointF &clickedPoint, double search
   if ( idx >= 0 )
   {
     // we found a point
-
     mActiveVertex = mVertices.at( idx );
 
     if ( mActiveVertex.type() == Vertex::Existing )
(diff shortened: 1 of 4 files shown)


app/maptools/recordingmaptool.cpp
```diff
@@ -568,19 +566,69 @@ void RecordingMapTool::releaseVertex( const QgsPoint &point )
...
+  updateVertex( mActiveVertex, point );
+
+  // if it is a first or last vertex of the line we go to the recording mode
+  if ( mRecordedGeometry.type() == QgsWkbTypes::LineGeometry )
   {
-    setState( MapToolState::View );
+    if ( mActiveVertex.type() != Vertex::Existing && mActiveVertex.vertexId().vertex == 0 )
+    {
+      setState( MapToolState::Record );
+      mActiveVertex.setCoordinates( QgsPoint() );
```
tomasMizera, line 615: I would say we should invalidate the active index altogether `setActiveVertex( Vertex() );`

app/maptools/recordingmaptool.cpp
```diff
@@ -568,19 +566,69 @@ void RecordingMapTool::releaseVertex( const QgsPoint &point )
...
+      return;
+    }
+    else if ( mActiveVertex.type() != Vertex::Existing && mActiveVertex.vertexId().vertex == vertexCount - 1 )
+    {
+      QgsVertexId id = mActiveVertex.vertexId();
+      id.vertex += 1;
+      mActiveVertex.setVertexId( id );
+      mActiveVertex.setCoordinates( QgsPoint() );
+      setState( MapToolState::Record );
+      emit activeVertexChanged( mActiveVertex );
```
tomasMizera, line 626: Again, In record state active vertex should be invalid

app/maptools/recordingmaptool.cpp
tomasMizera: And also here we want to call `mNewVertexOrder(NewVertexOrder::Start)` to start digitizing from the beginning

tomasMizera (changes requested):

app/maptools/recordingmaptool.cpp
```diff
@@ -568,19 +566,69 @@ void RecordingMapTool::releaseVertex( const QgsPoint &point )
...
   }
-  else
+
+  updateVertex( mActiveVertex, point );
+
+  // if it is a first or last vertex of the line we go to the recording mode
+  if ( mRecordedGeometry.type() == QgsWkbTypes::LineGeometry )
   {
-    setState( MapToolState::View );
+    if ( mActiveVertex.type() != Vertex::Existing && mActiveVertex.vertexId().vertex == 0 )
```
tomasMizera, line 612: Don't we actually want it to be existing? (and only existing?) 
```suggestion
    if ( mActiveVertex.type() == Vertex::Existing && mActiveVertex.vertexId().vertex == 0 )
```

alexbruy: In this case releasing start/end point after grabbing it will end up in recording state, no? I thought we want go to recording only after removing handle, so grabbing start/end will be used only to move it in the new place.

app/maptools/recordingmaptool.cpp
alexbruy: Hmm, not sure this is required. Recording works fine with valid acrive index, e.g. when handle is selected. Also when active vertex is set there is no need to set mNewVertexOrder.

tomasMizera (approved):
Looking good after last edits!

app/maptools/recordingmaptool.cpp
tomasMizera: I believe active vertex should only be valid in `Grab` state and nowhere else. Just like you changed it 🙂

app/maptools/recordingmaptool.cpp
tomasMizera: > With this change we will always enter recording mode when start/end vertex is released.

Yes, that is correct and expected 🙂

PR #2194, opened 2022-08-03
Refactoring

alexbruy:
Refactor createNodesAndHandles

### DIFF
--- a/app/maptools/recordingmaptool.cpp
+++ b/app/maptools/recordingmaptool.cpp
@@ -557,31 +557,31 @@ void RecordingMapTool::lookForVertex( const QPointF &clickedPoint, double search
   if ( idx >= 0 )
   {
     // we found a point
-    Vertex lastActiveVertex = mActiveVertex;
+    Vertex clickedVertex = mVertices.at( idx );
     mActiveVertex = Vertex();
 
-    if ( mActiveVertex.type() == Vertex::Existing )
+    if ( clickedVertex.type() == Vertex::Existing )
     {
       mActiveVertex = mVertices.at( idx );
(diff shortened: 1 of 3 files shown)


app/maptools/recordingmaptool.cpp
```diff
@@ -208,7 +207,7 @@ void RecordingMapTool::removePoint()
         if ( mRecordedGeometry.constGet()->partCount() > 1 )
         {
           QgsMultiLineString *ml = qgsgeometry_cast<QgsMultiLineString *>( mRecordedGeometry.get() );
-          r = ml->lineStringN( mActiveVertex.vertexId().part );
+          r = ml->lineStringN( mActiveVertex.vertexId().part ); // mActivePart?
```
tomasMizera, line 210: I think `mActiveVertex` is correctly used here

app/maptools/recordingmaptool.cpp
```diff
@@ -381,127 +380,182 @@ void RecordingMapTool::prepareEditing()
...
+      else if ( vertexId.vertex == vertexCount - 1 )
+      {
+        continue;
+      }
+    }
+    else if ( mRecordedGeometry.type() == QgsWkbTypes::LineGeometry )
     {
-      QgsVertexId id( vertexId.part, vertexId.ring, vertexId.vertex + 1 );
+      // if this is firt point in line (or part) we add handle start point first
+      if ( vertexId.vertex == 0 && vertexId.part != startPart )
```
tomasMizera, line 451: I believe we should also not create the handle when the line has only one point

tomasMizera (commented):
Just two notes, it looks great!!

PR #2198, opened 2022-08-05
Guideline & multipart fixes GEOEDIT

tomasMizera:
Fixes regarding:
 - guideline:: now it is just either a line constructed from points around crosshair or polygon constructed from first point, crosshair and last point of recorded polygon.
 - multipart:: grab vertex was reworked, it had multiple issues - mainly concerning multiparts and rings.


app/maptools/recordingmaptool.h
```diff
@@ -59,7 +59,7 @@ class Vertex
 
     bool operator==( const Vertex &other )
     {
-      return other.vertexId() == mVertexId && other.type() == mType;
+      return other.vertexId() == mVertexId && other.coordinates() == mCoordinates && other.type() == mType;
```
alexbruy, line 62: Maybe better use `InputUtils::equals()` to compare coordinates to avoid rounding issues?

alexbruy (approved):

PR #2200, opened 2022-08-08
redraw geometry

alexbruy:
Adds redraw geometry functionality. Original geometry is not hidden yet.

### DIFF
--- a/app/qml/map/MapWrapper.qml
+++ b/app/qml/map/MapWrapper.qml
@@ -98,15 +98,12 @@ Item {
       }
 
       case "edit": {
-        centerToPair( internal.featurePairToEdit )
-        howtoEditingBanner.show()
         editingGeometryStarted()
         hideHighlight()
         break
       }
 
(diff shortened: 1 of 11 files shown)


app/qml/map/MapWrapper.qml
```diff
@@ -883,12 +897,26 @@ Item {
   function edit( featurepair ) {
     __activeProject.setActiveLayer( featurepair.layer )
     centerToPair( featurepair )
+    howtoEditingBanner.show()
 
     internal.featurePairToEdit = featurepair
+    internal.startEditingFromScratch = false
```
tomasMizera, line 903: Can you please move this line in the end of `onCanceled` and `onDone` slots of `recordingToolsComponent`?

The thing is you would need to set it to false also in record() and recordInLayer() functions - possibly more in future. If we instead set it to false when editing is finished (after `state = "view"`), we are sure it is not false on the next edit / record / split / whatever :)

PR #2202, opened 2022-08-09
Fix TODOs and some minor improvements

alexbruy:
Add check for geometry validity which works also for multipart lines/polygons
Display "Done" button when editing multipoint layers

### DIFF
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -713,12 +713,10 @@ QgsPointXY InputUtils::transformPointXY( const QgsCoordinateReferenceSystem &src
         return srcPoint;
       }
     }
-    qDebug() << "CT IS NOT VALID";
   }
   catch ( QgsCsException &cse )
   {
-    qDebug() << "EXCEPTION" << cse.what();
-    //Q_UNUSED( cse )
+    Q_UNUSED( cse )
   }
 
   return QgsPointXY();
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -713,6 +713,7 @@ QgsPointXY InputUtils::transformPointXY( const QgsCoordinateReferenceSystem &src
         return srcPoint;
       }
     }
+    qDebug() << "CT IS NOT VALID";
   }
   catch ( QgsCsException &cse )
   {
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -716,7 +716,8 @@ QgsPointXY InputUtils::transformPointXY( const QgsCoordinateReferenceSystem &src
   }
   catch ( QgsCsException &cse )
   {
-    Q_UNUSED( cse )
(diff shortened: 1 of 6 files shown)


app/maptools/recordingmaptool.cpp
```diff
@@ -336,11 +349,41 @@ bool RecordingMapTool::hasValidGeometry() const
...
+          if ( ( *it )->nCoordinates() < 3 )
+          {
+            return false;
+          }
+        }
+        return true;
+      }
+      else
+      {
+        return mRecordedGeometry.constGet()->nCoordinates() >= 3;
```
tomasMizera, line 385: Shouldn't it actually be 4? (When ring is closed, with three points it is 4 in total - first duplicated)

```suggestion
        return mRecordedGeometry.constGet()->nCoordinates() >= 4;
```

app/maptools/recordingmaptool.cpp
```diff
@@ -336,11 +349,41 @@ bool RecordingMapTool::hasValidGeometry() const
...
     }
     else if ( mLayer->geometryType() == QgsWkbTypes::PolygonGeometry )
     {
-      return mRecordedGeometry.constGet()->nCoordinates() >= 3;
+      if ( mRecordedGeometry.isMultipart() )
+      {
+        const QgsAbstractGeometry *geom = mRecordedGeometry.constGet();
+        for ( auto it = geom->const_parts_begin(); it != geom->const_parts_end(); ++it )
+        {
+          if ( ( *it )->nCoordinates() < 3 )
```
tomasMizera, line 376: Same, 4 instead of 3

`nCoordinates < 4`

app/inpututils.cpp
```diff
@@ -1088,6 +1088,11 @@ bool InputUtils::isPointLayer( QgsVectorLayer *layer )
   return geometryFromLayer( layer ) == "point";
 }
 
+bool InputUtils::isMultiPointLayer( QgsVectorLayer *layer )
```
tomasMizera, line 1091: maybe better to have more generic `isMultiPart()` function

tomasMizera: then in QML we can use `__inputUtils.isPointLayer( __activeLayer.vectorLayer ) && !__inputUtils.isMultiPartLayer( __activeLayer.vectorLayer )`

tomasMizera (approved):

tomasMizera:
We have some issues with autotests here ..

tomasMizera:
can you please check why is that @alexbruy ?

alexbruy:
No idea why is that. Failed test is not related to the any of the changes in this PR and passes locally.

alexbruy:
@tomasMizera should be fixed now.

tomasMizera (approved):

PR #2206, opened 2022-08-10
Edit buffer support (pt1)

alexbruy:
Initial implementation of the "undo" and vector layer edit buffer support. Currently works only when editing geometry of the existing feature.


tomasMizera:
@alexbruy can you please merge master here? iOS build should work then :)

tomasMizera:
There is some crash in lookForVertex method

tomasMizera:
It crashes also for me when I identify 3rd / 4th feature

alexbruy:
No crashes here on identify. I clicked over 20 features and it still works.

alexbruy:
Regarding crash in test, it is expected as this PR does not support creating new features. Only editing existing geometries.

PR #2215, opened 2022-08-11
Edit buffer support (pt 2)

alexbruy:
Work in progress.

Support edit buffer also when recoding new feature.


app/maptools/recordingmaptool.cpp
```diff
@@ -45,8 +45,8 @@ void RecordingMapTool::addPoint( const QgsPoint &point )
 
     QgsPoint transformed = InputUtils::transformPoint(
                              PositionKit::positionCRS(),
-                             mFeatureLayerPair.layer()->sourceCrs(),
-                             mFeatureLayerPair.layer()->transformContext(),
+                             mActiveLayer->sourceCrs(),
```
tomasMizera, line 48: Missing sanity check for `mActiveLayer`  pointer (it can be null)
```suggestion
if ( mActiveLayer )
```

app/maptools/recordingmaptool.cpp
```diff
@@ -57,16 +57,25 @@ void RecordingMapTool::addPoint( const QgsPoint &point )
...
   QgsVertexId id( mActivePart, mActiveRing, 0 );
-  if ( mRecordedGeometry.isEmpty() )
+
+  if ( !mActiveFeature.isValid() )
   {
-    mRecordedGeometry = InputUtils::createGeometryForLayer( mFeatureLayerPair.layer() );
+    QgsAttributes attrs( mActiveLayer->fields().count() );
+    QgsExpressionContext context = mActiveLayer->createExpressionContext();
+    QgsGeometry geometry = InputUtils::createGeometryForLayer( mActiveLayer );
+
```
tomasMizera, line 66: You miss variables manager's context here

app/maptools/recordingmaptool.cpp
```diff
@@ -164,6 +174,7 @@ void RecordingMapTool::addPoint( const QgsPoint &point )
     }
   }
 
+  mActiveLayer->beginEditCommand( QStringLiteral( "Add vertex" ) );
```
tomasMizera, line 177: Why did we revert to approach of having `beginEditCommand` everywhere? 🙂

tomasMizera (commented):
Just some minor comments, I understand this is still just WIP.

app/maptools/recordingmaptool.cpp
alexbruy: Hmm, but we did not have this check here previously. Do we want to have such checks in every place where we access layer?

tomasMizera: Yeah, we need such checks everywhere, active layer can be null

tomasMizera (approved):

PR #2226, opened 2022-08-19
Edit buffer support (pt 3) 

alexbruy:
Fixes some issues in geometry editing:

- expose active vertex geometry as a map tool property
- hide midpoints around active vertex
- fix removal of points in single- and multipart point geometries
- fix saving geometry changes in multipoint layers
- fix creation of empty multipart polygons (fixes redrawing)
- when editing singlepart point layer start edit with grabbed point

### DIFF
--- a/app/maptools/recordingmaptool.cpp
+++ b/app/maptools/recordingmaptool.cpp
@@ -963,6 +963,17 @@ void RecordingMapTool::releaseVertex( const QgsPoint &point )
   setActiveVertex( Vertex() );
 }
 
+FeatureLayerPair RecordingMapTool::commitChanges()
+{
+  // TODO
+  return FeatureLayerPair();
+}
+
+void RecordingMapTool::rollbackChanges()
+{
+  // TODO
+}
+
 void RecordingMapTool::updateVertex( const Vertex &vertex, const QgsPoint &point )
 {
   if ( !mActiveLayer )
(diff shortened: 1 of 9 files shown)


app/maptools/recordingmaptool.cpp
```diff
@@ -693,7 +693,28 @@ void RecordingMapTool::updateVisibleItems()
     }
     else if ( v.type() == Vertex::MidPoint )
     {
-      midPoints->addGeometry( v.coordinates().clone() );
+      // for lines show midpoint if previous or next vertex is not active
+      if ( mRecordedGeometry.type() == QgsWkbTypes::LineGeometry )
+      {
+        Vertex prevVertex = mVertices.at( i - 1 );
+        Vertex nextVertex = mVertices.at( i + 1 );
```
tomasMizera, line 716: If something would be wrong with the `mVertices` structure, this might explode

app/qml/map/RecordingToolbar.qml
```diff
@@ -104,7 +104,6 @@ Item {
         Item {
             Layout.fillWidth: true
             height: parent.height
-            visible: root.pointLayerSelected ? false : true
```
tomasMizera, line 107: I'd prefer keeping remove button hidden for simple point geometries. It opens another level of possible bugs if we would show it here.

tomasMizera:
iOS build is fine, it is just a build number generation issue

tomasMizera (approved):

tomasMizera (approved):

PR #2250, opened 2022-09-01
Edit buffer support (pt 5) 

### LINKED ISSUE #2233: Feature is not redraw after cancel
When I cancel editing or redrawing, the feature is not re rendered to its previous (original) state. I need to move map to see the original state.

Scenario:
1) edit a line feature
2) add some point into it
3) then hit cancel, edit page appears and hit cancel again

On map canvas you can see feature with all the points added. When you move the map, the feature will rerender to its original state.

Expectation is, that after cancel of the editing, the feature will be in original state immediately.

https://user-images.githubusercontent.com/83582181/187130848-7f35e9bc-d076-40e2-a28a-4d227d2e0a9d.mp4

### LINKED ISSUE #2234: 3 points needed for linestring
When I try to save linestring feature with 1 point, I get notification, that 3 points are needed.
Actually only 2 points are needed. It would be good to change the number in the notification from 3 to 2.

image

### LINKED ISSUE #2239: Split geometry option is available for points features
Split geometry option is available for points features. I would expect, that for pints, splitting is not allowed
Also, both edit and redraw geometry is available for normal point features. They do basically the same thing. I am not sure we want it to have this way.

### LINKED ISSUE #2248: Crash during recording with default project
Scenario:
1) in input hit create new project
2) set a name and hit Create project (keep the default list of fields)
3) open the project
4) hit record, the form opens (or the app crashes)
5) hit done -> not the app crashes

alexbruy:
- enable previously disable tests
- fix issues revealed by tests
- add tests to move/delete vertex
- fix some crashes revealed by tests
- try to fix #2233
- try to fix #2234
- try to fix #2248
- try to fix #2239

### DIFF
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -1842,51 +1842,56 @@ QgsGeometry InputUtils::createGeometryForLayer( QgsVectorLayer *layer )
     return geometry;
   }
 
-  switch ( layer->wkbType() )
+  bool isMulti = QgsWkbTypes::isMultiType( layer->wkbType() );
+
+  switch ( layer->geometryType() )
   {
-    case QgsWkbTypes::Point:
+    case QgsWkbTypes::PointGeometry:
     {
-      QgsPoint *point = new QgsPoint();
-      geometry.set( point );
+      if ( isMulti )
+      {
+        QgsMultiPoint *multiPoint = new QgsMultiPoint();
+        geometry.set( multiPoint );
+      }
+      else
+      {
+        QgsPoint *point = new QgsPoint();
+        geometry.set( point );
+      }
       break;
     }
 
-    case QgsWkbTypes::LineString:
+    case QgsWkbTypes::LineGeometry:
     {
-      QgsLineString *line = new QgsLineString();
-      geometry.set( line );
+      if ( isMulti )
+      {
+        QgsMultiLineString *multiLine = new QgsMultiLineString();
+        geometry.set( multiLine );
+      }
+      else
+      {
+        QgsLineString *line = new QgsLineString();
+        geometry.set( line );
+      }
       break;
     }
 
-    case QgsWkbTypes::Polygon:
+    case QgsWkbTypes::PolygonGeometry:
     {
-      QgsLineString *line = new QgsLineString();
-      QgsPolygon *polygon = new QgsPolygon( line );
-      geometry.set( polygon );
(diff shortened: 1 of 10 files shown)


tomasMizera:
Can you @alexbruy please list all tickets from @jozef-budac that this PR fixes?

app/inpututils.cpp
```diff
@@ -1842,57 +1842,84 @@ QgsGeometry InputUtils::createGeometryForLayer( QgsVectorLayer *layer )
...
+  if ( layer->geometryType() == QgsWkbTypes::LineGeometry )
+  {
+    nPoints = 2;
+  }
+  else if ( layer->geometryType() == QgsWkbTypes::PolygonGeometry )
+  {
+    nPoints = 3;
+  }
+
+  return tr( "You need to add at least %1 point(s)." ).arg( nPoints );
```
tomasMizera, line 1924: Can you please add a check there for multipart geometry and print `You need to add at least %1 point(s) to every part.` if it is multipart?

app/inpututils.cpp
```diff
@@ -1842,57 +1842,84 @@ QgsGeometry InputUtils::createGeometryForLayer( QgsVectorLayer *layer )
     return geometry;
   }
 
-  switch ( layer->wkbType() )
+  bool isMulti = QgsWkbTypes::isMultiType( layer->wkbType() );
```
tomasMizera, line 1845: Can we have a test for the `createGeometryForLayer` and `invalidGeometryWarning` functions? Thaks

tomasMizera:
Only minor notes. Code-wise it looks good, but needs a thorough testing once merged

app/maptools/recordingmaptool.cpp
```diff
@@ -310,6 +310,21 @@ void RecordingMapTool::removePoint()
       {
         mRecordedGeometry.get()->deleteVertex( current );
       }
+
+      // if this was the last point in the ring and ring is interior
+      // we remove that ring completely
+      if ( r->isEmpty() && current.ring > 0)
+      {
+        QgsCurvePolygon *p = qgsgeometry_cast<QgsCurvePolygon *>( poly );
+        p->removeInteriorRing( current.ring - 1 );
```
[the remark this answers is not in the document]
alexbruy: Because interior rings numeration starts with 0. Added comment.

PR #2254, opened 2022-09-07
initialize controllers before calling their methods (fix #1938)

### LINKED ISSUE #1938: Default value is not applied for the position variable in the first instance
I have a project with the @position_latitude @position_longitude as defaults for lat and long in the table.
https://public.cloudmergin.com/projects/saber/12-gps-buff

When I try to record a point for the first time, the default values are not applied. I need to edit geometry or add a new point for the values to appear.

alexbruy:
Fixes #1938

### DIFF
--- a/app/qml/form/FeatureFormPage.qml
+++ b/app/qml/form/FeatureFormPage.qml
@@ -143,12 +143,11 @@ Item {
         project: root.project
 
         controller: AttributeController {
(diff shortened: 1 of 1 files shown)


tomasMizera (approved):
✔️

tomasMizera:
We need test from @jozef-budac or @PatrikMizera

PR #2277, opened 2022-09-20
Geometry editing support

alexbruy:
This adds support for editing/redrawing geometries in Input

To edit feature geometry one should select feature and edit it, then in the feature form click on the new button "Edit geometry".
image

Clicking on "Edit geometry" button will transfer user to the map where geometry of selected feature will be highlighted with visible vertices and segment midpoints.
image

Once vertex is selected user can move it to another position or completely remove.
image

Editing in action.
Peek 2022-09-20 16-38


tomasMizera (approved):
There are some minor bugs reported, but I believe we are ready to merge! :)

PR #2287, opened 2022-09-21
handle Z/M values when creating geometry (fix #2281)

### LINKED ISSUE #2281: Wrong polygons representation on the map on RC
Scenario
1) open test_digitizing project
2) record the feature
-> now you can see that feature is much bigger then it was recorded.

It happens on both iOS and android devices

https://user-images.githubusercontent.com/83582181/191471165-0080be07-444b-432e-9ac0-0c45bee98e39.MP4

alexbruy:
Fixes #2281.


tomasMizera (approved):
Looking good to me

PR #2297, opened 2022-09-22
fix reusing last enetered attribute values (fix #2285)

### LINKED ISSUE #2285: Reuse last value is not working
Scenario:
1) download project tc04_recording and open it
2) turn on reuse last value in settings
3) record a point feature, fill all fields and check reuse option for each field
4) save the feature
5) hit record to record new feature
Expectation is that all fields would have fields field with values from previous feature but they are empty

This issue is not present on 1.7.0 version of the app

alexbruy:
Fixes #2285.


tomasMizera:
Verified by our tester team! :)

PR #2303, opened 2022-09-26
Qt6 support (pt2)

alexbruy:
Fix QR code scanner.
Use new API and fix deprecations warnings.

### DIFF
--- a/app/android.pri
+++ b/app/android.pri
@@ -65,7 +65,7 @@ android {
     # and providers and auth methods
     LIBS += -lqgis_core
 
-    LIBS += -lqt5keychain -lqca-qt5
+    LIBS += -lqt6keychain -lqca-qt6
     LIBS += -lgdal -lpq -lspatialite
     LIBS += -lcharset
     LIBS += -ltasn1 -lproj
--- a/app/android.pri
+++ b/app/android.pri
@@ -33,7 +33,7 @@ android {
 
     INCLUDEPATH += $${INPUT_SDK_INCLUDE_PATH}
     INCLUDEPATH += $${INPUT_SDK_INCLUDE_PATH}/qgis
-    
+
     # by default QMake eats linked libs if they are mentioned multiple times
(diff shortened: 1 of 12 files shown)


app/androidutils.h
```diff
@@ -10,10 +10,9 @@
 #ifndef ANDROIDUTILS_H
 #define ANDROIDUTILS_H
 
-
 #ifdef ANDROID
-#include <QAndroidActivityResultReceiver>
-#include <QAndroidJniObject>
+#include <QtCore/private/qandroidextras_p.h>
```
PeterPetrik, line 14: uff, why this is needed? I thought androidextras are dead?

app/android.pri
```diff
@@ -74,11 +74,11 @@ android {
     LIBS += -lsqlite3 -liconv -lz -lzip
     LIBS += -ltiff
     LIBS += -lwebp -lcurl -ljpeg
-    
+
     QT += multimedia
     QT += printsupport
-    QT += androidextras
     QT += bluetooth
+    QT += core-private
```
PeterPetrik, line 81: looks scary? why it is needed?

app/androidutils.h
PeterPetrik: all of those classes should be part of QtCore now no?

alexbruy: According to https://www.qt.io/blog/qt-extras-modules-in-qt-6 and https://doc-snapshots.qt.io/qt6-6.3/qtandroidprivate.html no, they are not available in the QtCore. There are only few things in the https://doc-snapshots.qt.io/qt6-6.3/qnativeinterface-qandroidapplication.html

app/android.pri
alexbruy: According to https://www.qt.io/blog/qt-extras-modules-in-qt-6 and https://doc-snapshots.qt.io/qt6-6.3/qtandroidprivate.html we need it as we use lot of the functions which are not yet available in the QtCore

PR #2318, opened 2022-10-05
initialize record position (fix #2316)

### LINKED ISSUE #2316: Wrong first point during recorgind with turn off GPS
Scenario
1) disable GPS for the app
2) open the project with a layer
3) hit record button (notification of the missing GPS appears)
4) hit add button  (without moving the map!)
 -> if the feature is point, no point is show on the map
-> it the feature is line or polygon, first point appears on the top left of the map canvas

Now then you hit done and save, feature is not saved
When I move the map canvas, the first point is recorded on the pointer marker, which is correct

https://user-images.githubusercontent.com/83582181/193664381-fd6104ec-6c62-464f-b197-acf0c463e42b.MP4

alexbruy:
Fixes #2316.

### DIFF
--- a/app/maptools/recordingmaptool.cpp
+++ b/app/maptools/recordingmaptool.cpp
@@ -61,13 +61,14 @@ void RecordingMapTool::addPoint( const QgsPoint &point )
 
     QgsPoint transformed = InputUtils::transformPoint(
                              PositionKit::positionCRS(),
-                             mActiveLayer->sourceCrs(),
+                             mActiveLayer->crs(),
                              mActiveLayer->transformContext(),
                              pointToAdd
                            );
 
     pointToAdd.setX( transformed.x() );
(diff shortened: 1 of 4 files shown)


tomasMizera:
Explanation: `recordPosition` has not been initialised in the `SnapUtils`. We currently use `SnapUtils` as a source of crosshair position, even if snapping is not allowed. It converts a device centre (width/2, height/2) to map canvas' and active layer' CRS. 
Position in active layer CRS, so-called `recordPosition`, is then propagated further to the app. E.g., it is used to create new points in `recordingMapTool`. 

`SnapUtils` previously initialised `recordPosition` as an empty point and it was not re-evaluated until the map was moved. We now re-evaluate it after all necessary properties are assigned.

tomasMizera (approved):

PR #2328, opened 2022-10-18
do not add feature to the layer edit buffer if it is already there

alexbruy:
Fixes issue with duplicated parent features created when adding a relation by not adding feature to the layer second time in attempt to get its id.


jozef-budac:
Hi Alex. I believe two things needs to be fixed. 
In the projects there is relation 1-N with districts and district_naturals. The problems are:
1) when I record district_natural for new district, FID of the districts is not filled in the district natural
2) when I save district_natural, it is not visible between naturals of the district (only after closing and opening the form again)


https://user-images.githubusercontent.com/83582181/196628511-b834487d-0d3d-40c0-bebc-9dda9ab9d6de.mp4

jozef-budac:
Hi Alex, this fix looks good! Everything looks ok, just Back button doesn't work in one case. I am not sure if it is the same problem as 
Scenario:
1) open project tc04_relations
2) record a districs 
3) in the form hit Add district manual
4) now you endered record mode for district natural
5) hit Back button -> nothins happens

https://user-images.githubusercontent.com/83582181/196794142-590bea93-22f1-4e6a-b425-d7319eeac8e5.mp4

jozef-budac:
Also maybe this should be considered for fix: https://github.com/MerginMaps/input/issues/2329
But I believe, we have this already in production and that it is not major thing

tomasMizera:
Hi Alex, can you please update versions to 1.9.0 so that we can test this on ios? :)

tomasMizera (approved):
Looking good, thanks!

PR #2357, opened 2022-11-01
Random android qt6 fixes

wonder-sk:


tomasMizera:
Good stuff! Shall we also remove the patch altogether?

PR #2360, opened 2022-11-04
Replace unavailable qt5 shortcuts with standard paths in qt6 ~ fix opening of feature form

tomasMizera:
Fixes opening of feature form.
`shortcuts.pictures` was a Qt5 way to get a path to the gallery on different platforms. `QStandardPaths` is a way to do it in C++.

Note: it is used only on macOS/Linux/Windows, iOS and Android have a native implementation. Unfortunately, this code runs each time form is opened no matter the platform.

### DIFF
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -56,6 +56,7 @@
 #include <limits>
 #include <math.h>
 #include <iostream>
+#include <QStandardPaths>
 
 static const QString DATE_TIME_FORMAT = QStringLiteral( "yyMMdd-hhmmss" );
 static const QString INVALID_DATETIME_STR = QStringLiteral( "Invalid datetime" );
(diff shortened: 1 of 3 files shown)


app/inpututils.cpp
```diff
@@ -1970,3 +1971,8 @@ void InputUtils::updateFeature( const FeatureLayerPair &pair )
   pair.layer()->commitChanges();
   pair.layer()->triggerRepaint();
 }
+
+QString InputUtils::imageGalleryLocation()
+{
+  return QStandardPaths::standardLocations( QStandardPaths::PicturesLocation ).last();
```
wonder-sk, line 1977: In theory standardLocations() can also return empty list, so it may be worth adding a check and emit a warning + return empty string if no path is available...

wonder-sk (approved):

PR #2377, opened 2022-11-15
show error dialog if project could not be read (fix #2370)

### LINKED ISSUE #2366: QT6: Problem opening a project
I have problem with opening project tc06_data_providers
Project is not loaded or doesn't show map themes.
After opening this project, usually I can see map canvas from previous opened project

### LINKED ISSUE #2370: No error shown if project loading fails
If user tries to load a project and its QGIS project file fails to load, no error message is shown, and it seems like the project loaded correctly - map is shown, although it's completely broken.

The correct behavior would be to show error message (including details from QgsProject::error()) and do not switch to map view.

alexbruy:
Show an error dialog if Input could not read and load QGIS project. Help button in the dialog will open a corresponding documentation page (right now a dummy, non existent url is used, hence CI failure).

Peek 2022-11-15 17-18

Fixes #2370.


tomasMizera (commented):
Good job! Can we also have a test for opening an invalid project file? (It could check if `ActiveProject` emits the loading failed signal)

app/inputhelp.cpp
```diff
@@ -126,6 +126,11 @@ QString InputHelp::merginTermsLink() const
   return MerginApi::defaultApiRoot() + "assets/tos.html" + utmTagOther;
 }
 
+QString InputHelp::projectLoadingErrorHelpLink() const
+{
+  return MerginApi::defaultApiRoot() + "/manage/project_loading/" + utmTagHelp;
```
tomasMizera, line 131:
```suggestion
  return helpRoot + "/manage/project_loading/" + utmTagHelp;
```

tomasMizera:
Well, this does not really fix the #2366, it simply shows the error message. Users will still need to fix the project manually in QGIS.

alexbruy:
It fixes it in the sense that instead of showing canvas with the old project it shows a message and user can at least understand what is going on. I can drop this ticket from the commit message.

alexbruy:
Done

app/qml/main.qml
```diff
@@ -606,6 +606,26 @@ ApplicationWindow {
...
+
+      onButtonClicked: function(button, role) {
+        switch (button) {
+          case MessageDialog.Close:
+            projectLoadingScreen.visible = false
+            projectErrorDialog.close()
+            projectPanel.openPanel()
+            break;
+          case MessageDialog.Help:
+            Qt.openUrlExternally(__inputHelp.projectLoadingErrorHelpLink)
```
tomasMizera, line 623: We also need to close the project loading screen here. Otherwise, it remains visible and one must restart the app to get out.

alexbruy: We already do it, no? Line 618.

tomasMizera: That is when you click `Close`, but on ios also the `Help` button closes the dialog

tomasMizera: Probably because you are navigated outside of the app (docs)

alexbruy: Should be fixed now.

tomasMizera:
Docs link https://github.com/MerginMaps/docs/pull/244

PR #2390, opened 2022-11-25
fix failing test

alexbruy:
Fixes failing active project test

### DIFF
--- a/app/activeproject.cpp
+++ b/app/activeproject.cpp
@@ -153,11 +153,6 @@ bool ActiveProject::forceLoad( const QString &filePath, bool force )
       CoreUtils::log( QStringLiteral( "Project loading" ), QStringLiteral( "Could not read project file: " ) + error );
 
       mLocalProject = LocalProject();
-      if ( mQgsProject )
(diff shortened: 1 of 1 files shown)


app/activeproject.cpp
```diff
@@ -153,9 +153,16 @@ bool ActiveProject::forceLoad( const QString &filePath, bool force )
       CoreUtils::log( QStringLiteral( "Project loading" ), QStringLiteral( "Could not read project file: " ) + error );
 
       mLocalProject = LocalProject();
-      mQgsProject->clear();
-      QList< QgsMapLayer * > layers;
-      mMapSettings->setLayers( layers );
+      if ( mQgsProject )
```
tomasMizera, line 156: Hmm, I think there is no point in checking if `mQgsProject` is valid here. It is accessed directly a few lines above.

tomasMizera: It should always be valid as we assign the QGIS singleton to it in the constructor

tomasMizera (approved):

PR #2418, opened 2022-11-30
Expressions evaluation fixes

### LINKED ISSUE #1756: Position variables not updated when editing a point
I have created a simple form to capture [position variables](https://help.inputapp.io/howto/position_variables) such as `@position_altitude`, `@position_horizontal_accuracy` and `@position_vertical_accuracy`. 

image

When I "Record" a new point, everything works and the form is filled with the expected values. However, when I use the "Edit Feature" and edit a geometry, the fields are cleared out and there seems to be no way to populate the fields with updated values:

image

Just for comparison, if I use a simple expression like `second( now() )`, to populate a "seconds" field, the field is populated/updated as expected but this does not seem to work for position variables.

Not sure if this might be related to https://github.com/lutraconsulting/input/issues/1743

### LINKED ISSUE #1956: The calculation of the sum of two fields is not correct
When I configure a field where the default value is the sum of two fields at the end, what it does is add the second value. For example:

Field 1 + Field 2
1 + 2 = 3 Value to be expected
1 + 2 = 12 Value displayed in input
WhatsApp Image 2022-02-19 at 8 41 50 AM

### LINKED ISSUE #2331: Incorrect default value for new features in some cases
In Input 1.8.0 there's a new issue affecting users that have default value expression like this:
```
IF ("block_id" IS NULL, ... [ expression using $geometry ] ...  ,  "block_id")
```
(assuming this is the default value for `block_id` attribute)

If geometry is not used, things work correctly.

Currently the code flow goes like this:
1. in RecordingMapTool::addPoint() we create a new feature with empty geometry and default values are applied.
2. later in AttributeController::recalculateDefaultValues() we re-evaluate the default values

The problem is that the expression above first gets evaluated using empty geometry, producing incorrect value, and then in the step two, the attribute is not updated to the correct value, because it is not NULL anymore.

Workaround: do not wrap default value expression using `IF (abc IS NULL, [expr], abc)` - simply use the internal expression. There is no need to check whether the value is NULL. (Note: this only works when "Apply default value on update" is not checked)

### LINKED ISSUE #2352: x coordinates in test project are not being updated properly
in project tc05_dafault_values; there are 2 bugs in this project, also (https://github.com/MerginMaps/input/issues/2351)

1) open the project 
2) record a feature, you will get some coordinates in the last field 
3) change the geometry of the feature, the coordinates in the last field are not going to be changed 
4) close the feature and reopen it, now the coordinates are going to be updated 

Found on: 
iOS 15.7 and Android 11 

https://user-images.githubusercontent.com/94351263/198240198-90fe0d36-aa27-4ed4-9bd4-519e36e0fd70.MP4

### LINKED ISSUE #2371: Enumeration by expresion doesn't works
I have used this expresion in older versions of MerginMaps and worked all fine.
**if(count(@layer_id)=0,1,maximum("ID")+1)**

But now in v1.9.0. have errors in enumeration, the expresion doesn't works.
See the video [https://www.youtube.com/watch?v=V7FC-m7eefw](url)

alexbruy:
Fixes various issues with expressions evaluation:

- do not evaluate expressions when digitizing new feature, to avoid issues when they are re-evaluated in the attribute controller (fix #2331)
- re-calculate default values if feature geometry was changed (fix #2352)
- use correct variant type when setting field values (fix #1956)
- add form scope to the expression context to support expressions which use form variables (fix #2371)
- write attrbute values as is (fix #1756)


tomasMizera:
It probably fixes https://github.com/MerginMaps/input/issues/1756 too, right @alexbruy ?

app/qml/map/RecordingTools.qml
```diff
@@ -49,10 +49,6 @@ Item {
     positionKit: __positionKit
     activeLayer: __activeLayer.vectorLayer
     activeFeature: root.activeFeature
-    variablesManager: __variablesManager
-
-    // Bind variables manager to know if we are centered to GPS or not when evaluating position variables
-    onIsUsingPositionChanged: __variablesManager.useGpsPoint = isUsingPosition
```
tomasMizera, line 55: This line should most likely not be removed. It allows `variablesManager` to fill the `from_gps` variable.

tomasMizera (commented):
Really cool! 🚀  Great job @alexbruy in finding all of these scenarios

tomasMizera (approved):
Thanks! We now need to wait for some testing from @jozef-budac or @PatrikMizera

tomasMizera:
Verified on test projects! 🎉

PR #2429, opened 2022-12-03
use ctest instead of bash

PeterPetrik:
split CI for tests and use ctest for running it

tests can be now run as standard `ctest` suite, e.g

```
cmake .... -DENABLE_TESTS=TRUE ...
ninja
ctest --verbose 
```

see `ctest --help` for all options
for example to run single test `ctest --tests-regex ".*Links.*" --output-on-failure` , which is effectively calling `Input --testLinks`.


PeterPetrik:
merginLinks tests fails because of not released https://github.com/MerginMaps/docs/pull/244

tomasMizera (approved):
I like the CI separation 👍

PR #2448, opened 2022-12-05
fix auto tests

PeterPetrik:
- use test_data from build dir
- and fix crashes
- fix autotests - problem was that the INPUT_TEST_DATA_DIR (from qmake) was no longer defined in our cmake build -> we need to use TEST_DATA_DIR from inputconfig.h


tomasMizera (approved):
🔥

PR #2449, opened 2022-12-05
project/user name validation

alexbruy:
When creating a project/registering a user make sure that their name does not contain invalid characters or reserved names.


tomasMizera (commented):
Thanks! I left just minor notes

tomasMizera (approved):

PR #2452, opened 2022-12-07
Qt6 fixes/Fix swapped BT arguments

### LINKED ISSUE #2437: QT6: problem with Bluetooth connection
There are multiple problems with Bluetooth connection to external GPS receiver

1) when connected, the name is not display in the list of devices
2) when connected, the device doesn't have checked radio button
3) when I hit the radio button, connection fails
4) when I remove the device them the connected list, it is still used as a receiver
5) I have hard times to connect to the device

I use Samsung A53 with Android 13

Recording
https://drive.google.com/file/d/1oYQFEgLqc7rLbFc6XS3RoD6C5wBEoUa_/view?usp=sharing

tomasMizera:
`deviceAddress` and `deviceName` were simply swapped, which resulted in undefined behaviour 🙂 

#32qr2h2
Fixes #2437

### DIFF
--- a/app/qml/misc/PositionProviderPage.qml
+++ b/app/qml/misc/PositionProviderPage.qml
@@ -323,7 +323,7 @@ Page {
       height: root.height + header.height
       width: root.width
 
-      onInitiatedConnectionTo: function( deviceName, deviceAddress ) {
+      onInitiatedConnectionTo: function( deviceAddress, deviceName ) {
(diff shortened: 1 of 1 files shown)


alexbruy (approved):

jozef-budac:
Haha, I thought whis will be the worse thing to fix :smile:

PR #2458, opened 2022-12-09
Specify minimum and target API level and fix set of input version define

tomasMizera:
Our previous target SDK level was `30` and minimum SDK was `23`. We need to target `31` and have minimum of `24` (<-- due to our java code)

See https://doc.qt.io/qt-6.2/cmake-target-property-qt-android-target-sdk-version.html and https://doc.qt.io/qt-6.2/qt-android-generate-deployment-settings.html#qt6-android-generate-deployment-settings


app/CMakeLists.txt
```diff
@@ -257,7 +257,12 @@ if (ANDROID)
     ${CMAKE_CURRENT_SOURCE_DIR}/android/AndroidManifest.xml @ONLY
   )
   set_target_properties(
-    Input PROPERTIES QT_ANDROID_PACKAGE_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/android
+    Input
+    PROPERTIES QT_ANDROID_PACKAGE_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/android
+               QT_ANDROID_TARGET_SDK_VERSION 31
```
PeterPetrik, line 262: Should be expanded from variable , not hardcoded here (both 31 and 24) https://github.com/MerginMaps/input/blob/511fd2a7abca4554edc8e3f9abc6169c0971df4d/CMakeLists.txt#L21

wonder-sk: in any case it would be good to clean up those variables - I guess the issue is that those variables were defined, but not used anywhere, and the `QT_` prefix looked like it is something we set for Qt framework. Maybe in the main cmakelists file we could change the variable names to e.g. `INPUT_ANDROID_TARGET_SDK_VERSION` just to make it clear it is "our" variable

tomasMizera: addressed

Issue #2446, opened 2022-12-05
Input will become extremely slow or nonresponsive while recording or editing line/polygon 


PatrikMizera:
1. any project with lines and polygons 
2. start recording lines/polygons 
3. you can record the first 2 points without issues 
4. after that, the app will become very slow to record or will not record at all 
5. you can still save the feature 

iPhone: 11; iOS: 16.1.1.

(I am constantly pressing the record button)
https://user-images.githubusercontent.com/94351263/205653576-a7a0f5ce-b7ef-4cd3-acd7-373bfee849c7.mp4

tomasMizera:
@PatrikMizera Can you please pass me the project?

tomasMizera:
Should be fixed now with #2463

Issue #2441, opened 2022-12-05
QT6: editing of linestring 


jozef-budac:
Scenario:
1) open project https://app.merginmaps.com/projects/input-testing/tc16_geometry_editing/tree
2) create a line string with multiple pints and save
3) then hit edit geometry for it and try to remove all points
-> you can remove just one one two points -> 1st error

4) after hitting multiple times the remove and add point you are not able to add point to the end of the string -> 2nd error

https://user-images.githubusercontent.com/83582181/205616851-489eb957-eae2-46e6-9e30-44b31c058a96.mp4

tomasMizera:
Related to #2446

jozef-budac:
I retest it on the today's latest master and it looks like, it is the problem with multiple click on the same button.
When I hitting buttons very slowly, the issue is not observed. But when I hit for example add button multiple times, the buttons stops working. 
Same for Undo button, or remove button

tomasMizera:
Should be fixed now in #2463 . @jozef-budac reopen this one if it is not fixed

PR #2463, opened 2022-12-13
try to fix freezes when buttons are pressed too fast and too often

alexbruy:
Hopefully fixes issue with clicking on the "Create project" button and freezes during digitizing (#2441, #2446)

### DIFF
--- a/app/qml/components/MainPanelButton.qml
+++ b/app/qml/components/MainPanelButton.qml
@@ -39,10 +39,8 @@ Rectangle {
         enabled: root.enabled && handleClicks
 
         TapHandler {
-            onPressedChanged: {
-                if ( !pressed ) {
-                    root.activated()
-                }
+            onTapped: function(eventPoint, button) {
+                root.activated()
             }
 
             onLongPressed: {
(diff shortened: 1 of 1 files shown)


app/qml/components/MainPanelButton.qml
```diff
@@ -34,14 +34,20 @@ Rectangle {
...
         anchors.fill: parent
         enabled: root.enabled && handleClicks
-        onClicked: {
-            root.activated()
-        }
-        onPressAndHold: {
-            root.activatedOnHold()
+
+        TapHandler {
+            onPressedChanged: {
```
tomasMizera, line 42: Can we use `onTapped` instead? This way it would react even if one would press and release outside of the Item (if they wanted to cancel the action)

tomasMizera (approved):

app/qml/components/MainPanelButton.qml
alexbruy: If you say so. I just adopted one of the your previous fixes.

PeterPetrik:
whatabout too furious pressing?

PR #2501, opened 2023-01-17
Change docs to qt 6.4.2

tomasMizera:


INSTALL.md
```diff
@@ -175,7 +175,7 @@ Requirements Android:
 
   export ANDROID_SDK_ROOT=/home/martin/android;
   export ANDROID_NDK_ROOT=/home/martin/android/ndk/23.1.7779620;
-  export QT_BASE=/opt/Qt/6.4.1;
+  export QT_BASE=/opt/Qt/6.4.2;
   export INPUT_SDK_ANDROID_BASE=/home/martin/input-sdk/android-macOS-20221117-129;
```
PeterPetrik, line 179: the SDK numbers will be different for each qt bump, I am not sure if we want to update these when we update CI, or rather write there : look at github action to find it?

PeterPetrik (approved):

INSTALL.md
tomasMizera: I guess it is enough to just update docs once in a while

PR #2506, opened 2023-01-20
update i18n strings

alexbruy:


tomasMizera:
Let's do this only after we confirm that all texts are ok :)

tomasMizera:
We will do it from scratch instead

PR #2507, opened 2023-01-20
add signal to notify about server upgrade

alexbruy:
Emited if we use default server AND previous server type was OLD AND new server type is SAAS.

### DIFF
--- a/core/merginapi.cpp
+++ b/core/merginapi.cpp
@@ -3289,6 +3289,11 @@ void MerginApi::setServerType( const MerginServerType::ServerType &serverType )
 {
   if ( mServerType != serverType )
   {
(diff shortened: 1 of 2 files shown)


core/merginapi.cpp
```diff
@@ -3282,6 +3282,11 @@ void MerginApi::setServerType( const MerginServerType::ServerType &serverType )
 {
   if ( mServerType != serverType )
   {
+    if ( mApiRoot == defaultApiRoot() && mServerType == MerginServerType::OLD && serverType == MerginServerType::SAAS )
```
tomasMizera, line 3285: I was playing around a little bit with it and in the end, we can safely remove the check for `defaultApiRoot` as we would not be able to test this feature. Thanks!

```suggestion
    if ( mServerType == MerginServerType::OLD && serverType == MerginServerType::SAAS )
```

tomasMizera (approved):

PR #2508, opened 2023-01-20
don't show no workspace banner on old and ce server

alexbruy:


app/qml/ProjectPanel.qml
```diff
@@ -97,7 +97,7 @@ Item {
 
   NoWorkspaceBanner {
     id: noWorkspaceBanner
-    visible: __merginApi.userAuth.hasAuthData() && !__merginApi.userInfo.hasWorkspaces && (stackView.currentItem && stackView.currentItem.objectName !== "registrationFinishPanel" && stackView.currentItem.objectName !== "createWorkspacePanel")
+    visible: __merginApi.serverType !== MerginServerType.OLD && __merginApi.serverType !== MerginServerType.CE && __merginApi.userAuth.hasAuthData() && !__merginApi.userInfo.hasWorkspaces && (stackView.currentItem && stackView.currentItem.objectName !== "registrationFinishPanel" && stackView.currentItem.objectName !== "createWorkspacePanel")
```
tomasMizera, line 100: 2 suggestions here:
  1. let's create a `apiSupportsWorkspaces` property in MerginApi (returning true for SAAS and EE) - I believe we are going to use it on many other places as well
  2. that inline condition is hard to read, let's use JS expression instead

```suggestion
    visible: {
      if ( !__merginApi.apiSupportsWorkspaces ) {
        return false;
      }
      if ( !__merginApi.userAuth.hasAuthData() ) {
        return false;
      }
      
      // do not show the banner in case of accepting invitation or creating a workspace
      if (stackView.currentItem && stackView.currentItem.objectName !== "registrationFinishPanel" && stackView.currentItem.objectName !== "createWorkspacePanel") {
        return false;
      }
      
      return !__merginApi.userInfo.hasWorkspaces
    }
```

tomasMizera:
Right  now I just do not see the banner at all :/
Screenshot 2023-01-20 at 17 33 37
Screenshot 2023-01-20 at 17 33 40
Screenshot 2023-01-20 at 17 33 49

PR #2511, opened 2023-01-20
Fix displaying of projects where user is not an author (e.g. was invited)

alexbruy:
WE should not apply tag to list projects request if we are fetching projects from the specific workspace, as user might be not an author of the project.

Also re-enable use of order_params.

### DIFF
--- a/core/merginapi.cpp
+++ b/core/merginapi.cpp
@@ -123,35 +123,48 @@ MerginSubscriptionInfo *MerginApi::subscriptionInfo() const
   return mSubscriptionInfo;
 }
 
-QString MerginApi::listProjects( const QString &searchExpression, const QString &flag, const QString &filterTag, const int page )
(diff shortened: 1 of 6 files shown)


core/merginapi.cpp
```diff
@@ -147,11 +147,11 @@ QString MerginApi::listProjects( const QString &searchExpression, const QString
   {
     query.addQueryItem( "name", searchExpression.toUtf8().toPercentEncoding() );
   }
-  if ( !flag.isEmpty() )
+  if ( workspace.isEmpty() && !flag.isEmpty() )
```
tomasMizera, line 150: This break the listing of public projects on workspaces-enabled servers though

tomasMizera:
It also removes the unused filter_tag^

alexbruy:
Looks good to me.

PR #2516, opened 2023-01-24
when user creates a new workspace warn if workspace with this name already exists

alexbruy:
Needs fix on the server side: error response should have 409 code.


tomasMizera (approved):
Looking good, I vote for merging it and retesting with new server later

PR #2520, opened 2023-01-27
extract error message from the server reply containing automatic validation results

alexbruy:
For instance, displays meaningful error for the case when workspace name is too short.

### DIFF
--- a/core/merginapi.cpp
+++ b/core/merginapi.cpp
@@ -1351,14 +1351,27 @@ QString MerginApi::extractServerErrorMsg( const QByteArray &data )
   if ( doc.isObject() )
   {
     QJsonObject obj = doc.object();
-    QJsonValue vDetail = obj.value( "detail" );
-    if ( vDetail.isString() )
+    if ( obj.contains( QStringLiteral( "detail" ) ) )
     {
-      serverMsg = vDetail.toString();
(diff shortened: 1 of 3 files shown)


core/merginapi.cpp
```diff
@@ -1351,14 +1351,27 @@ QString MerginApi::extractServerErrorMsg( const QByteArray &data )
...
+      else if ( vDetail.isObject() )
+      {
+        serverMsg = QJsonDocument( vDetail.toObject() ).toJson();
+      }
     }
-    else if ( vDetail.isObject() )
+    else if ( obj.contains( QStringLiteral( "name" ) ) )
     {
-      serverMsg = QJsonDocument( vDetail.toObject() ).toJson();
+      QJsonArray errors = obj.value( "name" ).toArray();
```
tomasMizera, line 1368: What would happen if `name` is not an array?

tomasMizera:
Can we have an auto-test for this function? Like pass JSON doc to the func and check the result message?

core/merginapi.cpp
alexbruy: Do we have responces where `name` is not an array?

tomasMizera: I do not know what Flask is capable of :) Best to double-check

tomasMizera (approved):

PR #2522, opened 2023-01-27
Gps antenna height

### LINKED ISSUE #1710: GPS antenna height
For precise measurements with external GPS antenna, it is useful if users can set the height of the antenna above the ground, so that this height is subtracted from the measured elevation in order to get the correct elevation.

In case the antenna height is non-zero, it would be useful to show the height settings in the GUI when recording, so that users are aware of it, and in case they change the height for some reason, it's lower chance that they would forget to adjust it in the app.

There should be a new variable containing the antenna height configuration, in case users want to store that to an attribute.

### LINKED ISSUE #2261: Storing the number of satellites and PDOP value of measured points
I want to store the number of satellites as an attribute to the point, but it doesn’t work with **@position_used_satellites** like it works with for example **@position_timestamp.**
Am I using the wrong variable name? I guess it should be possible to store this property when its visible in the Mergin app?

The QField documentation tells that the number of satellites is only available when using NMEA receiver, but why is it then visible in the app when I use internal GNSS? 

**I would also like to get a PDOP value.**

mergin

### LINKED ISSUE #2524: Read other DOP values from NMEA source
We might need to change the QGIS NMEA parsing here.
Add these values then as variables to feature form.

List of values we are interested in (Feel free to add more):
- `vdop` 
- `pdop`

We can currently read HDOP and the number of (visible/used) satellites

alexbruy:
Adds new setting for GPS antenna height, which is applied to the Z coordinate for newly recorded geometries (fix #1710). Value is saved in the settings between application restarts.

Additionally:
- adds expression variable `@position_gps_antenna_height`
- adds `@position_satellites_visible`, `@position_satellites_used`, `@position_gps_fix` and `@position_hdop` variables (fix #2261)
- adds `@position_vdop` and `@position_pdop` position variables (fix #2524)
- current antenna height is displayed in the GPS information panel and in the accuracy button on the map
- vpod, hdop and pdop values are displayed in the GPS information panel

image

### DIFF
--- a/app/maptools/recordingmaptool.h
+++ b/app/maptools/recordingmaptool.h
@@ -194,9 +194,6 @@ class RecordingMapTool : public AbstractMapTool
     int recordingInterval() const;
     void setRecordingInterval( int newRecordingInterval );
 
-    double antennaHeight() const;
-    void setAntennaHeight( double newAntennaHeight );
-
     PositionKit *positionKit() const;
     void setPositionKit( PositionKit *newPositionKit );
 
@@ -252,7 +249,6 @@ class RecordingMapTool : public AbstractMapTool
     void recordedGeometryChanged( const QgsGeometry &recordedGeometry );
     void recordingIntervalChanged( int lineRecordingInterval );
     void recordingTypeChanged( const RecordingMapTool::RecordingType &recordingType );
-    void antennaHeightChanged( double antennaHeight );
 
     void existingVerticesChanged( const QgsGeometry &existingVertices );
     void midPointsChanged( const QgsGeometry &midPoints );
--- a/app/maptools/recordingmaptool.h
+++ b/app/maptools/recordingmaptool.h
@@ -84,6 +84,7 @@ class RecordingMapTool : public AbstractMapTool
     Q_PROPERTY( bool centeredToGPS READ centeredToGPS WRITE setCenteredToGPS NOTIFY centeredToGPSChanged )
     Q_PROPERTY( RecordingType recordingType READ recordingType WRITE setRecordingType NOTIFY recordingTypeChanged )
     Q_PROPERTY( int recordingInterval READ recordingInterval WRITE setRecordingInterval NOTIFY recordingIntervalChanged )
+    Q_PROPERTY( double antennaHeight READ antennaHeight WRITE setAntennaHeight NOTIFY antennaHeightChanged )
 
     Q_PROPERTY( QgsVectorLayer *activeLayer READ activeLayer WRITE setActiveLayer NOTIFY activeLayerChanged )
     Q_PROPERTY( PositionKit *positionKit READ positionKit WRITE setPositionKit NOTIFY positionKitChanged )
@@ -194,6 +195,9 @@ class RecordingMapTool : public AbstractMapTool
     int recordingInterval() const;
     void setRecordingInterval( int newRecordingInterval );
 
+    double antennaHeight() const;
+    void setAntennaHeight( double newAntennaHeight );
+
     PositionKit *positionKit() const;
     void setPositionKit( PositionKit *newPositionKit );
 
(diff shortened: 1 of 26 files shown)


app/maptools/recordingmaptool.h
```diff
@@ -194,6 +194,9 @@ class RecordingMapTool : public AbstractMapTool
     int recordingInterval() const;
     void setRecordingInterval( int newRecordingInterval );
 
+    double antennaHeight() const;
+    void setAntennaHeight( double newAntennaHeight );
+
```
tomasMizera, line 199: leftover code

app/maptools/recordingmaptool.h
```diff
@@ -249,6 +252,7 @@ class RecordingMapTool : public AbstractMapTool
     void recordedGeometryChanged( const QgsGeometry &recordedGeometry );
     void recordingIntervalChanged( int lineRecordingInterval );
     void recordingTypeChanged( const RecordingMapTool::RecordingType &recordingType );
+    void antennaHeightChanged( double antennaHeight );
```
tomasMizera, line 255: leftover code

app/qml/map/MapWrapper.qml
```diff
@@ -732,46 +733,67 @@ Item {
...
-            return qsTr( "Unknown accuracy" )
+            accuracyText = qsTr( "Unknown accuracy" )
+          }
+          else
+          {
+            accuracyText = __inputUtils.formatNumber( __positionKit.horizontalAccuracy, accuracyButton.accuracyPrecision ) + " m"
           }
-          return __inputUtils.formatNumber( __positionKit.horizontalAccuracy, accuracyButton.accuracyPrecision ) + " m"
+
+          return gpsText === "" ? accuracyText : gpsText + " / " + accuracyText
```
tomasMizera, line 796: Let's show the GPS antenna height only when this button is in a valid state ~ when accuracy is shown. Some of the error messages about connection to external GPS are wide and I am afraid it would not be visible. There is also no reason to show it if a user does not have a GPS connection.

app/qml/map/MapWrapper.qml
```diff
@@ -771,7 +779,17 @@ Item {
           {
             return qsTr( "Unknown accuracy" )
           }
-          return __inputUtils.formatNumber( __positionKit.horizontalAccuracy, accuracyButton.accuracyPrecision ) + " m"
+
+          let accuracyText = __inputUtils.formatNumber( __positionKit.horizontalAccuracy, accuracyButton.accuracyPrecision ) + " m"
+          if ( __appSettings.gpsAntennaHeight > 0 )
+          {
+            let gpsText = __inputUtils.formatNumber( __appSettings.gpsAntennaHeight, 3 ) + " m"
```
tomasMizera, line 786:
```suggestion
            let gpsText = Number( __appSettings.gpsAntennaHeight.toFixed( 3 ) ) + " m"
```

This will help you remove the trailing zeros image

PatrikMizera:
tested it: 
@position_gps_fix and @position_hdop were not showing 
@position_satellites_visible, @position_satellites_used shown correctly
@position_gps_antenna_height was showing me 0 and was unable to connect other gps device (other phone)

app/qml/SettingsPanel.qml
```diff
@@ -234,6 +236,33 @@ Item {
             }
           }
 
+          PanelItem {
+            height: root.rowHeight
+            width: parent.width
+            text: qsTr("GPS antenna height")
+
```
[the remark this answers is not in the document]
alexbruy: Added, but widget used here does not wrap long text for unknown reason.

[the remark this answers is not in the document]
alexbruy: Tried to add text with max lines and width set, but still no luck. Above the screenshot of what I get

tomasMizera (approved):
Let's continue with testing now, the code is looking good 👏

waiotahi:
Nice work guys - especially providing DOP values.
As you're probably aware, the higher the DOP value the less accurate the position
https://en.wikipedia.org/wiki/Dilution_of_precision_(navigation)
Some other apps allow you to set a DOP value so as to automatically reject (or at least notify the operator) coordinates with a poor DOP value - typically less that 5 or 6
Perhaps this could be done using an expression ?

tomasMizera:
Hi @waiotahi, glad to hear you like the feature. 
Yes, you can achieve that currently with expressions - e.g. set a hard constraint expression on a field

PR #2532, opened 2023-01-30
use invitations form the user info when populating manage invitations page

alexbruy:
When showing user invitations after login, we should not send another request to the server to list invitations, as user profile already has this information. Send request to the server only when switch workspace page is opened, because in this case we want to get also any new invitations.

**NOTE**: ~~this won't work correctly now, as user profile does not include expiration date for invitations~~

### DIFF
--- a/app/qml/ManageInvitationsPage.qml
+++ b/app/qml/ManageInvitationsPage.qml
@@ -104,7 +104,6 @@ Page {
         Repeater {
           id: invRepeater
 
-          // TODO: change model to userInfo -> invitations
           model: InvitationsProxyModel {
             invitationsSourceModel: InvitationsModel {
               merginApi: __merginApi
--- a/app/qml/ManageInvitationsPage.qml
+++ b/app/qml/ManageInvitationsPage.qml
@@ -108,6 +108,7 @@ Page {
           model: InvitationsProxyModel {
             invitationsSourceModel: InvitationsModel {
               merginApi: __merginApi
+              fetchFromServer: false
             }
           }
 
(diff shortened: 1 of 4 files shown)


tomasMizera:
@alexbruy Let's target this one also to `2.0.1 dev` branch

alexbruy:
Added workaround: invitations without date considered as valid.

app/qml/ManageInvitationsPage.qml
```diff
@@ -107,6 +107,7 @@ Page {
           model: InvitationsProxyModel {
             invitationsSourceModel: InvitationsModel {
               merginApi: __merginApi
+              fetchFromServer: false
```
tomasMizera, line 110: We can remove the todo in line `106`

tomasMizera, line 110: If you set `fetchFromServer` to false here, it will never call the `v1/invitations`, will it? My understanding is that you need to make a property in `ManageInvitationsPage` and change this flag based on that property

tomasMizera (changes requested):

tomasMizera:
Let's target `2.1.0` here as well

app/qml/ManageInvitationsPage.qml
alexbruy: Right, we don't want to query server again, as invitations already fetched as a part of the user info request.

tomasMizera: Oh, right. I've forgotten that we use a different component on the `switchWorkspace` page. My bad!

tomasMizera:
@alexbruy If we change the branch to 2.1.0 then I am happy to merge this

tomasMizera (approved):

PatrikMizera:
tested on Android, invitations are functional

PR #2541, opened 2023-01-31
hide create workspace button

alexbruy:
hide workspace button from manage invitations page


app/qml/ManageInvitationsPage.qml
```diff
@@ -185,7 +186,7 @@ Page {
           verticalAlignment: Qt.AlignVCenter
           horizontalAlignment: Qt.AlignHCenter
 
-          visible: !__purchasing.transactionPending
+          visible: root.showCreate && !__purchasing.transactionPending
```
tomasMizera, line 189: I wonder why in the world we had this check here 🤔 

```suggestion
          visible: root.showCreate
```

tomasMizera (approved):

PR #2544, opened 2023-02-01
Fix field validations

alexbruy:
Bring back field validations by storing value entered by the user as a "raw" value and running validations against this value instead of using value converted to the compatible form for the field.

### DIFF
--- a/app/attributes/attributecontroller.cpp
+++ b/app/attributes/attributecontroller.cpp
@@ -1112,6 +1112,8 @@ bool AttributeController::setFormValue( const QUuid &id, QVariant value )
     QgsField field = item->field();
     QVariant val( value );
 
+    item->setRawValue( val );
+
     if ( !field.convertCompatible( val ) )
     {
       QString msg( tr( "Value \"%1\" %4 could not be converted to a compatible value for field %2(%3)." ).arg( value.toString(), field.name(), field.typeName(), value.isNull() ? "NULL" : "NOT NULL" ) );
--- a/app/attributes/attributecontroller.cpp
+++ b/app/attributes/attributecontroller.cpp
@@ -458,7 +458,6 @@ void AttributeController::updateOnFeatureChange()
         {
           mFeatureLayerPair.featureRef().setAttribute( fieldIndex, rememberedValue );
           itemData->setRawValue( rememberedValue );
-          emit formDataChanged( itemData->id(), { AttributeFormModel::RawValue } );
         }
       }
     }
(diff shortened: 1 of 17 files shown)


app/attributes/attributecontroller.cpp
```diff
@@ -457,6 +457,8 @@ void AttributeController::updateOnFeatureChange()
         if ( shouldUseRememberedValue )
         {
           mFeatureLayerPair.featureRef().setAttribute( fieldIndex, rememberedValue );
+          itemData->setRawValue( rememberedValue );
+          emit formDataChanged( itemData->id(), { AttributeFormModel::RawValue } );
```
tomasMizera, line 461: It should (hopefully) not be necessary to emit the signal here as later we call the `derived`

app/qml/editor/inputrangeeditable.qml
```diff
@@ -88,14 +88,6 @@ AbstractEditor {
         onTextEdited: {
           let val = text.replace( ",", "." ).replace( / /g, '' ) // replace comma with dot
 
-          let endsWithDecimalSeparator = val.endsWith('.');
```
tomasMizera, line 91: We need to remove this hack also in `app/qml/editor/inputtextedit.qml`

app/attributes/fieldvalidator.cpp
```diff
@@ -35,7 +35,8 @@ FieldValidator::ValidationStatus FieldValidator::validate( const FeatureLayerPai
   ValidationStatus state = Valid;
 
   const QgsField field = item.field();
-  QVariant value = pair.feature().attribute( item.fieldIndex() );
+  //QVariant value = pair.feature().attribute( item.fieldIndex() );
```
tomasMizera, line 38: comment

app/attributes/attributecontroller.h
```diff
@@ -167,6 +167,7 @@ class  AttributeController : public QObject
     void featureIdChanged();
     void changesCommited();
     void commitFailed();
+    void rawValueChanged();
```
tomasMizera, line 170: What is the point of this signal?

tomasMizera:
This is a big change and needs proper testing -> let's have it in `2.1.0` instead

app/attributes/attributecontroller.h
alexbruy: Mayber I get it wrong, but we talked about it and have it in the diagrams. Removed.

tomasMizera (approved):
Code is looking good :) Let's test this properly

tomasMizera:
@alexbruy can you rebase this branch please in order to have ios testing candidates? Thanks!

PR #2550, opened 2023-02-08
fix refreshing of home projects after login

alexbruy:
As we now create UI dynamically we should explicitly set state for `pageContent` component after it is constructed, otherwise state will be an empty string and `refreshProjectList()` won't be executed.


app/qml/ProjectPanel.qml
```diff
@@ -351,6 +351,10 @@ Item {
           }
         ]
 
+        Component.onCompleted: {
+          pageContent.state = "local"
+        }
```
tomasMizera, line 356: Let's try as follows
```suggestion
        state: "local"
```

tomasMizera (approved):

PR #2559, opened 2023-02-15
evaluate default value expressions for fields outside form (fix #2548)

### LINKED ISSUE #2548: Default values in fields outside of the feature form are not (re-)evaluated
Mergin Maps Input 2.0.0 and 2.0.1 will not evaluate default expressions for fields in a layer that are not part of the feature form. This is because the `QgsVectorLayerUtils::createFeature` is no longer used in `RecordingMapTool` due to issues with expressions (https://github.com/MerginMaps/input/pull/2418/commits/489285f5709bafec2dfa8c32e866a823365bfb92). 

To fix this, the `AttributeController::recalculateDefaultValues` method should recalculate default values for all fields in the layer, not just those from the form. 

Note: this issue does not occur for no-geometry layers or when creating features through a relation. 

**Workaround:** add these fields to the feature form and set them as "Hidden."

alexbruy:
Before evaluating default values for the form fields we now evaluate default values for all fields with the default expression which are not included in the form.

Fixes #2548.


app/attributes/attributecontroller.cpp
```diff
@@ -526,6 +556,50 @@ bool AttributeController::recalculateDefaultValues(
   bool isFirstUpdateOfNewFeature
 )
 {
+  // update default values for fields which are not in the form
+  QSet<int>::iterator it = mExpressionFieldsOutsideForm.begin();
+  while ( it != mExpressionFieldsOutsideForm.end() )
```
tomasMizera, line 561: Let's use `for` instead and we are ready to merge

tomasMizera (approved):

PR #2560, opened 2023-02-19
fix entering negative values in range field

alexbruy:
Fixes enetering of minus sign in the range field, as per https://app.clickup.com/t/861mctcbe


tomasMizera (changes requested):
@alexbruy please provide the following
- description
- auto tests
- explanation why this is needed

This piece of code looks dangerous as it effectively returns `rawValue` when someone asks for `AttributeValue` in `attributeformmodel` https://github.com/MerginMaps/input/blob/1b7c002f0e4eefdf183222721f996c5f17353b75/app/attributes/attributeformmodel.cpp#L60 What is the reason to return the `rawValue`?

alexbruy:
Can you provide guidelines on how should I test UI behaviour?

tomasMizera (approved):

PR #2575, opened 2023-02-22
Disable HTTP/2 requests to work around a networking bug

### LINKED ISSUE #2572: Danish WMS server from SDFI does not work anymore
When using a particular WMS server (`https://api.dataforsyningen.dk/orto_foraar_DAF`), the initial GetCapabilities request fails to finish properly.

- Regression in Input >= 2.0
- This can be reproduced also outside of Input / QGIS, just by using Qt's networking classes.
- Only Qt6 is affected, it works fine on Qt5 build of the same code
- It works fine with other HTTP(S) URLs, only this server seems to be affected

wonder-sk:
Fixes #2572 

Qt5 has HTTP/2 disabled by default, Qt6 has HTTP/2 enabled by default.

Users have reported that WMS server of Danish SDFI is not working anymore and it turns out it is due to a bug in Qt ([QTBUG-111417](https://bugreports.qt.io/browse/QTBUG-111417)) so let's avoid HTTP/2 for the time being.

### DIFF
--- a/app/main.cpp
+++ b/app/main.cpp
@@ -31,6 +31,7 @@
 #include "qgsconfig.h"
 #include "qgsproviderregistry.h"
 #include "qgsmaplayerproxymodel.h"
+#include "qgsnetworkaccessmanager.h"
 
 #include "androidutils.h"
 #include "ios/iosutils.h"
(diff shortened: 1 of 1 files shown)


tomasMizera (approved):
Nice! Thanks for the fix @wonder-sk 
We could use this approach to automatically log all requests/responses in MerginApi too 👏🏻

PR #2630, opened 2023-03-13
fix endless spinner in project list

alexbruy:
Setting state directly does not work as expected, reverting back to use `onCompleted`.


tomasMizera:
Unfortunately, I can't see any difference in behaviour.

PR #2631, opened 2023-03-13
do not add current position twice in the streaming mode (fix #2612)

### LINKED ISSUE #2612: Time streaming mode problem
With time streaming mode and linestring layer, there is a long guideline to the topleft corner, which should not be there
Scenario:
1) open any project with linestring (ex. tc04_recording)
2) open it and in setting set streaming mode with time period
3) hit record on linestring layer
4) hold GPS button and guideline apperas
-> this is not present in 2.0.2 version

obrázok

https://drive.google.com/file/d/1_GzDdV4VOT15Wy8r7ohS_kQ9U1pX5rxG/view?usp=share_link

alexbruy:
This leads to strange results, as in #2612.

### DIFF
--- a/app/qml/map/RecordingTools.qml
+++ b/app/qml/map/RecordingTools.qml
@@ -197,9 +197,6 @@ Item {
       if ( mapTool.recordingType === RecordingMapTool.Manual )
       {
         mapTool.recordingType = RecordingMapTool.StreamMode
-
-        // add first point immediately
-        mapTool.addPoint( crosshair.recordPoint )
         root.map.mapSettings.setCenter( mapPositioning.mapPosition )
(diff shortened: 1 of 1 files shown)


app/qml/map/RecordingTools.qml
```diff
@@ -197,9 +197,6 @@ Item {
       if ( mapTool.recordingType === RecordingMapTool.Manual )
       {
         mapTool.recordingType = RecordingMapTool.StreamMode
-
-        // add first point immediately
-        mapTool.addPoint( crosshair.recordPoint )
```
tomasMizera, line 202: What is the other place where we add point, please?

PR #2633, opened 2023-03-14
fixes #2632

### LINKED ISSUE #2632: Preview panel does not adjust transformation
If I take a photo with rotation of the camera, the photo appears correctly in the form but not in the preview panel:
image
image

saberraz:
Since qt 6.2, it is possible to use autotransform to change the photo orientation:
https://doc.qt.io/qt-6.2/qml-qtquick-image.html#autoTransform-prop

This should fix the image orientation issue in the preview panel.

### DIFF
--- a/app/qml/form/PreviewPanel.qml
+++ b/app/qml/form/PreviewPanel.qml
@@ -172,6 +172,7 @@ Item {
                     source: controller.photo
                     sourceSize: Qt.size(width, height)
                     fillMode: Image.PreserveAspectFit
+                    autoTransform : true
                     anchors.fill: parent
                     anchors.topMargin: InputStyle.panelMargin
                 }


PeterPetrik:
@saberraz can you share the project from the bug report so we can test it? Have you tested it with the fix ?

saberraz:
@PeterPetrik yes, I have shared the project with support:
https://app.merginmaps.com/projects/saber/Photo-orientation

Here is before this PR:
image

Here is after:

PeterPetrik:
do you have photo with banana for size comparison? 

@jozef-budac maybe worth to add to testing?

PeterPetrik (approved):

PeterPetrik:
based on EXIF info; output is that if you take image landscape/portrait it will always be shown correct way in preview panel

jozef-budac:
@saberraz can you share the project also with jozebudac account? And can I create a clone of that project and make it public for testing purposes?

Tadeo98:
Tested and it indeed fixed the orientation issue.

PR #2641, opened 2023-03-22
fix endless spinner in the project panel

alexbruy:
Do not set state for project panel explicitly as this prevents fetching/refreshing project list on the first load of the component.


tomasMizera (approved):
yep, this looks reasonable

PR #2677, opened 2023-04-24
Apply Qt6 click propagation fix to desktop builds

tomasMizera:
PR applies the fix ( see https://github.com/MerginMaps/input/pull/2425 ) for map canvas clicks propagation to desktop builds.

-> fixes the issue that feature form closes itself randomly on Windows (and other desktop builds)

Fixes https://github.com/MerginMaps/input/issues/2556

### DIFF
--- a/app/qml/map/MapCanvas.qml
+++ b/app/qml/map/MapCanvas.qml
@@ -173,31 +173,6 @@ Item {
     }
   }
 
-  // Mouse, stylus and other pointer devices handler
-  TapHandler {
-    id: stylusClick
-
-    property bool longPressActive: false
-
-    enabled: !mouseAsTouchScreen
-    acceptedDevices: PointerDevice.AllDevices & ~PointerDevice.TouchScreen
-
-    onSingleTapped: {
-      root.clicked(point.position)
-    }
-
-    onLongPressed: {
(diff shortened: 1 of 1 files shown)


app/qml/map/MapCanvas.qml
```diff
@@ -297,6 +272,31 @@ Item {
         }
       }
 
+      // Mouse, stylus and other pointer devices handler
```
PeterPetrik, line 275: so the whole fix is just moving this code under `Component`?

PeterPetrik (approved):

app/qml/map/MapCanvas.qml
tomasMizera: Yes, but not any `Component`. This component is used in `Loader` with a variable height (see line 204 `height: root.height - root.mapExtentOffset`). The height is a real height of map that is visible and not covered by other elements like form preview, GPS data page or projects panel. This means, that the handler will work only on the part of the map that can be seen.
It is a massive hack, but unfortunately, we still wait for the fix for the upstream bug https://bugreports.qt.io/browse/QTBUG-108821

tomasMizera:
iOS build failed only due to the old version uploaded, will work on master

PR #2679, opened 2023-04-25
Add native iOS CLLocationManager tracking backend

tomasMizera:
Adds native `CLLocationManager`(https://developer.apple.com/documentation/corelocation/cllocationmanager?language=objc) implementation of position tracking backend. Used only on iOS devices (iPhones, iPads)

- supports background use
- notifies PositionTrackingManager about change in position. 
- uses `CLLocationManager distanceFilter` ( see https://developer.apple.com/documentation/corelocation/cllocationmanager/1423500-distancefilter?language=objc ) to update frequency of position updates
- does not use signal/slot connections as these are not delivered when app is minimised on iOS

### DIFF
--- a/app/position/tracking/iostrackingbackend.cpp
+++ b/app/position/tracking/iostrackingbackend.cpp
@@ -1,0 +1,62 @@
+/***************************************************************************
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+
+#include "iostrackingbackend.h"
+#include "coreutils.h"
+
+IOSTrackingBackend::IOSTrackingBackend( UpdateFrequency frequency, QObject *parent )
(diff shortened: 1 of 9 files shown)


app/position/tracking/iostrackingbackend.cpp
```diff
@@ -0,0 +1,62 @@
...
+{
+  CoreUtils::log( TAG, QStringLiteral( "Error:" ) + QString::fromNSString( message ) );
+}
+
+void IOSTrackingBackend::stopTrying( NSString *message )
+{
+  CoreUtils::log( TAG, QStringLiteral( "Stopping updates, error:" ) + QString::fromNSString( message ) );
+
+  //
+  // let user know that tracking could not start/continue for some reason
```
PeterPetrik, line 60: todo?

PeterPetrik (approved):
Am I right that on iOS the tracking is based on distance , but on desktop on time interval?

tomasMizera:
> Am I right that on iOS the tracking is based on distance , but on desktop on time interval?

Yes, iOS API does not have an option to use time interval

app/position/tracking/iostrackingbackend.cpp
tomasMizera: Yeah, it is, but I skipped it for the initial implementation

PR #2695, opened 2023-05-23
Fix: #861mtpj06 Updated calculation of sizes for Android devices. 

liiteam:
(Qt6.5.0 get crazy DPI)

### DIFF
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -1170,8 +1170,11 @@ qreal InputUtils::calculateScreenDpr()
 
     qreal realDpi = dpiX < dpiY ? dpiX : dpiY;
     realDpi = realDpi * screen->devicePixelRatio();
-
+#ifdef ANDROID // due to crazy values onphysicalDotsPerInch[XY] from Qt6.5 (https://bugreports.qt.io/browse/QTBUG-113751)
+    return realDpi / 8.;
+#else
     return realDpi / 160.;
+#endif
   }
 
   return 1;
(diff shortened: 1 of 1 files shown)


app/inpututils.cpp
```diff
@@ -1170,8 +1170,11 @@ qreal InputUtils::calculateScreenDpr()
 
     qreal realDpi = dpiX < dpiY ? dpiX : dpiY;
     realDpi = realDpi * screen->devicePixelRatio();
-
+#ifdef ANDROID // due to crazy values onphysicalDotsPerInch[XY] from Qt6.5 (https://bugreports.qt.io/browse/QTBUG-113751)
```
PeterPetrik, line 1173: Can you if-def the workaround for QT6.5.0 only? (so when we compile for other qt version  we do not have the workaround)

https://bugreports.qt.io/browse/QTBUG-112742 suggest the value is now in pixels and not in inches, so qt 6.5.0 is basically unusable for us for release with this bug

PeterPetrik: How you come up with the number 8?

PeterPetrik (approved):

PeterPetrik:
DPI will be correct after you rotate mobile to landscape and back - so I am even not sure if this workaround works if you do this or it breaks? I am wondering if there is some c++ code we can run to force update of DPI (e.g. make screen smaller and full screen immediately, or rotate there and back?)

app/inpututils.cpp
liiteam: Added ifdef for Qt6.5.0, will add unfortunately also Qt6.5.1

liiteam: > How you come up with the number 8?

Just by test. Why is there the number 160?

liiteam:
>

PR #2697, opened 2023-05-25
Registration error messages

liiteam:


PeterPetrik:
can you please change destination branch to dev-2.3 and not master?

PR #2702, opened 2023-05-30
Show layer attribution

alexbruy:
Show layer atribution information (if present) in the layer details panel
image


saberraz:
Nice, what do you think of having it under Layer info, instead of adding a new item to an already crowded lower panel?

image

PeterPetrik:
@alexbruy can you please redo the GUI as suggested by @saberraz . Thanks!

tomasMizera (commented):
Nice, thanks for updating the visual.
Can you please add a simple test for the utils function?
You can add attribution to the test "planes" project and check it in the test, thanks!

app/qml/layers/LayerDetail.qml
```diff
@@ -202,6 +202,33 @@ Page {
...
+            }
+
+            Text {
+              id: attributionText
+
+              width: parent.width
+              textFormat: Text.RichText
+              wrapMode: Text.WordWrap
+              text: __inputUtils.layerAttribution(layerDetailData.mapLayer)
+            }
```
wonder-sk, line 229: It would be good to show the whole section only when the attribution text is not empty...

app/qml/layers/LayerDetail.qml
```diff
@@ -202,6 +202,33 @@ Page {
...
+            Text {
+              id: attributionTitle
+
+              text: qsTr( "Attribution" )
+              font.bold: true
+              font.pixelSize: InputStyle.fontPixelSizeNormal
+            }
+
+            Text {
+              id: attributionText
```
wonder-sk, line 223: Shall we wrap the attribution into a gray rectangle like with the visibility and legend?

app/layer/layerdetaildata.h
```diff
@@ -56,6 +58,7 @@ class LayerDetailData : public QObject
     void nameChanged( const QString &name );
     void vectorLayerChanged( QgsVectorLayer *vectorLayer );
     void layerTreeNodeChanged( QgsLayerTreeNode *layerTreeNode );
+    void mapLayerChanged( QgsMapLayer *mapLayer );
```
PeterPetrik, line 61: shouldn't we emit this signal somewhere (`setLayerTreeNode` ?)

app/layer/layerdetaildata.cpp
```diff
@@ -76,6 +78,7 @@ void LayerDetailData::setLayerTreeNode( QgsLayerTreeNode *newLayerTreeNode )
   emit isVectorLayerChanged( mIsVectorLayer );
 
   emit vectorLayerChanged( vectorLayer() );
+  emit vectorLayerChanged( nodeLayer->layer() );
```
PeterPetrik, line 81: mapLayerChanged ?

PeterPetrik (approved):

PR #2710, opened 2023-06-06
fix autotests

PeterPetrik:
for testing API we need PRO subscription; still few purchasing tests …are skipped

### DIFF
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -22,7 +22,7 @@ env:
   IOS_PROVISIONING_PROFILE_UUID: 59aaa8d7-516a-4592-8c58-d7d1c1f81610
   KEYCHAIN: ${{ 'inputapp.keychain' }}
   CCACHE_DIR: /Users/runner/work/ccache
-  CACHE_VERSION: 0
+  CACHE_VERSION: 1
 
 concurrency:
   group: ci-${{github.ref}}-ios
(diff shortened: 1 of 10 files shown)


PeterPetrik:
unrelated iOS problem

tomasMizera:
Regarding these iOS issues, will it work once merged?

tomasMizera (commented):
I really appreciate your effort in tackling this!

Given that this is quite a tricky issue, I was thinking about approaching it from a slightly different angle. What if we had an option on our test server that allows us to ignore limits for specific workspaces? We could have an API call that marks a workspace as a testing workspace, and then we can proceed as we did before. It feels like including the entire "purchasing" process in tests dealing with synchronization is a bit excessive.

Adding to that, I believe tests on other clients, like mergin-py-client, might be broken in a similar way. Moreover, these clients don't even offer a way to automatically purchase a subscription. How should we handle that situation with those clients?

PeterPetrik:
Tomas, I am happy with introducing some new API for testing workspaces, if it is required. The "TestSubscription" API is very simple, you send JSON with "action + workspace_id", so arguably this existing testing API is doing what you suggest. 

For Plugin, arguably you want to implement in its tests the utility functions to buy test subscriptions - with that you can start testing situations with various subscription types and how the plugin handles it. 

For iOS, it should work, I think the cache is broken on this branch and QT is not installed correctly (missing). 

The bottomline of the whole task is that merginAPI tests must be always running for both Input and Plugin, otherwise it is very scary.

PeterPetrik:
Agreed with Tomas to merge this

PR #2713, opened 2023-06-13
remove Locale workaround

PeterPetrik:
Qt 6.x already tries Locale without country in QTranslator::load() function

fix https://github.com/MerginMaps/input/issues/1417

see Qt6 logic in https://doc.qt.io/qt-6/qtranslator.html#load which is good.

tested on Android with EN (us) + FR (Canada) languages.

### DIFF
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -513,52 +513,6 @@ void InputUtils::quitApp()
   }
 }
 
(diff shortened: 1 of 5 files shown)


tomasMizera (approved):
Looks good to me. Let's test it with the test steps from the regression test suite.

PR #2714, opened 2023-06-13
Add build number to diagnostic log

PeterPetrik:
fix CU-861mn4ge4
fix https://github.com/MerginMaps/input/issues/1965

Screenshot 2023-06-13 at 12 40 51


app/main.cpp
```diff
@@ -393,6 +392,7 @@ int main( int argc, char *argv[] )
   InputTests tests;
   tests.parseArgs( argc, argv );
 #endif
+  qDebug() << "Mergin Maps Input App" << version << InputUtils::appPlatform() << "(" << CoreUtils::appVersionCode() << ")";
   qDebug() << "Built with QGIS version " << VERSION_INT;
```
tomasMizera, line 396: Maybe both of these could be `CoreUtils::log` instead of `qDebug`?

tomasMizera (approved):
Looking good!

app/main.cpp
PeterPetrik: I am not sure, in log we have it in separate entry... I think here it is more useful for developers on top of the application output to see if they are compiling with correct QT and App version...

PR #2717, opened 2023-06-15
Fix app state when closing GPS info pane

iiLubos:
CU-861mrycj9


tomasMizera (changes requested):
We can not set the state this explicitly. It breaks another scenarios like when the app is in stakeout or in the recording state.

Let's have a chat and find a different solution to this

iiLubos:
This fix will be merged in https://github.com/MerginMaps/input/pull/2727

PR #2721, opened 2023-06-22
FIX: Do not clear auth cache when internet connection is lost

PeterPetrik:
Code in mergin shows that we get only code 401 in case we have invalid username/login or password. 
https://github.com/MerginMaps/mergin/blob/master/server/mergin/auth/controller.py#L215

So for ANY other network error in authorization, we can assume the username + password are fine, so do not clear them!

fix https://github.com/MerginMaps/input/issues/2523
fix https://github.com/MerginMaps/input/issues/2139

### DIFF
--- a/core/merginapi.cpp
+++ b/core/merginapi.cpp
@@ -1134,38 +1134,48 @@ void MerginApi::authorizeFinished()
     }
     else
     {
-      mUserAuth->blockSignals( true );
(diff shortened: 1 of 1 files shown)


PeterPetrik:
autotest error is not relevant to this PR

tomasMizera (approved):
Looking good, let's test it out :)

PR #2724, opened 2023-06-28
Expression-based names for photos

alexbruy:
Prepare ground for expression-based photo naming:
 - if new photo is taken or selected and the field hase expression set — rename file using expression
 - update field value accordingly and save changes.


app/attributes/attributecontroller.cpp
```diff
@@ -941,6 +941,78 @@ bool AttributeController::save()
...
+        }
+        else
+        {
+          QVariant val( value );
+          if ( !field.convertCompatible( val ) )
+          {
+            QgsMessageLog::logMessage( tr( "Value \"%1\" %4 could not be converted to a compatible value for field %2 (%3)." ).arg( value.toString(), field.name(), field.typeName(), value.isNull() ? "NULL" : "NOT NULL" ), QStringLiteral( "Input" ), Qgis::Warning );
+          }
+          else
+          {
```
tomasMizera, line 992: I would again prefer to `++formItemsIterator;` and `continue;` instead of nesting deeper

app/attributes/attributecontroller.cpp
```diff
@@ -941,6 +941,78 @@ bool AttributeController::save()
...
+    expressionContext << mVariablesManager->positionScope();
+
+  expressionContext.setFields( mFeatureLayerPair.feature().fields() );
+  expressionContext.setFeature( mFeatureLayerPair.featureRef() );
+
+  // check for new photos
+  QMap<QUuid, std::shared_ptr<FormItem>>::iterator formItemsIterator = mFormItems.begin();
+  while ( formItemsIterator != mFormItems.end() )
+  {
+    std::shared_ptr<FormItem> item = formItemsIterator.value();
```
tomasMizera, line 958: Can we check here if the `item` has the photo name format defined and if not, just `continue`? We will save quite a few checks this way.

Ideally, we can reuse the `photoFields` for the check to avoid asking for QgsProject's instance and `readEntry` many times.
Something like:

```
  const QgsField field = item->field();
  if ( !photoFields.contains( field.name ) )
  {
    // increment index..
    continue;
  }
```

tomasMizera (changes requested):
Thanks for the work!

Generally it looks good, my comments just try to make it a little faster (by saving some checks and lookups). This piece of code is kind-of crucial, and we need to ensure we do not keep the UI frozen for too long. In future (production will show how distant), we might want to make this entire `save` function run on a different thread (or at least this whole renaming mechanism).

app/attributes/attributecontroller.cpp
```diff
@@ -1276,3 +1278,82 @@ void AttributeController::onFeatureAdded( QgsFeatureId newFeatureId )
   setFeatureLayerPair( FeatureLayerPair( f, mFeatureLayerPair.layer() ) );
   emit featureIdChanged();
 }
+
+void AttributeController::renamePhotos()
+{
+  const QStringList photoNameFormat = QgsProject::instance()->entryList( QStringLiteral( "Mergin" ), QStringLiteral( "PhotoNaming/%1" ).arg( mFeatureLayerPair.layer()->id() ) );
+
```
tomasMizera, line 1285: We could skip the entire logic here in case a user does not have anything defined to save looping over all fields.

```suggestion
if (photoNameFormat.isEmpty())
  return?
```

tomasMizera (approved):

tomasMizera:
Let's merge this and keep fixing in additional PRs

PR #2736, opened 2023-07-24
Android tracking when app is minimised

tomasMizera:
### Makes tracking working in the background (app runs, but not visible) on the Android platform

Uses a file to pass position updates from `TrackingService (java, different thread - foreground service)` to `AndroidTrackingBackend (cpp)`. In case the main activity (cpp part) is stopped when in the background, the tracking service (java) still operates and keeps writing updates to the file. Once cpp is started again it reads the missed position updates from the file and continues to show the tracking.

image

This is how we check if the tracking service is running when the app (cpp) is maximised again:

image

---

 - _Some tweakings are still to be done with tracking, but will be tackled in the followup PRs_


app/position/tracking/positiontrackingmanager.h
```diff
@@ -69,14 +69,15 @@ class PositionTrackingManager : public QObject
...
+    Q_INVOKABLE void commitTrackedPath();
 
     QDateTime startTime() const;
 
     bool isTrackingPosition() const;
 
   public slots:
-    void addPoint( const GeoPosition &position );
+    void addPoint( const QgsPoint &position );
+    void addPoints( QList<QgsPoint> positions );
```
PeterPetrik, line 80: `const QList< QgsPoint > &`

app/inpututils.cpp
```diff
@@ -2013,3 +2013,33 @@ QString InputUtils::imageGalleryLocation()
 
   return galleryPaths.last();
 }
+
+QList<QgsPoint> InputUtils::parsePositionUpdates( const QString &data )
+{
+  QList<QgsPoint> parsedUpdates;
```
PeterPetrik, line 2019: Any reason why QList and not QVector is used?

wonder-sk (approved):
:clap: :clap: :clap:

app/inpututils.cpp
tomasMizera: I think QVector is just an alias for QList in Qt6 
image

tomasMizera: https://doc.qt.io/qt-6/qvector.html

PeterPetrik: good to know. thanks

PR #2739, opened 2023-07-28
Fixed feature text overflow

iiLubos:
FixTextOverflowSmall

### DIFF
--- a/app/qml/layers/FeaturesListPageV2.qml
+++ b/app/qml/layers/FeaturesListPageV2.qml
@@ -102,7 +102,7 @@ Page {
           Layout.fillWidth: true
 
           color: InputStyle.secondaryFontColor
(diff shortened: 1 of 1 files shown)


app/qml/layers/FeaturesListPageV2.qml
```diff
@@ -102,7 +102,7 @@ Page {
           Layout.fillWidth: true
 
           color: InputStyle.secondaryFontColor
-          text: model.Description + ( model.SearchResult ? ", " + model.SearchResult : "" )
+          text: model.Description + ( model.SearchResult ? ", " + model.SearchResult.substring(0, model.SearchResult.indexOf("\n")) : "" )
```
tomasMizera, line 105: LGTM; but can we add a comment here why do the substring? :)

tomasMizera (approved):

PR #2770, opened 2023-08-24
Show dialog when max. number of projects was reached

iiLubos:
Show this popup with a link to update the plan, instead of a simple toast, in case of max. number projects is reached
image

Fixed code to show this popup with a link to update the plan in case of max. size of projects is reached
MaxStorage

### DIFF
--- a/core/merginapi.cpp
+++ b/core/merginapi.cpp
@@ -1081,7 +1081,7 @@ void MerginApi::createProjectFinished()
   else
   {
     QByteArray data = r->readAll();
-    QString code = extractServerErrorStringValue( data, QStringLiteral( "code" ) );
+    QString code = extractServerErrorCode( data );
     QString serverMsg = extractServerErrorMsg( data );
     QString message = QStringLiteral( "FAILED - %1: %2" ).arg( r->errorString(), serverMsg );
     bool showLimitReachedDialog = EnumHelper::isEqual( code, ErrorCode::ProjectsLimitHit );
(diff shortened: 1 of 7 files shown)


core/merginapi.cpp
```diff
@@ -1079,14 +1079,29 @@ void MerginApi::createProjectFinished()
   }
   else
   {
-    QString serverMsg = extractServerErrorMsg( r->readAll() );
+    QByteArray data = r->readAll();
+    QVariant code = extractServerErrorValue( data, QStringLiteral( "code" ) );
+    QString serverMsg = extractServerErrorMsg( data );
     QString message = QStringLiteral( "FAILED - %1: %2" ).arg( r->errorString(), serverMsg );
-    CoreUtils::log( "create " + projectFullName, message );
+    bool showLimitReachedDialog = code.toString() == QStringLiteral( "ProjectsLimitHit" );
```
tomasMizera, line 1086: Can we define these error codes as a new enum type and compare against enum values instead?

Example:
`... code == ErrorCodes::ProjectsLimitHit`

core/merginapi.h
```diff
@@ -681,6 +695,23 @@ class MerginApi: public QObject
...
+    * Extracts string code of an error json. If its not json or value cannot be parsed, QString() is return;
+    * \param data Data received from mergin server on a request failed.
+    */
+    QString extractServerErrorCode( const QByteArray &data );
+    /**
+    * Extracts string value of an error json. If its not json or value cannot be parsed, QString() is return;
+    * \param data Data received from mergin server on a request failed.
+    * \param key Where should be a value from data
+    */
+    QString extractServerErrorStringValue( const QByteArray &data, const QString &key );
```
tomasMizera, line 708: This function is now obsolete when we have the `extractServerErrorCode`, right? I vote for removing it ✌🏻

tomasMizera (approved):

PR #2777, opened 2023-08-30
Add support for position tracking for Android versions lower than 12

tomasMizera:
FUSED_PROVIDER is available only from SDK level 31 (Android 12). 
Let's use GPS_PROVIDER (available for all versions) in case the FUSED one is not available.


tomasMizera:
For reference https://developer.android.com/reference/android/location/LocationManager#FUSED_PROVIDER

wonder-sk (approved):

PR #2788, opened 2023-09-12
Fixed showing multiline features

iiLubos:
Showing feature names and description in one row
CU-861n6vx85


tomasMizera (approved):
Thanks for the fix 👍🏻

PR #2791, opened 2023-09-14
Show migration dialogue 

iiLubos:
CU-861n8yh4d Show migration dialogue when Legacy CE servers version is lower than 2023.2.0

MigrationDialog

When the Ignore button is pressed this dialog will not be shown for the current version anymore


core/merginapi.cpp
```diff
@@ -1364,6 +1364,11 @@ void MerginApi::checkMerginVersion( QString apiVersion, bool serverSupportsSubsc
     {
       setApiVersionStatus( MerginApiStatus::INCOMPATIBLE );
     }
+
```
tomasMizera, line 1367: Can you add a comment here please to describe why we do this?

core/merginapi.h
```diff
@@ -370,6 +370,8 @@ class MerginApi: public QObject
 
     static const int MERGIN_API_VERSION_MAJOR = 2020;
     static const int MERGIN_API_VERSION_MINOR = 4;
+    static const int MERGIN_SERVER_VERSION_MAJOR = 2023;
+    static const int MERGIN_SERVER_VERSION_MINOR = 2;
```
tomasMizera, line 374:
```suggestion
    static const int MINIMUM_SERVER_VERSION_MAJOR = 2023;
    static const int MINIMUM_SERVER_VERSION_MINOR = 2;
```

iiLubos: solved

core/merginapi.cpp
iiLubos: added :)

core/merginapi.cpp
```diff
@@ -1364,6 +1364,12 @@ void MerginApi::checkMerginVersion( QString apiVersion, bool serverSupportsSubsc
...
+    // will be dropped support for old servers (mostly CE servers without workspaces)
+    if ( ( MINIMUM_SERVER_VERSION_MAJOR == major && MINIMUM_SERVER_VERSION_MINOR > minor ) || ( MINIMUM_SERVER_VERSION_MAJOR > major ) )
+    {
+      emit migrationRequested();
+    }
```
tomasMizera, line 1372: Unfortunately, this piece of code (`checkMerginVersion`) is executed each time the `GET /ping` request finishes. We need to check this only in the `GET /config` as the ping request is **called several times during the app run**. As a result, this dialogue would show up multiple times.

iiLubos: solved

core/merginapi.cpp
```diff
@@ -1347,14 +1347,8 @@ void MerginApi::checkMerginVersion( QString apiVersion, bool serverSupportsSubsc
...
-    QRegularExpression re;
-    re.setPattern( QStringLiteral( "(?<major>\\d+)[.](?<minor>\\d+)" ) );
-    QRegularExpressionMatch match = re.match( apiVersion );
-    if ( match.hasMatch() )
-    {
-      major = match.captured( "major" ).toInt();
-      minor = match.captured( "minor" ).toInt();
-    }
+
+    parseVersion( apiVersion, major, minor );
```
PeterPetrik, line 1351: Would it be possible to add server version + flavour (EE/CE/SaaS) to the "sticky" part of the diagnosis log https://github.com/MerginMaps/input/blob/780622dacf156cad443caa5fa6fd0089c21e2ae9/app/inputhelp.cpp#L184 ?

core/merginapi.cpp
```diff
@@ -1474,6 +1468,21 @@ ProjectDiff MerginApi::localProjectChanges( const QString &projectDir )
...
+  re.setPattern( QStringLiteral( "(?<major>\\d+)[.](?<minor>\\d+)" ) );
+  QRegularExpressionMatch match = re.match( version );
+  if ( match.hasMatch() )
+  {
+    major = match.captured( "major" ).toInt();
+    minor = match.captured( "minor" ).toInt();
+    return true;
+  }
+
+  return false;
```
PeterPetrik, line 1483: instead of return false/true (that is not used anyways in the parent scope), can we just log `version` to `coreutils:log` with some appropriate message?

core/merginapi.cpp
iiLubos: Added server version + flavour (EE/CE/SaaS) to the diagnosis log

core/merginapi.cpp
iiLubos: Currently used return value

PeterPetrik (approved):

tomasMizera (changes requested):
Hello @iiLubos , can you please fix conflicts? We can merge it then

app/qml/main.qml
```diff
@@ -623,6 +623,32 @@ ApplicationWindow {
         }
     }
 
+    MessageDialog {
+      id: migrationDialog
+
+      property string version
+
+      onAccepted: migrationDialog.close()
+      text: qsTr("Your server will soon be out of date \n\n Please contact your server administrator to upgrade your server to the latest version. Subsequent releases of our mobile app may not be compatible with your current server version.")
```
tomasMizera, line 632: Let's use `title` instead :)
```suggestion
      title: qsTr("Your server will soon be out of date")
      text: qsTr("Please contact your server administrator to upgrade your server to the latest version. Subsequent releases of our mobile app may not be compatible with your current server version.")
```

app/qml/main.qml
```diff
@@ -623,6 +623,32 @@ ApplicationWindow {
...
+      buttons: {
+        // show the Ignore button only when a server version is known
+        if( version.length > 0 )
+          return MessageDialog.Close | MessageDialog.Help | MessageDialog.Ignore
+        else
+          return MessageDialog.Close | MessageDialog.Help
+      }
```
tomasMizera, line 639: Why do we need such a comparison?
```suggestion
      buttons: MessageDialog.Close | MessageDialog.Help | MessageDialog.Ignore
```

tomasMizera (approved):

tomasMizera:
@jozef-budac Let me know when you are ready to test this, I can help you to set it up :)

tomasMizera:
We will merge this so that CI runs

Issue #2382, opened 2022-11-21
Project files' content cleared after an unsuccessful sync


tomasMizera:
We have seen a data loss during synchronisation of a huge project.

The pull step has removed content of local files and the following push overwritten those files' content on the server.

How it happened:
 - Project is downloaded and has local changes waiting to be synced
 - Start to sync
 - MM Input detects that pull must be performed first as the server has remote changes
 - MM Input starts to download (pull) files from the server to the `.temp` dir 
 - Some file chunks in the `.temp` directory **can not be opened for writing**  and remain empty (but sync is **not** aborted!)
 - The following update of local files with those from `.temp` most probably cleared the content
 - GEODIFF screams error `Missing 'base' file when opening sqlite driver` (but sync is **not** aborted!)
 - GEODIFF yells another error `Table names are not matching between the input databases` (but sync is **not** aborted!)
 - Pull finished, continued with push
 - Some more geodiff errors `Geodiff create changeset on Field Layer.gpkg FAILED with error 1 (will do full upload)`
 - All files were successfully sent to server
 - Files on server were cleared

Not all the files were cleared, some files went through the sync fine. 

_Maybe the device did not have enough storage._

Contact me to get the diagnostic log.

So far seen only on Apple devices 🍎

Hackster93:
Hi @tomasMizera,

We experienced a similar issue today. One of our consultants synced a bunch of projects, and some of them (3) ended up deleting the GPKG contents. I am sending a support ticket email as well.

The user was `annaline.smith` and the three WPs in question were:
1. agmc/Mountain View
2. agmc/Meulsrivier
3. agmc/Romansrivier

She uses both a phone and tablet and I have asked her to send diagnostic reports from both devices. Both are Android.

tomasMizera:
Hello @Hackster93 Chris, sorry to hear that and thanks for letting us know!
We will have a look at this one with priority and do our best to deliver fix soon.

PR #2793, opened 2023-09-20
Sync. problem when drive space or write permission issue

iiLubos:
Stop syncing and write a log message when problem with drive space or write permission.
Fixes: #2382 (hope)
Fixes: https://github.com/MerginMaps/input/issues/2808

### DIFF
--- a/core/merginapi.cpp
+++ b/core/merginapi.cpp
@@ -244,8 +244,6 @@ void MerginApi::downloadNextItem( const QString &projectFullName )
   Q_ASSERT( mTransactionalStatus.contains( projectFullName ) );
   TransactionStatus &transaction = mTransactionalStatus[projectFullName];
 
-  qInfo() << "xxx MerginApi::downloadNextItem" << projectFullName;
-
   if ( transaction.downloadQueue.isEmpty() )
   {
     // there's nothing to download so just finalize the pull
@@ -387,8 +385,38 @@ void MerginApi::downloadItemReplyFinished()
     QFile file( tempFilePath );
     if ( file.open( QIODevice::WriteOnly ) )
     {
-      file.write( data );
+      qint64 bytesWritten = file.write( data );
       file.close();
+
+      // maybe problem with permissions or drive space
+      if ( data.size() != bytesWritten )
+      {
+        QString serverMsg = extractServerErrorMsg( data );
+        if ( serverMsg.isEmpty() )
+        {
+          serverMsg = r->errorString();
+        }
+        CoreUtils::log( "pull " + projectFullName, "Failed to write content into: " + file.fileName() );
+
+        transaction.replyPullItem->deleteLater();
+        transaction.replyPullItem = nullptr;
+
+        // get rid of the temporary download dir where we may have left some downloaded files
+        QDir( getTempProjectDir( projectFullName ) ).removeRecursively();
+
+        if ( transaction.firstTimeDownload )
+        {
+          Q_ASSERT( !transaction.projectDir.isEmpty() );
+          QDir( transaction.projectDir ).removeRecursively();
+        }
+
+        int httpCode = r->attribute( QNetworkRequest::HttpStatusCodeAttribute ).toInt();
+        emit networkErrorOccurred( serverMsg, QStringLiteral( "Mergin API error: writeFile" ), httpCode, projectFullName );
+
+        finishProjectSync( projectFullName, false );
+
+        return;
+      }
     }
     else
     {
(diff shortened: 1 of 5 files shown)


core/merginapi.cpp
```diff
@@ -3178,9 +3217,16 @@ bool MerginApi::writeData( const QByteArray &data, const QString &path )
     return false;
```
tomasMizera, line 3217: Add coreutils::log here

core/merginapi.cpp
```diff
@@ -1693,7 +1723,15 @@ void MerginApi::finalizeProjectPullCopy( const QString &projectFullName, const Q
...
       return;
     }
-    f.write( fTmp.readAll() );
+    QByteArray data = fTmp.readAll();
+    qint64 bytesWritten = f.write( data );
+
+    // maybe problem with permissions or drive space
+    if ( data.size() != bytesWritten )
+    {
+      CoreUtils::log( "writeData ", "Failed to write content into: " + fTmp.fileName() );
```
tomasMizera, line 1732:
```suggestion
      CoreUtils::log( "finalizeProjectPullCopy ", "Failed to write content into: " + fTmp.fileName() );
```

core/merginapi.cpp
```diff
@@ -385,8 +385,38 @@ void MerginApi::downloadItemReplyFinished()
...
       file.close();
+
+      // maybe problem with permissions or drive space
+      if ( data.size() != bytesWritten )
+      {
+        QString serverMsg = extractServerErrorMsg( data );
+        if ( serverMsg.isEmpty() )
+        {
+          serverMsg = r->errorString();
+        }
```
tomasMizera, line 398: There is no error from server

core/merginapi.cpp
```diff
@@ -385,8 +385,38 @@ void MerginApi::downloadItemReplyFinished()
...
+
+        if ( transaction.firstTimeDownload )
+        {
+          Q_ASSERT( !transaction.projectDir.isEmpty() );
+          QDir( transaction.projectDir ).removeRecursively();
+        }
+
+        int httpCode = r->attribute( QNetworkRequest::HttpStatusCodeAttribute ).toInt();
+        emit networkErrorOccurred( serverMsg, QStringLiteral( "Mergin API error: writeFile" ), httpCode, projectFullName );
+
```
tomasMizera, line 415: This is not a network error

core/merginapi.cpp
```diff
@@ -385,8 +385,38 @@ void MerginApi::downloadItemReplyFinished()
     QFile file( tempFilePath );
     if ( file.open( QIODevice::WriteOnly ) )
     {
-      file.write( data );
+      qint64 bytesWritten = file.write( data );
       file.close();
+
+      // maybe problem with permissions or drive space
```
tomasMizera, line 391: It would be only not enough space left

core/merginapi.cpp
```diff
@@ -385,8 +385,38 @@ void MerginApi::downloadItemReplyFinished()
...
+        int httpCode = r->attribute( QNetworkRequest::HttpStatusCodeAttribute ).toInt();
+        emit networkErrorOccurred( serverMsg, QStringLiteral( "Mergin API error: writeFile" ), httpCode, projectFullName );
+
+        finishProjectSync( projectFullName, false );
+
+        return;
+      }
     }
     else
     {
```
tomasMizera, line 422: We need the same fix below

core/merginapi.cpp
```diff
@@ -1693,7 +1733,15 @@ void MerginApi::finalizeProjectPullCopy( const QString &projectFullName, const Q
...
     }
-    f.write( fTmp.readAll() );
+    QByteArray data = fTmp.readAll();
+    qint64 bytesWritten = f.write( data );
+
+    // not enough space on file system
+    if ( data.size() != bytesWritten )
+    {
+      CoreUtils::log( "pull " + projectFullName, "Failed to write content into: " + fTmp.fileName() );
+      return;
```
PeterPetrik, line 1743: when we return here
1/ file `f` is not closed
2/ file `f` is still present in the final folder in some random bad shape

When there is problem with sync (any), we need to guarantee that 
1/ all files are in the original/pre-update/previous version 
2/ there are no file leftovers anywhere, both in temporary folders and updated folders. 
3/ we have note in diagnostics logs about the issues

core/merginapi.cpp
```diff
@@ -3175,12 +3224,20 @@ bool MerginApi::writeData( const QByteArray &data, const QString &path )
...
 
-  file.write( data );
+  qint64 bytesWritten = file.write( data );
   file.close();
 
+  // not enough space on file system
+  if ( data.size() != bytesWritten )
+  {
+    CoreUtils::log( "writeData ", "Failed to write content into: " + path );
+    return false;
```
PeterPetrik, line 3238: https://github.com/MerginMaps/input/blob/780622dacf156cad443caa5fa6fd0089c21e2ae9/core/merginapi.cpp#L3178 does not check the return value of this function, so the sync is not failed.

PeterPetrik (commented):
We need to add a bunch of auto-test that simulates a various error conditions during sync (e.g. we can create write-only folder, we can simulate lack of disk space, etc).

tomasMizera:
To be continued by @wonder-sk

PR #2804, opened 2023-09-25
Removed Qt.inputMethod.commit() for tests

iiLubos:
Just for tests, if it is needed to use Qt.inputMethod.commit() method

### DIFF
--- a/app/qml/components/NumberInputField.qml
+++ b/app/qml/components/NumberInputField.qml
@@ -36,8 +36,6 @@ Item {
       }
     }
 
-    onPreeditTextChanged: if ( __androidUtils.isAndroid ) Qt.inputMethod.commit() // to avoid Android's uncommited text
-
     text: root.number
 
     height: parent.height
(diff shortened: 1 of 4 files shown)


tomasMizera:
@jozef-budac / @PatrikMizera, can we please test text and number inputs? We need to test different keyboards as well. The goal is to see if everything is working fine as previously (e.g. form validations)

jozef-budac:
TI @tomasMizera . I have tested on Android 11, 13, and iOS and on iOS keyboard, Gboad, and Samsung keyboard and found nothing problematic. Everything worked ok.

jozef-budac:
I just found one problem with relations. The app freezes when you create a parent and from it the child. 
Scenario:
1) Open project tc05_relations
2) Hit record on the district, add 3 points, and hit done so the form opens
3) Write a name on the form 
4) Hit +Add to add a child. On the map add 3 points and hit Done -> the app freezes

Screen recording:
https://drive.google.com/file/d/1lZa8aZelJV1xNAxHCXUJmB9W-Hr0pyOR/view?usp=sharing

tomasMizera:
This change seems very dangerous atm. We will revisit within redesign

PR #2813, opened 2023-09-27
add missing CRS to project load errors

PeterPetrik:
fix https://github.com/MerginMaps/input/issues/941
fix https://github.com/MerginMaps/input/issues/2362
related to https://github.com/MerginMaps/input/issues/2450
related to CU-862ke6jaw

added also test with one missing layer, one layer with invalid CRS and whole project has invalid CRS to test suite.

Screenshot from iPhone

IMG_BB728692FECB-1

Screencast from macOS (desktop)

https://github.com/MerginMaps/input/assets/804608/291b6f0e-c62f-46f2-af0b-9e625c05905e


test/test_data/project-with-missing-layer-and-invalid-crs/data.csv
tomasMizera, line 1: Do we need this file?

PeterPetrik: Yes, because we do not want report missing CRS for non-spatial layers - so in auto test we check it is not reported ..

jozef-budac:
Verified on 2.4.0 RC (Android: 376111 (2.4.0))

PR #2820, opened 2023-09-29
Speed improvements of home page / sync preparation speed

PeterPetrik:
CU-25cf6h3
CU-861mm3262
CU-862ke6mdb

fix https://github.com/MerginMaps/input/issues/1556
fix https://github.com/MerginMaps/input/issues/2806
fix https://github.com/MerginMaps/input/issues/1570
fix https://github.com/MerginMaps/input/issues/791

1/ Fixed problem with more than 50 projects downloaded from mergin - now the user is asked to remove some local projects instead of getting a completely unreadable error box. Note that internally we limit number of projects to send to mergin's API `listProjectsByName` to 50, so all extra projects simply do not have any valid/recent info from the server (so we do not know if they are up-to-date or not). 

@tomasMizera would be nice if we can get 50 projects from new API (by project UUID) in future so we do not need to change logic here ...

2/ Fixed speed of home screen
 - 2.0 / we still wait (sync) until we receive response from server - this is the majority of the time for waiting, similarly to other 2 screens. 
 - 2.1/ reverted order of checking if we need to sync - now first you check if the local "old" server version is lower than "new" remote version, before checking local files modification. This means that if on server there is new version, there is no computation of local modifications 
- 2.2./ created a new function to get if local files are modified or not - so it returns fast when it finds first modified file. 
- 2.3/ The major bottleneck of the home screen, especially with a lot of files (e.g. photos) is calculation of checksums. We want to get rid of them long-term https://gitlab.lutraconsulting.co.uk/mergin/mergin/-/issues/664 , but now we still need them (@varmar05) in sync protocol. Therefore it does not matter when we calculate them (which is costly for big projects), but more importantly we should not calculate it again and again.  Checksum cache is introduced - for each file path + modification time of file, the calculated checksum is stored in binary file `.mergin/checksum.cache`. 
- 2.4/ Initial load time will be same as before -it is when all files checksums are calculated and the cache file is stored. All subsequent loads will be more-less instant.
- 2.5/ Speed time for projects with massive geopackage file can be still determined by `geodiff` performance. 
 
3/ Improved merginapi sync speed 
 - see 2.3. Calculations of local files modifications is used also for pulls/pushes/etc. Therefore all this will be massively faster since the checksums will be taken from cache.  

See in action on desktop build:

https://github.com/MerginMaps/input/assets/804608/d40f7b26-b64f-4f40-bb33-05183c5bc067

4/ @jozef-budac for testing, it would be good to add to test suite project with a lot of photos (300?)
   - add a new point + photo
   - go to home screen - fast
   - switch between home - projects screen - fast

### DIFF
--- a/core/checksum.cpp
+++ b/core/checksum.cpp
@@ -26,6 +26,7 @@ Checksum::Checksum( const QString &projectDir )
 void Checksum::load()
 {
   mCache.clear();
+  mCacheModified = false;
 
   QFile f( mProjectDir + "/" + sCacheFile );
 
@@ -41,7 +42,6 @@ void Checksum::load()
     while ( stream.atEnd() == false )
     {
       stream >> path >> checksum >> mtime;
-
       entry.checksum = checksum;
       entry.mtime = mtime;
       mCache.insert( path, entry );
@@ -51,9 +51,12 @@ void Checksum::load()
 
 void Checksum::save()
 {
+  if ( !mCacheModified )
+    return;
+
   QFile f( mProjectDir + "/" + sCacheFile );
 
-  if ( f.open( QIODevice::WriteOnly ) )
+  if ( f.open( QIODevice::WriteOnly ) ) // implies Truncate
   {
     QDataStream stream( &f );
     stream.setVersion( QDataStream::Qt_6_5 );
@@ -80,6 +83,7 @@ QString Checksum::get( const QString &path )
     else
     {
       // invalid entry - remove from cache and recalculate
+      mCacheModified = true;
       mCache.remove( path );
     }
   }
@@ -91,6 +95,7 @@ QString Checksum::get( const QString &path )
   entry.checksum = localChecksum;
   entry.mtime = localLastModified;
   mCache.insert( path, entry );
+  mCacheModified = true;
 
   return localChecksum;
 }
--- a/core/checksum.cpp
+++ b/core/checksum.cpp
@@ -1,0 +1,115 @@
+/***************************************************************************
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+
+#include <QFile>
+#include <QCryptographicHash>
+#include <QFileInfo>
+#include <QDataStream>
+
+#include "checksum.h"
+
+static const int CHUNK_SIZE = 65536;
+const QString Checksum::sCacheFile = QStringLiteral( ".checksum.cache" );
+
+Checksum::Checksum( const QString &projectDir )
+  : mProjectDir( projectDir )
+{
+
+}
+
+void Checksum::load()
+{
+  mCache.clear();
+
+  QFile f( mProjectDir + "/" + sCacheFile );
+
+  if ( f.open( QIODevice::ReadOnly ) )
+  {
+    QDataStream stream( &f );
+    stream.setVersion( QDataStream::Qt_6_5 );
+    QString path;
+    QString checksum;
+    QDateTime mtime;
+    CacheValue entry;
+
+    while ( stream.atEnd() == false )
+    {
+      stream >> path >> checksum >> mtime;
+
+      entry.checksum = checksum;
+      entry.mtime = mtime;
+      mCache.insert( path, entry );
+    }
+  }
+}
+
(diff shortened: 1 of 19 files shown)


core/project.h
```diff
@@ -25,9 +25,7 @@ namespace ProjectStatus
   {
     NoVersion,  //!< the project is not downloaded
     UpToDate,   //!< both server and local copy are in sync with no extra modifications
-    OutOfDate,  //!< server has newer version than what is available locally (but the project is not modified locally)
-    Modified    //!< there are some local modifications in the project that need to be pushed (note: also server may have newer version)
-    // Maybe orphaned state in future
+    NeedsSync,  //!< server has newer version than what is available locally and/or the project is not modified locally
```
wonder-sk, line 28:
```suggestion
    NeedsSync,  //!< server has newer version than what is available locally and/or the project is modified locally
```

core/checksum.h
```diff
@@ -0,0 +1,58 @@
...
+     * Returns Sha1 checksum of file (with-caching)
+     * Recalculates checksum for all entries not in cache
+     */
+    QString get( const QString &path );
+
+    /**
+     * Returns Sha1 checksum of file (no-caching)
+     * This is potentially resourcing-costly operation
+     */
+    static QByteArray calculate( const QString &filePath );
```
wonder-sk, line 41: please add notes about relative/absolute paths for get/calculate

core/checksum.cpp
```diff
@@ -0,0 +1,119 @@
...
+{
+
+}
+
+void Checksum::load()
+{
+  mCache.clear();
+  mCacheModified = false;
+
+  QFile f( mProjectDir + "/" + sCacheFile );
```
wonder-sk, line 31: let's put it into .mergin subdir

core/checksum.cpp
```diff
@@ -0,0 +1,119 @@
...
+    for ( auto it = mCache.constBegin(); it != mCache.constEnd(); ++it )
+    {
+      stream << it.key() << it.value().checksum << it.value().mtime;
+    }
+  }
+}
+
+QString Checksum::get( const QString &path )
+{
+  QDateTime localLastModified = QFileInfo( mProjectDir + path ).lastModified();
```
wonder-sk, line 73: dir separator needed?

core/checksum.cpp
```diff
@@ -0,0 +1,119 @@
...
+  if ( match != mCache.end() )
+  {
+    if ( match.value().mtime == localLastModified )
+    {
+      return match.value().checksum;
+    }
+    else
+    {
+      // invalid entry - remove from cache and recalculate
+      mCache.remove( path );
```
wonder-sk, line 86: probably unneeded - replaced afterwards

core/checksum.h
```diff
@@ -0,0 +1,58 @@
...
+ */
+class Checksum
+{
+  public:
+    Checksum( const QString &projectDir );
+
+    //! Loads cache from mProjectDir/sCacheFile
+    void load();
+    //! Saves cache to mProjectDir/sCacheFile
+    void save();
```
wonder-sk, line 29: I would suggest doing load/save in constructor/destructor

core/checksum.h
```diff
@@ -0,0 +1,58 @@
...
+
+#include <QByteArray>
+#include <QString>
+#include <QDateTime>
+#include <QHash>
+
+/**
+ * Calculates the checksums of local files and store the results in the local binary file
+ */
+class Checksum
```
wonder-sk, line 21: would suggest renaming - e.g. ProjectChecksumCache ?

core/checksum.h
wonder-sk, line 41: move calculate to coreutils

core/merginapi.cpp
```diff
@@ -210,6 +212,19 @@ QString MerginApi::listProjectsByName( const QStringList &projectNames )
...
   }
 
+  const int listProjectsByNameApiLimit = 50;
+  QStringList projectNamesToRequest( projectNames );
+
+  if ( projectNamesToRequest.count() > listProjectsByNameApiLimit )
+  {
+    CoreUtils::log( "list projects by name", QStringLiteral( "Too many local projects: " ) + QString::number( projectNames.count(), 'f', 0 ) );
+    const int projectsToRemoveCount = projectNames.count() - listProjectsByNameApiLimit;
+    QString msg = tr( "Reached limit of local projects.\nRemove %1 projects from your device!" ).arg( projectsToRemoveCount );
```
wonder-sk, line 222: let's rephrase...

e.g. Please remove some projects as the app currently only allows up to 50 downloaded projects.

PR #2822, opened 2023-10-02
Revamp mapcanvas gesture handlers & fix propagation issues

### LINKED ISSUE #2549: Map sync button selects feature behind the button
When clicking the sync button in the map window, the feature behind the button gets selected as well. See the attached video.

Android running version 2.0.1.

https://user-images.githubusercontent.com/90187049/217281957-283b3940-8df9-406a-86c4-334ad552ac96.mp4

### LINKED ISSUE #2617: Adding multiple entries into field with value relation widget causes app to ask the user to edit geometry of parent feature
The situation is like this:
I create spatial parent layer, nonspatial child layer with 3 fields and one more nonspatial layer containing 3 fields with values.
I set relation between parent-child layer and for child layer's 3 fields I set their widgets to value relations, where each field corresponds to 1 field in the other nonspatial layer. Multiple selections is allowed.

Strange behavior:
In input when you create feature and within the child layer's form add values from the other table, manipulation with the second field makes the app to show the map window and edit the geometry (the one from parent layer that isn't even saved yet).

This ticket was created due to https://support.lutraconsulting.co.uk/a/tickets/1971 and testing project was also created (support/value_relation_test) and the issue replicated.

### LINKED ISSUE #2683: Tapping edit icon with pen triggers identify on map
I experience a strange issue when I am trying to edit a feature. When I tap the edit icon with my **finger**, it works as expected. When I tap the edit icon with the **pen**, a feature at the map behind the feature information panel is identified.

I have recorded a video that shows the issue.

https://user-images.githubusercontent.com/120651313/235968927-ca59312d-a1fc-453e-abb2-9481e7c817e1.mp4

### LINKED ISSUE #2684: White screen and unresponsive when adding or editing child (android)
We have an issue on mergin (android) that seems to be getting more and more common and renders the app unusable in many instances. In short, we have a point layer with a geometry-less child. The idea being we can re-survey the same point multiple times and add new 'surveys' to it without duplicating the points. This was all working fine until around the release of Mergin 2.0 and the following error has been intermittently present since.
When we click a point and then press the pencil to edit it opens attribute form. We then click to add or edit a child feature in that form and make some changes. When we click 'done' on this child feature it then instead jumps to the edit geomoetry view of the parent. From here, we have the options to 'cancel', 'release' or 'done' and all 3 options lead to a blank white sceen popping up which cannot be interacted with at all and requires a force-closing of the app and reloading. This in turn means the edits to the child are lost.
It seems to me that when clicking 'done' on the child feature, it is clicking through to the underlying 'edit geometry' button on the parent form perhaps?

tomasMizera:
This PR is a complete rewrite of MapCanvas gesture handlers. 
Due to multiple issues in Qt (like QTBUG-108689, QTBUG-108821 and more..), `TapHandlers`, `DragHandlers` and `PinchHandlers` were a pain to use and maintain.
 - they were fighting who should get an exclusive grab (often resulting in no one and frozen UI)
 - taps were propagated to parent scopes even though they should not

The new implementation is minimal and contains only `PinchArea` with nested `MouseArea`. Drags and double-clicks are implemented on our own now. 

**Important: we no longer support tap&drag gesture to zoom in/out the map**. 
If needed, it will not be hard to add it to the current implementation though 

There are still some bits and pieces needed to get rid of the propagation issues (will be a part of this PR):
 - [x] replace `TapHandler`s with `MouseArea`s everywhere (@iiLubos)
 - [x] get rid of the remaining code used as a workaround for the mentioned issues

We need quite a massive testing to make sure the map gestures work properly 😏

Fixes #2549 
Fixes #2683 
Fixes #2684 
Fixes #2617

### DIFF
--- a/app/qml/map/MapCanvas.qml
+++ b/app/qml/map/MapCanvas.qml
@@ -8,110 +8,50 @@
  ***************************************************************************/
 
 /*
- * The source code forked from https://github.com/qgis/QGIS on 25th Nov 2022
+ * Previous implementation was forked from https://github.com/qgis/QGIS on 25th Nov 2022
  * File: qgsquickmapcanvas.qml by (C) 2014 by Matthias Kuhn
  */
- 
+
 import QtQuick
+import QtCore
 import QtQuick.Controls
-import QtQml
 
 import Input 0.1 as Input
 
+//
+// MapCanvas QML object is a utility class for rendering map layers and handling user interaction
+//
+
 Item {
-  id: root
+  id: mapRoot
 
-  /**
-   * The mapSettings property contains configuration for rendering of the map.
-   *
-   * It should be used as a primary source of map settings (and project) for
-   * all other components in the application.
-   *
-   * This is a readonly property.
-   *
-   * See also InputMapCanvasMap::mapSettings
-   */
-  property alias mapSettings: mapCanvasWrapper.mapSettings
+  // Clicked signal is sent 350ms after click (to avoid mixing it with double click)
+  signal clicked( point p )
 
-  /**
-   * The isRendering property is set to true while a rendering job is pending for this map canvas map.
-   * It can be used to show a notification icon about an ongoing rendering job.
-   *
-   * This is a readonly property.
-   *
-   * See also InputMapCanvasMap::mapSettings
-   */
-  property alias isRendering: mapCanvasWrapper.isRendering
+  signal longPressed( point p )
 
-  /**
-   * When the incrementalRendering property is set to true, the automatic refresh of map canvas during rendering is allowed.
-   */
-  property alias incrementalRendering: mapCanvasWrapper.incrementalRendering
-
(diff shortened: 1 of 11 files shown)


app/qml/map/MapCanvas.qml
```diff
@@ -8,110 +8,50 @@
  ***************************************************************************/
 
 /*
- * The source code forked from https://github.com/qgis/QGIS on 25th Nov 2022
+ * Previous implementation was forked from https://github.com/qgis/QGIS on 25th Nov 2022
```
PeterPetrik, line 11: I would argue that if this implementation is complete revamp, we can remove this notice ( or you can name the file MapCanvasV2 and remove the original file )

tomasMizera:
Not ready for testing yet

PeterPetrik (commented):
LGTM

Issue #2675, opened 2023-04-18
IMPR MM: Swipe to other tab in the middle of my screen


JSoepenberg:
Hi, 

I would like to swipe to the other tab (from tree info to other data) on my screen. Not by clicking on the other data tab, but by swiping in the middle of my screen to the right. Could you make that possible? 

![image](https://user-images.githubusercontent.com/73890345/232802586-34c92e10-8711-4448-b596-4ec0a7e3eb09.png)

![image](https://user-images.githubusercontent.com/73890345/232800964-c31b0db3-feef-4ba6-9aa5-6a74b4aed31f.png)

tomasMizera:
Hi @JSoepenberg, thanks for your feedback! 
This used to be possible, but we needed to turn it off due to a bug in one of our core dependencies (Qt). See more here https://github.com/MerginMaps/input/issues/2473

We will enable it again once the bug is fixed :)

tomasMizera:
Hi @JSoepenberg, 2.4.0 version (upcoming one) will have the swipe gesture enabled again :)

PR #2828, opened 2023-10-05
Enable swipe gesture in forms to switch between tabs

### LINKED ISSUE #2473: QT6: tab sliding doesn't work
When you have tabs in the form, previously you could swipe left and right through it. Right now it is not possible and you have to use clicking to tabs

tomasMizera:
We disabled this gesture previously because of a QT bug, but it seems to be fixed now.

The issue was that previously `SwipeView` used to steal grab from `Slider`, see video in the related QT bug: https://bugreports.qt.io/browse/QTBUG-109124

Note: Even though the related QT bug is not closed, this seems to be working

https://github.com/MerginMaps/input/assets/22449698/77daf939-13ff-461a-bc4c-54e47676c120

Fixes #2473 
Related #2675

### DIFF
--- a/app/qml/form/FeatureForm.qml
+++ b/app/qml/form/FeatureForm.qml
@@ -276,12 +276,6 @@ Item {
       id: swipeView
       currentIndex: form.controller.hasTabs ? tabRow.currentIndex : 0
 
-      //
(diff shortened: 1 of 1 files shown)


PeterPetrik (approved):

jozef-budac:
Verified on iOS: 23.10.464711

PR #2834, opened 2023-10-05
Use new storage permissions on Android >13

tomasMizera:
Android introduced a new set of storage permission types. 
From API level 33 (Android version 13.0), we can not request `READ_EXTERNAL_STORAGE` to get photos from the gallery, but `READ_MEDIA_IMAGES` instead.

Fixes the issue that user can not pick image from gallery on android >13 (CU-862keem0n)

See https://developer.android.com/reference/android/Manifest.permission#READ_EXTERNAL_STORAGE

### DIFF
--- a/app/androidutils.cpp
+++ b/app/androidutils.cpp
@@ -157,7 +157,19 @@ void AndroidUtils::quitApp()
 bool AndroidUtils::requestStoragePermission()
 {
 #ifdef ANDROID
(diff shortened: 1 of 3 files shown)


PeterPetrik (approved):

jozef-budac:
Verified on Android 13, that gallery can be used again.

PR #2835, opened 2023-10-05
Remove subscription pages for iOS and call-to-actions

tomasMizera:
The iOS account page is not unified with Android. There is no subscription button. 
Once a user hits some limit (storage/projects count) they can click on "manage account" which will redirect them to the dashboard.

Unused code is removed

Screenshot 2023-10-05 at 14 00 55


app/inputhelp.cpp
```diff
@@ -63,12 +63,40 @@ QString InputHelp::merginWebLink() const
...
+  // Let's include UTM tags for production server
+  QString queryParams;
+
+  if ( !activeWorkspacePathPart.isEmpty() )
+  {
+    queryParams = activeWorkspacePathPart;
+
+    // URL can not have two question marks, let's remove the one in the utm tag
+    QString utms( utmTagAttention );
+    utms.replace( "?", "" );
```
PeterPetrik, line 91: shouldn't it be replaced by `&` ?

jozef-budac:
Verified on iOS: 23.10.464711

PR #2849, opened 2023-10-11
Remove of purchasing backend & fix tests

PeterPetrik:
- RIP in-app subscriptions
- also fix merginApi tests (use /v2/projects/uuid DEL for deletion)


.github/workflows/linux.yml
```diff
@@ -149,6 +149,8 @@ jobs:
       - name: run tests
         env:
           TEST_MERGIN_URL: https://test.dev.merginmaps.com/
+          TEST_API_USERNAME: test_mobileapp
+          TEST_API_PASSWORD: ${{ secrets.TEST_API_PASSWORD }}
```
tomasMizera, line 153: Why do we need to introduce this back?

tomasMizera: If we don't do that, tests should work just fine

PeterPetrik: before: we generated new user for each test run and assigned this user pro plan
now: we use the fixed user to run test (different to lnx and macos) and this user has assigned education/team plan. We create new user only in one or two tests where we test registration

PR #2854, opened 2023-10-12
revert symbol size

### LINKED ISSUE #2853: Simple Symbol marker size is half size
https://github.com/MerginMaps/input/pull/2823 introduced wrong size of symbols:

- QGIS project with simple symbol - size 20 mm

iOS version from App Store (v.2.3.0)
iPhone-release before

Release v2.4.0 RC
(QGIS on macos - external monitor; windows & ubuntu - primary monitor)

symbol-size

PeterPetrik:
fix #2853

### DIFF
--- a/app/map/inputmapcanvasmap.cpp
+++ b/app/map/inputmapcanvasmap.cpp
@@ -268,7 +268,7 @@ void InputMapCanvasMap::onScreenChanged( QScreen *screen )
     {
       mMapSettings->setDevicePixelRatio( screen->devicePixelRatio() );
     }
-    mMapSettings->setOutputDpi( screen->logicalDotsPerInch() );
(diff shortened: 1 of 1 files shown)


PeterPetrik:
@jozef-budac for test, you can create a simple project (or reuse some existing project) and set the simple symbol size to let say 20mm 

Screenshot 2023-10-12 at 10 04 12

and then when you load the project on Android and IOS devices the size of symbol should be around this number (let say in range 18-22mm ).

optional bonus points: also have a line feature and measure line width (again in QGIS setting for millimetres)

Would it be possible to add this test to our manual regression testing suite? thanks!

jozef-budac:
Verified on iOS and test was added: https://app.qase.io/case/INPUT-103

PR #2860, opened 2023-10-16
fix photo in relation preview 

### LINKED ISSUE #2058: Photo preview in parent relation feature is broken when photo field is not selected as layer display field
When viewing the form of a parent feature with a 1:N relation with a child layer for associating multiple photograph attachments with the parent feature, Input is only able to display the image preview when the attribute index of the field containing the attachment widget/file path on the child feature is in attribute index 1. If the attribute index for the file path is anything other than 1, the parent feature displays a broken image link icon instead of the photograph. Relationship associations all work properly and the image preview will work correctly in the child feature's form, the issue exists specifically on the parent feature form.

PeterPetrik:
fix #2058
fix https://github.com/MerginMaps/input/issues/2586

project `support/BugZapper2058` or any project where photo photo field is not set as layer display field (usually when user does does not have photo widget as second in the form)

Screenshot 2023-10-16 at 15 07 48

### DIFF
--- a/app/relationfeaturesmodel.cpp
+++ b/app/relationfeaturesmodel.cpp
@@ -112,12 +112,12 @@ QgsRelation RelationFeaturesModel::relation() const
 
 QVariant RelationFeaturesModel::relationPhotoPath( const FeatureLayerPair &featurePair ) const
 {
-  // Feature title used to get path of the referenced image.
-  QString path = featureTitle( featurePair ).toString();
-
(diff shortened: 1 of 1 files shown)


PeterPetrik:
unrelated test failure

tomasMizera (approved):

PeterPetrik:
@jozef-budac

jozef-budac:
Verified on the APK from the build

PR #2861, opened 2023-10-17
 conditional visibility of the relation widget

### LINKED ISSUE #2098: Conditional visibility is not applied to a group containing a 1-to-many relation
When using drag-n-drop form and having a group which contains a 1-to-many relation, Input ignores the conditional visibility and displays it all the time:
image

You can see this project for the issue:
AndyWalsh/WMBaRv2forLutra

### LINKED ISSUE #2712: Control Visibility by Expression not working in Combination with a relations widget
Hi, 

I found a bug when applying a 'Control Visibility by Expression' on a groupbox in a container in combination with a relations widget. So the control visibility has the expression column = true. Within the group box, there is a relation widget to a form. The relations widget is always visible, no matter the expression (true or false). Other parent columns, columns that are in the same table as the expression, will appear and disappear by the expression.

The bug only resides in the app. In QGIS the expression works as it should, the relation widget appears and disappears with the expression.

PeterPetrik:
fix #2712
fix #2098

before/released v2.3.0:
conditional_visibility-bad

after this fix:

https://github.com/MerginMaps/input/assets/804608/d48142a3-2d2b-4a3b-9838-267ce7628c0c

### DIFF
--- a/app/attributes/attributecontroller.cpp
+++ b/app/attributes/attributecontroller.cpp
@@ -302,6 +302,7 @@ void AttributeController::flatten(
               parentTabRow,
               FormItem::Relation,
               label,
+              parentVisibilityExpressions, // relation field doesn't have visibility expression itself
               associatedRelation
             )
           );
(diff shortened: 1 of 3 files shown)


jozef-budac:
Verified on the APK from the build

tomasMizera (approved):

PR #2864, opened 2023-10-19
Fix mapcanvas UX issues

tomasMizera:
**Update, ready for review**

These things are fixed/added:
 - The formula for calculating vector distance was missing `+` 🤦🏻‍♂️ 😃  , resulting in wrong distances calculated when you dragged the map vertically 
 - Added distance threshold for double clicks - now the second click must be within a _threshold_ from the first point.
 - Added time elapsed threshold for detecting drag movement - sometimes, if you move a map just by a tiny bit to fit the point, you might not exceed the distance threshold to detect the movement as a drag. It would be detected as a simple tap. This caused unexpected actions while recording new/updating existing geometries. Now to detect the drag gesture we use a combination of a distance threshold and a time threshold (_"how long the tap lasted"_) 
 - Thresholds are now read from Qt built-in `Application.styleHints.xyz`

---

 - [x] Added distance threshold for double clicks
 - [x] Make the `drag.threshold` smaller when digitizing -> when a user is digitizing, they need to pan the map by small distances that sometimes do not meet the `drag.threshold`. We need to find a better threshold and test it on all platforms.

### DIFF
--- a/app/qml/map/MMMapCanvas.qml
+++ b/app/qml/map/MMMapCanvas.qml
@@ -31,9 +31,6 @@ Item {
   property alias mapSettings: mapRenderer.mapSettings
   property alias isRendering: mapRenderer.isRendering
 
-  // Number of miliseconds to differentiate between normal click (select feature) and double-click (zoom)
-  property int doubleClickThresholdMilis: 350
-
   // Requests map redraw
   function refresh() {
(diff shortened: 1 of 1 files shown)


PeterPetrik (approved):
Tested iOS and it works nicely

PR #2867, opened 2023-10-20
HTML, Text, Spacer "other form widgets" + "show label"

PeterPetrik:
fix https://github.com/MerginMaps/input/issues/2643

Added:
- HTML widget (supporting https://doc.qt.io/qt-6/richtext-html-subset.html#supported-tags)
- Text widget
- Spacer widget
- (all widgets) allow "Show label" (so it shows/hides field name in format)

Screenshot 2023-10-20 at 15 54 47

https://github.com/MerginMaps/input/assets/804608/8c640301-0323-4ed0-89b2-6f1a17440225

TODO
- [x] make hline color gray and space similar to one in QGIS
- [x] fix when HTML/Text uses expressions 
- [x] tests

note for HTML we use QML Text, if we change rendering engine for preview-panel (https://github.com/MerginMaps/input/issues/1259) we should use it here.


PeterPetrik:
unrelated test failure

app/qml/editor/inputrichtext.qml
```diff
@@ -0,0 +1,54 @@
+/***************************************************************************
+ checkbox.qml
```
tomasMizera, line 2: checkbox?

app/attributes/attributedata.cpp
```diff
@@ -16,25 +16,26 @@
...
-  const int parentTabId,
-  FormItem::FormItemType type,
-  const QString &name,
-  bool isEditable,
-  const QgsEditorWidgetSetup &editorWidgetSetup,
-  int fieldIndex,
-  const QgsExpression &visibilityExpression,
-  const QgsRelation &relation
-)
+FormItem::FormItem( const QUuid &id,
```
tomasMizera, line 19: It would be nice for the future to have inheritance here instead of different constructors. Like `BaseFormItem` with common fields inherited by `RelationFormItem`, `FieldFormItem`, `SpacerFormItem`,...

app/qml/editor/inputrichtext.qml
tomasMizera, line 1: You can derive the field from `AbstractEditor.qml` so you do not need to do all the styling 🙂

app/qml/editor/inputspacer.qml
```diff
@@ -0,0 +1,30 @@
+/***************************************************************************
+ checkbox.qml
```
tomasMizera, line 2: checkbox?

app/qml/editor/inputspacer.qml
```diff
@@ -0,0 +1,30 @@
...
+import "../components"
+
+Rectangle {
+  id: spacer
+
+  /*required*/ property var config: parent.config
+
+  property bool isHLine: config["IsHLine"]
+
+  height: 1 // parent form's list inserts space between each 2 elements
```
tomasMizera, line 23:
```suggestion
  height: 1 * __dp? // parent form's list inserts space between each 2 elements
```

tomasMizera: Well, actually
```suggestion
  height: 1 * __dp < 1 ? 1 : 1 * __dp
```

app/qml/editor/inputrichtext.qml
tomasMizera: See e.g. `inputdatetime.qml`

app/qml/form/FeatureForm.qml
```diff
@@ -434,6 +434,8 @@ Item {
             id: fieldLabel
 
             text: Name
+            height: visible ? paintedHeight : 0
```
tomasMizera, line 437: Is it necessary to set the height to 0?

PeterPetrik: I think yes, on my iOS it was taking (white) place

app/attributes/attributedata.cpp
PeterPetrik: Would be nice for future, agree

tomasMizera (approved):

jozef-budac:
Verified on Android, that HTML, Spacer, and text widget work with expressions. Interestingly, expressions for these widgets don't work in QGIS 3.32.

PR #2880, opened 2023-10-27
Photo crash on project switch

### LINKED ISSUE #2879: Crash when switching between projects with photo attachments
1. You need two project with photo widget that uses expression (https://merginmaps.com/docs/layer/settingup_forms_photo/, e.g. just `@project_folder`)
2. Open first one - capture one image
3. Open second project that also has image widget with expression 
4. 🦀 💥 

https://github.com/MerginMaps/input/assets/804608/382d5282-2bd5-4614-bf85-0e2692fb8320

In google play console it shows tracebacks like:

```
QgsExpressionContextUtils::layerScope(QgsMapLayer const*)
InputUtils::evaluateExpression(FeatureLayerPair const&, QgsProject*, QString const&)+200) 
InputUtils::resolveTargetDir(Q...)
```
name of event on GPC:
```
[libInput_arm64-v8a.so] QgsCoordinateReferenceSystem::authid() const
[libInput_arm64-v8a.so] QgsObjectCustomProperties::value(QString const&, QVariant const&) const
[libQt6Core_arm64-v8a.so] QObject::metaObject() const
```

PeterPetrik:
fix #2879

When testing on test simple QML project and our implementation StackView, it looks like 
when `stackView.clear()` is called: 
- `StackView.onRemoved` is called first
- Some signals are passed and evaluated (e.g. QGIS project is loaded, layer in featurelayerpair contains dangling pointer and in evaluateExpression it crashes)
- Only after then   `Component.onDestruction` is called

Therefore we need to remove all connections to layer/project/etc. we are about to unload from all existing forms (there should be only one in the stack) before calling `clear()`

Other chances in the C++ code are probably because we are passing FeatureLayerPair with `nullptr` layer to the attribute controller, so we should make sure it does not crash by dereferencing null layer pointer.

Sample QML code
```
import QtQuick
import QtQuick.Controls
import QtQuick.Window

Window {
  width: 640
  height: 480
  visible: true
  title: qsTr("Hello World")

  Row {
    anchors.fill: parent
    spacing: 2

  Button {
    text: "add"
    onClicked: stackView.push(formComponent)
  }

  Button {
    text: "clean"
    onClicked: stackView.clear()
  }

  StackView {

    height: 100
    width: 100

    id: stackView

    focus: true

    // anchors.fill: parent
  }

  Component {
    id: formComponent

    Rectangle {
      id: wrapper

      height:100
      width: 100
      color: "red"

      StackView.onRemoved: {
        console.log("*****StackView.onRemoved")
      }

      Component.onCompleted: {
        console.log("*****Created form")
      }

      Component.onDestruction: {
        console.log("*****Destroying form")
      }
    }
  }

  }


}
```

### DIFF
--- a/app/qml/form/FormWrapper.qml
+++ b/app/qml/form/FormWrapper.qml
@@ -34,7 +34,7 @@ Item {
   property real previewHeight
   property real panelHeight
 
-  property bool isReadOnly: featureLayerPair ? featureLayerPair.layer.readOnly : false
+  property bool isReadOnly: featureLayerPair && featureLayerPair.layer ? featureLayerPair.layer.readOnly : false
 
   signal closed()
   signal editGeometry( var pair )
--- a/app/qml/form/FormWrapper.qml
+++ b/app/qml/form/FormWrapper.qml
@@ -34,7 +34,7 @@ Item {
   property real previewHeight
   property real panelHeight
 
-  property bool isReadOnly: featureLayerPair && featureLayerPair.layer ? featureLayerPair.layer.readOnly : false
+  property bool isReadOnly: featureLayerPair?.layer?.readOnly ?? false
 
   signal closed()
   signal editGeometry( var pair )
--- a/app/attributes/attributecontroller.cpp
+++ b/app/attributes/attributecontroller.cpp
@@ -359,58 +359,58 @@ void AttributeController::clearAll()
 void AttributeController::updateOnLayerChange()
 {
   clearAll();
+  QgsVectorLayer *layer = mFeatureLayerPair.layer();
+  if ( !layer )
+    return;
 
   // 1) DATA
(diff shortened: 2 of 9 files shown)


jozefbudac:
Verified on builded APK on Android OS

app/qml/form/FormWrapper.qml
```diff
@@ -34,7 +34,7 @@ Item {
   property real previewHeight
   property real panelHeight
 
-  property bool isReadOnly: featureLayerPair ? featureLayerPair.layer.readOnly : false
+  property bool isReadOnly: featureLayerPair && featureLayerPair.layer ? featureLayerPair.layer.readOnly : false
```
tomasMizera, line 37: Can be simplified as

```suggestion
  property bool isReadOnly: featureLayerPair?.layer?.readOnly ?? false
```

Issue #2871, opened 2023-10-23
Console error in features list


tomasMizera:
How to reproduce:
 - download and open `tc05_form_widgets`
 - open `layers` view
 - open `POI` layer
 - you will see the error

```
qrc:/com.merginmaps/imports/MMInput/layers/FeaturesListPageV2.qml:92: TypeError: Property 'replace' of object 8 is not a function
qrc:/com.merginmaps/imports/MMInput/layers/FeaturesListPageV2.qml:92: TypeError: Type error
qrc:/com.merginmaps/imports/MMInput/layers/FeaturesListPageV2.qml:92: TypeError: Type error
```

iiLubos:
iOS - version 23.10.476311 
Android - https://github.com/MerginMaps/input/actions/runs/6705390934?pr=2884

PR #2884, opened 2023-10-31
Fixed feature list - used replace function always for a string

iiLubos:
The problem was when the replace function was called from a number and not from a string
Fixes: #2871

### DIFF
--- a/app/qml/layers/FeaturesListPageV2.qml
+++ b/app/qml/layers/FeaturesListPageV2.qml
@@ -89,7 +89,7 @@ Page {
 
           color: InputStyle.fontColor
           font.bold: true
-          text: model.display.replace(/\n/g, ' ')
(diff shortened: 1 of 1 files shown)


app/qml/layers/FeaturesListPageV2.qml
```diff
@@ -89,7 +89,7 @@ Page {
 
           color: InputStyle.fontColor
           font.bold: true
-          text: model.display.replace(/\n/g, ' ')
+          text: model.display.toString().replace(/\n/g, ' ')
```
tomasMizera, line 92:
```suggestion
          text: model.display?.toString().replace(/\n/g, ' ') ?? ''
```

tomasMizera: In case `model.display` is undefined

tomasMizera (approved):

PR #2894, opened 2023-11-01
Fix download of files >= 2GB

### LINKED ISSUE #2808: Data corruption when there is a file with large size
I have created a project with a large file (~ 3 GB) (see Version 2)
https://app.merginmaps.com/projects/saber/problem_one_large_file/history

When I tried to download the project on the app, it shows it is all downloaded (strangely it took less than a second for a 3 GB file to be downloaded). Then I added a point and synced.

You can see file1.txt which was 3 GB is now changed to 0 MB in Version 3
https://app.merginmaps.com/projects/saber/problem_one_large_file/history/v3

wonder-sk:
Fixes #2808

- JSON parsing: use toInteger() instead of toInt() for big numbers (returns qint64 instead of int)
- store sizes and file offsets as qint64 instead of int

### DIFF
--- a/core/merginapi.cpp
+++ b/core/merginapi.cpp
@@ -2421,10 +2421,10 @@ void MerginApi::requestServerConfig( const QString &projectFullName )
 QList<DownloadQueueItem> MerginApi::itemsForFileChunks( const MerginFile &file, int version )
 {
   QList<DownloadQueueItem> lst;
-  int from = 0;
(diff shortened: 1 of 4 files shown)


PeterPetrik (approved):

jozef-budac:
@PeterPetrik this looks OK, I was able to download whole 3GB of the project. It took 40 min or so :)

PR #2895, opened 2023-11-01
WIP: Some investigations with Qt 6.5.3 [DO NOT MERGE]

PeterPetrik:
Some investigations with Qt 6.5.3 on iOS

https://github.com/MerginMaps/input/issues/2893
- `Repeater` misbehaves, when you filter in Q sort model, the change of rowCount() is not applied to repeater, so it will have some zombie elements. Interestigly `beginModelReset/endModelReset` fixes it, or also `invalidate()` calls
(fixed in this PR atm)

https://github.com/MerginMaps/input/issues/2892
- this is `ScollView -> ColumnLayout`, scrolling is possible but always pan back to top - not sure why
- it works if I remove `Repeater` and add there only simple elements 
- it looks `Repeater` may have 0 `implicitHeight`
(NOT fixed in this PR atm)
 
https://github.com/MerginMaps/input/issues/2887
- it looks like `ScrollView -> Column` is not possible to scroll on iOS, replacing with ColumnLayout enables scrolling
(somewhat fixed in this PR atm)
- we have also this is the project errors dialog and maybe somewhere else too

### DIFF
--- a/app/qml/PanelItem.qml
+++ b/app/qml/PanelItem.qml
@@ -12,8 +12,8 @@ import "."  // import InputStyle singleton
 
 Rectangle {
     id: root
-    height: InputStyle.rowHeight
-    width: parent.width
+    // height: InputStyle.rowHeight
+    // width: parent.width
     color: InputStyle.clrPanelMain
 
     property string text: ""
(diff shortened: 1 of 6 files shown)


PeterPetrik:
Reverting to qt652, closing this for now

PR #2917, opened 2023-11-15
[Qt66] qr code from QML + zxing

PeterPetrik:
remove workaround introduced in https://github.com/MerginMaps/mobile/pull/2380/files#diff-7c7c5aa42261561514066a886da0dea82b786582bb258a153349042951e8b889 because of bugs in QT framework related to QML Camera on Android/iOS

fix https://github.com/MerginMaps/mobile/issues/2669
fix https://github.com/MerginMaps/mobile/issues/2604


app/inpututils.cpp
```diff
@@ -544,7 +546,19 @@ bool InputUtils::acquireCameraPermission()
   {
     return mAndroidUtils->requestCameraPermission();
   }
-  return true;
+  else
+  {
+    auto status = qApp->checkPermission( QCameraPermission{} );
+
+    if ( Qt::PermissionStatus::Undetermined == status )
```
PeterPetrik, line 553: This will be handled by @tomasMizera in follow up PR properly (along with location permissions)

wonder-sk (approved):

PatrikMizera:
QR code reader is fully working on Android and iOS

PR #2930, opened 2023-11-30
Do not work with BluetoothPermission API on desktop builds

tomasMizera:
BluetoothPermission API on macOS is quite limited - requires `NSBluetoothAlwaysUsageDescription` in `Info.plist` file. We do have the key there, but with suffix `-macos` to indicate we do not use it on iOS devices.

This seems to be ignored though. The fix is to ignore the permission check on platforms other than Android.


tomasMizera:
I will merge this @PeterPetrik so that I can produce RC

PR #2932, opened 2023-12-01
Remove GUI handling old CE servers without workspaces 

PeterPetrik:
removed 
 - old account page (pre-workspace) from QML
 - old project page (pre-workspace) from QML (and associated C++ enums)
 - disabled registration for pre-workspace servers
 - fixed tests


app/projectsmodel.cpp
```diff
@@ -544,17 +543,6 @@ void ProjectsModel::onProjectAttachedToMergin( const QString & )
   listProjectsByName();
 }
 
-void ProjectsModel::onAuthChanged()
-{
-  if ( !mBackend->userAuth() || !mBackend->userAuth()->hasAuthData() ) // user logged out, clear created and shared lists
-  {
-    if ( mModelType == CreatedProjectsModel || mModelType == SharedProjectsModel )
-    {
-      clearProjects();
```
tomasMizera, line 553: Don't we need to do this for the `WorkspaceProjectsModel`?

PeterPetrik: I do not know.... if yes, it is bug in existing code?

PeterPetrik:
linux error is not connected - it is some problem with diff

tomasMizera (approved):
Looking good 💪

PR #2940, opened 2023-12-04
New design for Photo Gallery row

iiLubos:
https://github.com/MerginMaps/mobile/assets/72412563/efd1a05e-7a50-47f2-859e-5509bdce9ffd

```
MMPhotoGallery {
  title: "Gallery 3 photos"
  maxVisiblePhotos: 5
  warningMsg: "The size of image is too big"
  model: [
    "https://images.pexels.com/photos/615348/forest-fog-sunny-nature-615348.jpeg",
    "https://images.pexels.com/photos/955656/pexels-photo-955656.jpeg",
    "https://images.pexels.com/photos/559422/pexels-photo-559422.jpeg"
  ]
  onShowAll: console.log("Open Gallery")
  onClicked: function( path ) { console.log("Open " + path) }
}
```


app/qml/components/MMMorePhoto.qml
```diff
@@ -0,0 +1,105 @@
...
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+
+import QtQuick
+import QtQuick.Controls
+import Qt5Compat.GraphicalEffects
+import "."
```
tomasMizera, line 13: Given that style is now read from`__style` we do not need to `import "."`, right?

app/qml/components/MMPhoto.qml
```diff
@@ -0,0 +1,59 @@
...
+import Qt5Compat.GraphicalEffects
+import "."
+
+Image {
+  id: control
+
+  signal clicked( var path )
+
+  height: width
+  source: model.modelData
```
tomasMizera, line 21: If this component should only be used from the "MMPhotoGallery", please state it on top in comment. If you want to have it reusable, we can not reference model like this, but use properties instead

tomasMizera (approved):
Thanks! It will need some adjustments when using it in the feature form, but I would leave those for later. I missing piece is that there is no add button. Please see design system:

image

Let's add it in a different PR though

PR #2961, opened 2023-12-13
Onboarding adjusted for tablets

iiLubos:
Finished all onboard screens, also for tablets.
There are a few known visual issues, but they are not important for the implementation of the Input app.


PeterPetrik (approved):
LGTM

PR #2966, opened 2023-12-18
Replace login and registration panels

PeterPetrik:
Replace login and registration panels
- it is possible to sign in from new dialog


app/qml/onboarding/MMLogin.qml
```diff
@@ -19,12 +19,15 @@ Page {
...
   signal backClicked
-  signal signInClicked
+  signal signInClicked ( string username, string password )
   signal signUpClicked
-  signal changeServerClicked
+  signal changeServerClicked ( string newServer )
   signal forgotPasswordClicked
 
+  property string warningMsg //TODO
+  property bool canSignUp: true //TODO
```
tomasMizera, line 28: For EE servers there is a flag `user_self_registration` to determine if we should show or hide registration option. We have ignored it previously but might be a good time now to act according to it

app/qml/onboarding/MMLogin.qml
tomasMizera, line 1: @iiLubos @PeterPetrik Have you guys tested this page out on mobile devices? When user clicks on some input, the app should automatically scroll so that the input is not hidden behind the opened keyboard. This is automatically handled by Qt in case we follow their rules

app/qml/onboarding/MMLogin.qml
PeterPetrik: lets postpone it to next releases https://github.com/MerginMaps/mobile/issues/2969

app/qml/onboarding/MMLogin.qml
PeterPetrik: not tested on mobile so far

tomasMizera (approved):

PR #3013, opened 2024-01-16
New Position marker

iiLubos:
image

```
  MMPositionMarker {
    xPos: 200
    yPos: 100
    direction: 330
    screenAccuracy: 50
    trackingMode: true
  }
```

### DIFF
--- a/app/qml/map/MMPositionMarker.qml
+++ b/app/qml/map/MMPositionMarker.qml
@@ -1,0 +1,90 @@
+/***************************************************************************
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+
+import QtQuick
+import lc 1.0
+
+Item {
+  id: control
+
+  required property real xPos
+  required property real yPos
+  property real direction: 0
+  property int size: 24 * __dp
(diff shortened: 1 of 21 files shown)


app/qml/map/MMPositionMarker.qml
```diff
@@ -0,0 +1,90 @@
...
+    id: accuracyIndicator
+
+    x: control.xPos - width / 2
+    y: control.yPos - height / 2
+    width: control.screenAccuracy
+    height: width
+    visible: withAccuracy &&
+             control.hasPosition &&
+             (control.horizontalAccuracy > 0) &&
+             (accuracyIndicator.width > control.size / 2.0)
```
tomasMizera, line 38: Is this condition still valid? Sizes might be different now

tomasMizera (commented):
Good work! Just some minor comments and we are good to go :)

app/qml/map/MMPositionMarker.qml
```diff
@@ -8,35 +8,32 @@
...
-    y: control.yPos - height / 2
-    width: control.screenAccuracy
+    x: root.xPos - width / 2
+    y: root.yPos - height / 2
+    width: root.accuracyRingSize
     height: width
-    visible: withAccuracy &&
-             control.hasPosition &&
-             (control.horizontalAccuracy > 0) &&
-             (accuracyIndicator.width > control.size / 2.0)
```
tomasMizera: Why did we remove all of these lines? I though we would remove just the `root.hasPosition` part.

tomasMizera (approved):

PR #3014, opened 2024-01-17
New map blur label

iiLubos:
image

```
  MMMapBlurLabel {
    y: 20
    width: parent.width
    sourceItem: map
    text: "Mark the geometry on the map and click record"
  }
```

### DIFF
--- a/app/qml/components/MMMapBlurLabel.qml
+++ b/app/qml/components/MMMapBlurLabel.qml
@@ -1,0 +1,58 @@
+/***************************************************************************
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+
+import QtQuick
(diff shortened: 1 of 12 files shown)


gallery/qml/pages/MapPage.qml
```diff
@@ -12,6 +12,7 @@ import QtQuick.Controls
 import QtQuick.Controls.Basic
 import QtLocation
 import QtPositioning
+import Qt5Compat.GraphicalEffects
```
tomasMizera, line 15: This import should be in `MMMapBlurLabel.qml` instead

app/qml/components/MMMapBlurLabel.qml
```diff
@@ -0,0 +1,58 @@
...
+    anchors.fill: fastBlur
+    color: __style.forestColor
+    opacity: 0.8
+    radius: 21
+  }
+
+  FastBlur {
+    id: fastBlur
+
+    x: 20 * __dp
```
tomasMizera, line 30: It will be tricky to use it this way. I would say let's start from `0` and let the item that creates this object set margins instead (like with other map elements).

Is there any reason to set it up here?

tomasMizera (commented):
Great job! :)

tomasMizera (approved):

PR #3015, opened 2024-01-17
Notifications position

iiLubos:
Notifications are lined up on top of each other now. The newest is on the top. Visible is only the last notification.

### DIFF
--- a/app/qml/components/MMNotification.qml
+++ b/app/qml/components/MMNotification.qml
@@ -11,9 +11,8 @@ import QtQuick
 import notificationType 1.0
 
 Rectangle {
-  id: notification
+  id: root
 
-  width: listView.width - 40 * __dp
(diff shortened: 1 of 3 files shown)


app/qml/components/MMNotification.qml
```diff
@@ -30,6 +29,8 @@ Rectangle {
     }
   }
 
+  Component.onCompleted: scale = 0.0
```
tomasMizera, line 32: Why is this important?

tomasMizera: I'd prefer not using `scale` altogether

iiLubos: This is an initial value for the animation when notification starts

tomasMizera (approved):

PR #3022, opened 2024-01-22
Map scale bar

iiLubos:
image

```
  MMMapScaleBar {
    sourceItem: map
  }
```

### DIFF
--- a/app/qml/map/MMMapScaleBar.qml
+++ b/app/qml/map/MMMapScaleBar.qml
@@ -1,0 +1,94 @@
+/***************************************************************************
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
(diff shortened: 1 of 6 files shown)


app/qml/map/MMMapScaleBar.qml
```diff
@@ -0,0 +1,94 @@
...
+        anchors.horizontalCenter: parent.horizontalCenter
+
+        text: root.text
+        color: __style.forestColor
+        font: __style.t3
+        horizontalAlignment: Text.AlignHCenter
+        verticalAlignment: Text.AlignVCenter
+      }
+    }
+  }
```
tomasMizera, line 93: You are missing the opacity number animation, see https://github.com/MerginMaps/mobile/blob/master/app/qml/map/ScaleBar.qml#L86 :)

tomasMizera (changes requested):
Looking good, but let's add the animation :)

PeterPetrik (approved):

PR #3026, opened 2024-01-23
Add Image item in Photo Gallery.

iiLubos:
image
 
```
MMPhotoGallery {
  showAddImage: true
  onAddImage: console.log("Open camera")
  ...
}
```


app/icons/AddImage.svg
tomasMizera, line 1: Shouldn't this be white icon?

app/mmstyle.h
```diff
@@ -54,7 +54,7 @@ class MMStyle: public QObject
     // Colors - secondary palette
     Q_PROPERTY( QColor lightGreenColor READ lightGreenColor CONSTANT )
     Q_PROPERTY( QColor mediumGreenColor READ mediumGreenColor CONSTANT )
-    Q_PROPERTY( QColor greyColor READ greyColor CONSTANT )
```
tomasMizera, line 57: Heh, I changed it to `grey` in the previous PR. Looks like a war between US and UK english https://thisonevsthatone.com/grey-vs-gray/ :D

tomasMizera: I'd prefer to stick to `grey` though, it is how the design system names it. It makes it easier then to apply it

app/qml/components/MMPhotoGallery.qml
```diff
@@ -21,9 +21,11 @@ Item {
   property string warningMsg
   property string errorMsg
   property int maxVisiblePhotos: 5 // -1 for showing all photos
```
tomasMizera, line 23: Can we make `-1` default here please?

app/icons/AddImage.svg
iiLubos: All icons should be white. We can change all icons in one separate PR later

tomasMizera (approved):

PR #3028, opened 2024-01-24
Photo attachment

iiLubos:
image

```
        MMPhotoEditor {
          title: "MMPhotoEditor"
          width: parent.width
          photoUrl: "https://images.pexels.com/photos/615348/forest-fog-sunny-nature-615348.jpeg"

          onTrashClicked: console.log("Move to trash")
          onContentClicked: console.log("Open photo")
        }
```


app/icons/AddImage.svg
tomasMizera, line 1: We probably want white icon, right?

iiLubos: Yes, everywhere :)
Maybe we will change all icons in separate PR

PeterPetrik:
@iiLubos needs to be rebased

app/qml/components/MMIcon.qml
```diff
@@ -23,7 +24,8 @@ Item {
     id: icon
 
     source: __style.arrowLinkRightIcon
-    anchors.centerIn: control
+    anchors.fill: root.useCustomSize ? root : undefined
+    anchors.centerIn: root.useCustomSize ? undefined : root
```
tomasMizera, line 28: What is the point here? Let's directly create `Image` instead when you need something special :)

app/qml/components/MMPhoto.qml
```diff
@@ -38,19 +38,21 @@ Image {
     }
   }
 
-  MouseArea {
+  Rectangle {
     anchors.fill: parent
-    onClicked: control.clicked(control.photoUrl)
+    color: __style.whiteColor
+    z: -1
```
tomasMizera, line 44: What is the reason to use `z`? I highly recommend not using that

app/qml/inputs/MMPhotoEditor.qml
```diff
@@ -0,0 +1,67 @@
...
+
+  property url photoUrl
+
+  signal trashClicked()
+
+  contentItemHeight: 160 * __dp
+  spacing: 0
+  radius: 20 * __dp
+
+  content: MMPhoto {
```
tomasMizera, line 30: Sooo what if there is no photo added yet? :) There is an updated in the design system. We can add it in a different PR though

iiLubos: MMPhoto contains already an icon, when photo is not loaded by the design system

app/qml/components/MMPhoto.qml
iiLubos: I think, using z here in the base component is safe. This just adds a white background with the 'waiting' icon when the picture is not loaded yet.

app/qml/components/MMIcon.qml
tomasMizera: https://doc.qt.io/qt-6/qml-qtquick-image.html#fillMode-prop

tomasMizera (approved):

PR #3036, opened 2024-01-31
New Project Page

iiLubos:
image

Note: Upload icon is missing in the new design


PeterPetrik:
Can you add screenshots and rebase?

app/qml/project/MMProjectListPage.qml
```diff
@@ -0,0 +1,91 @@
...
+    anchors.fill: parent
+    color: __style.lightGreenColor
+  }
+
+  MMHeader {
+    id: header
+
+    title: root.headerTitle
+    onBackClicked: root.backClicked()
+  }
```
tomasMizera, line 36:
```suggestion
  header: MMHeader {
    id: header

    title: root.headerTitle
    onBackClicked: root.backClicked()
  }
```

app/qml/components/MMHeader.qml
```diff
@@ -73,6 +81,35 @@ Item {
     onClicked: root.backClicked()
   }
 
+  Rectangle {
```
PeterPetrik, line 84: This is not correct approach. Personal Icon should be bound from parent as right item

iiLubos: We can discuss it. This is a special item and should be extended by the new design, it can contain a red mark on the top right corner of this icon or can contain user initials instead of the icon.

PeterPetrik:
closing this in favour of https://github.com/MerginMaps/mobile/pull/3045

PR #3049, opened 2024-02-13
Send post-registration data when onboarding is completed

uclaros:
Send post-registration data when onboarding is completed (endpoint `/v1/post-register` is not yet available in test.dev server)

Success/failure is only logged and not shown to the user. He only cares that his workspace was created, so always show this.

The back button was also removed from `MMHowYouFoundUs` so that users can't skip _how you found us_ and _which industry_ questions.

### DIFF
--- a/app/qml/onboarding/MMHowYouFoundUs.qml
+++ b/app/qml/onboarding/MMHowYouFoundUs.qml
@@ -41,6 +41,7 @@ Page {
     id: header
 
     title: listView.contentY > -30 * __dp ? root.headerTitle : ""
+    backVisible: false
 
     onBackClicked: root.backClicked()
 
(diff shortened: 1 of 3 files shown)


app/qml/onboarding/MMHowYouFoundUs.qml
```diff
@@ -41,6 +41,7 @@ Page {
     id: header
 
     title: listView.contentY > -30 * __dp ? root.headerTitle : ""
+    backVisible: false
```
uclaros, line 44: Should I completely remove the root page's `backClicked` signal too?

PeterPetrik: lets keep it for now

PR #3068, opened 2024-02-21
Multi-feature selection initial implementation

uclaros:
Initial implementation of selection of overlapping features.

Use _tap and hold_ to bring up a list of features if at least one is under the tap point.
image

Pick from the list to open the feature preview drawer.
image

The current item delegate is too simple for this use case, we should use an item delegate that displays more info about the features (layer, fid?, icon)


PeterPetrik:
works nicely

PR #3084, opened 2024-02-29
Fix accept invitation dialog with revised logic

uclaros:
Accept invitation using revised logic:

Untitled-2024-02-26-0954


PeterPetrik (commented):
LGTM

PR #3092, opened 2024-02-29
Redesign of the MissingAuthDialog

VitorVieiraZ:
- ```MissingAuthDialog.qml``` to ```MMMissingAuthDialog.qml```
-  Sign In image included in the gallery.


app/qml/CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ set(MM_QML
     components/Symbol.qml
     components/TextWithIcon.qml
     dialogs/MigrateToMerginDialog.qml
-    dialogs/MissingAuthDialog.qml
+    dialogs/MMMissingAuthDialog.qml
```
PeterPetrik, line 10: I think you. need to use it in-app somewhere

VitorVieiraZ: Updated MissingAuthDialog to MMMissingAuthDialog in main.qml usage

PeterPetrik: Ah you are right!

PR #3099, opened 2024-03-04
Use MMBusyIndicator instead of Qt's one

PeterPetrik:
also fix Gallery's misc page

### DIFF
--- a/app/qml/components/MMBusyIndicator.qml
+++ b/app/qml/components/MMBusyIndicator.qml
@@ -12,12 +12,13 @@ import QtQuick
 Item {
   id: root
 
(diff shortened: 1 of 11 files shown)


app/qml/components/MMBusyIndicator.qml
```diff
@@ -12,12 +12,13 @@ import QtQuick
 Item {
   id: root
 
-  property bool running: false
+  required property real size
```
tomasMizera, line 15: Can we have a default size? I would expect it to have one set size that I won't need to edit

tomasMizera (approved):

PR #3100, opened 2024-03-04
Notifications

iiLubos:
- Used convenient methods to save typing of some parameters, instead of add( parameters )
- removed unnecessary rowCount
- fixed icons in notifications
- support to invoke an action when clicked on notification


app/qml/components/MMNotification.qml
```diff
@@ -38,15 +38,15 @@ Rectangle {
     anchors.verticalCenter: parent.verticalCenter
     anchors.left: parent.left
     anchors.leftMargin: 20 * __dp
-    size: model.icon === MM.NotificationType.Check ? __style.icon18 : __style.icon24
```
PeterPetrik, line 41: I think it has size.icon16

iiLubos: by the design, icon is 24px with margins. 18px without margins.
16px is not good readable

app/notificationmodel.h
```diff
@@ -73,41 +73,32 @@ class NotificationModel : public QAbstractListModel
...
+    Q_ENUM( NotificationModelRoles )
 
     NotificationModel( QObject *parent = nullptr );
     ~NotificationModel();
 
-    QHash<int, QByteArray> roleNames() const override;
-    int rowCount( const QModelIndex &parent = QModelIndex() ) const override;
-    QVariant data( const QModelIndex &index, int role = Qt::DisplayRole ) const override;
-
-    Q_PROPERTY( int rowCount READ rowCount NOTIFY rowCountChanged );
```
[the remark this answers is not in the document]
iiLubos: ok

app/qml/components/MMNotification.qml
iiLubos: by design document, icon with margins are 24px, without margins (circle) are 18px.
16px are not good readable

PeterPetrik: in that case change in mmstyle.h icon16 -> icon18

PeterPetrik (approved):

PR #3102, opened 2024-03-04
move onboarding to account and rename files

PeterPetrik:


app/qml/account/MMWhichIndustryPage.qml
```diff
@@ -12,6 +12,7 @@ import QtQuick.Controls
 
 import mm 1.0 as MM
 
+import "./components"
 import "../components"
```
tomasMizera, line 16: Hmm, just wondering if we should name the subfolders like this... I know we use it in other places too, I am just a bit afraid that you can import "../components" as well as "./components" 😵‍💫

tomasMizera (approved):
Only one question :)

PR #3112, opened 2024-03-05
MMFormEditFailedDialog and MMFormDeleteFeatureDialog

VitorVieiraZ:
MMFormEditFailedDialog

MMFormDeleteFeatureDialog


app/qml/dialogs/MMFormDeleteFeatureDialog.qml
```diff
@@ -0,0 +1,33 @@
...
+import QtQuick
+
+import "../components"
+
+MMDrawerDialog {
+  id: root
+
+  signal deleteFeature()
+
+  picture: __style.negativeMMSymbolImage
```
PeterPetrik, line 19: this is not an error = use positiveMMSymbol

PR #3123, opened 2024-03-07
Formfixes1

PeterPetrik:
all fixed


tomasMizera (approved):
🚀 

Let's also make sure that https://github.com/MerginMaps/mobile/issues/2870 is fixed when releasing (relation form editors should be available for readonly layers if the linked layer is editable).

PR #3136, opened 2024-03-12
Photo preview fix

iiLubos:
- [x] Show photo preview only when a photo is ready
- [x] Fix: Show preview of the whole photo, not only a cropped view


tomasMizera:
I'll test out and merge :)

PR #3152, opened 2024-03-18
"Take a picture" feature not available on desktops

VitorVieiraZ:
Use of the ```hasCameraCapability``` property in ```MMPhotoViewer.qml``` to control the ```'Take a Picture'``` feature.

| hasCameraCapability: false |
|--------|
|| 

| hasCameraCapability: true |
|--------|
| |


app/qml/form/editors/MMFormPhotoViewer.qml
```diff
@@ -99,6 +99,8 @@ MMBaseInput {
       height: root.contentItemHeight
       visible: root.state === "notSet"
 
+      canDeviceTakePictures: hasCameraCapability
```
PeterPetrik, line 102: can you use the same name all around the code? it improves readability a lot.

PR #3153, opened 2024-03-18
Get the correct closing of MMStakeoutPanel and correct state update

VitorVieiraZ:
When the ESC key or back button is pressed, or when the closeButton is clicked, the MMStakeoutPanel closes correctly.

### DIFF
--- a/app/qml/gps/MMStakeoutPanel.qml
+++ b/app/qml/gps/MMStakeoutPanel.qml
@@ -58,19 +58,6 @@ MMDrawer {
     root.open()
   }
 
-  Item {
-    focus: true
-
-    Keys.onPressed: {
-      if ( event.key === Qt.Key_Back || event.key === Qt.Key_Escape ) {
-        root.endStakeout()
-        event.accepted = true
-      }
-    }
-  }
-
-  closePolicy: Popup.NoAutoClose
-
   Behavior on height {
     SequentialAnimation {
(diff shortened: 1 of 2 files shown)


app/qml/gps/MMStakeoutPanel.qml
```diff
@@ -58,15 +58,26 @@ MMDrawer {
...
+  Item {
+    focus: true
+
+    Keys.onPressed: {
+      if ( event.key === Qt.Key_Back || event.key === Qt.Key_Escape ) {
+        root.endStakeout()
+        event.accepted = true
+      }
+    }
+  }
```
PeterPetrik, line 70: This is already implemented in https://github.com/MerginMaps/mobile/blob/master/app/qml/components/MMDrawer.qml#L65

so either that MMDrawer implementation doesn't work and we need to fix it (e.g. move it to separate Item as here) or it is not needed here at all if you connect correctly onBackClicked?

app/qml/gps/MMStakeoutPanel.qml
```diff
@@ -58,15 +58,26 @@ MMDrawer {
...
+
+  closePolicy: Popup.NoAutoClose
```
PeterPetrik, line 72: This should be part of MMDrawer, in case we want it. we want same behaviour for all drawers

PeterPetrik (commented):
talked with Vitor - lets remove back signal from drawer and bind it to onClose()

PeterPetrik (approved):

PR #3156, opened 2024-03-19
fix form editors and geometry

PeterPetrik:


app/qml/form/editors/MMFormTextMultilineEditor.qml
```diff
@@ -38,7 +38,7 @@ MMBaseInput {
   property bool _fieldRememberValueState: parent.fieldRememberValueState
 
   property alias placeholderText: textArea.placeholderText
-  property alias text: textArea.text
+  property string text: _fieldValue === undefined || _fieldValueIsNull ? '' : _fieldValue
```
tomasMizera, line 41: Why not having this directly in the `textArea.text` property?

tomasMizera (approved):

app/qml/form/editors/MMFormTextMultilineEditor.qml
PeterPetrik: if we use alias we get runtime warnings 
`qrc:/com.merginmaps/imports/MMInput/form/MMFormPage.qml:261:7: QML Loader: Binding loop detected for property "fieldValue"`

without alias it works 🤷

PR #3159, opened 2024-03-19
Add layer name and layer geometry icon to multi-identify results

uclaros:
Add an icon according to layer's geometry type, add layer name as a secondary _description_ text.

image

Also fixes scrolling in MMListDrawer, it would always snap-return to top of list.

issues:
Spacing between primary and secondary text probably needs some adjusting to match the design docs.
List items overlap the Header when scrolled, as seen on the screenshot above (was the case before this pr too)

### DIFF
--- a/app/qml/components/MMListDrawerItem.qml
+++ b/app/qml/components/MMListDrawerItem.qml
@@ -50,14 +50,14 @@ Item {
       height: parent.height
       width: parent.width
 
-      Text {
+      MMText {
         width: parent.width
         text: control.text
         color: __style.nightColor
         font: __style.t3
       }
 
-      Text {
+      MMText {
         width: parent.width
         text: control.description
         color: __style.nightColor
--- a/app/qml/components/MMListDrawerItem.qml
+++ b/app/qml/components/MMListDrawerItem.qml
@@ -15,10 +15,11 @@ import QtQuick.Layouts
 Item {
   id: control
 
-  signal clicked(type: string)
+  signal clicked(var type)
 
-  property string type
+  property var type
   property string text
+  property string description
   property var iconSource
 
   property bool isActive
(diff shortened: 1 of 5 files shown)


app/qml/components/MMListDrawerItem.qml
```diff
@@ -45,14 +46,24 @@ Item {
...
     }
 
-    Text {
-      Layout.fillWidth: true
+    Column {
       height: parent.height
-      Layout.alignment: Qt.AlignVCenter
+      width: parent.width
+
+      Text {
```
PeterPetrik, line 53: MMText

app/qml/components/MMListDrawerItem.qml
```diff
@@ -45,14 +46,24 @@ Item {
...
+        width: parent.width
+        text: control.text
+        color: __style.nightColor
+        font: __style.t3
+      }
 
-      text: control.text
-      color: __style.nightColor
-      font: __style.t3
+      Text {
```
PeterPetrik, line 60: MMText

app/qml/main.qml
```diff
@@ -774,17 +774,18 @@ ApplicationWindow {
...
 
-  MMDropdownDrawer {
+  MMListDrawer {
     id: featurePairSelection
 
-    title: qsTr( "Select feature" )
-    withSearchbar: false
-    model: MM.FeaturesModel {}
+    drawerHeader.title: qsTr( "Select feature" )
+    maxHeight: ApplicationWindow.window?.height * 2 / 3 ?? 0
```
PeterPetrik, line 781: why it cannot be full screen?

app/qml/main.qml
```diff
@@ -774,17 +774,18 @@ ApplicationWindow {
...
+    listModel: MM.FeaturesModel {}
```
PeterPetrik, line 782: cannot we rename listModel -> model?

PR #3162, opened 2024-03-20
use red button when hard constrains are not met

PeterPetrik:
Screenshot 2024-03-20 at 9 04 45

Screenshot 2024-03-20 at 9 07 12

Screenshot 2024-03-20 at 9 07 04

### DIFF
--- a/app/qml/form/MMFormPage.qml
+++ b/app/qml/form/MMFormPage.qml
@@ -16,8 +16,8 @@ import QtQuick.Dialogs
 import mm 1.0 as MM
 import qgs 1.0 as QGS
 
-import "../components"
-import "./components"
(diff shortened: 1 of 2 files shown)


app/qml/form/MMFormPage.qml
```diff
@@ -443,7 +442,6 @@ Page {
   function save() {
     if ( controller.hasValidationErrors )
     {
-      console.log( qsTr( 'Can not save the form, there are validation errors' ) )
```
tomasMizera, line 446: How about `inpututils.log()` here?

PeterPetrik: Not sure if worth

PR #3168, opened 2024-03-22
Adjustment of the text displayed during synchronization

VitorVieiraZ:
From "Synchronising..." to "Please don't close the app."


tomasMizera:
Hi @VitorVieiraZ 

 - Would it be possible to adjust it like this

image

- You need to add the text "Synchronising project changes" to `ProjectsModel::data` when returning `ProjectDescription`. There is a flag `ProjectSyncPending` that you can use to differentiate if the project is currently being synchronised

tomasMizera (changes requested):
See my comment

PeterPetrik (approved):

PR #3169, opened 2024-03-22
WIP: Add MMListDrawer and MMListMultiselectDrawer

tomasMizera:
Todo:
 - [x] handle invisible state
 - [ ] add lists to the gallery
 - [ ] fix bottom margin with multiselect button

Examples:
- Use with `ListModel` or any kind of cpp model

image

 - Use with `ObjectModel`

image


PeterPetrik:
@tomasMizera I worked a bit on the replacement of drawers in the app - it works really nicely 

only issue I had with MMMapController > activeLayerPanel, where I had to set height to delegate otherwise the panel was empty (minimised)

PeterPetrik (approved):

PR #3172, opened 2024-03-22
Fixed rotation of photos in the MMFormPhotoViewer

VitorVieiraZ:
Bug fix: ISS-102

Fixed: Photos attached from gallery / captured are rotated in the form
The issue was occurring on iOS only.

 

Similar bug also fixed: image rotation in MMPhotoPreview:
| Before | After |
|--------|--------|
||  |

### DIFF
--- a/app/qml/components/MMMorePhoto.qml
+++ b/app/qml/components/MMMorePhoto.qml
@@ -31,6 +31,7 @@ Row {
 
     source: control.photoUrl
     asynchronous: true
+    autoTransform: true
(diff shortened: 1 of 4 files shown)


saberraz:
I remember I fixed this a while back! https://github.com/MerginMaps/mobile/pull/2633

tomasMizera:
> I remember I fixed this a while back! #2633

Indeed! I remembered there was a fix for that but could not find it!

PR #3175, opened 2024-03-25
LayersPanel with MMPage and MMListDrawer

PeterPetrik:
it looks like pageSpacing on MMPage was not used @tomasMizera ?

### DIFF
--- a/app/qml/components/MMPage.qml
+++ b/app/qml/components/MMPage.qml
@@ -16,7 +16,7 @@ Page {
   property alias pageHeader: mmheader
   property alias pageContent: contentGroup.children
 
-  property double pageSpacing: __style.spacing20 // Change this to 20 if using searchbar
+  property double pageSpacing: __style.spacing20 // Usually 20 if using searchbar and 40 without search bar
 
   property real pageBottomMargin: __style.margin20
   property int pageBottomMarginPolicy: MMPage.BottomMarginPolicy.UseMargin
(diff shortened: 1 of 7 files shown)


app/qml/components/MMPage.qml
```diff
@@ -71,16 +71,17 @@ Page {
...
       else {
-        return parent.height
+        return parent.height - root.pageSpacing
       }
     }
     width: parent.width - leftPadding - rightPadding
 
     // center the content
     x: leftPadding
+    y: root.pageSpacing
```
tomasMizera, line 84: Let's instead remove that pageSpacing altogether

app/qml/layers/MMFeaturesListPage.qml
```diff
@@ -25,132 +25,80 @@ Page {
...
 
-  ListView {
-    id: listView
+  Timer {
+    id: searchDelay
+    interval: 500
+    running: false
+    repeat: false
+    onTriggered: featuresModel.searchExpression = searchbox.text
+  }
```
tomasMizera, line 43: This is already a part of MMSearchInput - just use its API

app/qml/layers/MMLayerDetailPage.qml
```diff
@@ -13,9 +13,10 @@ import QtQuick.Layouts
 
 import mm 1.0 as MM
 
-import "../components"
+import "../components" as MMComponents
 import "../inputs"
 
+// Do not use MMPage here
```
tomasMizera, line 19: why?

tomasMizera (changes requested):

app/qml/layers/MMLayerDetailPage.qml
```diff
@@ -113,21 +99,18 @@ Page {
```
tomasMizera, line 99: We should extract the following into `MMSwitchInput`

app/qml/layers/MMLayerDetailPage.qml
tomasMizera: Ah, I see now, wow.

tomasMizera (approved):

PR #3177, opened 2024-03-25
Randomizing order of "where did you hear about us" and "industry" options

VitorVieiraZ:
| MMHowYouFoundUsPage |
|----------------|
|  |

| MMWhichIndustryPage |
|----------------|
|  |

Obs: Option "other" always last.

### DIFF
--- a/app/qml/account/MMHowYouFoundUsPage.qml
+++ b/app/qml/account/MMHowYouFoundUsPage.qml
@@ -24,6 +24,39 @@ MMPage {
 
   signal howYouFoundUsSelected( var selectedText )
 
+  ListModel {
+    id: sourceListModel
+
+    Component.onCompleted: {
+      var items = [
+        {name: qsTr("Search engine (Google, ...)"), key: "search_engine", icon: __style.searchIcon, submenu: false},
+        {name: qsTr("Blog"), key: "blog", icon: __style.termsIcon, submenu: false},
+        {name: qsTr("Mouth"), key: "mouth", icon: __style.mouthIcon, submenu: false},
+        {name: qsTr("QGIS website"), key: "qgis_website", icon: __style.qgisIcon, submenu: false},
+        {name: qsTr("Application store"), key: "app_store", icon: __style.subscriptionsIcon, submenu: false},
+        {name: qsTr("Teacher"), key: "teacher", icon: __style.teacherIcon, submenu: false},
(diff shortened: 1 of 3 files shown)


saberraz:
I would use `Word of Mouth` instead of `Mouth`.

VitorVieiraZ:
> I would use `Word of Mouth` instead of `Mouth`.


Adjustment applied.

app/qml/account/MMHowYouFoundUsPage.qml
```diff
@@ -24,6 +24,38 @@ MMPage {
...
+        { name: qsTr( "Teacher" ), key: "teacher", icon: __style.teacherIcon, submenu: false },
+        { name: qsTr( "Conference" ), key: "conference", icon: __style.briefcaseIcon, submenu: false },
+        { name: qsTr( "Social media" ), key: "social", icon: __style.socialMediaIcon, submenu: false }
+      ];
+      var otherItem = { name: qsTr( "Other" ), key: "other", icon: __style.otherIcon, submenu: false };
+
+      for ( var i = items.length - 1; i > 0 ; i-- ) {
+        var j = Math.floor( Math.random() * ( i + 1 ) );
+        var temp = items[ i ];
+        items[ i ] = items[ j ];
```
tomasMizera, line 46: I can see it is motivated by this [algo](https://stackoverflow.com/a/12646864/7875594) which if fine, but I struggle to understand how come this won't lead to exceptions.
If you consider this scenario:

```
array length = 8

for loop ( i == 7 )
  j = Math.floor( 0.999 * ( 7 + 1 ) ) // j == 8 here
  ..
  items [j] // <<-- here you go outside of bounds of the input array
```

For that reason, I would prefer a simpler and more readable version: https://stackoverflow.com/a/46545530/7875594 :)

tomasMizera (changes requested):

tomasMizera:
> > > I would use `Word of Mouth` instead of `Mouth`.
> > 
> > 
> > Adjustment applied.
> 
> It needs to be the same on mobile and web, and definitely the codes sent to backend must match (cc @tomasMizera )

The keys should be the same on the server and here. It was an agreement between @MarcelGeo and @uclaros. Texts should be the same as here https://docs.google.com/spreadsheets/d/1VSrdlT0JdnI4BnMlzVeUaaI9AICFnwVk0m9Rr-MKJZ8/edit#gid=0 - feel free to suggest changes there first

PeterPetrik (approved):

PR #3180, opened 2024-03-26
Add combobox input and use it in the project wizard page

tomasMizera:
Screenshot 2024-03-26 at 09 04 21
Screenshot 2024-03-26 at 09 04 20


app/qml/project/MMProjectWizardPage.qml
```diff
@@ -151,4 +132,32 @@ MMPage {
...
+
+  function indexFromWidgetType( widgetType ) {
+    for ( let i = 0; i < typesmodel.count; i++ ) {
+      let item = typesmodel.get(i)
+      if ( widgetType === item.type ) {
+        return i
+      }
+    }
+    return -1
+  }
```
PeterPetrik, line 162: wondering if more generic version of this function shouldn't be part of comboboxinput ? since we will probably need to use it in other situations too?

tomasMizera:
@PeterPetrik Addressed your comments

PeterPetrik (approved):

PR #3184, opened 2024-03-27
always try to show 4 buttons in toolbar

PeterPetrik:
if screen size is >= 240px 4 buttons will be shown now (I hope there are no smaller mobile phones :) )

https://github.com/MerginMaps/mobile/assets/804608/f89b9215-c871-4a3a-a763-b06201d69067

### DIFF
--- a/app/qml/components/MMToolbar.qml
+++ b/app/qml/components/MMToolbar.qml
@@ -144,8 +144,19 @@ Rectangle {
       if (btnData.visible) ++visibleButtonsCount
     }
 
+    // Find how many short buttons we can fit into toolbar:
+    // We want to show minimumToolbarButtonsInMainToolbar buttons
+    // if possible, even at the cost that they will be shorter.
+    var maxButtonsInToolbar = Math.floor( toolbar.width / internal.preferredToolbarButtonWidth )
+    if (maxButtonsInToolbar < internal.minimumToolbarButtonsInMainToolbar) {
+      maxButtonsInToolbar = Math.floor( toolbar.width / internal.minimumToolbarButtonWidth )
+      if (maxButtonsInToolbar > internal.minimumToolbarButtonsInMainToolbar) {
+        // we do not want to show more buttons than minimum even they would fit
+        maxButtonsInToolbar = internal.minimumToolbarButtonsInMainToolbar
+      }
+    }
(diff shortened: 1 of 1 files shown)


app/qml/components/MMToolbar.qml
```diff
@@ -144,8 +144,19 @@ Rectangle {
       if (btnData.visible) ++visibleButtonsCount
     }
 
+    // Find how many short buttons we can fit into toolbar:
+    // We want to show minimumToolbarButtonsInMainToolbar buttons
+    // if possible, even at the cost that they will be shorter.
+    var maxButtonsInToolbar = Math.floor( toolbar.width / internal.preferredToolbarButtonWidth )
```
tomasMizera, line 150: Please replace all `var`s with `let`

```suggestion
    let maxButtonsInToolbar = Math.floor( toolbar.width / internal.preferredToolbarButtonWidth )
```

app/qml/components/MMToolbar.qml
```diff
@@ -144,8 +144,19 @@ Rectangle {
...
+    if (maxButtonsInToolbar < internal.minimumToolbarButtonsInMainToolbar) {
+      maxButtonsInToolbar = Math.floor( toolbar.width / internal.minimumToolbarButtonWidth )
+      if (maxButtonsInToolbar > internal.minimumToolbarButtonsInMainToolbar) {
+        // we do not want to show more buttons than minimum even they would fit
+        maxButtonsInToolbar = internal.minimumToolbarButtonsInMainToolbar
+      }
+    }
```
tomasMizera, line 157: I was thinking about something like this - would that work? 🤔 

```suggestion
    let maxButtonsInToolbar = Math.floor( toolbar.width / internal.preferredToolbarButtonWidth )
    
    if ( Screen.primaryOrientation === Qt.PortraitOrientation || Qt.InvertedPortraitOrientation ) {
      maxButtonsInToolbar = Math.max( maxButtonsInToolbar, internal.minimumToolbarButtonsInMainToolbar)
    }
```

PeterPetrik: this would work on mobiles , but when you will play with width on desktop it will be screwed... and we have windows production version at minimum

PR #3189, opened 2024-03-27
Up to date image adjustment

VitorVieiraZ:
Adjust: ISS-120
In main.qml, changed the icon from WaitingIcon to CheckIcon if it is already up to date.

### DIFF
--- a/app/qml/main.qml
+++ b/app/qml/main.qml
@@ -903,7 +903,7 @@ ApplicationWindow {
     {
       if ( projectFullName === __activeProject.projectFullName() )
       {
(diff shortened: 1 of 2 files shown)


tomasMizera:
Hi Vitor! Thanks for these changes. I believe though that ISS-120 actually refers to something else - it is notification icon when the app says "Up to date" (after one clicked on the sync button). See https://github.com/MerginMaps/mobile/blob/8dc9cbc9bd0e68b2c3100f8b12bd99e5a3825e86/app/qml/main.qml#L906

tomasMizera:
I quite liked the image that was here previously :)

PR #3196, opened 2024-04-02
open subscription page directly

PeterPetrik:


app/inputhelp.cpp
```diff
@@ -99,6 +99,18 @@ QString InputHelp::merginDashboardLink() const
   return MerginApi::defaultApiRoot() + "dashboard" + queryParams;
 }
 
+QString InputHelp::merginSubscriptionLink() const
+{
+  if ( mMerginApi && mMerginApi->apiSupportsSubscriptions() )
+  {
+    return MerginApi::defaultApiRoot() + "subscription" + utmTagSubscription;
```
tomasMizera, line 106: You should include the workspace id directly in the url so that the user ends up in the correct workspace. [activeWorkspacePathPart = QStringLiteral( "?workspace=%1" ).arg( activeWS );](https://github.com/MerginMaps/mobile/blob/5bba56e896913d0e09e1179a14f25b5ad11db0f0/app/inputhelp.cpp#L73-L74)

PeterPetrik: added

tomasMizera (approved):

PR #3204, opened 2024-04-03
show notification for drawer on close account

PeterPetrik:


tomasMizera (approved):
Not the best way forward but given that only a minority of people use it, we can keep it like that for now

PR #3205, opened 2024-04-03
show invitations on startup

PeterPetrik:


tomasMizera (commented):
Is this going to work when you are in the map?

PeterPetrik:
> Is this going to work when you are in the map?

yes, it works when you are on map

tomasMizera (approved):

PR #3230, opened 2024-04-03
Enable qml debugging

uclaros:
This allows stepping through code in qml files.


PeterPetrik:
we need to add the same to gallery/main.cpp

Issue #3221, opened 2024-04-03
NoGeo layer problems


jozef-budac:
There are 3 problems for the NoGeo layer. 

1. NoGeo has its visibility switch in the layer list. I believe it would be better not to show it at all.
2.  When adding a new feature for Nogeo, there is an "Edit Geometry" button on the bottom panel, even then NoGeo has no geometry
3. When creating a new feature and I hit the bin icon to delete the feature and confirm the deletion, I get a "Saving failed" error.

You can use the `tc03_layer_legend_tree` project for the scenarios

https://github.com/MerginMaps/mobile/assets/83582181/0e314ce8-429c-47bf-9296-af07264563aa

PeterPetrik:
1 and 2 addressed in https://github.com/MerginMaps/mobile/pull/3231

tomasMizera:
4. make the delete button red 
<img width="300" alt="image" src="https://github.com/MerginMaps/mobile/assets/22449698/dc9287be-10c9-449b-a37b-07f74947ce38">

PR #3231, opened 2024-04-03
Nogeo

PeterPetrik:
- fixes 1/ and 2/ from #3221
- also fixes toolbar when visibility dynamically changes and when image source is QUrl


tomasMizera:
Let's deal with the issue of "one-button" toolbar in the separate PR

tomasMizera (approved):

PR #3251, opened 2024-04-04
Fix no geometry layers feature deletion

### LINKED ISSUE #3221: NoGeo layer problems

PeterPetrik:
fix #3221

### DIFF
--- a/app/attributes/attributecontroller.cpp
+++ b/app/attributes/attributecontroller.cpp
@@ -1061,8 +1061,19 @@ bool AttributeController::deleteFeature()
     return false;
   }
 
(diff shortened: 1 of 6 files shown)


tomasMizera (approved):
🚀

PR #3252, opened 2024-04-04
Maptheme dialog fixes

PeterPetrik:
fix https://github.com/MerginMaps/mobile/issues/3220
fix https://github.com/MerginMaps/mobile/issues/3244

### DIFF
--- a/app/qml/components/MMListDrawer.qml
+++ b/app/qml/components/MMListDrawer.qml
@@ -30,7 +30,7 @@ MMDrawer {
       width: parent.width
       height: Math.min( root.drawerContentAvailableHeight, contentHeight )
 
-      enabled: contentHeight > height
+      enabled: emptyStateDelegateLoader.visible
 
       Loader {
         id: emptyStateDelegateLoader
--- a/app/qml/main.qml
+++ b/app/qml/main.qml
@@ -493,7 +493,7 @@ ApplicationWindow {
   MMMapThemeDrawer {
     id: mapThemesPanel
 
-    height: ( window.height / 2 )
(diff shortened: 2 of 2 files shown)


app/qml/components/MMListDrawer.qml
```diff
@@ -30,7 +30,7 @@ MMDrawer {
       width: parent.width
       height: Math.min( root.drawerContentAvailableHeight, contentHeight )
 
-      enabled: contentHeight > height
+      enabled: emptyStateDelegateLoader.visible
```
tomasMizera, line 33: That condition was important - you would be scrolling otherwise even when you do not need to (content fits the screen)
```suggestion
      enabled: contentHeight > height || emptyStateDelegateLoader.visible
```

tomasMizera (changes requested):

app/qml/components/MMListDrawer.qml
PeterPetrik: Enabled = false blocks all events (like mouse click) on all children

PeterPetrik: Which is exactly source of reported bug

tomasMizera: That's fine, so it could be like:
```qml
enabled: emptyStateDelegateLoader.visible ? emptyStateDelegateLoader.visible : contentHeight > height
```

PeterPetrik: scrollbar is only for no-items element, so if emptyStateDelegate is not shown whole scrollbar should be off (which is the fix in the PR?)

PR #3265, opened 2024-04-05
rework center to gps

### LINKED ISSUE #3257: When GPS is centred, map canvas should follow your position; also during recording
ISS-138

uclaros:
Fixes #3257 

Also fixes:
- Stakeout line is now rendered below cursor
- Jump-to is used when in `view` state and `centerToGPS` is enabled 
- ^ Jump-to is triggered when moving outside the central 20% of the screen
- Jump-to animation is now straight and not split in two axes
- In stakeout mode the gps accuracy and center to gps buttons are visible


tomasMizera (approved):
Looking good 🚀 Does it behave well when feature is selected from the layers panel?

PR #3293, opened 2024-04-10
fix centering when preview panel is opened

### LINKED ISSUE #3260: Center map extent when feature is identified
ISS-150, reported by @PeterPetrik 

When I click on identify - the map is centered, but the drawer overflows the center point, the selected point should be probably in the top part so is visible?

uclaros:
Fixes #3260

- Preview panel height is taken into account when centering to the identified feature
- Animation is used while centering
- If _follow gps_ was enabled, map pans back to gps position after closing the preview panel

### DIFF
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -281,7 +281,7 @@ QPointF InputUtils::geometryCenterToScreenCoordinates( const QgsGeometry &geom, 
 {
   QPointF screenPoint;
 
-  if ( geom.isNull() || !geom.constGet() )
+  if ( !mapSettings || geom.isNull() || !geom.constGet() )
     return screenPoint;
 
   QgsRectangle bbox = geom.boundingBox();
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -277,6 +277,19 @@ void InputUtils::setExtentToFeature( const FeatureLayerPair &pair, InputMapSetti
   mapSettings->setExtent( currentExtent );
 }
 
+QPointF InputUtils::geometryCenterToScreenCoordinates( const QgsGeometry &geom, InputMapSettings *mapSettings )
+{
+  QPointF screenPoint;
+
+  if ( geom.isNull() || !geom.constGet() )
+    return screenPoint;
+
+  QgsRectangle bbox = geom.boundingBox();
+  screenPoint = mapSettings->coordinateToScreen( QgsPoint( bbox.center() ) );
+
+  return screenPoint;
+}
+
 double InputUtils::convertCoordinateString( const QString &rationalValue )
 {
   QStringList values = rationalValue.split( "," );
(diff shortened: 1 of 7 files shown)


app/inpututils.cpp
```diff
@@ -277,6 +277,19 @@ void InputUtils::setExtentToFeature( const FeatureLayerPair &pair, InputMapSetti
   mapSettings->setExtent( currentExtent );
 }
 
+QPointF InputUtils::geometryCenterToScreenCoordinates( const QgsGeometry &geom, InputMapSettings *mapSettings )
+{
+  QPointF screenPoint;
+
```
tomasMizera, line 283: Please add check for validity of the `mapSettings` pointer

app/qml/main.qml
tomasMizera, line 1: Do we still need to calculate `takenVerticalSpace` property of `MMFormsStackController` and consider it when evaluating the `mapExtentOffset` of `MMMapController`?

app/qml/map/MMMapController.qml
```diff
@@ -1183,6 +1188,11 @@ Item {
       }
 
       case "view": {
+        if ( identifyHighlight.geometry !== null )
```
tomasMizera, line 1191: Can you add an explanation comment here please?

app/inpututils.cpp
uclaros: whoops!

app/qml/map/MMMapController.qml
uclaros: sure

app/qml/main.qml
uclaros: This is also used for the conditional visibility of the accuracy button:
```
          visible: {
            if ( root.mapExtentOffset > 0 && root.state !== "stakeout" ) return false
```

the actual value is not used, so it could be replaced by some other means. However, it's probably a good idea to keep the mapExtentOffset updated so that it always holds the correct map's offset.

uclaros:
What's also not used any more is the `considerMapExtentOffset` parameter in 
```qml
  function centerToPair( pair, considerMapExtentOffset = false ) {
    if ( considerMapExtentOffset )
      var mapExtentOffsetRatio = mapExtentOffset / mapCanvas.height
    else
      mapExtentOffsetRatio = 0

    __inputUtils.setExtentToFeature( pair, mapCanvas.mapSettings, mapExtentOffsetRatio )
  }
```
and the relevant code in `InputUtils::setExtentToFeature()` 
Should I remove those?

tomasMizera:
> What's also not used any more is the `considerMapExtentOffset` parameter in
> 
> ```qml
>   function centerToPair( pair, considerMapExtentOffset = false ) {
>     if ( considerMapExtentOffset )
>       var mapExtentOffsetRatio = mapExtentOffset / mapCanvas.height
>     else
>       mapExtentOffsetRatio = 0
> 
>     __inputUtils.setExtentToFeature( pair, mapCanvas.mapSettings, mapExtentOffsetRatio )
>   }
> ```
> 
> and the relevant code in `InputUtils::setExtentToFeature()` Should I remove those?

Sure, if it is not used anymore, let's remove it

tomasMizera (approved):

PR #3298, opened 2024-04-11
Fix download button

### LINKED ISSUE #3206: App freezes on Explorer page
When I do some stress tests on the Explorer page I can get into some problems:
1) After hitting the download button, the button is still active and I can hit it again to start downloading again. It should be disabled I think
2) I can get to a problem, I can't download a project at all (first part of the video with the project tc04_photo_rename)
3) I can get into a situation where the application freezes (the last seconds of the video and project tc04_snapping).

https://drive.google.com/file/d/1JyiklxyPlpYDttztFaJrsZzFtLEW4K87/view?usp=sharing

Also on the Home page, I can see multiple projects with same name and download failed:
image

wonder-sk:
Fixes #3206

Do not show download button while download is in progress... Debug build would crash on assert anyway.
This fixes the problem (1) and it should not be possible to get problems (2) and (3) anymore

https://github.com/MerginMaps/mobile/assets/193367/cc6042ec-732e-46e4-92db-1c4a8087831e

### DIFF
--- a/app/qml/project/components/MMProjectDelegate.qml
+++ b/app/qml/project/components/MMProjectDelegate.qml
@@ -146,6 +146,8 @@ Control {
           iconSource: root.state === "OnServer" ? __style.downloadIcon : __style.moreVerticalIcon
           bgndColor: root.projectIsOpened ? __style.polarColor : __style.lightGreenColor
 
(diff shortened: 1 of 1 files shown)


PeterPetrik:
invisible button is not possible to press, right?
right?
right? :)

PR #3307, opened 2024-04-12
Fix photo editor

tomasMizera:
Fixes multiple issues with the photo editor:
 - show trash icon even when the image is not available (previously it showed the icon only when the Image was loaded successfully)
 - update the text of the `MMRemovePhotoDialog` to better represent what it does
 - photos were not removed from the project even if the `Delete photo` was selected. The issue was that the callback function `onFormSaved` was placed in the private `QtObject` and thus could not be called from the form
 - we used to call `root.editorValueChanged( "", false )` where the boolean value represents if the new value is null. We were passing `false` even when removing the photo. Changed it to `true`.

Screenshot 2024-04-12 at 13 12 41

### DIFF
--- a/app/qml/components/MMPhotoPreview.qml
+++ b/app/qml/components/MMPhotoPreview.qml
@@ -38,7 +38,7 @@ Popup {
 
       height: root.height / 2
 
-      autoTransform: true;
+      autoTransform: true
       focus: true
       asynchronous: true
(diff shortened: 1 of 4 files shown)


PeterPetrik:
shouldn't unlink photo button has `grape` border (as secondary button)?

wonder-sk (approved):

PR #3311, opened 2024-04-12
Try Qt 6.6.3

### LINKED ISSUE #3296: App crashes when screen is rotated during QR code scanning
To reproduce:
 - open QR code scanner
 - rotate screen
   - 💥 

Reproduced on Android, iOS works fine.

I managed to get the same stacktrace (crash) when doing the following:

 - Open form that has an image field and a QR code field
 - Add image
 - Immediately after the image is captured open the QR code scanner
  -  💥 

A Quick Google search showed this Qt bug: https://bugreports.qt.io/browse/QTBUG-119445 <-- the same stacktrace

I tried to apply the fix mentioned in this comment https://bugreports.qt.io/browse/QTBUG-119445?focusedId=760746&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-760746 but that only led to other issues with the camera.

One possible solution could be to lock the screen orientation when opening the QR code scanner, see https://stackoverflow.com/questions/34310349/temporarily-programmatically-disable-screen-rotation-in-qt

I did not test the current 2.5.0, so not sure if it is a regression or not.

wonder-sk:
- let's see if Qt 6.6.3 will help with #3296
- small fix in `update_qt_version.bash` (with `\'` sed does not work for me on linux)

Fixes #3296

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -23,7 +23,7 @@ jobs:
     if: ( github.repository == 'MerginMaps/mobile' ) && (!contains(github.event.head_commit.message, 'Translate '))
     runs-on: macos-13
     env:
-      QT_VERSION: '6.6.0' # use scripts/update_qt_version.bash to change
+      QT_VERSION: '6.6.3' # use scripts/update_qt_version.bash to change
       NDK_VERSION: r25
       NDK_VERSION_FULL: r25b
       JDK_VERSION: 11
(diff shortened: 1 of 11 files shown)


PeterPetrik:
Don’t you need new SDK too ?

wonder-sk:
> Don’t you need new SDK too ?

I don't know :smile:

wonder-sk:
This fixes a crash for me as described in https://github.com/MerginMaps/mobile/issues/3296#issuecomment-2051899124

Best if @tomasMizera checks if it also fixes the other crashes that I was unable to replicate

tomasMizera:
@wonder-sk I confirm this fixes the mentioned camera issue for me (in both cases I reported) 👍🏻

PR #3312, opened 2024-04-12
Don't emit networkErrorOccurred when cancelling operations

### LINKED ISSUE #3217: Canceling of project upload makes a little mess
When I upload a new project and then cancel the upload it

1. show notification about "Failed to communicate with the server" which is not true
2. Both upload and synchronize buttons are missing for the project even when it was not uploaded fully (till I refresh the page)
3. After refreshing, there is a "Not uploaded" notification for the project but when I hit it, it says, the project with the name exists - which is true, because the project was created but the files were not synchronized.

I am not sure how big a problem it is, but I would address at least the first problem

https://github.com/MerginMaps/mobile/assets/83582181/6cf81d73-012e-4f35-a69c-de8991fb3748

uclaros:
Partially fixes #3217 : 1. which was happening on all operations that can be cancelled

2. and 3. are not fixed and are also present in the currently released version

### DIFF
--- a/core/merginapi.cpp
+++ b/core/merginapi.cpp
@@ -38,6 +38,7 @@ const QSet<QString> MerginApi::sIgnoreExtensions = QSet<QString>() << "gpkg-shm"
 const QSet<QString> MerginApi::sIgnoreImageExtensions = QSet<QString>() << "jpg" << "jpeg" << "png";
 const QSet<QString> MerginApi::sIgnoreFiles = QSet<QString>() << "mergin.json" << ".DS_Store";
 const int MerginApi::UPLOAD_CHUNK_SIZE = 10 * 1024 * 1024; // Should be the same as on Mergin server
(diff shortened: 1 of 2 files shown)


PeterPetrik:
let me look at this

PeterPetrik (changes requested):
This way we miss handling of failed request in `SynchronizationManager::onProjectSyncCancelled`

core/merginapi.cpp
```diff
@@ -2102,6 +2105,9 @@ void MerginApi::pushStartReplyFinished()
   {
     QByteArray data = r->readAll();
     QString serverMsg = extractServerErrorMsg( data );
+    if ( r->error() == QNetworkReply::OperationCanceledError )
+      serverMsg = tr( "Synchronisation canceled" );
```
PeterPetrik, line 2109: maybe pull it somewhere to the top of the file as cons string so it is just 1 string to translate?

PeterPetrik (approved):

PR #3317, opened 2024-04-16
remove syncInProgressAnimation leftover

### LINKED ISSUE #3216: Upload a new project with sync button when user is logged out
When a user is logged out and tries to upload a project with the Sync button in the project, a new dialog with the "Upload project?" notification appears. After hitting Yes, nothing happens. The expected behavior is the app to show Log-in page. 

https://github.com/MerginMaps/mobile/assets/83582181/51255b5b-d389-4f71-aad0-85ea62d7216d

uclaros:
Call to non existing `syncInProgressAnimation.stop()` was breaking the sync button logic

Fixes #3216

### DIFF
--- a/app/qml/main.qml
+++ b/app/qml/main.qml
@@ -814,7 +814,7 @@ ApplicationWindow {
     {
       if ( projectFullName === __activeProject.projectFullName() )
       {
-        syncInProgressAnimation.stop()
+        syncButton.iconRotateAnimationRunning = false
       }
     }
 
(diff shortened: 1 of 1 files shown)


app/qml/main.qml
```diff
@@ -898,7 +898,6 @@ ApplicationWindow {
     {
       if ( projectFullName === __activeProject.projectFullName() )
       {
-        syncInProgressAnimation.stop()
```
tomasMizera, line 901: Is this not needed anymore? `syncButton.iconRotateAnimationRunning = false`

uclaros: Apparently not. `syncButton.iconRotateAnimationRunning` is only set to true by the sync manager when `__syncManager::onSyncStarted` and when `__syncManager::hasPendingSync`, which can't happen without authing first.

PeterPetrik (approved):

PR #3325, opened 2024-04-17
Fix annoying error message when starting app with expired token

### LINKED ISSUE #3261: When app starts there is a notification that auth token is missing - but then the account loads OK in another moment
ISS-155, reported by @PeterPetrik 

When app starts there is a notification that auth token is missing - but then the account loads OK in another moment

wonder-sk:
Fixes #3261 

Ugly but works :zany_face:


tomasMizera (approved):

PeterPetrik (approved):

PeterPetrik:
now we are talking

PR #3332, opened 2024-04-18
Rework MMFormPhotoEditor

tomasMizera:
image
image

I moved two photo editor-related files from `./components` to `./form/components/photo`.
There are a bunch of things that could be changed but I think it is fine for now :)

### DIFF
--- a/app/qml/form/components/photo/MMPhotoAttachment.qml
+++ b/app/qml/form/components/photo/MMPhotoAttachment.qml
@@ -9,13 +9,14 @@
 
 import QtQuick
 import QtQuick.Controls
-import "."
+
+import "../../../components" as MMComponents
 
 Rectangle {
   id: root
 
   color: __style.polarColor
-  radius: 20 * __dp
+  radius: __style.radius20
 
   property bool hasCameraCapability: true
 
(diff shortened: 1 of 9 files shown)


app/qml/form/editors/MMFormPhotoViewer.qml
```diff
@@ -53,50 +47,58 @@ MMBaseInput {
...
+          rightMargin: __style.margin10
+          bottomMargin: __style.margin10
         }
 
         bgndColor: __style.negativeColor
         iconSource: __style.deleteIcon
         iconColor: __style.grapeColor
-        visible: root.allowEditing && root.state !== "notSet"
+
+        visible: root.editState === "enabled" && root.state !== "notSet"
```
uclaros, line 90: is "notSet" a leftover? It's not in the base class but is used elsewhere in this file too

app/qml/form/components/photo/MMPhotoAttachment.qml
```diff
@@ -42,13 +43,13 @@ Rectangle {
...
+        MMComponents.MMIcon {
           anchors.horizontalCenter: parent.horizontalCenter
           source: __style.addImageIcon
           color: __style.forestColor
         }
 
-        Text {
+        MMComponents.MMText {
           width: parent.width
           height: 24 * __dp
```
uclaros, line 54: Shouldn't the height be removed for MMTexts?

uclaros (commented):
A couple of minor comments, otherwise looks good!

app/qml/form/editors/MMFormPhotoViewer.qml
tomasMizera: It is a part of this `MMFormPhotoViewer` file - but good catch, maybe it should not be set as a state of base class. I will add it to a sub-state group instead 👍🏻

tomasMizera:
@uclaros your comments should be addressed now

uclaros (approved):

PR #3333, opened 2024-04-18
Rework MMPasswordInput

uclaros:


app/qml/inputs/MMPasswordInput.qml
```diff
@@ -8,46 +8,26 @@
...
-    placeholderTextColor: __style.darkGreyColor
-    font: __style.p5
-    hoverEnabled: true
-    echoMode: eyeButton.pressed ? TextInput.Normal : TextInput.Password
-    inputMethodHints: Qt.ImhNoPredictiveText | Qt.ImhNoAutoUppercase
-    background: Rectangle {
-      color: __style.transparentColor
-    }
-  }
+  textField.inputMethodHints: textField.inputMethodHints | Qt.ImhNoAutoUppercase
```
tomasMizera, line 20: Won't this produce a binding loop? 🤔

app/qml/inputs/MMPasswordInput.qml
```diff
@@ -8,46 +8,26 @@
...
+  rightContent: MMComponents.MMIcon {
     id: eyeButton
 
     property bool pressed: false
 
-    anchors.verticalCenter: parent.verticalCenter
-
     size: __style.icon24
     source: pressed ? __style.hideIcon : __style.showIcon
     color: root.enabled ? __style.forestColor : __style.mediumGreenColor
```
tomasMizera, line 29:
```suggestion
    color: root.editState === "enabled" ? __style.forestColor : __style.mediumGreenColor
```

tomasMizera (commented):
Can you please add these two additional input method hints 
image

app/qml/inputs/MMPasswordInput.qml
uclaros: Hmm it does.. What's the proper way to do |= in qml?

Also why do we have `inputMethodHints: Qt.ImhNoPredictiveText` in the base class?

tomasMizera: We want to keep `inputMethodHints: Qt.ImhNoPredictiveText` for all inputs in the app for now. Predictive text was giving us a lot of issues.

I have not found a way to do `|=` in QML yet, but If you find a way, I will be excited! :)

tomasMizera (approved):

tomasMizera (approved):

PR #3340, opened 2024-04-18
fix setting the background for a MMSearchInput

uclaros:


app/qml/components/MMListMultiselectDrawer.qml
```diff
@@ -52,7 +52,8 @@ MMDrawer {
 
         placeholderText: qsTr( "Search" )
 
-        bgColor: __style.lightGreenColor
+        textFieldBackground: __style.lightGreenColor
```
tomasMizera, line 55: Shouldn't it be like this? 🤔  `textFieldBackground` aliases to Rectangle, not its color
```suggestion
        textFieldBackground.color: __style.lightGreenColor
```

PR #3351, opened 2024-04-19
Tracking tweaks

wonder-sk:
- add tracking layer to new projects
- show elapsed time when tracking is active (instead of time when tracking started)
- fixed formatting of hours/minutes/seconds for tracking time

https://github.com/MerginMaps/mobile/assets/193367/ff1ac626-125d-4426-b578-d2c56037e9f6


tomasMizera (approved):
Looking good :)

PR #3383, opened 2024-04-25
Fixing zoom to project behavior while the GPS location is locked

VitorVieiraZ:
Following [Issue 3359](https://github.com/MerginMaps/mobile/issues/3359), the fix now ensures that when the ```Zoom to Project``` button is clicked while the GPS location is locked, the GPS location will be unlocked and the ```Zoom to Project``` function will work normally.

### DIFF
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -1834,6 +1834,8 @@ bool InputUtils::isPointLayerFeature( const FeatureLayerPair &feature )
 
 void InputUtils::zoomToProject( QgsProject *qgsProject, InputMapSettings *mapSettings )
 {
+  //change gps lock status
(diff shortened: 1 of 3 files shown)


app/inpututils.cpp
```diff
@@ -1834,6 +1834,8 @@ bool InputUtils::isPointLayerFeature( const FeatureLayerPair &feature )
 
 void InputUtils::zoomToProject( QgsProject *qgsProject, InputMapSettings *mapSettings )
 {
+  //change gps lock status
+
```
PeterPetrik, line 1838: remove this comment? or we need it here?

uclaros (approved):

PR #3386, opened 2024-04-26
send uuid in HTTP request

PeterPetrik:
Add GUID to each HTTP request and also show it in the diagnosis log. GUID is generated on first run or the app.
We use `X-Device-Id` with uuid as custom header

### DIFF
--- a/app/appsettings.cpp
+++ b/app/appsettings.cpp
@@ -13,13 +13,12 @@
 #include <QSettings>
 #include <QFileInfo>
 
-const QString AppSettings::INPUTAPP_GROUP_NAME = QStringLiteral( "inputApp" );
 const QString AppSettings::POSITION_PROVIDERS_GROUP = QStringLiteral( "inputApp/positionProviders" );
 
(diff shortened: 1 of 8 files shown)


tomasMizera:
The implementation looks good, let's see where to add it

mostlyAtNight:
Hi @PeterPetrik ,

Regarding how to send it, I would also be fine with it going in its own header, we should be able to accomodate that in the backend.  [This SE page](https://stackoverflow.com/questions/3561381/custom-http-headers-naming-conventions) may be useful for how to name a new one if we cannot find something appropriate to put it into.

Re the implementation itself, I have a question.  I assume the UUID is automatically generated when the app is installed - is it subsequently readable and editable by the user?  If so, I'm not sure this would be wise and I would instead opt for it to be stored in scoped storage or similar so it's protected from editing/cloning.

PeterPetrik:
Looks like non standart headers might be blocked by firewall, what name you suggest?

wonder-sk (approved):

PeterPetrik:
@varmar05 can you please secure server implementation? thanks

PR #3401, opened 2024-04-30
Unabling project reloading after project sync in home/projects page

### LINKED ISSUE #3358: Synchronizing upstream change opens the project
When I am on the home page, and there is an upcoming change to the project, I hit 3 dots, and then synchronize. The changes are successfully synchronized and immediately, the project is opened automatically. 

I believe this is not intended and the app should stay on the home page after synchronization.

VitorVieiraZ:
When syncing from the Home/Projects page, the project does not open automatically after the synchronization is completed.

Fixes #3358

### DIFF
--- a/app/qml/main.qml
+++ b/app/qml/main.qml
@@ -961,11 +961,6 @@ ApplicationWindow {
     function onProjectReloaded( project ) {
       map.clear()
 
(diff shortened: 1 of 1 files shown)


tomasMizera (approved):
This is looking good 👍🏻

PR #3413, opened 2024-05-07
Conditioning closing through dragging in MMListDrawer and MMListMultiselectDrawer

### LINKED ISSUE #3412: Glitchy scrolling in drawers
Currently, in MMDrawer with ListView, the drawer closes when dragged from the top. We need to modify this behavior to only allow closing the drawer using the close button, rather than by dragging.

VitorVieiraZ:
MMListDrawer and MMListMultiselectDrawer closure via dragging is now disabled when the listViewContent's interactive property is set to true (contentHeight > height). This change impacts drawers like MMMapThemeDrawer and featurePairSelection (features list drawer).

Fixes #3412

### DIFF
--- a/app/qml/components/MMListDrawer.qml
+++ b/app/qml/components/MMListDrawer.qml
@@ -18,7 +18,7 @@ MMDrawer {
   property bool showFullScreen: false
   property alias emptyStateDelegate: emptyStateDelegateLoader.sourceComponent
 
-  interactive: false
+  interactive: !listViewComponent.interactive
 
   drawerContent: Item {
     width: parent.width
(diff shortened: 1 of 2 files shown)


app/qml/components/MMListDrawer.qml
```diff
@@ -18,6 +18,8 @@ MMDrawer {
   property bool showFullScreen: false
   property alias emptyStateDelegate: emptyStateDelegateLoader.sourceComponent
 
+  interactive: false
```
tomasMizera, line 21: Let's make it not interactive only when the list view is interactive

tomasMizera (changes requested):

tomasMizera (approved):

PR #3421, opened 2024-05-13
Update code layout according to the new astyle version

tomasMizera:
There is a new astyle version installed by `brew` -> `Artistic Style Version 3.4.15`


tomasMizera:
Scratch that, let's revert to previous astyle version

PR #3447, opened 2024-05-19
Android: fused location provider

wonder-sk:
Initial version of the support for fused location provider:

- fused provider is opt-in only: nothing should change for the existing users if they do not switch to the "fused" provider. The app remembers whether user wants fused provider or not. The usual position provider is called just "Internal", the fused one is called "Internal (fused)"
- if the fused location provider is not available (due to Google Play services missing, outdated, disabled or invalid), we detect that and fall back to the internal gps-only provider (based on Qt Positioning)
- in a future release, we may want to switch to fused location as the default
- the implementation also adds "Internal (gps)" provider, which is based on Android's `LocationManager` just like the implementation in Qt Positioning - it is currently left disabled to avoid too many options.

| Default (fused available) | Default (fused not available) | Using fused |
| ------------- | ------------- | -- |
| Screenshot_20240604_145711 | Screenshot_20240604_145022 | Screenshot_20240604_145720 |

### DIFF
--- a/app/android/src/uk/co/lutraconsulting/MMAndroidPosition.java
+++ b/app/android/src/uk/co/lutraconsulting/MMAndroidPosition.java
@@ -1,0 +1,190 @@
+package uk.co.lutraconsulting;
+
+import android.Manifest;
+import android.content.Context;
+import android.content.pm.PackageManager;
+import android.location.GnssStatus;
+import android.location.Location;
+import android.location.LocationListener;
+import android.location.LocationManager;
+import android.os.Looper;
+import android.os.Handler;
+import android.util.Log;
+
+import androidx.annotation.NonNull;
+import androidx.core.app.ActivityCompat;
+
+import com.google.android.gms.common.ConnectionResult;
+import com.google.android.gms.common.GoogleApiAvailability;
+import com.google.android.gms.location.FusedLocationProviderClient;
+import com.google.android.gms.location.LocationCallback;
+import com.google.android.gms.location.LocationRequest;
+import com.google.android.gms.location.LocationResult;
+import com.google.android.gms.location.LocationServices;
+import com.google.android.gms.location.Priority;
+
+
+
+public class MMAndroidPosition {
+
+    static public abstract class Callback {
+        public void onPositionChanged(@NonNull Location location, GnssStatus gnssStatus) {
+        }
+    }
+
+    private static native void jniOnPositionUpdated(int instanceId, Location location, GnssStatus gnssStatus);
+
+    static public MMAndroidPosition createWithJniCallback(Context context, boolean useFused, int instanceId) {
+        Log.i("CPP", "[java] createWithJniCallback");
+
+        MMAndroidPosition.Callback callback = new MMAndroidPosition.Callback() {
+            @Override
+            public void onPositionChanged(@NonNull Location location, GnssStatus gnssStatus) {
+                jniOnPositionUpdated(instanceId, location, gnssStatus);
+            }
+        };
+
+        return new MMAndroidPosition(context, callback, useFused);
+    }
+
+    private final Context mContext;
+    private final LocationManager mLocationManager;
+    private final boolean mUseFused;
+    private FusedLocationProviderClient mFusedLocationClient = null;
+    private final LocationCallback mLocationCallback;
+    private final LocationListener mLocationManagerCallback;
+    private final GnssStatus.Callback mGnssStatusCallback;
+    private final MMAndroidPosition.Callback mClientCallback;
+    private boolean mFusedAvailable = false;
+    private boolean mGpsProviderAvailable = false;
+    private boolean mIsStarted = false;
+    private String mErrorMessage;
+    private GnssStatus mLastGnssStatus;
+
+    public MMAndroidPosition(Context context, MMAndroidPosition.Callback clientCallback, boolean useFused) {
+        mContext = context;
+        mClientCallback = clientCallback;
+        mUseFused = useFused;
+
+        Log.i("CPP", "[java] constructor!");
+
+        mLocationManager = (LocationManager) context.getSystemService(Context.LOCATION_SERVICE);
+
+        if (mUseFused) {
+            GoogleApiAvailability googleApiAvailability = GoogleApiAvailability.getInstance();
+            mFusedAvailable = googleApiAvailability.isGooglePlayServicesAvailable(context) == ConnectionResult.SUCCESS;
+            Log.i("CPP", "[java] fused available: " + mFusedAvailable);
+            if (mFusedAvailable) {
+                mFusedLocationClient = LocationServices.getFusedLocationProviderClient(context);
+            }
+        } else {
+            mGpsProviderAvailable = mLocationManager.isProviderEnabled(LocationManager.GPS_PROVIDER);
+            Log.i("CPP", "[java] gps provider available: " + mGpsProviderAvailable);
+        }
+
+        mLocationCallback = new LocationCallback() {
+            @Override
+            public void onLocationResult(@NonNull LocationResult locationResult) {
+                for (Location location : locationResult.getLocations()) {
+                    Log.i("CPP", "[java] FLP " + location.getLatitude() + " " + location.getLongitude());
+
+                    // call the native function!
+                    mClientCallback.onPositionChanged(location, mLastGnssStatus);
+                }
+            }
+        };
+
+        mGnssStatusCallback = new GnssStatus.Callback() {
+            @Override
+            public void onSatelliteStatusChanged(@NonNull GnssStatus status) {
+                //Log.v("LOC", "GNSS Status: " + status.getSatelliteCount() + " satellites.");
+
+                // store the satellite info
+                mLastGnssStatus = status;
+            }
+        };
+
+        mLocationManagerCallback = new LocationListener() {
+            @Override
+            public void onLocationChanged(@NonNull Location location) {
+                Log.i("CPP", "[java] GPS " + location.getLatitude() + " " + location.getLongitude());
+
+                mClientCallback.onPositionChanged(location, mLastGnssStatus);
+            }
+        };
+
+        Log.i("CPP", "[java] constructor end");
+
+    }
+
+    public String errorMessage() {
+        return mErrorMessage;
+    }
+
+    public boolean start() {
+        Log.i("CPP", "[java] start()");
+
+        if (mIsStarted)
+            return false;
+
+        Log.e("CPP", "[java] here 0");
+
+        if (mUseFused && !mFusedAvailable) {
+            mErrorMessage = "FUSED_NOT_AVAILABLE";
+            Log.e("CPP", "[java] FUSED_NOT_AVAILABLE");
+            return false;
+        }
+
+        if (!mUseFused && !mGpsProviderAvailable) {
+            mErrorMessage = "GPS_NOT_AVAILABLE";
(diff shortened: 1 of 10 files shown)


app/android/src/uk/co/lutraconsulting/MMAndroidPosition.java
```diff
@@ -0,0 +1,190 @@
...
+import androidx.core.app.ActivityCompat;
+
+import com.google.android.gms.common.ConnectionResult;
+import com.google.android.gms.common.GoogleApiAvailability;
+import com.google.android.gms.location.FusedLocationProviderClient;
+import com.google.android.gms.location.LocationCallback;
+import com.google.android.gms.location.LocationRequest;
+import com.google.android.gms.location.LocationResult;
+import com.google.android.gms.location.LocationServices;
+import com.google.android.gms.location.Priority;
```
tomasMizera, line 24: Can we even import this on Huawei? 🤔

app/android/src/uk/co/lutraconsulting/MMAndroidPosition.java
```diff
@@ -0,0 +1,190 @@
...
+
+        mLocationManager = (LocationManager) context.getSystemService(Context.LOCATION_SERVICE);
+
+        if (mUseFused) {
+            GoogleApiAvailability googleApiAvailability = GoogleApiAvailability.getInstance();
+            mFusedAvailable = googleApiAvailability.isGooglePlayServicesAvailable(context) == ConnectionResult.SUCCESS;
+            Log.i("CPP", "[java] fused available: " + mFusedAvailable);
+            if (mFusedAvailable) {
+                mFusedLocationClient = LocationServices.getFusedLocationProviderClient(context);
+            }
```
tomasMizera, line 79: Shall we fallback to GPS if fused is not available?

app/android/src/uk/co/lutraconsulting/MMAndroidPosition.java
```diff
@@ -0,0 +1,190 @@
...
+                    // call the native function!
+                    mClientCallback.onPositionChanged(location, mLastGnssStatus);
+                }
+            }
+        };
+
+        mGnssStatusCallback = new GnssStatus.Callback() {
+            @Override
+            public void onSatelliteStatusChanged(@NonNull GnssStatus status) {
+                //Log.v("LOC", "GNSS Status: " + status.getSatelliteCount() + " satellites.");
```
tomasMizera, line 100: Commented out code

app/android/src/uk/co/lutraconsulting/MMAndroidPosition.java
```diff
@@ -0,0 +1,190 @@
...
+
+                // store the satellite info
+                mLastGnssStatus = status;
+            }
+        };
+
+        mLocationManagerCallback = new LocationListener() {
+            @Override
+            public void onLocationChanged(@NonNull Location location) {
+                Log.i("CPP", "[java] GPS " + location.getLatitude() + " " + location.getLongitude());
```
tomasMizera, line 110: I see a bunch of logs, do you want to keep them there?

tomasMizera (commented):
Brilliant 💎

app/android/src/uk/co/lutraconsulting/MMAndroidPosition.java
wonder-sk: I believe that should be possible. The check whether the play services are actually available is in the run-time code...

app/android/src/uk/co/lutraconsulting/MMAndroidPosition.java
wonder-sk: Updated the code to fall back to Qt implementation if fused is not available.

app/android/src/uk/co/lutraconsulting/MMAndroidPosition.java
wonder-sk: I would keep them, at least for now until we fine tune this :slightly_smiling_face:

app/android/src/uk/co/lutraconsulting/MMAndroidPosition.java
wonder-sk: Removed

app/position/providers/androidpositionprovider.cpp
```diff
@@ -0,0 +1,277 @@
...
+          this->setState( tr( "Approximate location only!" ), State::NoConnection );
+        else
+          this->setState( tr( "No location permissions" ), State::NoConnection );
+      }
+    } );
+    return;
+  }
+
+  // TODO: this should not be needed?
+  AndroidPositionProvider::startUpdates();
```
tomasMizera, line 206: Todo

app/position/providers/androidpositionprovider.cpp
```diff
@@ -0,0 +1,277 @@
...
+  jobject context = QNativeInterface::QAndroidApplication::context();
+
+  mAndroidPos = QJniObject::callStaticObjectMethod( "uk/co/lutraconsulting/MMAndroidPosition", "createWithJniCallback",
+                "(Landroid/content/Context;ZI)Luk/co/lutraconsulting/MMAndroidPosition;", context, mFused, mInstanceId );
+
+  // Request permissions if needed
+
+  QLocationPermission perm;
+  perm.setAccuracy( QLocationPermission::Precise );
+  if ( qApp->checkPermission( perm ) != Qt::PermissionStatus::Granted )
```
tomasMizera, line 176: Will this now prompt for permissions again at app startup? It currently prompts user when a project is opened.

app/position/providers/androidpositionprovider.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+  if ( !res )
+  {
+    QJniObject errMsgJni = mAndroidPos.callObjectMethod( "errorMessage", "()Ljava/lang/String;" );
+    QString errMsg = errMsgJni.toString();
+    __android_log_print( ANDROID_LOG_INFO, "CPP", "[c++] error: %s", errMsg.toUtf8().constData() );
+    if ( errMsg == "MISSING_PERMISSIONS" )
+      setState( tr( "No location permissions" ), State::NoConnection );
+    else if ( errMsg == "FUSED_NOT_AVAILABLE" )
+      setState( tr( "Fused location not available" ), State::NoConnection );
```
tomasMizera, line 253: Shall we have these strings `MISSING_PERMISSIONS` and `FUSED_NOT_AVAILABLE` defined as `const strings`? Alternatively they could be in the `AbstractPositionProvider`

app/position/positionkit.cpp
```diff
@@ -105,6 +109,25 @@ AbstractPositionProvider *PositionKit::constructProvider( const QString &type, c
       QQmlEngine::setObjectOwnership( provider, QQmlEngine::CppOwnership );
       return provider;
     }
+#ifdef ANDROID
+    else if ( id == QStringLiteral( "android_fused" ) || id == QStringLiteral( "android_gps" ) )
+    {
+      bool fused = ( id == QStringLiteral( "android_fused" ) );
+      if ( fused && !AndroidPositionProvider::isFusedAvailable() )
+      {
+        // TODO: inform user + use AndroidPositionProvider::fusedErrorString() output?
```
tomasMizera, line 118: Todo - for later?

app/position/providers/androidpositionprovider.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+  __android_log_print( ANDROID_LOG_INFO, "CPP", "[c++] stop updates res: %d", res );
+
+}
+
+void AndroidPositionProvider::closeProvider()
+{
+  stopUpdates();
+
+  mAndroidPos = QJniObject();
```
tomasMizera, line 276: So in Java we do not need to destroy all the created objects ourselves, right?

app/position/providers/positionprovidersmodel.cpp
```diff
@@ -20,13 +24,40 @@ PositionProvidersModel::PositionProvidersModel( QObject *parent ) : QAbstractLis
...
+
+#ifdef ANDROID
+  PositionProvider internalFused;
+  internalFused.name = tr( "Internal (fused)" );
+  if ( AndroidPositionProvider::isFusedAvailable() )
+    internalFused.description = tr( "Using GPS, Wifi and sensors" );
+  else
+    internalFused.description = tr( "Not available (%1)" ).arg( AndroidPositionProvider::fusedErrorString() );
+  internalFused.providerType = "internal";
+  internalFused.providerId = "android_fused";
```
tomasMizera, line 46: How about not even including the fused provider if there is no way to get it working?

tomasMizera (commented):
Wonderful PR 🥇

app/position/providers/androidpositionprovider.cpp
```diff
@@ -0,0 +1,277 @@
...
+#include <QTimeZone>
+
+
+int AndroidPositionProvider::sLastInstanceId = 0;
+QMap<int, AndroidPositionProvider *> AndroidPositionProvider::sInstances;
+
+
+void jniOnPositionUpdated( JNIEnv *env, jclass clazz, jint instanceId, jobject locationObj, jobject gnssStatusObj )
+{
+  AndroidPositionProvider *inst = AndroidPositionProvider::sInstances[instanceId];
```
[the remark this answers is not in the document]
wonder-sk: yeah the return value would be nullptr. It should never happen, but I can change it to only get the value, without possibly inserting anything.

app/position/providers/androidpositionprovider.cpp
wonder-sk: Todo intentionally left there - the position provider interface probably could get a bit of a cleanup. It seems that in some cases, the startUpdates() is called by the client code, but in other times it is not (app startup vs when changing providers). It would be good to unify it.

app/position/positionkit.cpp
wonder-sk: Yes - we will need some updates of QML + probably how setState() in position providers works, and did not want to expand the scope of this PR too much :slightly_smiling_face:

app/position/providers/positionprovidersmodel.cpp
wonder-sk: There are a couple of error states - some of them look like things are broken and can't be fixed easily (SERVICE_MISSING, SERVICE_INVALID), and then there are other that seem fixable (SERVICE_UPDATING, SERVICE_VERSION_UPDATE_REQUIRED, SERVICE_DISABLED) ... play services API even has a bunch of calls to open an error dialog and guide user towards fixing problems (again maybe something to look into later).

I think it makes sense to have it there even if the fused provider is permanently unavailable, so users at least know about it, and can get a better device next time :smile:

app/position/providers/androidpositionprovider.cpp
wonder-sk: My understanding is that while we have QJniObject, it holds reference to the Java object, and when it releases the reference, then the garbage collector may eventually destroy the object at some point later.

app/position/providers/androidpositionprovider.cpp
wonder-sk: Fixed in adfa992 ... (it was also causing a crash when prompting permissions upon start)

wonder-sk:
@tomasMizera hopefully the last bits should be sorted

tomasMizera (approved):
Looking goooood

tomasMizera:
CI failures are unrelated

jozef-budac:
Verified, I can see the fused location provided according to the screenshot. I also verified that I can switch to it and that the application remembers the user's preference.

PR #3450, opened 2024-05-21
Handling attempt to upload a project after reaching the project limit

### LINKED ISSUE #3403: Empty entries created when creating and deleting a local project that could not be uploaded due to project limits
RC4 for 2024.1.0

There are two existing projects in the workspace (the limit for projects in this workspace).
What I did:
1. Created a new project
2. Tried to upload it to the server - there is the message about the maximum number of projects, so cannot upload
3. Removed the local project from the device
4. Switched to the **Projects** tab and back the **Home** tab
In the Home tab, there is now an empty project entry in **Downloaded projects** - when I click on the three dots, the message about the maximum number of projects appears again. When clicking the three dots again,nothing happens.

5. Switched between the **Projects** and **Home** tabs again, there are now 2 empty entries.

The empty entries dissapear when closing the app and opening it back again.

https://github.com/MerginMaps/mobile/assets/94905350/e62a708a-03b7-4548-9e6c-0cc1c54a13f7

VitorVieiraZ:
When attempting to upload a local only project, if the workspace project limit has been reached, the project is now removed from the synchronization pending list and is completely deleted if the project is removed by the user.

https://github.com/MerginMaps/mobile/assets/155513369/73233041-f390-4cc7-b0c3-9a7b7402fe09

Fixes #3403

### DIFF
--- a/app/synchronizationmanager.cpp
+++ b/app/synchronizationmanager.cpp
@@ -233,8 +233,8 @@ void SynchronizationManager::onProjectSyncProgressChanged( const QString &projec
 
 void SynchronizationManager::onProjectCreated( const QString &projectFullName, bool result )
 {
-  // 'projectFullName' is in the format "namespace/projectName" and
-  // 'mSyncProcess' stores projects in the format "projectName" only
+  // 'projectFullName' is in the format "namespace/projectName" and 'mSyncProcess' stores
+  // projects that were not previously uploaded to the server in the format "projectName".
   QString projectNamespace, projectName;
   MerginApi::extractProjectName( projectFullName, projectNamespace, projectName );
 
--- a/app/synchronizationmanager.cpp
+++ b/app/synchronizationmanager.cpp
@@ -233,9 +233,11 @@ void SynchronizationManager::onProjectSyncProgressChanged( const QString &projec
 
 void SynchronizationManager::onProjectCreated( const QString &projectFullName, bool result )
 {
-  if ( !result && mSyncProcesses.contains( projectFullName ) )
+  QString projectName = projectFullName.section( '/', 1, 1 );
+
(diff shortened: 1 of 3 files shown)


tomasMizera (commented):
Do we know why the project is not valid in the first place? Can you check that please?

core/merginapi.cpp
```diff
@@ -1181,7 +1181,7 @@ void MerginApi::createProjectFinished()
 
     CoreUtils::log( "create " + projectFullName, message );
 
-    emit projectCreated( projectFullName, false );
+    emit projectCreated( projectName, false );
```
tomasMizera, line 1185: This does not sound like a good idea, the signal is used in other handlers too and we might break them this way. I would keep the `projectFullName` and parse the `projectName` out of it in your handler instead :)

tomasMizera (changes requested):

tomasMizera (commented):
Just a small comment, otherwise looks good to me!

app/synchronizationmanager.cpp
```diff
@@ -230,6 +231,19 @@ void SynchronizationManager::onProjectSyncProgressChanged( const QString &projec
 
 }
 
+void SynchronizationManager::onProjectCreated( const QString &projectFullName, bool result )
+{
+  // 'projectFullName' is in the format "namespace/projectName" and
+  // 'mSyncProcess' stores projects in the format "projectName" only
```
tomasMizera, line 237: Just a minor adjustment, `mSyncProcesses` usually stores `namespace/projectName`, it is only when the project was previously not on the server altogether (went through the `createProject` method) that we use just the `projectName` form.

```suggestion
  // 'mSyncProcess' stores projects that were not previously uploaded to the server in the format "projectName".
```

PR #3455, opened 2024-05-23
Fixing override of relations editor bug

### LINKED ISSUE #3452: Override of relations editor does not work
We recently introduced an override for relations to be the word mode instead of the gallery mode when 'nogallery' string is a part of the editor title.

The condition is wrong though, we cannot check `field` here as this is a relation and not `QgsField`. We need to check `FormItem::name` instead.

See https://github.com/MerginMaps/mobile/blob/baf5c4c940386eb91b88bbbf5f4cfaabf85599bb/app/inpututils.cpp#L1128-L1134

VitorVieiraZ:
Fixing check for whether `FormItem::name` has `nogallery` flag in the form's label name in `Relation Editor` widget type.

Fixes #3452

### DIFF
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -1046,7 +1046,7 @@ const QUrl InputUtils::getThemeIcon( const QString &name )
   return QUrl( path );
 }
 
-const QUrl InputUtils::getFormEditorType( const QString &widgetNameIn, const QVariantMap &config, const QgsField &field, const QgsRelation &relation )
+const QUrl InputUtils::getFormEditorType( const QString &widgetNameIn, const QVariantMap &config, const QgsField &field, const QgsRelation &relation, const QString &fieldName )
 {
   QString widgetName = widgetNameIn.toLower();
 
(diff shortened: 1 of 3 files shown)


app/inpututils.cpp
```diff
@@ -1129,7 +1129,7 @@ const QUrl InputUtils::getFormEditorType( const QString &widgetNameIn, const QVa
     }
 
     // Mind this hack - fields with `no-gallery-use` won't use gallery, but normal word tags instead
-    if ( field.name().contains( "nogallery", Qt::CaseInsensitive ) || field.alias().contains( "nogallery", Qt::CaseInsensitive ) )
+    if ( fieldName.contains( "nogallery", Qt::CaseInsensitive ) || field.alias().contains( "nogallery", Qt::CaseInsensitive ) )
```
tomasMizera, line 1132: Do not check `field.alias` as that one is empty too

```suggestion
    if ( fieldName.contains( "nogallery", Qt::CaseInsensitive ) )
```

app/inpututils.cpp
```diff
@@ -1046,7 +1046,7 @@ const QUrl InputUtils::getThemeIcon( const QString &name )
   return QUrl( path );
 }
 
-const QUrl InputUtils::getFormEditorType( const QString &widgetNameIn, const QVariantMap &config, const QgsField &field, const QgsRelation &relation )
+const QUrl InputUtils::getFormEditorType( const QString &widgetNameIn, const QVariantMap &config, const QgsField &field, const QgsRelation &relation, const QString &fieldName )
```
tomasMizera, line 1049: Let's rename to `editorTitle` so that it does not sound the same as `field.name` :) 

```suggestion
const QUrl InputUtils::getFormEditorType( const QString &widgetNameIn, const QVariantMap &config, const QgsField &field, const QgsRelation &relation, const QString &editorTitle )
```

tomasMizera (changes requested):
Just a small comments, otherwise it looks allright!

tomasMizera:
As regards to the fix itself, I tested it and it looks all good :) 
image

tomasMizera (approved):
Thanks for adjusting the code! :)

PR #3462, opened 2024-05-24
Fix ANR when opening project with searchbar focus

### LINKED ISSUE #3027: Downloading and then opening projects causes app to freeze
At least one other user with a Samsung A53 has reported similar issues and I have been able to replicate this on my own device of the same model. 

When downloading a project from the server in the Mergin Maps app, the app will frequently freeze or become unresponsive after the user clicks on the project to open it ([see video](https://drive.google.com/file/d/1AA2QT9wtDNJW41cjZav5fEDW9lyyKYuR/view?usp=sharing)). The steps to reproduce this behaviour are as follows:

1. Open the project menu 
2. Go to the search for projects tab
3. Download a new project
4. Click on project when it finishes downloading
5. App crashes
6. Close app
7. Open app
8. Go to project menu
9. Open recently downloaded project from the home tab

This behaviour is random and does not appear to be project specific. When testing this on publicly available projects from the documentation I get the following results:

- sarah/Basic survey: crash
- documentation/forms_one-to-many-relations: no crash
- documentation/form_setup: crash
- documentation/forms_multiple_photos: no crash

However, if I remove one of the projects that did not crash during the first test, sometimes it will crash on the second attempt.

Creating a new project using the mobile app does not cause the app to crash but creating a new project using the default template in QGIS and then downloading and opening it on the mobile device will sometimes cause the app to crash.

tomasMizera:
Weird bug here.. When we were trying to set `visible` property of the entire project controller (`MMProjectController.qml`) to false while `searchBar` had focus, the app would freeze (ANR). I spent a few hours debugging and have not found anything interesting. I believe this is coming from Qt (most likely https://bugreports.qt.io/browse/QTBUG-123876) as it started to appear in Qt 6.6.0 and even more in 6.6.3.
Ideally, the best fix would be to have the `MMProjectController` instantiated dynamically when needed and not play around with the `visible` property. However, that would be a massive change for now.

Resolves #3027 

Thanks @raherin for all the details in the bug report, they helped a lot! Can you please try this build and see if it fixes the issue also for your phone?

### DIFF
--- a/app/qml/project/MMProjectHomeTab.qml
+++ b/app/qml/project/MMProjectHomeTab.qml
@@ -142,6 +142,19 @@ Item {
     }
 
     onOpenProjectRequested: function( projectFilePath ) {
+
+      //
+      // There was an issue when closing the projects controller while having the searchbar textField focused
+      // (not neccessarily with keyboard opened). This is a kind-of workaround and hotfix.
+      // It might be related to https://bugreports.qt.io/browse/QTBUG-123876 as it started
+      // to occur more frequently when we upgraded to Qt 6.6.3
+      //
+      // See https://github.com/MerginMaps/mobile/issues/3027
+      //
+      if ( searchBar.textField.activeFocus ) {
+        searchBar.textField.focus = false
+      }
+
       root.openProjectRequested( projectFilePath )
     }
     onShowLocalChangesRequested: function( projectId ) {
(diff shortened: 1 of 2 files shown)


tomasMizera:
@raherin you will find the build here https://github.com/MerginMaps/mobile/actions/runs/9223933873?pr=3462

tomasMizera:
Tested by @raherin

wonder-sk (approved):

VitorVieiraZ (approved):

PR #3469, opened 2024-05-27
Open sync failed dialog when "willRetry" notification is clicked

### LINKED ISSUE #3451: In maintenance mode, the appr after restart doesn't show a notification about it.
In maintenance mode, the appr after restart doesn't show a notification about it.
When the server gets into maintenance mode, firstly it shows a red notification about maintenance mode. However, after the app restarts, the notification is not displayed. Only on the Signup page it is said, that the server is unavailable. 

Also there is a small problem, when synchronization fails, there is the message learn more, which can't be clicked

image

This but is in not released, because the maintenance mode was not released yet :)

VitorVieiraZ:
When a sync error of type "willRetry" (maintenance mode) occurs and its notification is displayed, the syncFailedDialog opens upon clicking it.

https://github.com/MerginMaps/mobile/assets/155513369/f00a57e0-015d-4f23-a54c-0fa4818e7606

Fixes #3451

### DIFF
--- a/app/notificationmodel.cpp
+++ b/app/notificationmodel.cpp
@@ -150,6 +150,12 @@ void NotificationModel::onNotificationClicked( uint id )
           emit showSwitchWorkspaceActionClicked();
           break;
         }
+        case NotificationType::ActionType::ShowSyncFailedDialog:
+        {
+          remove( id );
+          emit showSyncFailedDialogClicked();
+          break;
+        }
         default: break;
       }
     }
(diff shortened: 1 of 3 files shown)


tomasMizera:
Looking good, is this still WIP (work in progress) @VitorVieiraZ? :)

VitorVieiraZ:
> Looking good, is this still WIP (work in progress) @VitorVieiraZ? :)

All ready now :)

PR #3470, opened 2024-05-27
Custom server UX enhancement

### LINKED ISSUE #3437: Better UX for custom server - URL validation
Let's add a validation rule when setting up a custom server URL:
 - [ ] URL must end with `/` -> or even better, add it automatically if not present
 - [ ] Add example of how the URL should look like, e.g. `https://my-server-app.com/`

VitorVieiraZ:
| Default Server | Placeholder Text After Deleting Default Server |
|--------|--------|
||  |

| Server Name Without "/" Then "Confirm" | Auto-Fixed Path With "/" At The End |
|--------|--------|
||  |

Fixes #3437

### DIFF
--- a/app/qml/account/MMLoginPage.qml
+++ b/app/qml/account/MMLoginPage.qml
@@ -204,7 +204,7 @@ MMPage {
             textFieldBackground.color: __style.lightGreenColor
 
             text: root.apiRoot
-            placeholderText: qsTr( "https://my-server-app.com/" )
+            placeholderText: "https://my-server-app.com/"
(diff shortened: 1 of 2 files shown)


app/qml/account/MMLoginPage.qml
```diff
@@ -204,6 +204,7 @@ MMPage {
             textFieldBackground.color: __style.lightGreenColor
 
             text: root.apiRoot
+            placeholderText: qsTr( "https://my-server-app.com/" )
```
tomasMizera, line 207: No need of `qsTr`, we do not want this string translated :)

```suggestion
            placeholderText: "https://my-server-app.com/"
```

tomasMizera (changes requested):
Looking good, just one change please :)

tomasMizera (approved):

PR #3480, opened 2024-05-30
Fixing section headers' wrapping and eliding 

### LINKED ISSUE #3463: Form section headers do not wrap
Form section headers do not wrap, nor elide (...). Let's make sure they wrap to 3 lines and then elide (add 3 dots to the end).

image

VitorVieiraZ:
Fixes #3463

### DIFF
--- a/app/qml/form/MMFormPage.qml
+++ b/app/qml/form/MMFormPage.qml
@@ -233,7 +233,7 @@ Page {
         color: __style.lightGreenColor;
       }
 
-      Text {
+      MMComponents.MMText {
         id: sectionTitle
 
         text: section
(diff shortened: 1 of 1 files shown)


app/qml/form/MMFormPage.qml
```diff
@@ -240,8 +240,14 @@ Page {
         font: __style.h3
```
tomasMizera, line 240: Can you please change the `Text` component to `MMText`? It will set the correct line height then. Thanks! :)

tomasMizera (commented):
Looking good! :)

tomasMizera (approved):
Looking good!

PR #3490, opened 2024-06-03
Tracking service: initial prototype with fused location

wonder-sk:
- service does not run in a separate process anymore
- using FLP to get location
- seems to track well even with screen turned off

to-do:
- make the code more robust & finish things
- test with various android versions
- maybe record track to a file anyway, to prevent data loss? (commit the recorded track on app restart)
- better android notification - regularly update tracking time and distance


tomasMizera:
- Update interval -> should we keep project settings or just use a fixed one. Compare battery-efficiency on different update  frequencies
- Simulate missing Google Play (in system settings) -> Huawei phones
- Use position provider that is selected in the app

tomasMizera:
Another reason to get this sorted sooner than later: https://medium.com/@malikfarooq4321/changes-to-broadcastreceiver-in-android-14-710cb0df2133

PR #3502, opened 2024-06-11
Fixing app freeze in MMFormPage

### LINKED ISSUE #3483: App freezes during recording child feature in relations
In the current 2024.2.0 RC, the recording of the child feature of a parent makes the app freeze. This problem is also in production. The app doesn't' freeze when you don't set the name into the district. 

Project:
https://app.merginmaps.com/projects/input-testing/tc05_relations/tree

Scenario:
1) open the project
2) Hit the record button and choose the "district" layer
3) add three points for it and hit "Record". 
4) On the form set the name of the district -> this is needed for the app to freeze. 
5) Hit the + button on the Naturals field
6) record 3 points for Natural and hit "Record" -> Now the app freezes. 

Here is the recording:
https://drive.google.com/file/d/16e6UfrBq8jlYOVIMtbsXJztHy2KFFV0r/view?usp=sharing

VitorVieiraZ:
Switching between different forms on `FormPage` during feature addition or editing, and clicking a button that changes the FormPage state or opens a new drawer over the existing one (e.g., the 'Edit Geometry' button or the '+' button in relations form), would cause the application to freeze.

Fixes #3483

### DIFF
--- a/app/qml/components/MMToolbar.qml
+++ b/app/qml/components/MMToolbar.qml
@@ -120,7 +120,7 @@ Rectangle {
 
       onClicked: {
         // Prevent freezing by clearing focus from all elements
-        // See more at https://github.com/MerginMaps/mobile/issues/3389
+        // Issues #3389 and #3483
         root.forceActiveFocus()
         root.index = model.index
         model.btnData.clicked()
--- a/app/qml/form/MMFormController.qml
+++ b/app/qml/form/MMFormController.qml
@@ -91,10 +91,6 @@ Item {
       ]
 
       onStateChanged: {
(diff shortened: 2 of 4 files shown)


tomasMizera:
Would be really nice to understand why this is happening and get a minimal reproducible code that we could send to Qt guys. It will help us to see if the issue is fixed in any future Qt versions or not.

tomasMizera:
The current situation - we are still exploring all possible places for these ANRs. Unfortunately, it is not ready for testing at the moment.

tomasMizera (approved):

PR #3509, opened 2024-06-13
Reintroduction of centering to GPS when starting to record new feature

### LINKED ISSUE #3506: Reintroduction of centering to GPS when starting to record new feature
As a follow-up after https://github.com/MerginMaps/mobile/issues/3008 we want  to keep the functionality to center to GPS when going to the recording mode. There should be a new setting in the app to disable this default behavior if one wills.

How it should work:
 - when a user clicks on the "Add" button in the main toolbar, then the map canvas should center to the GPS position (if GPS position is available), unless this behavior is disabled in the app settings

IMG_6B1CBA11F9A7-1

 - there should be a new entry in the settings page under the "recording" group with title "Auto-lock position" and description "Each time you start recording, the app centers to GPS". It should be switch component and the value should be true by default. It must be stored in the `AppSettings` too.

---

See how it was implemented previously here ⬇️ https://github.com/MerginMaps/mobile/blob/4b59718bda38cb2215134ca6f713fc944f2367b8/app/qml/map/RecordingTools.qml#L303

VitorVieiraZ:
https://github.com/MerginMaps/mobile/assets/155513369/d7778a9e-b9ae-4637-b5e7-801d9840088e

Development and functionality of the feature based on: https://github.com/MerginMaps/mobile/issues/3506

Resolves  #3506

### DIFF
--- a/app/qml/map/MMMapController.qml
+++ b/app/qml/map/MMMapController.qml
@@ -895,6 +895,7 @@ Item {
       map: mapCanvas
       positionMarkerComponent: positionMarker
       recordingMapTool.centeredToGPS: root.centeredToGPS
+      centerToGPSOnStartup: __appSettings.autolockPosition
 
       activeFeature: root.state === "edit" ? internal.featurePairToEdit.feature : __inputUtils.emptyFeature()
 
--- a/app/qml/map/MMMapController.qml
+++ b/app/qml/map/MMMapController.qml
@@ -895,7 +895,6 @@ Item {
       map: mapCanvas
       positionMarkerComponent: positionMarker
       recordingMapTool.centeredToGPS: root.centeredToGPS
-      centerToGPSOnStartup: __appSettings.autolockPosition
 
       activeFeature: root.state === "edit" ? internal.featurePairToEdit.feature : __inputUtils.emptyFeature()
 
--- a/app/qml/map/MMMapController.qml
+++ b/app/qml/map/MMMapController.qml
@@ -935,6 +935,8 @@ Item {
 
         root.state = "view"
       }
+
+      onAutoLockGps: root.centeredToGPS = true
     }
   }
 
--- a/app/qml/map/MMMapController.qml
+++ b/app/qml/map/MMMapController.qml
@@ -92,15 +92,20 @@ Item {
   ]
 
   onStateChanged: {
+    // We call this first because when previous state is 'view' and `centeredToGPS` is true
+    // the map may still not be centered. It's only centered after a certain threshold is exceeded.
+    updatePosition()
+
     switch ( state ) {
       case "record": {
         if ( __appSettings.autolockPosition ) { // center to GPS
           if ( gpsStateGroup.state === "unavailable" ) {
-              __notificationModel.addError( "GPS currently unavailable." )
-              return;
+            __notificationModel.addError( "GPS currently unavailable." )
           }
-
-          root.centeredToGPS = true
+          else {
+            root.centeredToGPS = true
+            updatePosition()
+          }
         }
 
         root.recordingStarted()
@@ -140,10 +145,6 @@ Item {
         break
       }
     }
-
-    // We call this first because when previous state is 'view' and `centeredToGPS` is true
-    // the map may still not be centered. It's only centered after a certain threshold is exceeded.
-    updatePosition()
   }
 
   state: "view"
--- a/app/qml/map/MMMapController.qml
+++ b/app/qml/map/MMMapController.qml
@@ -92,59 +92,58 @@ Item {
   ]
 
   onStateChanged: {
+    switch ( state ) {
+      case "record": {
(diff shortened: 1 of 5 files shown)


app/qml/map/MMMapController.qml
```diff
@@ -895,6 +895,7 @@ Item {
       map: mapCanvas
       positionMarkerComponent: positionMarker
       recordingMapTool.centeredToGPS: root.centeredToGPS
+      centerToGPSOnStartup: __appSettings.autolockPosition
```
tomasMizera, line 898: Redundant information here, it is already set to the value of `__appSettings.autolockPosition` in `MMRecordingTools`.

tomasMizera (commented):
This is looking good, thanks @VitorVieiraZ  :rocket: 

@uclaros can I ask you to have a look at this PR too? You were adjusting the centering logic recently :)

app/qml/map/MMRecordingTools.qml
```diff
@@ -314,6 +315,19 @@ Item {
...
 
+  Component.onCompleted: {
+    if ( root.centerToGPSOnStartup ) {
+      // center to GPS
+      if ( gpsStateGroup.state === "unavailable" ) {
+          __notificationModel.addError( "GPS currently unavailable." );
+          return;
+      }
+
+      root.map.mapSettings.centeredToGPS = true;
```
uclaros, line 326: This looks strange, there's no `InputMapSettings::centeredToGPS`. Did you mean `mapTool.centeredToGPS = true`?

uclaros (commented):
What should the behavior be if location is not available yet? Just a message or should the map jump as soon as a location is available?

What should happen if the user moves before clicking record?

My take on those is that we should be setting `MMMapController.centerToGPS = true`, which means we get the green dot on the GPS button plus map position is updated when location is updated.
`recordingMapTool.centeredToGPS` seems to be bound to that so should be updated.

tomasMizera (approved):
Hi @uclaros, can you check once again please? It looks good to me. There is one change of logic - `updatePosition()` is now called after we change the state, not before. However, I could not think of a scenario where this would be an issue

app/qml/map/MMMapController.qml
```diff
@@ -92,50 +92,58 @@ Item {
...
-    // the map may still not be centered. It's only centered after a certain threshold is exceeded.
-    updatePosition()
-
     switch ( state ) {
+      case "record": {
+        if ( __appSettings.autolockPosition ) { // center to GPS
+          if ( gpsStateGroup.state === "unavailable" ) {
+              __notificationModel.addError( "GPS currently unavailable." )
+              return;
+          }
```
uclaros, line 101: We completely disable recording when GPS is unavailable and autolock is on?
In that case maybe we should also handle the case when GPS is lost while recording?

app/qml/map/MMMapController.qml
```diff
@@ -92,50 +92,58 @@ Item {
...
+      }
+
+      case "inactive": {
+        break
+      }
     }
+
+    // We call this first because when previous state is 'view' and `centeredToGPS` is true
+    // the map may still not be centered. It's only centered after a certain threshold is exceeded.
+    updatePosition()
```
uclaros, line 146: This comment is no longer true: we don't call this first, we actually call this last!
Does that mean we'll get a first point when `root.recordingStarted()` is called which may not be on the actual GPS position?
Why not call `updatePosition()` right after `root.centeredToGPS = true` ?

app/qml/map/MMMapController.qml
tomasMizera: Good point, no, we need to continue even without GPS signal

app/qml/map/MMMapController.qml
```diff
@@ -97,44 +97,53 @@ Item {
     updatePosition()
 
     switch ( state ) {
+      case "record": {
+        if ( __appSettings.autolockPosition ) { // center to GPS
+          if ( gpsStateGroup.state === "unavailable" ) {
+            __notificationModel.addError( "GPS currently unavailable." )
+          }
+          else {
+            root.centeredToGPS = true
```
uclaros, line 106: if we moved this out of the GPS availability check, does it mean the map would re-center as soon as gps position is actually available? If yes, that could be useful, but I'm OK with merging it as it is now!

VitorVieiraZ: In this scenario, it would center even if the user decides to uncheck the "auto-lock" property in the recording settings.

tomasMizera:
Unrelated test failure

app/qml/map/MMMapController.qml
uclaros: No, I mean like:
```qml
        if ( __appSettings.autolockPosition ) { // center to GPS
          root.centeredToGPS = true
          if ( gpsStateGroup.state === "unavailable" ) {
            __notificationModel.addError( "GPS currently unavailable." )
          }
          else {
            updatePosition()
          }
        }

PR #3514, opened 2024-06-20
Enhancing photo renaming with custom expressions

### LINKED ISSUE #3415: Photo renaming does not work when multiple photo fields have custom expressions
I have been able to recreate an issue reported by at least one other user where multiple photo naming expressions are used in 'Project>Properties', only the first listed expression will be applied in the app. 

I have a project with two point survey layers. The first survey layer has two photo attachment fields and the second layer has one photo attachment field. In my project directory, I have a 'photos/' folder to store photo attachments. In the project properties, I have a photo naming expression set for each attachment field, the same expression is used for each field:

`'photos/'||@mergin_username||'-'||@layer_name||to_string("count")||'_'||format_date("date",'yyyy-MM-dd_Thh.mm.ss')`

image

In the app, I add a feature to `Survey` layer and add two photo attachments
Screenshot_20240509_163349

I then add a feature to `survey2` layer and add one photo attachment
Screenshot_20240509_163428

I then synchronise the changes in the mobile app and go back to sync in QGIS, however I can see that the photo naming expression was only applied to one of the three photos I just recorded in the app
image

If I go to the project directory, I can see that two of the photos did not get saved to the /photos folder and the photo in the `Survey.photo` field is in the /photos folder with the correctly applied custom name

image

photos/
image

VitorVieiraZ:
Setting `Photo name format` in `Photo attachments configuration` previously did not accept formats beginning with '/', such as `'/photos/'||@mergin_username||'-'||@layer_name||to_string("count")||'_'||format_date("date",'yyyy-MM-dd_Thh.mm.ss')`. Now, it is possible to configure expression with leading slashes. Mergin Maps will automatically remove any slashes at the beginning of the photo name format during data processing.

Fixes #3415


tomasMizera (commented):
Looks good. Can we add a test here please?

app/attributes/attributecontroller.cpp
```diff
@@ -1511,6 +1511,11 @@ void AttributeController::renamePhotos()
         const QString src = InputUtils::getAbsolutePath( mFeatureLayerPair.feature().attribute( item->fieldIndex() ).toString(), prefix );
         QFileInfo fi( src );
         QString newName = QStringLiteral( "%1.%2" ).arg( val.toString() ).arg( fi.completeSuffix() );
+
+        // Remove leading slashes from newName following issue #3415
+        static const QRegularExpression leadingSlashes( "^/+" );
```
tomasMizera, line 1516: Can we not make this static? If you'd like to make it static, make it a member of this class instead.

```suggestion
       const QRegularExpression leadingSlashes( "^/+" );
```

test/test_data/test_photo_rename/photo2.jpg
tomasMizera, line 1: All is looking good, can you just please give here a much smaller file? It is 1.5MB, you can simply add an empty file with `*.jpg` or some tiny photo instead :)

tomasMizera (changes requested):
Almost there! :)

tomasMizera (approved):
Thanks for addressing my comments! :)

PR #3520, opened 2024-06-24
Fixing "Unique" and "Enforce Unique constraint" error message

### LINKED ISSUE #3457: "Unique" and "Enforce Unique constraint"  does not work
I have a sampling layer the auto creates the field station_id off of a prefix, geologist and station_number . And then I have set this field to be "Unique" and "Enforce Unique constraint" in QGIS. In the screenshot below (I had already taken a sample named "SM2024_Elliot_001") it appropriately will not let me save this point (note the red check mark in top right corner!), however it does not give me any warning that the station_id is the culprit! When I add station number to 2 the station_id changes appropriately, and I can save the point. But it would be helpful to get a warning pop up about Unique values just like there are warnings for Null values!
image

VitorVieiraZ:
'Unique' and 'Enforce Unique Constraint' error message would not work when 'Apply Default Value on Update' option was checked after setting a 'Default Value' in 'Defaults' tab.

Fixes #3457


app/qml/form/MMFormPage.qml
```diff
@@ -281,7 +281,11 @@ Page {
         property var fieldWidget: model.EditorWidget
         property var fieldConfig: model.EditorWidgetConfig
 
-        property bool fieldIsReadOnly: root.state === "readOnly" || !AttributeEditable
+        // Following issue #3457, error message can be displayed even for non-editable attributes
+        // Thus, we split attributes indicating edit mode for formPage and attribute editability
+        property bool fieldIsReadOnly: root.state === "readOnly"
```
tomasMizera, line 286: Let's keep the naming consistency here and rename the `fieldIsReadOnly` to `fieldFormIsReadOnly`

app/qml/form/MMFormPage.qml
```diff
@@ -281,7 +281,11 @@ Page {
...
+        property bool isAttributeEditable: AttributeEditable
```
tomasMizera, line 287: Let's keep the naming convention here - all properties injected into this loader start with `fieldXYZ`

```suggestion
        property bool fieldIsEditable: AttributeEditable // Field is not editable when "Apply default value on update" is set
```

app/qml/form/MMFormPage.qml
```diff
@@ -281,7 +281,11 @@ Page {
         property var fieldWidget: model.EditorWidget
         property var fieldConfig: model.EditorWidgetConfig
 
-        property bool fieldIsReadOnly: root.state === "readOnly" || !AttributeEditable
+        // Following issue #3457, error message can be displayed even for non-editable attributes
+        // Thus, we split attributes indicating edit mode for formPage and attribute editability
```
tomasMizera, line 285: I don't think we need this comment when we rename the properties appropriately :)

app/qml/form/editors/MMFormCalendarEditor.qml
```diff
@@ -58,7 +59,7 @@ MMPrivateComponents.MMBaseSingleLineInput {
   hasCheckbox: _fieldRememberValueSupported
   checkboxChecked: _fieldRememberValueState
 
-  readOnly: _fieldIsReadOnly
+  readOnly: _fieldFormIsReadOnly
```
tomasMizera, line 62: Sorry @VitorVieiraZ, I had a second thought about it and I believe we need to approach it from a slightly different perspective. 
Our current solution would keep fields with "Apply default value on update" in editable state. This would show icons / actions of the field as if one could click on them, see 
image

But in reality, they should be disabled.

We need to keep such fields disabled though.
image

---

Thus, I suggest to do the following:
 - add `property bool shouldShowValidationMsg` to `MMBaseInput` and use it as a condition here https://github.com/MerginMaps/mobile/blob/1c823dc8c732936392fda0f17b7d8ea056215716/app/qml/components/private/MMBaseInput.qml#L173 instead of the `root.editState === "enabled"`
 - then set the `shouldShowValidationMsg` in every form editor to `!_fieldFormIsReadOnly`
 - then revert the value of the `readOnly` in all editors to `readOnly: _fieldFormIsReadOnly || !_fieldIsEditable`
 - with this, we do not need to change anything in `MMBaseSingleLineInput.qml` and validation messages will be visible for all fields, no matter if they are editable or not.

This solution is more flexible, does not change already working workflows and should not bring any unwanted changes 🤞🏻.

tomasMizera (changes requested):
See my comments

tomasMizera (changes requested):
We are getting there, there should be only two things left to do. 

1. Fix gallery, please open the gallery - page 'editors'. There are some errors due to these new properties
2. When `shouldShowValidationMsg` is `false`, we should keep the `validation` state `valid`.  For this we simply need to adjust the condition for the `Valid` state of the `validationStateGroup` to consider also the `shouldShowValidationMsg` flag. ... Maybe `shouldShowValidation` would be a better name then :)

tomasMizera (approved):
Looking good! 🚀

PR #3523, opened 2024-06-26
Report available storage before syncing in diagnostic logs

### LINKED ISSUE #3522: Report available storage before syncing in diagnostic logs
Let's report available storage to the diagnostic log when sync is starting. Bonus points if we can report also how much data are we about to download.

Would help a lot in support.

VitorVieiraZ:
Available storage is now added to diagnostic logs at the start of syncing.

Download and upload information is already available and logged in:
https://github.com/MerginMaps/mobile/blob/33ff5a8da0b04a55a83056cced9a5ed9952a8a1b/core/merginapi.cpp#L2418

https://github.com/MerginMaps/mobile/blob/33ff5a8da0b04a55a83056cced9a5ed9952a8a1b/core/merginapi.cpp#L2746

Resolves #3522

### DIFF
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -535,27 +535,7 @@ QString InputUtils::filesToString( QList<MerginFile> files )
 
 QString InputUtils::bytesToHumanSize( double bytes )
 {
-  const int precision = 1;
-  if ( bytes < 1e-5 )
-  {
-    return "0.0";
-  }
-  else if ( bytes < 1024.0 * 1024.0 )
-  {
-    return QString::number( bytes / 1024.0, 'f', precision ) + " KB";
-  }
-  else if ( bytes < 1024.0 * 1024.0 * 1024.0 )
-  {
-    return QString::number( bytes / 1024.0 / 1024.0, 'f', precision ) + " MB";
-  }
-  else if ( bytes < 1024.0 * 1024.0 * 1024.0 * 1024.0 )
-  {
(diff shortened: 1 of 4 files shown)


tomasMizera (commented):
Hi @VitorVieiraZ, thanks for working on this one 👀  However, it seems that I did not give a good description of the issue. What I meant by the available storage is the _device storage_ - how much storage is left on the device. The reason for that is to know whether we have enough space to do the sync or not. Imagine having 64GB mobile that already has 61GB used. User could not sync a project that is 5GB in size..

We do not handle such a case neatly nowadays, it will help us a lot in understanding some failed syncs.

app/synchronizationmanager.cpp
```diff
@@ -311,3 +314,9 @@ void SynchronizationManager::onProjectReloadNeededAfterSync( const QString &proj
   }
 
 }
+
+QString SynchronizationManager::getAvailableStorage()
+{
+  double availableStorage = mMerginApi->workspaceInfo()->storageLimit() - mMerginApi->workspaceInfo()->diskUsage();
+  return InputUtils::bytesToHumanSize( availableStorage );
+}
```
tomasMizera, line 322: You might want to put this function into `MerginApi` when the only thing it does is calling its API :)

tomasMizera:
We should be able to use this class https://doc.qt.io/qt-6/qstorageinfo.html#details

tomasMizera (approved):
This is looking good 👍🏻  Have you had a chance to test it on your phones?

tomasMizera (approved):
Looking good! :)

PR #3524, opened 2024-06-28
Handling XLSX and DOCX files in file viewing system for newer Android versions

VitorVieiraZ:
Changes in `AndroidManifest.xml` to ensure that newer versions of Android can open more types of documents, such as XLSX, DOCX, and PNG, provided there is an application that can open these documents. See [reference.](https://stackoverflow.com/questions/11068648/launching-an-intent-for-file-and-mime-type)


tomasMizera:
Thanks Vitor, do we know why we need to set this explicitly only for these two extensions?

tomasMizera (approved):
Let's see what testing brings 🚀

PR #3532, opened 2024-07-09
Handling accidental double-taps on buttons with MMSingleClickMouseArea

### LINKED ISSUE #3012: Accidental double-taps on buttons in relation or image widgets
A user discovered a problem when they accidentally double-tap the `Add` button in a relation widget in the MM mobile app - two features get added and only one of them has reference to the parent feature. This can be a problem when pens or gloves need to be used - accidental double-taps are likely to happen.

The question: is it possible to disable this double-taps in the sense when two taps are found to be too close in time to each-other, one should be ignored?

The same behaviour can be reproduced when adding a photo (either from camera or gallery) - if you double-tap on the cam or gal button, two images are added but only one (the latter) is properly referenced to the feature.

VitorVieiraZ:
New component `MMSingleClickMouseArea` for single-click detection. It temporarily disables further clicks using a Timer to prevent multiple triggers shortly after the first click.

Fixes #3012

### DIFF
--- a/app/qml/CMakeLists.txt
+++ b/app/qml/CMakeLists.txt
@@ -50,6 +50,7 @@ set(MM_QML
     components/MMText.qml
     components/MMToolbar.qml
     components/MMToolbarButton.qml
+    components/MMSingleClickMouseArea.qml
     components/private/MMBaseInput.qml
     components/private/MMBaseSingleLineInput.qml
     components/private/MMToolbarLongButton.qml
--- a/app/qml/components/MMSingleClickMouseArea.qml
+++ b/app/qml/components/MMSingleClickMouseArea.qml
@@ -12,6 +12,7 @@ import QtQuick
 /**
   * MMSingleClickMouseArea enhances MouseArea by preventing multiple clicks within a
   * specified time frame, ensuring consistent behavior and avoiding unintended actions
+  * Similar to MouseArea, but it should be used with onSingleClicked signal handling instead of onClicked
 */
 
 MouseArea {
--- a/app/qml/components/MMSingleClickMouseArea.qml
+++ b/app/qml/components/MMSingleClickMouseArea.qml
@@ -15,7 +15,10 @@ MouseArea {
   signal singleClicked()
 
   onClicked: {
-    if ( !root.enabled ) return;
+    if ( !root.enabled ) {
+      mouse.accepted = true;
+      return;
+    }
 
     root.enabled = false;
     singleClicked()
--- a/app/qml/components/MMSingleClickMouseArea.qml
+++ b/app/qml/components/MMSingleClickMouseArea.qml
@@ -1,0 +1,22 @@
+import QtQuick
+
+MouseArea {
+  id: mouseArea
+  property alias clicked: onClicked
+  signal singleClicked()
+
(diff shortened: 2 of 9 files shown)


app/qml/components/MMSingleClickMouseArea.qml
```diff
@@ -0,0 +1,34 @@
...
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+
+import QtQuick
+
```
tomasMizera, line 11: Can you please add a comment about why we want such a class? + how to use it :)

app/qml/CMakeLists.txt
```diff
@@ -50,6 +50,7 @@ set(MM_QML
     components/MMText.qml
     components/MMToolbar.qml
     components/MMToolbarButton.qml
+    components/MMSingleClickMouseArea.qml
```
tomasMizera, line 53: I believe you also need to add this class to the gallery `.qrc` file

app/qml/components/MMSingleClickMouseArea.qml
```diff
@@ -0,0 +1,40 @@
...
+  signal singleClicked()
+
+  onClicked: {
+    if ( !root.enabled ) {
+      mouse.accepted = true;
+      return;
+    }
+
+    root.enabled = false;
```
tomasMizera, line 29: If you set `enabled` property to `false`, you should not receive the `clicked` signal at all, do you?

app/qml/form/editors/MMFormRelationEditor.qml
tomasMizera, line 1: We need to use `MMSingleClickMouseArea` also when creating a new child feature via the gallery mode

tomasMizera (commented):
This is looking good now, thanks! I found one more place that would benefit from this new component. Can we please add it when opening child feature in relation form editors? It is again in both relation form editors when calling `openLinkedFeature` signal.

app/qml/form/editors/MMFormGalleryEditor.qml
tomasMizera, line 1: Heh, sorry again, but we should use this also when opening the child feature (similarly to how you added it in `MMFormRelationEditor`, here: https://github.com/MerginMaps/mobile/pull/3532/files#diff-6fe89969eb421213908ef7fc7a22ecd2460a61f19d367a06887f2b049fbf84f6R138).

tomasMizera (commented):
One last remaining place and I think we are good to go :)

app/qml/form/editors/MMFormGalleryEditor.qml
tomasMizera: What I meant here is this place https://github.com/MerginMaps/mobile/blob/f0b69c6e58f30fb85ce7467c5b985e9d877e3c99/app/qml/form/editors/MMFormGalleryEditor.qml#L65

tomasMizera (approved):
Thanks for all the updates!

PR #3543, opened 2024-07-22
Introduce a FeaturesProxyModel to conditionally sort features list and value relations

### LINKED ISSUE #1199: Sorting of the list in "Browse features"
Currently we do not have any order for layer features in the list.

It would be good to alphabetically order the features, so it is easier for people to navigate through the list (when not using search).

Or maybe it should be upon a request to sort the list ascending/descending based on a column?

### LINKED ISSUE #1482: Value Relation Widget: Order by Value not working
When the Order by Value option is ticked within the widget settings, the list of options are sorted in QGIS, but within input it is not sorted. Example below:
SortingIssue

uclaros:
- `MMFeaturesListPage` is sorted according to the layer's `QgsAttributeTableConfig`
- `MMFormValueRelationEditor` is sorted by value when the widget has the `OrderByValue` config set

Fixes #1199 
Fixes #1482

~needs some tests~


app/featuresmodel.cpp
```diff
@@ -359,10 +370,27 @@ void FeaturesModel::setLayer( QgsVectorLayer *newLayer )
     }
 
     emit layerFeaturesCountChanged( layerFeaturesCount() );
+    setupSorting();
```
tomasMizera, line 373: Can we call the `setupSorting()` method from the `FeaturesProxyModel` instead of calling it here? That way we would not need to clear the search expression in other models that should not be sorted.

tomasMizera:
Thanks for this @uclaros! I have a couple of minor comments, but it looks good overall

app/featuresmodel.cpp
uclaros: That's a great suggestion! I really didn't like having to override `setupSorting()` for models that don't use it...

app/featuresproxymodel.cpp
```diff
@@ -0,0 +1,41 @@
...
+}
+
+FeaturesModel *FeaturesProxyModel::featuresSourceModel() const
+{
+  return mModel;
+}
+
+void FeaturesProxyModel::setFeaturesSourceModel( FeaturesModel *sourceModel )
+{
+  if ( mModel == sourceModel )
```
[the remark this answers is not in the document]
uclaros: Hmm, we probably should, indeed.
The logic was copied from `ProjectsProxyModel`, which also doesn't disconnect.
https://github.com/MerginMaps/mobile/blob/0f8a80119054816a5ff9cecc1ec6e8a4a6a24cf7/app/projectsproxymodel.cpp#L61-L68

I guess this didn't matter so far as we're only calling `setProjectsSourceModel`/`setFeaturesSourceModel` once.

app/featuresmodel.cpp
uclaros: However, this will not work correctly _if_ we use a`FeaturesProxyModel` with `RelationFeaturesModel` or `RelationReferenceFeaturesModel` that should not be sorted (which we shouldn't do anyway)

tomasMizera:
Thanks for addressing my comments @uclaros, how about tests? 😇

uclaros:
Test data were already pretty much sorted, so I modified the `subsub` gpkg layer to look like:
image

app/featuresproxymodel.cpp
```diff
@@ -0,0 +1,49 @@
...
+    return;
+
+  if ( mModel )
+    disconnect( mModel, nullptr, this, nullptr );
+
+  mModel = sourceModel;
+  connect( mModel, &FeaturesModel::fetchingResultsChanged, this, [ = ]( bool pending ) { if ( !pending ) initialize(); } );
```
tomasMizera, line 48: Do we need to call `initialize` all the time? Wouldn't it be enough to call `sort()`? :)

tomasMizera (approved):
Thanks for such thought-through test cases!

tomasMizera:
Looking good, thanks for all the adjustments @uclaros! I'll mark this as frozen to wait until 2024.3.1 is out

wonder-sk:
Just in case someone wonders how the sorting order works (will be useful for docs): it is set in QGIS, by either clicking a column to sort using it in ascending/descending order, or it is even possible to set a custom sort expression from a popup menu. Sorting order gets saved in QGIS project file and should be respected by the app.

image

uclaros:
Also available in the attribute table when using _Form_ view:

image

PR #3545, opened 2024-07-24
Avoid running autotests in multiple workflows at once

tomasMizera:
This PR adds concurrency for autotests so that two Linux or macOS tests do not run simultaneously. GitHub will keep at most one job running and one in a queue (pending). If someone else requests a third run, their run will be cancelled.

It helps to avoid overwriting data between two simultaneous runs. 
> Note: One macOS and one Linux workflow can run at the same time, because they use different users

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -32,7 +32,7 @@ jobs:
     if: ( github.repository == 'MerginMaps/mobile' ) && (!contains(github.event.head_commit.message, 'Translate '))
     runs-on: ubuntu-22.04
     concurrency: 
-      group: mobileapp-autotests # Avoid running autotests at the same time across multiple workflows, because the same user is used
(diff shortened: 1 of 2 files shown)


.github/workflows/linux.yml
```diff
@@ -22,6 +22,7 @@ env:
   CMAKE_VERSION: '3.29.0'
   CACHE_VERSION: 0
       
+# Execute only the latest CI run on a branch
 concurrency:
   group: ci-${{github.ref}}-linux
```
PeterPetrik, line 27: Shouldn't we have the concurency defined right here and cancel only if branch is not master?

concurrency:
  group: mobileapp-autotests-linux # Avoid running autotests at the same time across multiple linux workflows
  cancel-in-progress: ${{ !contains(github.ref, 'master')}} # Cancel in progress for branched outside master. Always run on master

tomasMizera:
Ok, let's park it for now and introduce a proper fix on the server side

PR #3546, opened 2024-07-25
Fix TabBar issues with multiple tabs

tomasMizera:
We kept having issues with TabBar when project had multiple tabs, for example:
Screenshot 2024-07-25 at 13 37 46

Solution: use `ListView` instead of `TabBar` 🙃 

ListView is set to mimic the behaviour of TabBar (horizontal orientation), the same project now looks like this:
Screenshot 2024-07-25 at 14 59 08

There is a behavioural change for sections - I limited it to use only `ViewSection.InlineLabels`, because `CurrentLabelAtStart` was also producing some issues (see https://doc.qt.io/qt-6/qml-qtquick-listview.html#section-prop)

### DIFF
--- a/app/mmstyle.h
+++ b/app/mmstyle.h
@@ -289,6 +289,7 @@ class MMStyle: public QObject
     Q_PROPERTY( double row24 READ number24 CONSTANT )
     Q_PROPERTY( double row36 READ number36 CONSTANT )
     Q_PROPERTY( double row40 READ number40 CONSTANT )
+    Q_PROPERTY( double row45 READ number45 CONSTANT )
     Q_PROPERTY( double row49 READ number49 CONSTANT )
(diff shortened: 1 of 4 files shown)


VitorVieiraZ (approved):
I tested it in some projects that use tabs, and everything seems good here.

PeterPetrik (approved):
LGTM - sad to see another Qt bug in core components like TabBar

tomasMizera:
@jozef-budac note for testing: we will need to test tabs behaviour - dynamic hiding and so on

jozef-budac:
We have this in the regression run. So I believe it should be covered. One issue is probably related to it: https://github.com/MerginMaps/mobile/issues/3552. It looks like an iOS problem. I can see it on android.

PR #3553, opened 2024-07-31
Fix #2320 Split geometry visual problems

### LINKED ISSUE #2320: Split geometry visual problems
There are a few problems with the split geometry in the latest version:

1. No (or sometimes wrong-coloured) guideline (❌)
2. ~~Floating `Back` button is there (correct ✅), but has no action (❌)~~
3. ~~`Cancel` button is there, even though it should not be there anymore (❌)~~

Screenshot 2022-10-06 at 15 05 40

ValentinBuira:
This PR fix #2320 where the guideline from last point to cursor wasn't showing in the splitting tool

The root issue was that the `GuidelineController::buildGuideline` was cancelled at the beginning of the function, because of the `allowed` property. So I turned allowed to be always true for the splitting tool

I wanted to show clearly to the users that the actual splitting line were the one to the last point and not  to the cursor.
So I updated the visual to match the way it works when adding a new line geometry
Let me know if it's working for the user experience otherwise I will revert that part

### DIFF
--- a/app/guidelinecontroller.h
+++ b/app/guidelinecontroller.h
@@ -100,7 +100,7 @@ class GuidelineController : public QObject
     QgsGeometry mRealGeometry;
     InputMapSettings *mMapSettings = nullptr; // not owned
     Vertex mActiveVertex;
-    bool mAllowed;
+    bool mAllowed = true;
     RecordingMapTool::InsertPolicy mInsertPolicy;
     int mActivePart = 0;
     int mActiveRing = 0;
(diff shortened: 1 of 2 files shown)


tomasMizera (changes requested):
The code is looking good, I have just one suggestion. The first recorded point for the split line has a different color than the line. Can we unify that please?

ValentinBuira:
I changed the marker color to match the color of the line 

Should be ready to merge once unfrozen

tomasMizera (approved):

PR #3554, opened 2024-07-31
Fixing text wrapping in MMToolbar

### LINKED ISSUE #3547: Word Wrapping Issue in MMToolbar
In some translations, button names can become too long and cause word wrapping in `MMToolbar`:

VitorVieiraZ:
Text font for `MMToolbarShortButton` can change from t4 to t5 when the button text is too long and needs a better fit, resolvin text wrapping issue and also setting a maximum number of lines to 2, eliding it right after that.

| Portrait Before | Portrait After  |
|--------|--------|
||  |

| Landscape Before| Landscape After |
|--------|--------|
||  |

Fixes #3547

### DIFF
--- a/app/qml/components/private/MMToolbarShortButton.qml
+++ b/app/qml/components/private/MMToolbarShortButton.qml
@@ -72,14 +72,17 @@ Item {
     Text {
       id: text
 
+      width: parent.width
       text: root.text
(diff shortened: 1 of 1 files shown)


tomasMizera:
Let's keep it like this for now and I'll double check with Andrej

tomasMizera (approved):

tomasMizera:
Let's keep it like this for now :)

Issue #3525, opened 2024-06-30
Unsaved changes dialog not displayed when using back button


wonder-sk:
When I open edit form for a feature, do some edits and then press android's system back button, it will show the "Unsaved changes" dialog as expected - but if I use the top-left back button in the form, no warning gets shown and the changes are discarded...

jozef-budac:
This doesn't work well on 2024.3.1 RC at least in two scenarios

1) when the feature form is empty, the android and top left back button should close the window without any notification or confirmation, but

- top left button shows "Unsaved changes" window - bug
- The Android button does nothing - bug

2)  when I fill out the form and hit the top left back button, the "Unsaved changes" window appears. When I click the 'X' button on it to close the modal windows, from not the back Android button does nothing - bug

For the second point, here is the video:
https://github.com/user-attachments/assets/c3457b63-9b83-416f-8e69-1a7678a3897a

tomasMizera:
> This doesn't work well on 2024.3.1 RC at least in two scenarios
> 
>     1. when the feature form is empty, the android and top left back button should close the window without any notification or confirmation, but
> 
> 
>     * top left button shows "Unsaved changes" window - bug
> 
>     * The Android button does nothing - bug
> 
> 
>     2. when I fill out the form and hit the top left back button, the "Unsaved changes" window appears. When I click the 'X' button on it to close the modal windows, from not the back Android button does nothing - bug
> 
> 
> For the second point, here is the video: https://github.com/user-attachments/assets/c3457b63-9b83-416f-8e69-1a7678a3897a

Hi @jozef-budac thanks for that, we will have a look at the android button issue. The "unsaved changes" window should be there when creating a new form all the time (even if you did not change anything) as going back will discard the entire captured geometry.

ValentinBuira:
Hi @jozef-budac  thanks for the catch 

For 1, the android button bug doing nothing  when adding the feature form is empty is already present in last release 
And for 2, now the issue also appear after closing "Unsaved changes" dialog

Both got the same root issue; the *back event is handled only when the form is in focus*, so if you tap on a field again it works again

I'm working on a fix

tomasMizera:
> @tomasMizera , the "Unsaved changes" window appears when hitting the back button even for nongeo feature with an empty. form. Are we OK with that?
> Screen_Recording_20240807_082826.mp4

Yeah, I think that is fine for now, thanks for double-checking @jozef-budac

saberraz:
[...]

PR #3556, opened 2024-08-05
Fix not able to close drawer using system back button

ValentinBuira:
This PR try to fix an error where in some case you couldn't close a drawer using the system back button

This happened because the control focus was not transferred correctly 

This bug was found during release testing and relate to #3525

### DIFF
--- a/app/qml/components/MMDrawer.qml
+++ b/app/qml/components/MMDrawer.qml
@@ -48,13 +48,6 @@ Drawer {
       height: parent.height / 2
       y: parent.height / 2
     }
-
-    Keys.onReleased: function( event ) {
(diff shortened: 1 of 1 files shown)


tomasMizera (approved):

ValentinBuira:
I also fixed the same issue for MMFormPage

I have been pondering and in the end I moved the handling event to the root of MMFormPage, and I removed this part of the code which as no used anymore AFAIK: 

```
-    onVisibleChanged: function( visible ) {
-      if ( visible )
-        backHandler.forceActiveFocus()
-    }
```

tomasMizera (approved):
This is looking good, let's see what testing brings

PR #3562, opened 2024-08-09
Adding Loaders to Highlights in Recording Tools

VitorVieiraZ:
In `MMRecordingTools`, `MMHighlights` are loaded only when they match the layer's geometry, improving performance by eliminating the need to load unnecessary components.

https://github.com/user-attachments/assets/f8cbac6d-de6b-4bfe-a71a-933513a3146b

Partially Resolves #[2980](https://github.com/MerginMaps/mobile/issues/2980)

### DIFF
--- a/app/qml/map/MMRecordingTools.qml
+++ b/app/qml/map/MMRecordingTools.qml
@@ -111,96 +111,29 @@ Item {
   }
 
   Loader {
-    active: __inputUtils.geometryFromLayer( __activeLayer.vectorLayer ) === "linestring" ||
-            __inputUtils.geometryFromLayer( __activeLayer.vectorLayer ) === "polygon"
+    active: __inputUtils.isLineLayer( __activeLayer.vectorLayer ) ||
+            __inputUtils.isPolygonLayer( __activeLayer.vectorLayer )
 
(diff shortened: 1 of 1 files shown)


tomasMizera (changes requested):
A few changes, but it seems like we are close to merge :)

app/qml/map/MMRecordingTools.qml
```diff
@@ -110,56 +110,30 @@ Item {
...
-
-    mapSettings: root.map.mapSettings
-    geometry: __inputUtils.transformGeometryToMapWithLayer( mapTool.existingVertices, __activeLayer.vectorLayer, root.map.mapSettings )
-
-    markerType: MMHighlight.MarkerTypes.Circle
-    markerSize: MMHighlight.MarkerSizes.Bigger
+  Loader {
+    active: true
+    sourceComponent: existingVerticesHighlightComponent
   }
```
tomasMizera, line 137: We do not need to have the loader altogether if we want to show it all the time :)

tomasMizera (changes requested):

tomasMizera (approved):

PR #3564, opened 2024-08-09
Adjusting MMProjectLoadingPage text to respect safe area

VitorVieiraZ:
Before:
 

After
| Landscape | Portrait |
|--------|--------|
||  |


alexd0793:
Hi. I have tested this ticket,all good, but i found an issue when i landscape mode, more options is not complete
Screenshot_2024-09-16-11-43-41-422_uk co lutraconsulting

PR #3574, opened 2024-08-13
Measurement Tool

### LINKED ISSUE #1400: Add measuring tool for length and area
See the roadmap ticket for more details: https://github.com/MerginMaps/roadmap/issues/25

---

I found Google Earth app has implemented a very nice and intuitive way of doing it:
https://www.youtube.com/watch?v=Dxpp6Yfj9AU

VitorVieiraZ:
### Measurement Tool

https://github.com/user-attachments/assets/2e804597-c87e-475e-ab45-c11663668d6a

Enhancements to existing Mergin Maps Components:
- MMDrawer: Added support for a top-left button with an optional icon.
- Added support for area formatting in project units within Utils.
- MMButton now offers size options: Regular (traditional mode) and Small (used in the measurement tool).
- MMMapHidingLabel (used in `showInfoTextMessage` in `MMMapController`) now supports at most three lines.

Resolves #1400


app/qml/map/components/MMMeasureCrosshair.qml
tomasMizera, line 1: Please rewrite to inherit the `MMCrosshair` as a base item
```qml
// MMMeasureCrosshair

MMCrosshair {

  id: root

  // add the custom code for the distance label
}
```

tomasMizera (changes requested):

app/inpututils.cpp
```diff
@@ -189,6 +189,28 @@ QString InputUtils::formatDistanceInProjectUnit( const double distanceInMeters,
...
+  Qgis::AreaUnit areaUnit = destUnit;
+
+  if ( areaUnit == Qgis::AreaUnit::Unknown )
+  {
+    areaUnit = QgsProject::instance()->areaUnits();
+  }
+
+  if ( areaUnit == Qgis::AreaUnit::Unknown )
+  {
+    return QString::number( areaInSquareMeters, 'f', precision );
```
PeterPetrik, line 203: This should return units similarly to last return in the function, not just a number

app/inpututils.h
```diff
@@ -76,6 +76,8 @@ class InputUtils: public QObject
     Q_INVOKABLE QString formatProjectName( const QString &fullProjectName );
     Q_INVOKABLE QString formatNumber( const double number, int precision = 1 );
     Q_INVOKABLE QString formatDistanceInProjectUnit( const double distanceInMeters, int precision = 1, Qgis::DistanceUnit destUnit = Qgis::DistanceUnit::Unknown );
+    Q_INVOKABLE QString formatAreaInProjectUnit( const double areaInSquareMeters, int precision = 1, Qgis::AreaUnit destUnit = Qgis::AreaUnit::Unknown );
```
PeterPetrik, line 79: we should have unit tests for this function.

app/qml/components/MMButton.qml
```diff
@@ -15,8 +15,10 @@ Button {
   id: root
 
   enum Types { Primary, Secondary, Tertiary }
+  enum Sizes { Small, Regular }
```
PeterPetrik, line 18: add small button to gallery

PeterPetrik (approved):
LGTM

app/qml/map/MMMapController.qml
```diff
@@ -568,7 +595,7 @@ Item {
           }
 
           visible: {
-            if ( root.mapExtentOffset > 0 && root.state !== "stakeout" ) return false
+            if ( root.mapExtentOffset > 0 && (root.state === "stakeout") ) return false
```
VitorVieiraZ, line 598: Revert

app/maptools/measurementmaptool.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+void MeasurementMapTool::checkCanCloseShape()
+{
+  if ( mRecordedGeometry.isEmpty() || mPoints.count() < 3 )
+  {
+    setCanCloseShape( false );
+    return;
+  }
+
+  if ( mPoints.count() >= 3 )
```
VitorVieiraZ, line 67: If not needed

app/maptools/measurementmaptool.cpp
```diff
@@ -0,0 +1,274 @@
...
+  {
+    QgsPoint firstPoint = mPoints.first();
+    QPointF firstPointScreen = mapSettings()->coordinateToScreen( firstPoint );
+    double distanceToFirstPoint = std::hypot( mCrosshairPoint.x() - firstPointScreen.x(), mCrosshairPoint.y() - firstPointScreen.y() );
```
tomasMizera, line 71: If you want to have it cleaner, you can create a new function in inpututils, something like `pixelDistanceBetween(QPointF p1, QPointF p2)`

app/maptools/measurementmaptool.cpp
```diff
@@ -0,0 +1,274 @@
...
+  setArea( area );
+
+  double perimeter = mDistanceArea.measurePerimeter( polygonGeometry );
+  setPerimeter( perimeter );
+
+  setCanCloseShape( false );
+  setCloseShapeDone( true );
+}
+
+void MeasurementMapTool::reset()
```
tomasMizera, line 98: I have a feeling I asked you previously to rename this function to `reset`, but looking at it now, I think it makes more sense to name it `resetMeasurement` (or something similar ofc). The thing is `reset` functions are commonly used to reset everything in the object to its initial state - including member pointers and so on. Here you only try to reset the abacus :) Sorry for the confusion!

app/maptools/measurementmaptool.h
```diff
@@ -0,0 +1,132 @@
...
+
+  protected:
+    void rebuildGeometry();
+    void checkCanCloseShape();
+
+  public slots:
+    void updateDistance();
+
+  private slots:
+    void onMapSettingsChanged( InputMapSettings *newMapSettings );
```
tomasMizera, line 116: You _should_ be able to declare this as `override`, I believe even without declaring it as a slot. Then work with it as a derived method from base class.

```suggestion
    void onMapSettingsChanged( InputMapSettings *newMapSettings ) override;
```

app/maptools/measurementmaptool.cpp
```diff
@@ -0,0 +1,274 @@
...
+    setCanUndo( false );
+  }
+
+  emit existingVerticesChanged( mExistingVertices );
+  setRecordedGeometry( geometry );
+}
+
+void MeasurementMapTool::onMapSettingsChanged( InputMapSettings *newMapSettings )
+{
+  if ( newMapSettings )
```
tomasMizera, line 141: You might need to `disconnect` from the previous mapSettings. You can do so here (in case you override the base function) or disconnect in the base `AbstractMapTool` - I _think_ that should work

app/qml/components/MMButton.qml
```diff
@@ -177,14 +179,14 @@ Button {
...
-      property real maxWidth: parent.width - 2 * __style.margin20
+      property real maxWidth: parent.width - 2 * ( root.size === MMButton.Sizes.Small ? __style.margin16 : __style.margin20 )
 
       x: ( parent.width - width ) / 2
 
       width: Math.min( paintedChildrenWidth, maxWidth )
       height: Math.max( buttonContent.paintedHeight, buttonIconRight.height )
 
-      spacing: buttonIconRight.visible || buttonIconLeft.visible ? __style.spacing12 : 0
+      spacing: ( root.size === MMButton.Sizes.Small ) ? 2 : ( buttonIconRight.visible || buttonIconLeft.visible ? __style.spacing12 : 0 )
```
tomasMizera, line 189: Please reformat to multiple lines instead of two nested inline `if`'s

app/qml/components/MMDrawerHeader.qml
tomasMizera, line 1: You need to elide text in case it would overflow the content on the left side. See `MMPageHeader` how to do it

tomasMizera (changes requested):
Where do you handle the `done` button?

app/maptools/measurementmaptool.h
tomasMizera: Wrong suggestion! We need to override the setter, not the signal 🤦🏻

tomasMizera:
Please merge master here so the iOS is running :)

app/qml/gps/MMStakeoutDrawer.qml
```diff
@@ -110,7 +109,7 @@ MMDrawer {
         width: ( parent.width + parent.spacing ) / 2
 
         title: qsTr( "Distance" )
-        value: remainingDistance >= 0 ?__inputUtils.formatDistanceInProjectUnit( remainingDistance, 2 ) : qsTr( "N/A" )
+        value: remainingDistance >= 0 ?__inputUtils.formatDistanceInProjectUnit( remainingDistance, 2, __activeProject.qgsProject ) : qsTr( "N/A" )
```
tomasMizera, line 112: Mind the parameters (number and order) - as discussed on the call

app/qml/map/components/MMMapHidingLabel.qml
```diff
@@ -40,6 +42,8 @@ MMHidingBox {
     font: __style.t3
     horizontalAlignment: Text.AlignHCenter
     verticalAlignment: Text.AlignVCenter
+    maximumLineCount: 3
```
tomasMizera, line 45: Unsure about 3 lines, but we can test and tune later :)

app/qml/map/components/MMMeasureCrosshair.qml
```diff
@@ -0,0 +1,134 @@
...
+    signal closeShapeClicked
+
+    MM.SnapUtils {
+      id: snapUtils
+
+      centerPosition: root.center
+      mapSettings: root.mapSettings
+      qgsProject: root.qgsProject
+      useSnapping: root.shouldUseSnapping
+      destinationLayer: __activeLayer.vectorLayer
```
tomasMizera, line 47: Ah cool, so we support snapping?

tomasMizera: By the way, `MMCrosshair` is creating its own `SnapUtils`, do we need to duplicate it here?

app/qml/main.qml
```diff
@@ -326,6 +335,15 @@ ApplicationWindow {
         }
       }
 
+      MMToolbarButton {
+        text: qsTr("Measure")
+        iconSource: __style.measurementToolIcon
+        onClicked: {
+          stateManager.state = "map"
```
tomasMizera, line 342: Is it mandatory to change the state here?

app/qml/map/MMMeasurementTools.qml
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  MMMeasureCrosshair {
+    id: crosshair
+
+    anchors.fill: parent
+    qgsProject: __activeProject.qgsProject
+    mapSettings: root.map.mapSettings
+
+    text: __inputUtils.formatDistanceInProjectUnit( mapTool.lengthWithGuideline, __activeProject.qgsProject )
```
tomasMizera, line 94: Mind that this is missing the `precision` parameter

app/qml/map/MMMeasurementTools.qml
```diff
@@ -0,0 +1,113 @@
...
+    canCloseShape: mapTool.canCloseShape
+
+    onCloseShapeClicked: finalizeMeasurement( true )
+  }
+
+  function finalizeMeasurement( closeShapeClicked )
+  {
+    guidelineController.allowed = false
+    crosshair.visible = false
```
tomasMizera, line 103: If you want to be more qml-like, you can bind crosshair`s visible property as:
```
  visible: !mapTool.measurementFinalized
```

app/qml/map/MMMeasurementTools.qml
```diff
@@ -0,0 +1,113 @@
...
+
+    text: __inputUtils.formatDistanceInProjectUnit( mapTool.lengthWithGuideline, __activeProject.qgsProject )
+    canCloseShape: mapTool.canCloseShape
+
+    onCloseShapeClicked: finalizeMeasurement( true )
+  }
+
+  function finalizeMeasurement( closeShapeClicked )
+  {
+    guidelineController.allowed = false
```
tomasMizera, line 102: If you want to be more qml-like, you can bind this property as:

  allowed: !mapTool.measurementFinalized

app/qml/map/MMMapController.qml
```diff
@@ -35,6 +35,10 @@ Item {
   property bool isStreaming: recordingToolsLoader.active ? recordingToolsLoader.item.recordingMapTool.recordingType === MM.RecordingMapTool.StreamMode : false
   property bool centeredToGPS: false
 
+  property var mapToolComponent: {
+    state === "measure" ? measurementToolsLoader.item : null
```
tomasMizera, line 39: I'd recommend binding this as 
```suggestion
    measurementToolsLoader.active ? measurementToolsLoader.item : null
```

In this case you will be sure that the loader will first get active and only after that its item will be assigned here.

app/qml/map/components/MMMeasureCrosshair.qml
tomasMizera, line 1: Please make sure to use pixel values from `mmstyle` instead of typing in specific values like `5 * __dp`

app/qml/map/components/MMMeasureCrosshair.qml
```diff
@@ -0,0 +1,134 @@
...
+      }
+
+      function onProjectReloaded( project ) {
+        // We need to re-assign qgs project to snaputils, because
+        // even though we loaded a different project,
+        // internally we keep the same pointer for QgsProject.
+        snapUtils.qgsProject = __activeProject.qgsProject
+        snapUtils.mapSettings = root.mapSettings
+      }
+    }
```
tomasMizera, line 133: We won't need this if we use crosshair's snap utils

app/qml/components/MMButton.qml
```diff
@@ -177,14 +179,21 @@ Button {
...
 
       x: ( parent.width - width ) / 2
 
       width: Math.min( paintedChildrenWidth, maxWidth )
       height: Math.max( buttonContent.paintedHeight, buttonIconRight.height )
 
-      spacing: buttonIconRight.visible || buttonIconLeft.visible ? __style.spacing12 : 0
+      spacing: {
+        if ( ( root.size === MMButton.Sizes.Small ) )
+          return 2;
```
tomasMizera, line 191: Please use `mmstyle` instead

app/qml/main.qml
```diff
@@ -596,6 +614,32 @@ ApplicationWindow {
...
+  }
+
+  Component {
+    id: measurePanelComponent
+
+    MMMeasureDrawer {
+      id: measurePanel
+
+      width: window.width
+      mapCanvas: map
```
tomasMizera, line 634: I'd suggest passing in the maptool instead:
```
mapTool: map.mapToolComponent
```

app/qml/map/MMMapController.qml
tomasMizera: I think we are only interested in the mapTool (the cpp class), aren't we? In that case, we should pass just `measurementToolsLoader.item.mapTool`

tomasMizera (changes requested):
Quite a few comments, but they are rather small. Thanks for addressing my previous ones!

app/qml/map/MMMapController.qml
VitorVieiraZ: For `finalizeMeasurement` and `repeatMeasure` methods inside `MMMeasurementTools`, we need to modify the mapTool, guidelineController and crosshair. Therefore, we need access to both `mapTool` and `MMMeasurementTools` instance.

tomasMizera:
Hi @uclaros, can I ask you to review the usage of QGIS API please? Thanks! You will find it in `measurementmaptool.cpp/h`

VitorVieiraZ:
> When you measure a line (not a closed shape) it still says it is `Perimeter` instead of just `Length`

Hi @erpas, thanks for the comment. That’s the expected behavior following the tool design

app/maptools/measurementmaptool.h
```diff
@@ -0,0 +1,139 @@
...
+
+  private:
+    QVector<QgsPoint> mPoints;
+    QgsGeometry mRecordedGeometry;
+    QgsGeometry mExistingVertices;
+    QgsDistanceArea mDistanceArea;
+    QPointF mCrosshairPoint;
+    double mLengthWithGuideline = 0;
+    double mPerimeter = 0;
+    double mArea = 0;
```
uclaros, line 132: this feels unnecessary complicated to me:
- Why do we need 3 sets of points (`mPoints`, `mRecordedGeometry`, `mExistingVertices`)? Can't we just have one?
- Why do we need to calculate and store the derived attributes (`mArea`, `mPerimeter`, `mLengthWithGuideline`)? Can't we directly fetch them from the `QgsGeometry` methods? I'm afraid our stored values could eventually get out of sync with the underlying geometry

uclaros (commented):
Nice work!
I've got a few comments though :)

- I'd also like to have the units selectable in the measurements panel. While the project default units is a good default start, I can see users wanting to change that temporarily for a single measurement)
- It would be nice to have 2 decimal digits when the measured distance is small enough (less than a meter? 10 meter? other?)
- It would be nice if there was a button to toggle closing/opening the geometry. If a user is trying to measure the area of a big square field, we shouldn't force him to re-digitize the first point in order to get an area! It would be way easier to just hit a button after digitizing the 3rd point.

app/inpututils.cpp
```diff
@@ -168,27 +168,50 @@ QString InputUtils::formatNumber( const double number, int precision )
   return QString::number( number, 'f', precision );
 }
 
-QString InputUtils::formatDistanceInProjectUnit( const double distanceInMeters, int precision, Qgis::DistanceUnit destUnit )
+QString InputUtils::formatDistanceInProjectUnit( const double distanceInMeters, int precision, QgsProject *activeProject )
```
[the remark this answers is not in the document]
VitorVieiraZ: I have updated `activeProject` to `project`. About the usage of QgsProject::instance(), we have considered it as anti-pattern, and passing the project as a parameter allows better modularity.

tomasMizera (approved):

tomasMizera (approved):
Looks good to me! Let me remove the changes to the android CI job

PR #3582, opened 2024-09-04
Fix CI tests

tomasMizera:
We dropped test.dev server env. I updated credentials for app.dev. See https://github.com/MerginMaps/python-api-client/commit/621c5ffe2bbfc3103894ee6f63b26a88ea963551


PeterPetrik:
no idea what is going on here, not sure if I am the best for review :)

PR #3589, opened 2024-09-10
WIP - Fix binding loop in MMListDelegate height calculation

VitorVieiraZ:
Moved height assignment to Component.onCompleted to break circular dependency between height and implicitHeight. 
`MMListDelegate.qml:43:3: QML MMListDelegate: Binding loop detected for property "height"` fixed.

Notes:
Setting `opacity` instead of `height` to hide invisible items.

Screenshot 2024-10-17 at 10 23 59


tomasMizera:
Did this solution help the issue Vitor? Can you please check if buttons on the map are behaving correctly? Especially the one with the streaming mode - when you navigate to edit line/polygon feature geometry

tomasMizera (approved):

tomasMizera (changes requested):
Hi @VitorVieiraZ while the solution works for the binding loop, I found an issue during testing - height of list delegates is somehow not preserved, resulting in views like the following:

| Before | Now (with the fix) |
|--------|--------|
|  |  |

tomasMizera:
Let's close this one and retry later

PR #3606, opened 2024-09-18
Fix FeaturesProxyModel sorting when model layer is set after setting the sourceModel

### LINKED ISSUE #3604: Sorting of the list in "Browse features" doesn t work
Complete workflow:
open attribute table
alter/set sort order
save qgis project
sync on pc
sync on mobile

after sync  on mobile the features are not sorted , they have the same order like before sorting

uclaros:
In _MMFeaturesListPage_, when `FeaturesProxyModel::setFeaturesSourceModel( &model );` is called, the model layer is not yet defined, 
This led to `FeaturesModel::sortingEnabled()` always returning `false` and no sorting taking place.

With this PR, when the model's layer is changed, `FeaturesModel::setupSorting()` is also called followed by `FeaturesProxyModel::updateSorting()`

Test was adjusted to handle this.

@tomasMizera do you think it would be better if `FeaturesModel::sortingEnabled()` would implicitly call `FeaturesModel::setupSorting()` first instead?

Fixes #3604

### DIFF
--- a/app/featuresmodel.cpp
+++ b/app/featuresmodel.cpp
@@ -363,6 +363,7 @@ void FeaturesModel::setLayer( QgsVectorLayer *newLayer )
     }
 
     mLayer = newLayer;
+    setupSorting();
     emit layerChanged( mLayer );
 
     if ( mLayer )
--- a/app/featuresproxymodel.cpp
+++ b/app/featuresproxymodel.cpp
@@ -45,5 +45,4 @@ void FeaturesProxyModel::setFeaturesSourceModel( FeaturesModel *sourceModel )
   setSourceModel( mModel );
   mModel->setupSorting();
   connect( mModel, &FeaturesModel::fetchingResultsChanged, this, [ = ]( bool pending ) { if ( !pending ) updateSorting(); } );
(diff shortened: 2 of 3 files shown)


tomasMizera:
Hi Stefanos, this looks reasonable, one thing that I'd probably like to see is to introduce a new signal like `sortingUpdated` or `sortingChanged` and connect the `SortFilterProxyModel` to it instead of the `layerChanged` signal. The thing is that if we at some point change the order from:
```cpp
  setupSorting();
  emit layerChanged();
```
to:
```cpp
emit layerChanged();
setupSorting();
```

then the sorting would not work again.

tomasMizera (commented):
See my comment ⬆️

VitorVieiraZ (approved):
It's all set and working well for me.

tomasMizera (approved):
Thanks for the update @uclaros

tomasMizera:
There was a hiccup with GH actions yesterday (https://www.githubstatus.com/history), I reran the actions

tomasMizera:
@alexd0793 this is the build to test: https://github.com/MerginMaps/mobile/actions/runs/11035810053?pr=3606

PR #3635, opened 2024-10-02
Fix inconsistent UI/UX for read-only non-spatial layers

### LINKED ISSUE #3633: Read-only non-spatial layers: inconsistent UI/UX
I have a project with a non-spatial layer as a reference for my form. I have made it read-only. Users can open the layer from Layers tab and then Add feature. But it blocks them to save the feature.

Similar to the spatial layer, if a non-spatial layer is Read-only, we should remove the options to add or edit the data.

VitorVieiraZ:
Added `layerIsReadOnly` property as a condition for visibility of `Add Feature` button. If a layer is non-spatial and not read-only, the button will be visible.

| Read-only | Editable |
|--------|--------|
||  |

Fix #3633

### DIFF
--- a/app/qml/layers/MMFeaturesListPage.qml
+++ b/app/qml/layers/MMFeaturesListPage.qml
@@ -83,20 +83,6 @@ MMComponents.MMPage {
       anchors.centerIn: parent
       running: featuresModel.fetchingResults
     }
-
-    MMComponents.MMButton {
-      id: addButton
-
-      width: parent.width
-      anchors.bottom: parent.bottom
-      anchors.bottomMargin: root.hasToolbar ? __style.margin20 : ( __style.safeAreaBottom + __style.margin8 )
-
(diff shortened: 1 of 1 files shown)


app/qml/layers/MMFeaturesListPage.qml
```diff
@@ -21,6 +21,7 @@ MMComponents.MMPage {
 
   property var selectedLayer: null
   property bool hasToolbar: false
+  property bool layerIsReadOnly: featuresModel.layer?.readOnly ?? false
```
tomasMizera, line 24: It seems like we could go just with this
```suggestion
  property bool layerIsReadOnly: selectedLayer?.readOnly ?? false
```
that way we do not add dependency for `featuresModel`

tomasMizera (changes requested):

tomasMizera (approved):
Looks good!

tomasMizera:
Please rebase/merge master to fix the iOS build

PR #3638, opened 2024-10-04
optimize feature model sorting

### LINKED ISSUE #3637: App lags or freezes when viewing layers with many features in Layer view since update to 2024.4.0
Since the addition of the new sorting feature, layers that have many features (5,000+) will lag or cause the app to freeze when opening the layer in the 'Layer' view. This has been reported by multiple users since the update to 2024.4.0. I have created an example project [here](https://app.merginmaps.com/projects/support/layer_sort_test/tree).

To recreate the problem:

- Download the project in the app
- Open project and select 'Layers'
- Open the 'random_points' layer, the layer will take some time to load
- Try to search for a feature

Sometimes the app will freeze when searching for a feature or other times it will just lag excessively.

uclaros:
fix #3637


tomasMizera (changes requested):
As discussed on the call, marking it here for reference

tomasMizera (commented):
As agreed
- add flag to not follow attribute table sorting
- see if we can get info if QGIS will order features or a data provider itself

tomasMizera (approved):
Looking good now, thanks! We will start testing

PR #3665, opened 2024-10-25
Do not center to geometry when highlighting when geometry does not fit the screen

### LINKED ISSUE #3559: Do not center to geometry if feature is outside of screen bounds
We currently try to center the extent to identified feature. This is not ux-friendly in multiple scenarios:
 - when the feature is a C-shaped line (we end up in the center, not showing the feature at all)
 - when the feature is "too big" compared to your current scale - your extent changes a lot

The idea is to _not_ move the extent at all (if the geometry is _too big_) and just highlight the part of the geometry that is currently visible on the screen.

https://github.com/user-attachments/assets/8e459da6-07a0-498f-891f-a5992d465437

ValentinBuira:
This PR fix the ux-unfriendly behavior when in some scenario (concave shaped, or too big) the user would highlight a feature and the extend would  either not move to the feature or move quite a lot.

The new behaviors works by not jumping to the feature  if the feature can not be *fully* contained within the current extend size 

Example both jumping and not jumping in the same video:
[Screencast from 2024-10-25 11-46-47.webm](https://github.com/user-attachments/assets/a3b88506-afcf-4029-8607-d01ffd4e482e)

Fix #3559


VitorVieiraZ (approved):
Looks good to me. I have tested it, and it works just fine. There are just some code-related details left to cover in the review.

ValentinBuira:
There is one things left that bother me, it's that in some cases you, since we don't offset the view, you can hide the highlighted feature with the pop-up, see the attached video

[Screencast from 2024-10-28 15-34-54.webm](https://github.com/user-attachments/assets/2c8aa226-f7a0-4684-9949-51c2359a0d7d)


I am working on a fix but I can do it in a separate PR

app/inpututils.cpp
```diff
@@ -316,6 +316,25 @@ QPointF InputUtils::geometryCenterToScreenCoordinates( const QgsGeometry &geom,
   return screenPoint;
 }
 
+bool InputUtils::canExtentContainGeometry( const QgsGeometry &geom, InputMapSettings *mapSettings )
```
PeterPetrik, line 319: I suggest to rename `canExtentContainGeometry` to  `extentContainGeometry`

app/inpututils.h
```diff
@@ -89,6 +89,13 @@ class InputUtils: public QObject
      */
     Q_INVOKABLE QPointF geometryCenterToScreenCoordinates( const QgsGeometry &geom, InputMapSettings *mapSettings );
 
+
+    /**
+     * Returns the true if the geometry could fully be contained in the current screen otherwise false
+     * Geometry must be in canvas CRS
+     */
+    Q_INVOKABLE bool canExtentContainGeometry( const QgsGeometry &geom, InputMapSettings *mapSettings );
```
PeterPetrik, line 97: maybe worth to have simple unit test ?

PeterPetrik (approved):

app/qml/map/MMMapController.qml
```diff
@@ -1210,8 +1210,14 @@ Item {
   function jumpToHighlighted( mapOffset ) {
     if ( identifyHighlight.geometry === null )
       return
+    let screenPt
+    if ( __inputUtils.extentContainGeometry( identifyHighlight.geometry, mapCanvas.mapSettings ) ){
+      screenPt = __inputUtils.geometryCenterToScreenCoordinates( identifyHighlight.geometry, mapCanvas.mapSettings )
+    }
+    else{
+      screenPt = __inputUtils.onScreenGeometryCenterToScreenCoordinates( identifyHighlight.geometry, mapCanvas.mapSettings )
+    }
```
tomasMizera, line 1219: All the three `InputUtils` methods are used in just this one place. How about we simplify the qml part and do just:
```js
let screenPt = __inputUtils.getRelevantGeometryCenter(...)
```

And we combine the logic of the three methods in InputUtils?
The thing is we want to keep QML/JS logic minimal (for multiple reasons - no autotests, no types, runtime errors) and instead use C++ whenever possible :)

ValentinBuira: Combinied the three functions under a single function. For the name I want for the middle man `relevantGeometryCenterToScreenCoordinates`

The test are based on input data from my device let me know if it suits the test convention

app/test/testutilsfunctions.cpp
```diff
@@ -946,3 +946,20 @@ void TestUtilsFunctions::testFormatAreaInProjectUnit()
   area2str = mUtils->formatAreaInProjectUnit( 7000, 1, project );
   QVERIFY( area2str == "1.7 ac" );
 }
+
+void TestUtilsFunctions::testOnScreenGeometryCenterToScreenCoordinates()
```
tomasMizera, line 950: As the method name changed in `InputUtils`
```suggestion
void TestUtilsFunctions::testRelevantGeometryCenterToScreenCoordinates()
```

tomasMizera (approved):
This is looking very goood, thanks for the adjustments @ValentinBuira! @harminius let's proceed with testing now :)

app/test/testutilsfunctions.cpp
ValentinBuira: Thanks for the catch, I renamed it

harminius (approved):
Works great on Android :green_book:

PR #3673, opened 2024-11-04
Retry failed download requests implementation

### LINKED ISSUE #3644: Retry failed download requests
Syncing big projects in areas with low data coverage or often network issues can be challenging.

We want to implement a mechanism to retry requests after network error or different server errors.
Requests should be retried in case of network error and server 5xx http codes.

This ticket is only about _project download_, not about uploading chunks. Uploading is a bit more tricky and will be part of a different ticket.

Note: we already have sync retries in place, but this ticket applies to individual requests during sync.

Let's not retry more than 5 times for now (as a total in transaction).

Related issues https://github.com/MerginMaps/mobile/issues/885, https://github.com/MerginMaps/mobile/issues/1308

VitorVieiraZ:
Implements a retry mechanism for failed downloads during project downloads, with a maximum of 5 retries per transaction. Individual file downloads now retry on network issues, being re-added to the download queue. Added a `downloadItemRetried` signal to track retry attempts. Added a getter/setter for the Mergin API's QNetworkAccessManager to support testing.

Retry system for downloads diagram:
image

Transaction download queue replies handling:
image

Resolves #3644

### DIFF
--- a/core/merginapi.cpp
+++ b/core/merginapi.cpp
@@ -49,6 +49,7 @@ MerginApi::MerginApi( LocalProjectsManager &localProjects, QObject *parent )
   , mWorkspaceInfo( new MerginWorkspaceInfo )
   , mSubscriptionInfo( new MerginSubscriptionInfo )
   , mUserAuth( new MerginUserAuth )
+  , mManager( new QNetworkAccessManager( this ) )
 {
   // load cached data if there are any
   QSettings cache;
@@ -199,7 +200,7 @@ QString MerginApi::listProjects( const QString &searchExpression, const QString 
 
   QString requestId = CoreUtils::uuidWithoutBraces( QUuid::createUuid() );
 
-  QNetworkReply *reply = mManager.get( request );
+  QNetworkReply *reply = mManager->get( request );
   CoreUtils::log( "list projects", QStringLiteral( "Requesting: " ) + url.toString() );
   connect( reply, &QNetworkReply::finished, this, [this, requestId]() {this->listProjectsReplyFinished( requestId );} );
 
@@ -248,7 +249,7 @@ QString MerginApi::listProjectsByName( const QStringList &projectNames )
 
   QString requestId = CoreUtils::uuidWithoutBraces( QUuid::createUuid() );
 
-  QNetworkReply *reply = mManager.post( request, body.toJson() );
+  QNetworkReply *reply = mManager->post( request, body.toJson() );
   CoreUtils::log( "list projects by name", QStringLiteral( "Requesting: " ) + url.toString() );
   connect( reply, &QNetworkReply::finished, this, [this, requestId]() {this->listProjectsByNameReplyFinished( requestId );} );
 
@@ -296,7 +297,7 @@ void MerginApi::downloadNextItem( const QString &projectFullName )
     request.setRawHeader( "Range", range.toUtf8() );
   }
 
-  QNetworkReply *reply = mManager.get( request );
+  QNetworkReply *reply = mManager->get( request );
   connect( reply, &QNetworkReply::finished, this, [this, item]() { downloadItemReplyFinished( item ); } );
 
   transaction.replyPullItems.insert( reply );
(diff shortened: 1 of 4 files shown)


core/merginapi.cpp
```diff
@@ -428,6 +422,19 @@ void MerginApi::downloadItemReplyFinished()
       // no more requests to start, but there are pending requests - let's do nothing and wait
     }
   }
+  else if ( transaction.retryCount < transaction.MAX_RETRY_COUNT && isRetryableNetworkError( r ) )
+  {
+    transaction.retryCount++;
+    transaction.downloadQueue.append( item );
+
+    CoreUtils::log( "pull " + projectFullName, QStringLiteral( "Retrying download (attempt %1 of 5)" ).arg( transaction.retryCount ) );
```
PeterPetrik, line 430: attempt ... of transaction.MAX_RETRY_COUNT (not 5)

core/merginapi.h
```diff
@@ -573,6 +577,17 @@ class MerginApi: public QObject
...
     bool apiSupportsWorkspaces();
 
+    /**
+     * Returns the network manager used for Mergin API requests
+     */
+    QNetworkAccessManager *networkManager() const { return mManager; }
+
+    /**
+     * Sets the network manager to be used for Mergin API requests
+     * \param QNetworkAccessManager
```
PeterPetrik, line 587: remove undocumented \param from docs ?

core/merginapi.cpp
```diff
@@ -3222,8 +3228,8 @@ ProjectDiff MerginApi::compareProjectFiles(
   /*
   for ( MerginFile file : oldServerFilesMap )
   {
-    // R-D/L-D
-    // TODO: need to do anything?
+  // R-D/L-D
+  // TODO: need to do anything?
```
PeterPetrik, line 3232: I would rather keep the old formatting so it does not show as your change in the history :)

core/merginapi.h
```diff
@@ -785,11 +804,18 @@ class MerginApi: public QObject
...
+     * \returns True if the error should trigger a retry, false otherwise
+     */
+    bool isRetryableNetworkError( QNetworkReply *reply );
+
     QNetworkRequest getDefaultRequest( bool withAuth = true );
 
     bool projectFileHasBeenUpdated( const ProjectDiff &diff );
 
-    QNetworkAccessManager mManager;
+    QNetworkAccessManager *mManager = nullptr;
```
PeterPetrik, line 818: Any reason it is changed to pointer?

VitorVieiraZ: We use a pointer to allow setting as the MockNetworkManager during tests, enabling us to simulate network connection errors.

core/merginapi.cpp
```diff
@@ -406,13 +403,10 @@ void MerginApi::downloadItemReplyFinished()
     {
       CoreUtils::log( "pull " + projectFullName, "Failed to open for writing: " + file.fileName() );
     }
-
     transaction.transferedSize += data.size();
     emit syncProjectStatusChanged( projectFullName, transaction.transferedSize / transaction.totalSize );
-
     transaction.replyPullItems.remove( r );
     r->deleteLater();
-
```
tomasMizera, line 415: Why all these removed lines? :)

core/merginapi.h
```diff
@@ -650,14 +665,18 @@ class MerginApi: public QObject
     void apiSupportsWorkspacesChanged();
 
     void serverWasUpgraded();
+    void networkManagerChanged();
+
+    void downloadItemRetried( const QString &projectFullName, int retryCount );
+    void downloadItemsStarted();
```
tomasMizera, line 671: There is already a signal called `pullFilesStarted` do we need additional one?

tomasMizera:
We will need to retest once the comments are incorporated

core/merginapi.h
VitorVieiraZ: Yes, we need this additional signal to received the info on when individual items have started, so we can test the retry mechanism.

tomasMizera (approved):
Let's move to the testing phase now :)

PR #3676, opened 2024-11-14
Show dialog when active contributors per month limit is hit 

ValentinBuira:
This PR add a new dialog when user try to sync while they have already hit the active contributor per month limit

Screenshot from 2024-11-14 12-32-41

Note for reviewer the error live and is triggered in `synchronizationerror` and `synchronizationmanager`

Resolve https://github.com/orgs/MerginMaps/projects/15/views/1?pane=issue&itemId=86203521

### DIFF
--- a/app/qml/dialogs/MMMonthlyContributorsLimitDialog.qml
+++ b/app/qml/dialogs/MMMonthlyContributorsLimitDialog.qml
@@ -15,9 +15,11 @@ import "../components" as MMComponents
 MMComponents.MMDrawerDialog {
   id: root
 
+  property int contributorsQuota
+
   signal manageAccountClicked()
 
-  title: qsTr( "You've reached the maximum number of active monthly contributors for your current subscription." )
+  title: qsTr( "You've reached the maximum number of active monthly contributors (%1) for your current subscription." ).arg( contributorsQuota )
   imageSource: __style.reachedMonthlyContributorLimitImage
 
   description: qsTr( "Upgrade your subscription or wait until next month for the limit to reset." )
--- a/app/qml/main.qml
+++ b/app/qml/main.qml
@@ -901,7 +901,8 @@ ApplicationWindow {
       storageLimitDialog.open()
     }
 
-    function onMonthlyContributorLimitReached( uploadSize ) {
+    function onMonthlyContributorLimitReached( quota, serverMsg ) {
+      monthlyContributorsLimitDialog.contributorsQuota = quota
       monthlyContributorsLimitDialog.open()
     }
 
--- a/core/merginapi.cpp
+++ b/core/merginapi.cpp
@@ -2145,7 +2145,11 @@ void MerginApi::pushStartReplyFinished()
     }
     else if ( EnumHelper::isEqual( code, ErrorCode::MonthlyContributorsLimitHit ) )
     {
-      emit monthlyContributorLimitReached( serverMsg );
+      int quota = 0;
+      QVariant maximunMonthlyContributor = extractServerErrorValue( data, "projects_quota" );
+      if ( maximunMonthlyContributor.isValid() )
+        quota = maximunMonthlyContributor.toInt();
+      emit monthlyContributorLimitReached( quota, serverMsg );
     }
     else
     {
--- a/core/merginapi.h
+++ b/core/merginapi.h
@@ -598,7 +598,7 @@ class MerginApi: public QObject
     );
 
     void storageLimitReached( qreal uploadSize );
-    void monthlyContributorLimitReached( const QString &message );
+    void monthlyContributorLimitReached( int contributorsQuota, const QString &message );
     void projectLimitReached( int maxProjects, const QString &message );
     void migrationRequested( const QString &version );
     void notifySuccess( const QString &message );
--- a/gallery/qml/pages/DrawerPage.qml
+++ b/gallery/qml/pages/DrawerPage.qml
@@ -297,6 +297,8 @@ Page {
   MMMonthlyContributorsLimitDialog {
     id: monthlyContributorsLimitDialog
 
+    contributorsQuota: 15
+
     onPrimaryButtonClicked: {
       console.log("Manage workspace clicked")
     }


app/qml/dialogs/MMMonthlyContributorsLimitDialog.qml
```diff
@@ -0,0 +1,38 @@
...
+  required property bool apiSupportsSubscription
+
+  signal manageAccountClicked()
+
+  title: qsTr("You've reached the maximum number of active monthly contributors for your current subscription.")
+  imageSource: __style.reachedMonthlyContributorLimitImage
+
+  description: qsTr( "Upgrade your subscription or wait until next month for the limit to reset." )
+
+  primaryButton.text: apiSupportsSubscription ? qsTr("Manage account") : ""
```
VitorVieiraZ, line 27: Maybe we could have text as "Upgrade" and link to https://www.merginmaps.com/pricing when false?

app/qml/main.qml
```diff
@@ -859,6 +867,9 @@ ApplicationWindow {
           // just banner that we will try again
           __notificationModel.addInfo( qsTr( "Somebody else is syncing, we will try again later" ) )
         }
+        else if ( errorType === MM.SyncError.MonthlyContributorsLimitHit){
```
VitorVieiraZ, line 870: Small code styling detail to fix.

VitorVieiraZ (commented):
Looks good to me. I’ve left only comments about code styling and a suggestion for when apiSupportsSubscriptions is false.

app/qml/dialogs/MMMonthlyContributorsLimitDialog.qml
ValentinBuira: Solved in latest

VitorVieiraZ (approved):

VitorVieiraZ (changes requested):
Fix tests

app/qml/dialogs/MMMonthlyContributorsLimitDialog.qml
```diff
@@ -0,0 +1,38 @@
...
+  required property bool apiSupportsSubscription
+
+  signal manageAccountClicked()
+
+  title: qsTr( "You've reached the maximum number of active monthly contributors for your current subscription." )
+  imageSource: __style.reachedMonthlyContributorLimitImage
+
+  description: qsTr( "Upgrade your subscription or wait until next month for the limit to reset." )
+
+  primaryButton.text: apiSupportsSubscription ? qsTr( "Manage account" ) : qsTr( "Upgrade" )
```
tomasMizera, line 27: Other texts here also do not make sense when the API does not support subscription, so let's leave it out
```suggestion
  primaryButton.text: qsTr( "Upgrade" )
```

app/qml/main.qml
```diff
@@ -715,6 +715,21 @@ ApplicationWindow {
...
+    apiSupportsSubscription: __merginApi.apiSupportsSubscriptions
+
+    onManageAccountClicked: function() {
+      if (__merginApi.apiSupportsSubscriptions) {
+        Qt.openUrlExternally(__inputHelp.merginSubscriptionLink)
+      }
+      else {
+        Qt.openUrlExternally(__inputHelp.merginPricingLink)
+      }
+    }
```
tomasMizera, line 730: This should be enough, if the server does not support subscription, we do not need to send them to our pricing page.

```suggestion
    onManageAccountClicked: Qt.openUrlExternally( __inputHelp.merginSubscriptionLink )
```

app/qml/main.qml
```diff
@@ -715,6 +715,21 @@ ApplicationWindow {
     onManageAccountClicked: Qt.openUrlExternally(__inputHelp.merginSubscriptionLink)
   }
 
+  MMMonthlyContributorsLimitDialog {
+    id: monthlyContributorsLimitDialog
+
+    apiSupportsSubscription: __merginApi.apiSupportsSubscriptions
```
tomasMizera, line 721: We do not have this limit in other servers apart from Cloud, so this info is not needed
```suggestion

```

app/inputhelp.cpp
```diff
@@ -116,6 +116,11 @@ QString InputHelp::merginSubscriptionLink() const
   }
 }
 
+QString InputHelp::merginPricingLink() const
+{
+  return inputWeb + "/pricing" + utmTagSubscription;
+}
```
tomasMizera, line 122: I do not think we'll need this link

app/synchronizationerror.cpp
```diff
@@ -37,6 +37,11 @@ SynchronizationError::ErrorType SynchronizationError::errorType( int errorCode,
     // Project no longer exists / is on different server
     return ErrorType::ProjectNotFound;
   }
+  else if ( errorCode == 422 )
+  {
+    // Hit the maximun number of contributor per month
+    return ErrorType::MonthlyContributorsLimitHit;
+  }
```
tomasMizera, line 44: `422` error code can be returned in multiple scenarios, not only when the contributors limit is hit. Please see how we handle the storage limit here:

https://github.com/MerginMaps/mobile/blob/404ac2b379289be2d1c5b387e1e2a11bc69dcdb2/core/merginapi.cpp#L2119-L2121

In fact, these two places should be unified 🙈

tomasMizera (changes requested):

app/inputhelp.cpp
ValentinBuira: Removed the need for support subscription and link to pricing page

app/qml/dialogs/MMMonthlyContributorsLimitDialog.qml
ValentinBuira: Removed the need for support subscription and link to pricing page

app/qml/main.qml
ValentinBuira: Removed the need for support subscription and link to pricing page

app/synchronizationerror.cpp
ValentinBuira: Agreed it should be unified both of these place got advantage over the other

In the meantime I just added an extra safeguard in the sync error to parse the message for appropriate server error

app/synchronizationerror.cpp
```diff
@@ -37,6 +37,16 @@ SynchronizationError::ErrorType SynchronizationError::errorType( int errorCode,
     // Project no longer exists / is on different server
     return ErrorType::ProjectNotFound;
   }
+  else if ( errorCode == 422 )
+  {
+    // Hit the maximun number of contributor per month
+    if ( errorMessage.contains( QStringLiteral( "Maximum number of workspace contributors" ) ) )
```
tomasMizera, line 43: This option is possible, but it does not give us the flexibility to change the description on server later on (as clients would expect this exact string). Instead, we need to check for the error code `MonthlyContributorsLimitHit` ([see server code](https://github.com/MerginMaps/server-private/blob/fda7fc34bdbad91827391ebf4f525f4abd0c8e5b/server/src/service/errors.py#L70)).

You probably do not have that code available in this place, so we have two options, either
 1. pass the error code to this method from `merginapi`, or
 2. (less preferable) emit the failure signal directly from `merginapi`, similarly to how we do it for the storage limit, see https://github.com/MerginMaps/mobile/blob/404ac2b379289be2d1c5b387e1e2a11bc69dcdb2/core/merginapi.cpp#L2135

tomasMizera (changes requested):

app/qml/main.qml
```diff
@@ -895,6 +901,10 @@ ApplicationWindow {
       storageLimitDialog.open()
     }
 
+    function onMonthlyContributorLimitReached( uploadSize ) {
```
tomasMizera, line 904: We do not use this `uploadSize` anywhere. What would be nice though is to include the actual limit - similarly to what we did on plugin (see https://github.com/MerginMaps/qgis-plugin/pull/640/commits/f1ea460fb94c5c45d1d4e88218bbbd451dad624b#diff-d5f841eda10b7504474461cbece2f5a718c04995de1956f64f1674138655c675R28)

```suggestion
    function onMonthlyContributorLimitReached( contributorsQuota ) {
```

tomasMizera (approved):
Just a tiny change, feel free to merge once adjusted 👍🏻  Thanks!

tomasMizera:
> Fix tests

... aand yes, we will need to fix tests in case they will still be failing once the CI is re-run

app/qml/main.qml
ValentinBuira: Added in the last commit, similarly as the plugin

tomasMizera:
We need revert here @ValentinBuira

PR #3682, opened 2024-11-18
Improving UX by hiding editing buttons for readers  of a project

### LINKED ISSUE #3174: Users unable to sync project after a reader user makes some local changes
A user reported a problem when one of their Reader collaborators made some changes to a feature and later when tried to sync the project they get `No rights to access the project` and effectively can't get the project synced (get new changes from others).
The only option is to remove such a project locally and re-download it.

Related #1463

VitorVieiraZ:
While in `Reader` role, editing buttons are no longer visible to user. These buttons include:

Main `Add` button and the one within the features list.
`Add feature` button and its variations.
`Edit feature` button and its variations.
`Edit geometry` button.

| Reader | Other Roles |
|--------|--------|
||  |
||  |
||  |

Resolves #3174


tomasMizera (changes requested):

app/activeproject.cpp
```diff
@@ -31,13 +31,15 @@ ActiveProject::ActiveProject( AppSettings &appSettings
                               , ActiveLayer &activeLayer
                               , LayersProxyModel &recordingLayerPM
                               , LocalProjectsManager &localProjectsManager
+                              , MerginApi *merginApi
```
tomasMizera, line 34: Let's get rid of the dependency according to what we discussed

tomasMizera (changes requested):

core/coreutils.h
```diff
@@ -130,6 +130,11 @@ class CoreUtils
      */
     static QString getProjectMetadataPath( QString projectDir );
 
+    /**
+     * Updates a value in a JSON file at the specified top-level key
+     */
+    static bool replaceValueInJson( const QString &filePath, const QString &key, const QJsonValue &value );
```
tomasMizera, line 136: Let's add some unit tests for the new method please

app/test/testcoreutils.cpp
```diff
@@ -272,3 +272,66 @@ void TestCoreUtils::testNameAbbr()
...
+  QString testFilePath = QDir::tempPath() + "/test_replace_value.json";
+
+  // Test case 1: Basic replacement in valid JSON
+  {
+    QFile file( testFilePath );
+    QVERIFY( file.open( QIODevice::WriteOnly ) );
+    file.write( R"({"name": "test", "value": 123})" );
+    file.close();
+
+    QVERIFY( CoreUtils::replaceValueInJson( testFilePath, "value", 456 ) );
```
tomasMizera, line 288: I would add another key to test strings (now you test int)

core/merginapi.h
```diff
@@ -789,6 +795,12 @@ class MerginApi: public QObject
 
     bool projectFileHasBeenUpdated( const ProjectDiff &diff );
 
+    void reloadProjectRoleReplyFinished();
+
+    bool updateCachedProjectRole( const QString &projectFullName, const QString &newRole );
+
+    QString getCachedProjectRole( const QString &projectFullName ) const;
```
tomasMizera, line 802: Would be nice to have some explanation comments here

app/test/testmerginapi.cpp
```diff
@@ -2942,3 +2942,31 @@ void TestMerginApi::testParseVersion()
...
+  QCOMPARE( cachedRole, QString( "owner" ) );
+
+  // Test 2: Update cached role to 'reader'
+  QString newRole = "reader";
+  bool updateSuccess = mApi->updateCachedProjectRole( MerginApi::getFullProjectName( mWorkspaceName, projectName ), newRole );
+  QVERIFY( updateSuccess );
+
+  // Verify role was updated in cache
+  cachedRole = mApi->getCachedProjectRole( MerginApi::getFullProjectName( mWorkspaceName, projectName ) );
+  QCOMPARE( cachedRole, newRole );
```
tomasMizera, line 2968: This is fine, but we now need to test the actual behaviour - communication with the server. Let's call the `reloadProjectRole` method and check if the cached role will be back to `owner` as it should :)

tomasMizera: Just a tip - it is handy to qcompare to real values instead of variables (if possible). This way unit tests are more readable.

```suggestion
  QCOMPARE( cachedRole, "reader" );
```

tomasMizera (commented):
Just small adjustments and we are ready to go 🚀

app/test/testmerginapi.cpp
```diff
@@ -2943,6 +2943,33 @@ void TestMerginApi::testParseVersion()
...
+
+  // Test 2: Update cached role to 'reader'
+  QString newRole = "reader";
+  bool updateSuccess = mApi->updateCachedProjectRole( MerginApi::getFullProjectName( mWorkspaceName, projectName ), newRole );
+  QVERIFY( updateSuccess );
+
+  // Verify role was updated in cache
+  cachedRole = mApi->getCachedProjectRole( MerginApi::getFullProjectName( mWorkspaceName, projectName ) );
+  QCOMPARE( cachedRole, QString( "reader" ) );
+
```
tomasMizera, line 2968: Shouldn't we call server to update our project role here?

VitorVieiraZ: We could, but the result would be unpredictable, since we can't manipulate the role in the server from here.

tomasMizera: But that is what we want, no? You manipulated it locally to be "reader" and now we want to make sure that the next time we update it from server, the role goes back to what it should be - "owner". Otherwise we are just testing a small subset of the functionality (changing json value)

tomasMizera (approved):
🚀

tomasMizera:
Merging, CI will pass on master 🙃

Issue #3671, opened 2024-10-31
Unnecessary Confirmation Dialog


VitorVieiraZ:
A confirmation dialog appears when navigating back on unchanged, read-only layers. To reproduce: open one feature form, hit back (no confirmation), then open a second feature's form—subsequent backs trigger the confirmation. A binding loop is observed from the second feature form onward and may be contributing to this issue.

![Image](https://github.com/user-attachments/assets/91714b95-49a3-4d33-8d78-7aaeb2f155e1)

Reported by @uclaros.

ValentinBuira:
This is two bug in a trench coat : 
* One in Mergin Maps where the way we constructed the form dialog didn't trigger the function to check for changes, should be fixed in #3690 
* The second is directly in  QGIS and mutiline form using the is html checkbox, QGIS (or maybe Qt) insert a boiler plate html code to simulate richtext editing, thus creating a difference if the field was simple text before and trigger the save dialog. But once the field is saved again in html it should be good.

PR #3690, opened 2024-12-03
Fix uneccessary confirmation dialog when reopenning a form

ValentinBuira:
Previously we kept the left most form component in the stackview for performance but it prevent the call to `recalculateDerivedItems` and the appropriate `HasAnyChanges` signal

This PR solve the issue by destroying the last form component and rebuild it each time a form is open

May partially solves #3671

### DIFF
--- a/app/qml/form/MMFormStackController.qml
+++ b/app/qml/form/MMFormStackController.qml
@@ -258,7 +258,7 @@ Item {
       if ( formsStack.depth > 1 ) {
         formsStack.pop()
       }
-      else{
+      else {
         formsStack.clear()
       }
 


app/qml/form/MMFormStackController.qml
```diff
@@ -257,11 +255,16 @@ Item {
     id: formsStack
 
     function popOneOrClose() {
-      formsStack.pop()
+      if ( formsStack.depth > 1 ) {
+        formsStack.pop()
+      }
+      else{
```
tomasMizera, line 261:
```suggestion
      else {
```

tomasMizera (commented):
With this fix in place, we should be safe to revert the fix introduced in https://github.com/MerginMaps/mobile/pull/2880, @alexd0793 When you test this, please test also this issue: https://github.com/MerginMaps/mobile/issues/2879

@ValentinBuira let's also bear in mind that `activeFormIndex` will now often (when no form is opened) be `-1` (previously `0`). I checked the code and even though it works a bit differently now, I do not see any issue with it.

PeterPetrik:
please rebase with master (cppcheck fix)

tomasMizera (approved):

harminius (approved):
I verified this fixes the original issue. #3671 :green_book: 

However, @VitorVieiraZ provided me with another project to test (https://app.merginmaps.com/projects/Vitor%20Tests/VV-readOnlyUnecessaryConf/tree) and it keeps asking for unnecessary confirmation :red_envelope: 

Apparantly, "partially solves" is the best description, @ValentinBuira

tomasMizera:
> I verified this fixes the original issue. #3671 📗
> 
> However, @VitorVieiraZ provided me with another project to test (https://app.merginmaps.com/projects/Vitor%20Tests/VV-readOnlyUnecessaryConf/tree) and it keeps asking for unnecessary confirmation 🧧
> 
> Apparantly, "partially solves" is the best description, @ValentinBuira 📙

Thanks @harminius - we need to check if this is happening only because of the HTML content. Can you please double-check if the project contains the `text multiline with HTML` content? That should be the root cause.

harminius:
Indeed, that is the case of the project  @tomasMizera 
image

ValentinBuira:
The line between a bug and a misuse of the html widget in really fine in this case, I'll make a bug report in QGIS see what they think

But I think we can merge as is

PR #3700, opened 2025-01-02
Hiding tracking layer while recording

### LINKED ISSUE #3687: When recording, tracking layer is presented as an user editable layer
When I press record to capture a feature in one of my editable layers and I also have tracking option on, I can see the tracking layer also as an option to edit. 

I think tracking layer should be treated as a system layer and only  should be changed through the intended method.

### LINKED ISSUE #3699: Item in MMListDelegate still occupies space when invisible
Even when an item in MMListDelegate is set as not visible, it occupies space:

VitorVieiraZ:
Add a new filter condition to the recording layer model to ensure a layer is added only if it is not the position tracking layer.

| 

Resolves #3687


app/layersproxymodel.cpp
```diff
@@ -153,3 +153,17 @@ QVariant LayersProxyModel::getData( QModelIndex index, int role ) const
 {
   return sourceModel()->data( index, role );
 }
+
+bool LayersProxyModel::isPositionTrackingLayer( QgsMapLayer *layer ) const
+{
+  if ( !layer )
+    return false;
+
+  QgsProject *project = QgsProject::instance();
```
tomasMizera, line 162: Everything looks good, the only bit is that we should not read the QGIS project instance as singleton. Let's pass it in as parameter instead (via QML?)

tomasMizera (changes requested):

app/qml/main.qml
```diff
@@ -274,10 +282,12 @@ ApplicationWindow {
       }
 
       MMToolbarButton {
+        id: addButton
+
         text: qsTr("Add")
         iconSource: __style.addIcon
         onClicked: {
-          if ( __recordingLayersModel.rowCount() > 0 ) {
+          if ( recordingLayersModel.rowCount() > 0 ) {
```
tomasMizera, line 290: Would be better to not instantiate `LayersProxyModel` just for this check, but instead add a function like `__activeProject.projectHasRecordingLayers()`

tomasMizera (changes requested):

app/inpututils.cpp
```diff
@@ -2222,3 +2222,44 @@ double InputUtils::pixelDistanceBetween( const QPointF &p1, const QPointF &p2 )
...
 }
+
+bool InputUtils::layerHasGeometry( const QgsVectorLayer *layer )
+{
+  if ( !layer || !layer->isValid() )
+    return false;
+  return layer->wkbType() != Qgis::WkbType::NoGeometry && layer->wkbType() != Qgis::WkbType::Unknown;
+}
+
+bool InputUtils::layerVisible( QgsMapLayer *layer )
```
tomasMizera, line 2233: Missing check for pointer validity

app/layersproxymodel.cpp
```diff
@@ -13,22 +13,10 @@
 #include "qgsproject.h"
 #include "qgslayertree.h"
 
-LayersProxyModel::LayersProxyModel( LayersModel *model, LayerModelTypes modelType ) :
-  mModelType( modelType ),
-  mModel( model )
+LayersProxyModel::LayersProxyModel( QObject *parent ) :
+  QgsMapLayerProxyModel{ parent },
+  filterFunction( []( QgsMapLayer * layer ) { return ( layer && layer->isValid() ); } ) // default filter
```
tomasMizera, line 18: Would it be possible to assign this default behaviour directly in the `layersproxymodel.h`?

tomasMizera: It's quite complicated to read here

tomasMizera: Or optionally just call the `applyFilterFunction` method directly in the constructor as previously. It will work, because now you also "reset" the filtering function when the model type changes

app/layersproxymodel.cpp
```diff
@@ -153,3 +114,68 @@ QVariant LayersProxyModel::getData( QModelIndex index, int role ) const
...
+{
+  if ( mModel != model )
+  {
+    mModel = model;
+    setSourceModel( mModel );
+    emit modelChanged();
+  }
+}
+
+void LayersProxyModel::applyFilterFunction()
```
tomasMizera, line 164: I suggest
```suggestion
void LayersProxyModel::updateFilterFunction()
```

Would be a better name for this purpose

app/inpututils.cpp
tomasMizera, line 1: We need autotests here :)

app/activeproject.cpp
```diff
@@ -492,22 +488,26 @@ void ActiveProject::updateActiveLayer()
...
-      defaultLayer = mRecordingLayerPM.firstUsableLayer();
+      QgsMapLayer *layer = it.value();
+
+      // If it's a vector layer and visible, let's choose it
+      QgsVectorLayer *vectorLayer = qobject_cast<QgsVectorLayer *>( layer );
+      if ( vectorLayer && layerVisible( layer ) )
+      {
+        defaultLayer = layer;
+        break;
+      }
```
tomasMizera, line 504: The whole point of moving the functions to `InputUtils` was that we can use them here.. We need to check if the layer can be used for recording (`InputUtils::recordingAllowed`). Otherwise we end up in a situation like now:
image

app/activeproject.cpp
```diff
@@ -492,22 +488,26 @@ void ActiveProject::updateActiveLayer()
 {
   if ( !layerVisible( mActiveLayer.layer() ) )
```
tomasMizera, line 489: You moved the function to `InputUtils`, let's use it and remove the one defined here.

```suggestion
  if ( !InputUtils::layerVisible( mActiveLayer.layer() ) )
```

app/inpututils.cpp
```diff
@@ -2222,3 +2222,44 @@ double InputUtils::pixelDistanceBetween( const QPointF &p1, const QPointF &p2 )
...
+{
+  QgsLayerTree *root = QgsProject::instance()->layerTreeRoot();
```
tomasMizera, line 2235: Mind that `root` can also be null. Best to follow the previous implementation from: https://github.com/MerginMaps/mobile/blob/f7b663121365f62011f0d5fbeaa82c6ba1008b3f/app/activeproject.cpp#L388-L410

app/qml/map/MMMapController.qml
```diff
@@ -800,7 +800,13 @@ Item {
 
     drawerHeader.title: qsTr( "Choose Active Layer" )
```
tomasMizera, line 802: There is one thing that I realized while reviewing the PR is that we need to instantiate this drawer dynamically (wrap it in Loader/Component) - the reason is that it holds references (pointers) to the opened qgs project and layers, and in case we open a different project, we would need to invalidate all the pointers. It will be easier to have this drawer dynamically instantiated only when requested.

tomasMizera (changes requested):

app/inpututils.cpp
```diff
@@ -2222,3 +2222,65 @@ double InputUtils::pixelDistanceBetween( const QPointF &p1, const QPointF &p2 )
...
+    return false;
+  return layer->wkbType() != Qgis::WkbType::NoGeometry && layer->wkbType() != Qgis::WkbType::Unknown;
+}
+
+bool InputUtils::layerVisible( QgsMapLayer *layer )
+{
+  if ( !layer || !layer->isValid() )
+    return false;
+
+  QgsLayerTree *root = QgsProject::instance()->layerTreeRoot();
```
tomasMizera, line 2238: We should not use `QgsProject::instance()` --> pass it in to the function instead

app/activeproject.cpp
```diff
@@ -483,31 +456,31 @@ void ActiveProject::setMapTheme( const QString &themeName )
...
+    for ( auto it = layers.cbegin(); it != layers.cend(); ++it )
     {
-      defaultLayer = mRecordingLayerPM.firstUsableLayer();
+      QgsMapLayer *layer = it.value();
+      if ( InputUtils::recordingAllowed( layer, mQgsProject ) )
+      {
+        defaultLayer = layer;
+        break;
+      }
     }
```
tomasMizera, line 478: Why do you read the default layer from the app settings if you do not use it then? If we have some layer in the app settings, we want to pick that one (ofc, only if it is visible and can be used for recording)! :)

app/qml/map/MMMapController.qml
```diff
@@ -795,39 +798,56 @@ Item {
...
+
+        rightContent: MMIcon {
+          source: __style.doneCircleIcon
+          visible: __activeLayer.layerId === model.layerId
+        }
+
+        onClicked: {
+          __activeProject.setActiveLayer( recordingLayersModel.layerFromLayerId( model.layerId ) )
+          activeLayerPanel.close()
+          activeLayerPanelLoader.active = false
```
tomasMizera, line 841: We also need to deactivate the loader when closing the drawer without clicking on any delegate

tomasMizera (changes requested):

app/qml/map/MMMapController.qml
```diff
@@ -795,39 +798,57 @@ Item {
     }
   }
 
-  MMListDrawer {
-    id: activeLayerPanel
+  Loader {
```
tomasMizera, line 801: Looks good, thank you for making this dynamic. Just a note for future - we could create a new type of drawer, something like `dynamic-drawer` that would have `Loader` already incorporated 🤔 💡

tomasMizera (approved):

PR #3701, opened 2025-01-07
Bump upload-artifact action to v4

ValentinBuira:
This PR bump the upload-artifact action to v4 because previous version are deprecated at the end of the month

In the case of the mobile repository, the naive bump to v4 worked without change. and the artifacts content are similar on the few I checked (android, windows, linux, and the gallery app artifacts)


tomasMizera (approved):
Let's test it out :)

PR #3703, opened 2025-01-28
Fix incorrect missing authentication dialog trigger

VitorVieiraZ:
In `reloadProjectRole`, we check if user is logged with new method `isLoggedIn` before proceeding, preventing `missingAuthorizationError` and `MMMissingAuthDialog` from popping up.


tomasMizera (approved):
Looks good, let's test that :)

PR #3711, opened 2025-02-19
Online image preview support

### LINKED ISSUE #2718: Allow preview of online images in feature forms
Currently, in feature forms users can only preview images stored on the device.
It would be great if they could also preview images stored online, for example in minio.

It works on desktop QGIS but not in the mobile app. Example project (public):
https://app.merginmaps.com/projects/support/save_images_to_web

VitorVieiraZ:
When loading a photo, we now also verify if it's a url. If the url is valid, we consider it a valid image and proceed to load it. Since some urls might take longer to load, which is beyond our control, a `MMBusyIndicator` has been added to `MMPhoto` to enhance user experience in cases where loading images from urls might take a while.

https://github.com/user-attachments/assets/810a6e98-955f-4b70-b66a-2bc64040d051

| Remote image loading with error |
|--------|
 |

Resolves #2718

### DIFF
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -2293,8 +2293,3 @@ bool InputUtils::isValidUrl( const QString &link )
   QUrl url( link );
   return url.isValid();
 }
-
-bool InputUtils::isLocalFile( const QString &filePath )
-{
-  return filePath.startsWith( QStringLiteral( "file://" ) );
-}
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -2289,8 +2289,4 @@ bool InputUtils::isValidUrl( const QString &link )
 {
   QUrl url( link );
   return url.isValid();
-
-  // should we check for https scheme?
-  // ( url.scheme().toLower() == QStringLiteral( "http" ) ||
-  // url.scheme().toLower() == QStringLiteral( "https" ) );
 }
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -2288,6 +2288,9 @@ QgsMapLayer *InputUtils::mapLayerFromName( const QString &layerName, QgsProject 
 bool InputUtils::isValidUrl( const QString &link )
 {
   QUrl url( link );
-  return url.isValid() && ( url.scheme().toLower() == QStringLiteral( "http" ) ||
-                            url.scheme().toLower() == QStringLiteral( "https" ) );
+  return url.isValid();
+
+  // should we check for https scheme?
+  // ( url.scheme().toLower() == QStringLiteral( "http" ) ||
+  // url.scheme().toLower() == QStringLiteral( "https" ) );
 }
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -2284,3 +2284,10 @@ QgsMapLayer *InputUtils::mapLayerFromName( const QString &layerName, QgsProject 
 
   return nullptr;
 }
+
+bool InputUtils::isValidUrl( const QString &link )
+{
+  QUrl url( link );
+  return url.isValid() && ( url.scheme().toLower() == QStringLiteral( "http" ) ||
+                            url.scheme().toLower() == QStringLiteral( "https" ) );
+}
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -2287,6 +2287,9 @@ QgsMapLayer *InputUtils::mapLayerFromName( const QString &layerName, QgsProject 
 
 bool InputUtils::isValidUrl( const QString &link )
 {
+  if ( link.isEmpty() )
+    return false;
+
   QUrl url( link );
   return url.isValid();
 }
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -2293,3 +2293,8 @@ bool InputUtils::isValidUrl( const QString &link )
   QUrl url( link );
   return url.isValid();
 }
+
+bool InputUtils::isLocalFile( const QString &filePath )
+{
+  return filePath.startsWith( QStringLiteral( "file://" ) );
+}
--- a/app/inpututils.h
+++ b/app/inpututils.h
@@ -620,11 +620,6 @@ class InputUtils: public QObject
      */
     Q_INVOKABLE static bool isValidUrl( const QString &link );
 
-    /**
-     * Checks whether the given file path starts with "file://"
-     */
-    Q_INVOKABLE static bool isLocalFile( const QString &filePath );
-
   public slots:
     void onQgsLogMessageReceived( const QString &message, const QString &tag, Qgis::MessageLevel level );
 
--- a/app/inpututils.h
+++ b/app/inpututils.h
@@ -615,6 +615,11 @@ class InputUtils: public QObject
      */
     static QgsMapLayer *mapLayerFromName( const QString &layerName, QgsProject *project );
 
+    /**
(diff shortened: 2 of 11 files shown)


app/qml/form/editors/MMFormPhotoEditor.qml
```diff
@@ -208,16 +208,22 @@ MMFormPhotoViewer {
...
-      else if ( root._fieldValue && __inputUtils.fileExists( absolutePath ) ) {
+
+      let absolutePath = __inputUtils.getAbsolutePath( root._fieldValue, internal.prefixToRelativePath )
+
+      if ( __inputUtils.fileExists( absolutePath ) ) {
         root.photoState = "valid"
         absoluteImagePath = "file://" + absolutePath
       }
+      else if ( __inputUtils.isValidUrl( absolutePath ) ) {
+          root.photoState = "valid";
```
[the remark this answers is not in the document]
VitorVieiraZ: Regarding point 1, as soon as we have the image ready, we will make that change, and for 2. As discussed, we will apply this thunbmain idea on a future occasion.

app/inpututils.cpp
```diff
@@ -2284,3 +2284,13 @@ QgsMapLayer *InputUtils::mapLayerFromName( const QString &layerName, QgsProject
 
   return nullptr;
 }
+
+bool InputUtils::isValidUrl( const QString &link )
+{
+  QUrl url( link );
+  return url.isValid();
+
+  // should we check for https scheme?
```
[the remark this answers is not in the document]
VitorVieiraZ: In Image, if the URL points to something that is not a valid image, the loader will fail with Image.Error and the content won't be executed, therefore we wouldn't be executing or downloading any malicious element.

app/test/testutils.cpp
```diff
@@ -324,3 +324,19 @@ void TestUtils::testMapLayerFromName()
...
 }
+
+void TestUtils::testIsValidUrl()
+{
+  // valid urls
+  QVERIFY( InputUtils::isValidUrl( "http://www.example.com" ) );
+  QVERIFY( InputUtils::isValidUrl( "https://example.com/path?query=1" ) );
+  QVERIFY( InputUtils::isValidUrl( "ftp://ftp.example.com/resource" ) );
+  QVERIFY( InputUtils::isValidUrl( "www.example.com" ) );
+  QVERIFY( InputUtils::isValidUrl( "" ) ); // empty url is considered valid by QUrl
```
PeterPetrik, line 335: we should probably test for empty urls and reject them in our inpututils

PeterPetrik:
please rebase with master (cppcheck fix)

app/qml/components/MMPhoto.qml
```diff
@@ -49,7 +49,7 @@ Image {
 
     MMIcon {
       anchors.centerIn: parent
-      source: ( root.photoUrl && root.photoUrl.startsWith( "file://" ) ) ? __style.morePhotosIcon : __style.morePhotosIcon //__style.loadingImageErrorIcon
+      source: ( root.photoUrl && __inputUtils.isLocalFile( root.photoUrl ) ) ? __style.morePhotosIcon : __style.remoteImageLoadErrorIcon
```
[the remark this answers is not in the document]
VitorVieiraZ: I have renamed `__style.loadingImageErrorIcon` to `__style.remoteImageLoadErrorIcon`. To determine whether it's a file, we check if it starts with file://; otherwise, we assume it's a URL, aligning with the approach used here: https://github.com/MerginMaps/mobile/blob/f88d9c741ff9ee9b95a99ffc654185e7f5549c80/app/qml/form/editors/MMFormPhotoEditor.qml#L211.

tomasMizera:
In case of error, can we use the `error` text message to say that we could not load the image?

app/qml/components/MMPhoto.qml
```diff
@@ -68,6 +68,7 @@ Image {
 
   onStatusChanged: {
     if (status === Image.Error) {
+      __notificationModel.addError( "Could not load the image. The file may not exist, could be invalid, or the URL might be incorrect: " + root.photoUrl );
       console.error("MMPhoto: Error loading image: " + root.photoUrl);
```
PeterPetrik, line 72: 2 questions: 
1/ if there is notification than we co not need console.error 
2/ if we are offline (no internet)  I would not show any error . Also if you have broken image (wrong URL) then every time you open form you got notification error? I am not sure that is good UX

app/inpututils.h
```diff
@@ -615,6 +615,16 @@ class InputUtils: public QObject
...
 
+    /**
+     * Checks if the given string is a valid URL
+     */
+    Q_INVOKABLE static bool isValidUrl( const QString &link );
+
+    /**
+     * Checks whether the given file path starts with "file://"
+     */
+    Q_INVOKABLE static bool isLocalFile( const QString &filePath );
```
tomasMizera, line 626: Do local files always start with the `file://` on all platforms? I have a feeling it was not like this on all platforms (I might be wrong). Have you tried all platforms please?

VitorVieiraZ: In this scenario we always have file:// due to https://github.com/MerginMaps/mobile/blob/f88d9c741ff9ee9b95a99ffc654185e7f5549c80/app/qml/form/editors/MMFormPhotoEditor.qml#L211.  
However, in some scenarios, this function might not be applicable. Should we use startsWith directly in QML?

app/qml/components/MMPhoto.qml
VitorVieiraZ: With `errorMsg` property:

tomasMizera (changes requested):
Just this one change please 😇

tomasMizera (approved):

harminius:
Cannot make it working :red_circle: 

https://github.com/user-attachments/assets/d0cd9482-417e-46bf-acc0-48d751917eae

harminius (changes requested):
Cannot make it working :red_circle: 


Uploading screen-20250319-101112.mp4…

harminius (approved):
Tested on Android  :green_book: 
Tested for 
- a public image link (https://www.northrivergeographic.com/wp-content/uploads/2024/05/mergin-logo-small-2793196391.png) :green_circle: 
- S3 link (https://rp-test-media-sync.s3.eu-west-1.amazonaws.com/public/Screenshot+from+2025-03-18+15-36-24.png) :green_circle: 
- Google drive :red_circle: (https://drive.google.com/file/d/1Xenct9UGIubJNgawErWaUCfln7edqrc8/view?usp=drive_link)
    - for QGIS preview see video :open_mouth: 
    - app gives `Could not load the image error`


[Screencast from 20-03-25 08:19:37.webm](https://github.com/user-attachments/assets/ff1afdbb-0995-4e81-a9ae-f9796618032c)

tomasMizera:
>     * Google drive 🔴 (https://drive.google.com/file/d/1Xenct9UGIubJNgawErWaUCfln7edqrc8/view?usp=drive_link)
>       * app gives `Could not load the image error`

Thanks @harminius! For Google drive, I suspect it is because GDrive does a redirect when visiting the image. Could you try it with the final link please (https://drive.google.com/file/d/1Xenct9UGIubJNgawErWaUCfln7edqrc8/view)? 

> Screencast.from.20-03-25.08.19.37.webm

That's sick 🙃 I guess QGIS is using webview then

harminius (commented):
I tried the final link, too. Not sure what QGIS showed but the mobile app **didn't** loaded the image.

PR #3720, opened 2025-02-28
Hide position tracking button for readers

### LINKED ISSUE #3719: Position tracking is available to reader (but cannot sync)
I have a project where I have reader permission. In the mobile app, I can use Position tracking.

As I only have reader permission, I am obviously not able to sync the project and get the error message. However, there is no way how to resolve this error message, as I cannot delete features that were created from the tracking layer (I do not have permissions). Thus, the error message pops up everytime I press the sync button.

Should the Position tracking feature be available to readers in the mobile app?

app version: v2025.1.0, Android

VitorVieiraZ:
Hide position tracking button for readers, completing https://github.com/MerginMaps/mobile/pull/3682.

Resolves #3719

### DIFF
--- a/app/activeproject.h
+++ b/app/activeproject.h
@@ -116,11 +116,17 @@ class ActiveProject: public QObject
 
     //! Returns true if the project has at least one layer that allows recording
     Q_INVOKABLE bool projectHasRecordingLayers() const;
+
     /**
      * Returns role/permission level of current user for this project
      */
     Q_INVOKABLE QString projectRole() const;
     void setProjectRole( const QString &role );
+
+    /**
+     * Returns true if user role is different than reader
+     */
+    Q_INVOKABLE bool userHasEditableRole() const;
 
   signals:
     void qgsProjectChanged();
--- a/app/qml/form/MMFormPage.qml
+++ b/app/qml/form/MMFormPage.qml
@@ -201,7 +201,7 @@ Page {
 
   footer: MMComponents.MMToolbar {
 
-    visible: !root.layerIsReadOnly && __activeProject.projectRole !== "reader"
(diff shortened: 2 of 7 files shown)


PeterPetrik (approved):

tomasMizera (changes requested):

app/qml/form/MMFormPage.qml
```diff
@@ -231,7 +231,7 @@ Page {
         id: editGeometry
         text: qsTr( "Edit geometry" )
         iconSource: __style.editIcon
-        visible: root.layerIsSpatial && __activeProject.projectRole !== "reader"
+        visible: root.layerIsSpatial && __activeProject.userHasEditableRole
```
tomasMizera, line 234: `userHasEditableRole` is now defined as `Q_INVOKABLE` method, not property. This code must be resulting in errors..

tomasMizera (changes requested):

app/activeproject.h
```diff
@@ -116,12 +116,18 @@ class ActiveProject: public QObject
...
     /**
      * Returns role/permission level of current user for this project
      */
     Q_INVOKABLE QString projectRole() const;
     void setProjectRole( const QString &role );
 
+    /**
+     * Returns true if user role is different than reader
+     */
+    Q_INVOKABLE bool userHasEditableRole() const;
```
tomasMizera, line 129: If you want to use this method instead of the `projectRole`, then I suggest creating a new property `userHasEditRights`. Using method is not ideal as you won't be notified about change - your property bindings in QML won't work

VitorVieiraZ:
This PR is deprecated after position tracking refactoring.

PR #3727, opened 2025-03-07
Fix dangling pointer in LayersProxyModel

VitorVieiraZ:
`LayersProxyModel` previously did not reset its `mQgsProject` reference upon a project change, resulting in dangling pointers. To address this, the solution involves connecting to `projectChanged` signal emitted by `ActiveLayer` and resetting `mQgsProject` accordingly.

### DIFF
--- a/app/qml/map/MMMapController.qml
+++ b/app/qml/map/MMMapController.qml
@@ -853,10 +853,6 @@ Item {
       Connections {
         target: __activeProject
 
-        function onProjectWillBeReloaded() {
-          recordingLayersModel.model.reset()
-        }
-
         function onProjectReloaded( qgsProject ) {
           recordingLayersModel.qgsProject = __activeProject.qgsProject
         }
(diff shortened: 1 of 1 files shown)


app/qml/map/MMMapController.qml
```diff
@@ -849,6 +849,18 @@ Item {
         linkText: qsTr( "See how to enable digitizing in your project." )
         link: __inputHelp.howToEnableDigitizingLink
       }
+
+      Connections {
+        target: __activeProject
+
+        function onProjectWillBeReloaded() {
+          recordingLayersModel.model.reset()
+        }
```
tomasMizera, line 858: This seems to call Qt's `reset` method - no need to do that, the code below will do the trick.
```suggestion

```

tomasMizera (changes requested):

tomasMizera (approved):

Issue #3654, opened 2024-10-23
Categorized symbology does not render the "all other values" category


tomasMizera:
When I use the "categorized" symbology type, the category "all other values" is not rendered on the map.

Example:
<img src="https://github.com/user-attachments/assets/1fc6c24b-7caa-4c80-9b96-073a46266048" width=700>

Rendering in QGIS and the mobile app - note that the purple line is not visible on the map in the mobile app.
<img src="https://github.com/user-attachments/assets/fcb3fce1-6e7d-4432-9bd2-4accc108b7f9" width=500>

Test project: `SurveyCorp/city_survey` (on dev)

wonder-sk:
It looks like https://github.com/qgis/QGIS/pull/56369 (QGIS 3.36) changed the way how the fallback value of categorized renderer is read and stored in an incompatible way, so the mobile app (currently on QGIS 3.34.x) is unable to load it properly.

PR #3728, opened 2025-03-07
use qt682 mobile-SDK

PeterPetrik:
see https://github.com/MerginMaps/mobile-sdk/pull/115

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -21,16 +21,16 @@ concurrency:
 jobs:
   android_build:
     if: ( github.repository == 'MerginMaps/mobile' ) && (!contains(github.event.head_commit.message, 'Translate '))
-    runs-on: macos-13
+    runs-on: macos-15
     env:
-      QT_VERSION: '6.6.3' # use scripts/update_qt_version.bash to change
+      QT_VERSION: '6.8.2' # use scripts/update_qt_version.bash to change
       NDK_VERSION: r25
       NDK_VERSION_FULL: r25b
       JDK_VERSION: 11
       SDK_PLATFORM: android-34
       SDK_BUILD_TOOLS: 34.0.0
-      INPUT_SDK_VERSION_ARM: arm-android-20240517-224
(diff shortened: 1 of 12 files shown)


PeterPetrik:
TODO: 
- @tomasMizera fix android build (some android context passing changed between Qt version)
- @ValentinBuira test windows build - if it does not start it will be with missing qt/deps here https://github.com/MerginMaps/mobile/blob/f502da4521259f725a9f150fcdf58e9445a75d0c/cmake/InstallStep.cmake#L171 - one can try to see what is missing with DependencyWalker win utility or some other means on inspecting load of Qt libraries on windows 

I am leaving this with you guys :D

PeterPetrik:
@tomasMizera maybe we need to bump android SDK too ?

tomasMizera:
> @tomasMizera maybe we need to bump android SDK too ?

There are two things to it - `target SDK` and `min SDK`. I believe we need to bump the `min SDK`. `Target SDK` I'd leave for summer.

ValentinBuira:
I tried to build locally on Windows and use `windeployqt ` see https://doc.qt.io/qt-6/windows-deployment.html to detect missing but I got cmake error that prevent me to build.


I tried the build from CI on windows, it does not work out of the box, there is at least missing `Qt6QmlMeta.dll` but I'd like to be able to catch them all at once 

image

PeterPetrik:
6.8.3 should be out 18.March - so if we need those we can wait week and redo the PR with bumped Qt

tomasMizera:
Upgrade of QGIS should also fix https://github.com/MerginMaps/mobile/issues/3654

tomasMizera:
> Upgrade of QGIS should also fix #3654

I double-checked and it does 🎉

tomasMizera:
I'll give it a quick test and we can merge it then 👍🏻

tomasMizera (approved):
Managed to build desktop and Android with 6.8.3 👍🏻

PR #3735, opened 2025-03-18
Added default feature title format when display expression is empty

tomasMizera:
In case `displayExpression` is empty or QGIS heuristic fails to give some default feature title, we now set feature title in format `layer_name (feature_id)`. It was previously set as empty.

Screenshot 2025-03-18 at 17 27 05


app/attributes/attributepreviewcontroller.cpp
```diff
@@ -135,7 +135,19 @@ QString AttributePreviewController::featureTitle( )
...
-  return expr.evaluate( &context ).toString();
+
+  QString title = expr.evaluate( &context ).toString();
+
+  if ( title.isEmpty() )
+  {
+    // let's use a default format with layer name as user did not set any specific display expression
+    if ( mFeatureLayerPair.layer() && mFeatureLayerPair.layer()->isValid() && mFeatureLayerPair.feature().isValid() )
+    {
+      title = QStringLiteral( "%1 (%2)" ).arg( mFeatureLayerPair.layer()->name() ).arg( mFeatureLayerPair.feature().id() );
```
VitorVieiraZ, line 146: Should we implement a fallback, such as `Unnamed Layer`, in cases where `mFeatureLayerPair.layer()->name()` is empty?

app/attributes/attributepreviewcontroller.cpp
```diff
@@ -143,7 +143,13 @@ QString AttributePreviewController::featureTitle( )
     // let's use a default format with layer name as user did not set any specific display expression
     if ( mFeatureLayerPair.layer() && mFeatureLayerPair.layer()->isValid() && mFeatureLayerPair.feature().isValid() )
     {
-      title = QStringLiteral( "%1 (%2)" ).arg( mFeatureLayerPair.layer()->name() ).arg( mFeatureLayerPair.feature().id() );
+      QString layerName = mFeatureLayerPair.layer()->name();
+      if ( layerName.isEmpty() )
+      {
+        layerName = QStringLiteral( "Unnamed Layer" );
```
tomasMizera, line 149: This should be translatable string

tomasMizera (commented):
Looks good, just that one comment and we can test/merge :)

harminius (changes requested):
Does not work well with Null values for integer field. :red_envelope: 
It shows `0` when the field is empty. Even when the feature is opened it shows `0` while displaying an empty field.

Screenshot_20250324-113313

Screenshot_20250324-111049

Screenshot from 2025-03-24 11-33-55

tomasMizera:
> Does not work well with Null values for integer field. 🧧 It shows `0` when the field is empty. Even when the feature is opened it shows `0` while displaying an empty field.

Thanks @harminius for the tests! However, I am not sure if there is anything to do about this now, as this is generated by some logic in QGIS.. 

By the way, have you set the display expression yourself or is it QGIS-generated?

PR #3736, opened 2025-03-20
Fix handle highlight issue

### LINKED ISSUE #3234: Guideline to 0:0 posittion
When you record a line feature and you hit the add button twice without moving on the map, the guideline appears from 0:0 position to the point. This is present also in current production

https://github.com/MerginMaps/mobile/assets/83582181/1f1560bf-b609-499b-bea1-1c4393a462fb

ValentinBuira:
This PR  fix the handle not being displayed in line string in case user would place multiple point at the same place 

Fix #3234


ValentinBuira:
Cleaned up the PR and added a test case for this edge case

app/maptools/recordingmaptool.cpp
tomasMizera, line 1: This is a good possible solution. However, couldn’t the same issue occur at the end of the line as well?

I was wondering - what if we instead ignored adding duplicate points?

ValentinBuira: I reworked the PR in this way, and it's way more simple now.

in the record tool we add a safegard in the `addPoint` method to early return if the added point vertex  is the same as the last recorded point vertex


NB: The PR avoid inserting duplicate point in line geometry but should we also do it in polygon geometry ?

tomasMizera: Sure, let's do it for polygons too.

app/maptools/recordingmaptool.cpp
```diff
@@ -77,8 +77,6 @@ void RecordingMapTool::addPoint( const QgsPoint &point )
     pointToAdd.setZ( pointToAdd.z() - mPositionKit->antennaHeight() );
   }
 
-  mLastRecordedPoint = pointToAdd;
```
tomasMizera, line 80: How about checking if the `pointToAdd == mLastRecordedPoint` right in the beginning of this method, let's say here https://github.com/MerginMaps/mobile/pull/3736/files#diff-f59bbc2a03f7f398a6d05be79a023a47d92ecbd9cbad0c5e74302fe76125d912R53 ? Can't we just skip the whole method?

tomasMizera (changes requested):
We need to make sure the `mLastRecordedPoint` is cleared when layer changes

app/maptools/recordingmaptool.cpp
```diff
@@ -622,6 +630,8 @@ void RecordingMapTool::prepareEditing()
   {
     setRecordedGeometry( QgsGeometry() );
   }
+
+  mLastRecordedPoint = QgsPoint();
```
ValentinBuira, line 634: I am not sure what is the best syntax to reset `mLastRecordedPoint` ? If it was a pointer I would just use `nullptr` but mLastRecordedPoint is a non pointer and I would get a build warning ?

app/test/testmaptools.cpp
```diff
@@ -1123,7 +1149,9 @@ void TestMapTools::testAddVertexPolygonLayer()
   mapTool.addPoint( pointsToAdd[0] );
 
   QVERIFY( !mapTool.hasValidGeometry() );
+  qDebug() << "mapTool.recordedGeometry().constGet()->nCoordinates()" << mapTool.recordedGeometry().constGet()->nCoordinates();
```
tomasMizera, line 1152: Is the debug intentional here?

tomasMizera (approved):
Looks good to me

PR #3761, opened 2025-03-24
Refactor Recording Layers Logic and Rename LayersProxyModel to RecordingLayersProxyModel

VitorVieiraZ:
Refinement on recording layers handling by:
- Renaming `LayersProxyModel` to `RecordingLayersProxyModel`
- Decoupling `QgsProject` from `RecordingLayersProxyModel`
- Removing `filterFunction` in favor of using `setFilters` and `exceptedLayerIds`
- Moving `recordingAllowed` from `InputUtils` to `ActiveProject`
- Adding new helper methods on `Active Project`: `getVisibleLayers()` and `positionTrackingLayerId()`
- Adding a new role (`LayerVisible`) in `LayersModel` to support visibility filtering


app/activeproject.cpp
```diff
@@ -465,31 +450,28 @@ void ActiveProject::setMapTheme( const QString &themeName )
 
 void ActiveProject::updateActiveLayer()
 {
+  QList< QgsMapLayer * > visibleLayers = getVisibleLayers();
+
+  if ( !mQgsProject || visibleLayers.isEmpty() )
```
uclaros, line 455: Checking `mQgsProject` is not really needed, as `visibleLayers` will be empty if `!mQgsProject`.
Not bad checking again, though!

app/activeproject.cpp
```diff
@@ -465,31 +450,28 @@ void ActiveProject::setMapTheme( const QString &themeName )
...
-      setActiveLayer( defaultAppSettingsLayer );
-      return;
-    }
+    QgsMapLayer *defaultLayer = InputUtils::mapLayerFromName( mAppSettings.defaultLayer(), mQgsProject );
 
-    // default layer from app settings has no recording allowed => let's try to search for a new one
-    QgsMapLayer *defaultLayer = nullptr;
-    const QMap<QString, QgsMapLayer *> layers = mQgsProject->mapLayers();
-    for ( auto it = layers.cbegin(); it != layers.cend(); ++it )
+    if ( !recordingAllowed( defaultLayer ) || !defaultLayer )
```
uclaros, line 462: Also here, `!defaultLayer` is checked within `recordingAllowed` first, so it's practically unreachable.

app/activeproject.cpp
```diff
@@ -465,31 +450,28 @@ void ActiveProject::setMapTheme( const QString &themeName )
 
 void ActiveProject::updateActiveLayer()
 {
+  QList< QgsMapLayer * > visibleLayers = getVisibleLayers();
+
+  if ( !mQgsProject || visibleLayers.isEmpty() )
+    return;
+
   if ( !InputUtils::layerVisible( mActiveLayer.layer(), mQgsProject ) )
```
uclaros, line 458: We already have `visibleLayers` here, so we could directly check if it contains `mActiveLayer.layer()`

app/activeproject.cpp
```diff
@@ -567,3 +549,49 @@ void ActiveProject::setProjectRole( const QString &role )
...
   }
 }
+
+bool ActiveProject::recordingAllowed( QgsMapLayer *layer ) const
+{
+  if ( !layer )
+    return false;
+
+  if ( layer->readOnly() )
+    return false;
```
uclaros, line 559: Can you add a comment that this is only needed because of a bug in `QgsMapLayerProxyModel::layerMatchesFilters`?
Having `Qgis::LayerFilter::WritableLayer` should normally be enough

app/activeproject.cpp
```diff
@@ -567,3 +549,49 @@ void ActiveProject::setProjectRole( const QString &role )
...
+    return QList<QgsMapLayer *>();
+
+  QgsLayerTree *root = mQgsProject->layerTreeRoot();
+
+  if ( !root )
+    return QList<QgsMapLayer *>();
+
+  // Get list of all visible and valid layers in the project
+  QList<QgsMapLayer *> visibleLayers;
+  foreach ( QgsLayerTreeLayer *nodeLayer, root->findLayers() )
```
uclaros, line 584: I'm used from qgis and dislike the qt macros :)
```suggestion
  const QList<QgsLayerTreeLayer *> nodeLayers = root->findLayers();
  for ( QgsLayerTreeLayer *nodeLayer : nodeLayers )
```

app/activeproject.h
```diff
@@ -116,12 +116,22 @@ class ActiveProject: public QObject
...
     Q_INVOKABLE QString projectRole() const;
     void setProjectRole( const QString &role );
 
+    //! Returns if project layer allows recording (has geometry, editable, not position tracking layer) regardless of visibility
+    bool recordingAllowed( QgsMapLayer *layer ) const ;
+
+    //! Returns position tracking layer ID if exists
+    Q_INVOKABLE QString positionTrackingLayerId() const;
+
+    //! Returns all visible and valid layers in the project
```
uclaros, line 132:
```suggestion
    //! Returns all visible valid layers in the project
```

app/qml/map/MMMapController.qml
```diff
@@ -814,11 +814,10 @@ Item {
 
       onClosed: activeLayerPanelLoader.active = false
 
-      list.model: MM.LayersProxyModel {
+      list.model: MM.RecordingLayersProxyModel {
         id: recordingLayersModel
 
-        qgsProject: __activeProject.qgsProject
-        modelType: MM.LayersProxyModel.ActiveLayerSelection
+        exceptedLayerIds: __activeProject.positionTrackingLayerId()
```
uclaros, line 820: `exceptedLayerIds` expects a `QStringList`. Does QML just take care of that?

app/recordinglayersproxymodel.h
```diff
@@ -93,8 +81,6 @@ class LayersProxyModel : public QgsMapLayerProxyModel
      * In future will allow dependency injection of custom filter functions.
      */
     std::function<bool( QgsMapLayer * )> filterFunction;
```
uclaros, line 83: `filterFunction` can now be removed too, right?

uclaros (commented):
Nice!

One more note:
`ActiveProject::updateMapSettingsLayers()` is always called right after `ActiveProject::updateActiveLayer()`.
If `ActiveProject::updateMapSettingsLayers()` was always called first instead, then `ActiveProject::updateActiveLayer()` could avoid calling `getVisibleLayers()` once more and could just fetch the list of visible layers stored in settings (during `ActiveProject::updateMapSettingsLayers()`)
Would that make sense?

app/qml/map/MMMapController.qml
VitorVieiraZ: QML indeed performs the implicit conversion, as seen in [this documentation](https://doc.qt.io/qt-6.7/qstringlist.html#QStringList). However, I enclosed it in square brackets to clearly indicate that it is a list.

uclaros (approved):

app/recordinglayersproxymodel.cpp
```diff
@@ -8,21 +8,19 @@
...
-LayersProxyModel::LayersProxyModel( QObject *parent ) :
+RecordingLayersProxyModel::RecordingLayersProxyModel( QObject *parent ) :
   QgsMapLayerProxyModel{ parent }
 {
-  QObject::connect( this, &LayersProxyModel::rowsInserted, this, &LayersProxyModel::countChanged );
-  QObject::connect( this, &LayersProxyModel::rowsRemoved, this, &LayersProxyModel::countChanged );
-
-  updateFilterFunction();
+  QObject::connect( this, &RecordingLayersProxyModel::rowsInserted, this, &RecordingLayersProxyModel::countChanged );
+  QObject::connect( this, &RecordingLayersProxyModel::rowsRemoved, this, &RecordingLayersProxyModel::countChanged );
```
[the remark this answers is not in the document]
VitorVieiraZ: Indeed, there is not direct usage for it nor indirect usage in QML code, worked fine without it and count property.

app/recordinglayersproxymodel.h
```diff
@@ -21,17 +21,15 @@
...
     Q_PROPERTY( int count READ rowCount NOTIFY countChanged )
-    Q_PROPERTY( QgsProject *qgsProject READ qgsProject WRITE setQgsProject NOTIFY qgsProjectChanged )
-    Q_PROPERTY( LayerModelTypes modelType READ modelType WRITE setModelType NOTIFY modelTypeChanged )
     Q_PROPERTY( LayersModel *model READ model WRITE setModel NOTIFY modelChanged )
 
   public:
-    Q_INVOKABLE explicit LayersProxyModel( QObject *parent = nullptr );
+    Q_INVOKABLE explicit RecordingLayersProxyModel( QObject *parent = nullptr );
 
     enum LayerModelTypes
```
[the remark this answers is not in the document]
VitorVieiraZ: From this list, the only one to be kept is `layerFromLayerId`, used in `MMMapController`.

app/layersmodel.h
```diff
@@ -15,6 +15,7 @@
 #include <QPointer>
 
 #include "inputconfig.h"
+#include "inpututils.h"
```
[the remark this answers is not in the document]
VitorVieiraZ: We use it when `case LayerVisible: return ( mProject && InputUtils::isLayerVisible( layer, mProject ) );`.

app/recordinglayersproxymodel.cpp
```diff
@@ -45,65 +43,6 @@ void RecordingLayersProxyModel::refreshData()
   invalidate();
```
tomasMizera, line 47: It seems that this `refreshData` is not used too

tomasMizera (changes requested):
Please check the usage of all other members/methods of the proxy model

harminius (approved):
Tested as suggested: :green_book: 
- Behavior on closing and reopening the app, preferably multiple times
- Behavior on switching projects and adding features of different layer geometry type between switches
- Behavior on adding multiple features across different layers and observing the "Choose Active Layer" panel
- Switching between multiple layers on the "Choose Active Layer" panel
- Ensuring tracking layer is not present on "Choose Active Layer," and performing the tests on both types of projects (with and without position tracking layer)

I once managed to record to the tracking layer manually (which should not be possible) -see empty field on the screenshot, but I cannot replicate that :man_shrugging: 

Screenshot_20250416_101951

PR #3785, opened 2025-03-26
Fix 3783 Broken identification when feature is off the screen

### LINKED ISSUE #3783: Broken identification when feature is off the screen
When you try to identify a feature through the `Layers` menu - <Layer> - <Feature> while the feature is not within the screen you are centered on the Null island with the highest zoom.

https://github.com/user-attachments/assets/7cf8d788-aa35-4d94-a58c-6c4e7b195fbe

ValentinBuira:
In the first implementation we would check if the geometry could fit inside the current extent however it could be theoretically possible for the extent to be larger than the geometry YET not panned to right place thus result in false and it would intersect and returns an empty rectangle

This PR solves the issue by always trying to clip the geometry to the current extent all the time and if the result is null then we pan to the whole geometry

Fix #3783


app/inpututils.cpp
```diff
@@ -310,22 +310,22 @@ QPointF InputUtils::relevantGeometryCenterToScreenCoordinates( const QgsGeometry
...
   }
   else
   {
-    // Cut the geometry to current extent
-    QgsGeometry currentExtentAsGeom = QgsGeometry::fromRect( currentExtent );
-    QgsGeometry intersectedGeom = geom.intersection( currentExtentAsGeom );
-    QgsRectangle bbox = intersectedGeom.boundingBox();
-    target = QgsPoint( bbox.center() );
+    // The geometry is outside the current viewed extent
+    mapSettings->setExtent( geom.boundingBox() );
```
tomasMizera, line 327: Would this change the zoom level too? :)

ValentinBuira: Yes it does.

But this make the method name a bit of a lie now, but it's not a const  method so I guess that's fine

tomasMizera (approved):
Let's test :)

harminius (changes requested):
Unfortunately, map canvas freezes when I try to identify a feature which is not within the current displayed extent. Bottom menu does respond, "just the map is unresponsive". :red_circle: 

https://github.com/user-attachments/assets/04acbd51-36cb-4f58-875f-97e160a849b8

harminius:
ready for retesting @ValentinBuira ? Feel free to remove WIP from branch name and re-request me for the review - testing

ValentinBuira:
In the last PR iteration I refactored  the code used to set the extent so it also handles geometry with no bounding box like a point.

This also has the nice side effect, when you start to edit the geometry in the app the extent is now set to the whole geometry instead of only around the geometry center

app/inpututils.cpp
```diff
@@ -289,17 +289,42 @@ void InputUtils::setExtentToFeature( const FeatureLayerPair &pair, InputMapSetti
...
+
+    currentExtent.setXMinimum( currentExtent.xMinimum() - offsetX );
+    currentExtent.setXMaximum( currentExtent.xMaximum() - offsetX );
+    currentExtent.setYMinimum( currentExtent.yMinimum() - offsetY );
+    currentExtent.setYMaximum( currentExtent.yMaximum() - offsetY );
+  }
+  else
+  {
+    // Add a small offset to encompass handles etc..
+    double h = 15 * mapSettings->mapSettings().mapUnitsPerPixel();
```
ValentinBuira, line 320: For reviewer, I tried to add a little offset around the geometry, but I am not 100% confident of this code, to handle different screen size resolution etc...

tomasMizera (approved):
Looks good to me!

uclaros (commented):
As in the comment, I am not sure what the desired behavior is.
I think we need to have this defined. My suggestion:
| Geom type | No geom | Point | MultiPoint | line/poly/multi |
|---|---|---|---|---|
| Clicked | n/a | Nothing | ? | ? |
| Clicked (partially in extent/close to border) | n/a | Pan | ? | ? |
| Selected from list (in extent) | Nothing | Pan | ? | ? |
| Selected from list (partially in extent) | Nothing | Pan | ? | ? |
| Selected from list (outside extent) | Nothing | Pan | ? | ? |
| Edit geometry clicked | Nothing | Pan | ? | ? |

harminius (approved):
It works. :green_book: 
I assume you agreed on this behaviour: don't change the zoom level when a selected line doesn't fit the screen but change the zoom if polygon doesn't fit, right?

app/inpututils.cpp
```diff
@@ -289,17 +289,43 @@ void InputUtils::setExtentToFeature( const FeatureLayerPair &pair, InputMapSetti
...
-  QgsPointXY featureCenter = bbox.center();
-
-  double offsetX = currentExtentCenter.x() - featureCenter.x();
-  double offsetY = currentExtentCenter.y() - featureCenter.y();
-  currentExtent.setXMinimum( currentExtent.xMinimum() - offsetX );
-  currentExtent.setXMaximum( currentExtent.xMaximum() - offsetX );
-  currentExtent.setYMinimum( currentExtent.yMinimum() - offsetY );
-  currentExtent.setYMaximum( currentExtent.yMaximum() - offsetY );
+  geom = transformGeometryToMapWithLayer( geom, pair.layer(), mapSettings );
+  setExtentToGeom( geom, mapSettings );
```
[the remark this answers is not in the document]
ValentinBuira: Indeed the two behaviors right now:
When highlight on click -> Crop to geometry currently displayed screen of fall back to full extent
When start editing, select on the list of feature... -> set to full extent of the geometry

Either we avoid panning to the center of the geometry, because on some concave shape we would end up with nothing displayed on screen

ValentinBuira:
>  change the zoom level when a selected line doesn't fit the screen but change the zoom if polygon doesn't fit, right?

No it should not, in both case should work similarly. Does it happends when you clicked on the feature or when you select it on the feature list ?

ValentinBuira:
Adressed the comments for code style no functional changes excepted   

@uclaros  Not sure how to read your chart, I think the method should works on all these type of geometry since we only check if they have a bounding box?

harminius:
> > change the zoom level when a selected line doesn't fit the screen but change the zoom if polygon doesn't fit, right?
> 
> No it should not, in both case should work similarly. Does it happends when you clicked on the feature or when you select it on the feature list ?

Ahh, now it is a bit clearer to me. It does work well for a normal polygon but it doesn't like the shape of the polygon I used for testing :cold_sweat: :shrug: 

https://github.com/user-attachments/assets/82767066-e98b-491c-aa6f-c21a2a933007

tomasMizera:
Let's squash this now and merge and we'll make it better in the next release

PR #3788, opened 2025-03-27
Fix Reuse last entered value tool and QGIS's reuseLastValue support 

### LINKED ISSUE #3784: Reusing last entered value is broken
The feature for auto-filling field with value from the previously recorded feature does not work for me.

Hope I get how it should work
Scenario:
1. Enable Reuse last entered value feature for a project in its Settings
2. Record a feature, check the checkbox for an attribute and fill it in
3. Record another feature which should have the checked fields prefilled

https://github.com/user-attachments/assets/76f26e01-c415-400a-b8d5-7690ed459214

VitorVieiraZ:
Support for QGIS's native `reuseLastValue` option and refactoring of `RememberAttributesController` to store attribute values persistently in QSettings instead of in-memory, making them survive reinstantiation on each new form open. This adds project-specific storage with `ActiveProject` integration. On project deletion, stored settings from `RememberAttributesController` are cleared.

Fix #3784

### DIFF
--- a/app/attributes/attributecontroller.cpp
+++ b/app/attributes/attributecontroller.cpp
@@ -183,8 +183,6 @@ void AttributeController::setRememberAttributesController( RememberAttributesCon
   if ( mRememberAttributesController != rememberAttributes )
   {
     mRememberAttributesController = rememberAttributes;
-    if ( mRememberAttributesController && mFeatureLayerPair.layer() )
-      mRememberAttributesController->storeLayerFields( mFeatureLayerPair.layer() );
     emit rememberAttributesChanged();
   }
 }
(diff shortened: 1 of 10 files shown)


VitorVieiraZ:
Testing looks good to me. The only scenario to confirm, that is set this way now: if we have a fresh project with `QGIS's reuseLastValue` set for a field and our tool `on` in Settings, then open the form, we encounter the checkbox for that field unchecked, although this field with `QGIS's reuseLastValue` it works fine when our tool is `off`. If this behavior is intentional, then everything is okay.

uclaros:
Untitled-2023-10-31-1527

tomasMizera:
Hi @uclaros - is this one ready for review pleasE?

PR #3792, opened 2025-04-01
Fix click area not filling whole parent bug

### LINKED ISSUE #3732: Enlarge project item's click area
In the image below you can see the current (red) click area for project item. This is quite limiting as clicking outside of the red area, but still in the project item results in no action.

Let's make the click area work within the blue zone instead.

Image

Withalion:
Fixes #3732

### DIFF
--- a/app/qml/project/components/MMProjectDelegate.qml
+++ b/app/qml/project/components/MMProjectDelegate.qml
@@ -56,6 +56,13 @@ Control {
   background: Rectangle {
     color: root.projectIsOpened ? __style.forestColor : __style.polarColor
     radius: __style.radius12
+    MouseArea {
+      anchors.fill: parent
+      onClicked: function( mouse ) {
+        mouse.accepted = true
+        root.openRequested()
+      }
+    }
   }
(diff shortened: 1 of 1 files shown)


tomasMizera:
Looks good to me from the code perspective, @VitorVieiraZ can you have additional look please?

VitorVieiraZ (changes requested):
> Looks good to me from the code perspective, @VitorVieiraZ can you have additional look please?

All good with tests. To finalize, I suggest removing previous and redundant MouseArea implementation: https://github.com/MerginMaps/mobile/blob/874902313c26a63f78a045ac1d1113c3eef01777/app/qml/project/components/MMProjectDelegate.qml#L76

tomasMizera:
> All good with tests. To finalize, I suggest removing previous and redundant MouseArea implementation:

I do not think that would work - I believe the text would not be clickable then.

VitorVieiraZ (approved):
Tests passed.

tomasMizera (approved):

PR #3793, opened 2025-04-01
Make Mergin Maps unusable for oil companies

ValentinBuira:
This PR make Mergin Maps unusable for a set of company/institution harmful to the environment 

This work by self sabotage the method in charge of adding a vertex point, by adding a random offset in case the user email belong to a blacklisted company name or institution (also include the us army 🤫)

The sabotage itself should be high enough to be disruptive but not high enough to be noticeable at first sight


PeterPetrik:
While I believe this was intended as an April Fool's joke, I do not find it appropriate or funny. The pull request and its content do not reflect the views of Lutra Consulting or the Mergin Maps project.

PR #3794, opened 2025-04-01
Add form button to read only feature previews

### LINKED ISSUE #3697: Adjust preview panel for read-only layers
If you have read-only layer, the "edit" button is hidden and the feature preview looks like this:

| Editable feature | Read-only feature |
|--------|--------|
| Image | Image | 

This is fine, but makes it hard for users to realize they can still open the form by clicking anywhere in the panel. We found this out on a workshop on UC.

There is also a possibility that users will completely miss the option to open a form in read-only mode - as this requires you to click anywhere in the preview panel, but the "edit" button.

---

Suggestions: Let's add a new primary button "Open form" (text to be confirmed); visual:

https://www.figma.com/design/d78a4ehBBgK6MyxCTuCAqg/Mobile-App?node-id=4074-19617&t=21E8Z32xyqWfRNdD-4

Image

We have two options here:

1. replace the current `edit` button with the new `open form` button
  - Pros 👍🏻 : the UI will be the same for all features (read-only and editable), the option to edit the form will be in the form toolbar; One could argue that this is how it should work - you first open the form, then hit the "Edit" option.
  - Cons 👎🏻 : on the flip side, we add an additional click when someone wants to edit the feature

2. as a new button (only for read-only features)
  - Pros 👍🏻 : The existing flow to edit feature will stay the same
  - Cons 👎🏻 : Different preview buttons for read-only and editable features

Image

> Note: we have a "edit feature" button in the form toolbar in case an editable feature is opened for reading

Withalion:
Fixes #3697 

image

### DIFF
--- a/app/mmstyle.h
+++ b/app/mmstyle.h
@@ -171,6 +171,7 @@ class MMStyle: public QObject
     Q_PROPERTY( QUrl cloudIcon READ cloudIcon CONSTANT )
     Q_PROPERTY( QUrl measurementToolIcon READ measurementToolIcon CONSTANT )
     Q_PROPERTY( QUrl closeShapeIcon READ closeShapeIcon CONSTANT )
+    Q_PROPERTY( QUrl formIcon READ formIcon CONSTANT )
 
     // Filled Icons - for visualizing of selected item in toolbar
     Q_PROPERTY( QUrl projectsFilledIcon READ projectsFilledIcon CONSTANT )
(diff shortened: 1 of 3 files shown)


VitorVieiraZ (commented):
With the new button, width overflows and `Stake out` button gets cut off by margins. We could either use property `size: MMButton.Sizes.Small` for both `Open form` and `Stake out` when `visible: !internal.showEditButton` or change text to `Form`. What do you think, @tomasMizera?

uclaros:
> With the new button, width overflows and Stake out button gets cut off by margins.

_Stake out_ and the upcoming _Select more_ will be hidden behind a hamburger button combo box, so I don't think this needs to be addressed yet.

tomasMizera:
> _Stake out_ and the upcoming _Select more_ will be hidden behind a hamburger button combo box, so I don't think this needs to be addressed yet.

I agree with @uclaros, let's keep this one as is for now :)

tomasMizera:
btw. I believe you should be able to scroll the list of buttons now :)

uclaros:
> btw. I believe you should be able to scroll the list of buttons now :)

unfortunately not, at least on Linux build.
I can only scroll the row of buttons if I enable the ScrollView's scrollbar and drag that.

tomasMizera:
> unfortunately not, at least on Linux build. I can only scroll the row of buttons if I enable the ScrollView's scrollbar and drag that.

Ah, right, we intentionally disabled that. Anyways, we can make the `disable` flag dynamic in the future

tomasMizera (approved):

tomasMizera:
Let's merge this now :) 

> With the new button, width overflows and Stake out button gets cut off by margins. We could either use property size: MMButton.Sizes.Small for both Open form and...

It was a good review catch @VitorVieiraZ!

PR #3798, opened 2025-04-02
Fix password hint

### LINKED ISSUE #3734: Password hint is not accurate
Might be worth the simplicity, but you can actually choose a password while not respecting the instructions.

Image

https://github.com/user-attachments/assets/e199c41c-05c5-4948-9586-8d87db206f93

The exact rule is (probably) on the server. However, it might sound cumbersome.

Image

Withalion:
Fixes #3734 

We replace the outdated hint with new hint.

image

Also part of this PR will be seeing this hint, when the password check fails on server (which actually does the whole check) and not just notification that password is not strong enough.

### DIFF
--- a/core/merginapi.cpp
+++ b/core/merginapi.cpp
@@ -1311,6 +1311,12 @@ void MerginApi::registrationFinished( const QString &login, const QString &passw
         emit registrationFailed( msg, RegistrationError::RegistrationErrorType::EMAIL );
         emit notifyError( tr( "Registration failed" ) );
       }
+      else if ( serverErrorCode == "ExistingEmail" )
+      {
+        const QString msg = tr( "This email address is already registered" );
+        emit registrationFailed( msg, RegistrationError::RegistrationErrorType::EMAIL );
+        emit notifyError( tr( "Registration failed" ) );
+      }
       else if ( serverErrorCode == "InvalidPassword" )
       {
         const QString msg = tr( "Password not strong enough!%1"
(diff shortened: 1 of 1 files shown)


VitorVieiraZ (approved):

core/merginapi.cpp
```diff
@@ -829,12 +829,15 @@ void MerginApi::registerUser( const QString &email,
...
-                  .arg( "<br />  -" );
+    const QString msg = tr( "%1%3 Password must be at least 8 characters long.%4"
+                            "%3 Include at least 3 character categories among the following:"
+                            "<ul type=\"disc\">"
+                            "%3 Lowercase characters (a-z)%4"
+                            "%3 Uppercase characters (A-Z)%4"
+                            "%3 Digits (0-9)%4"
+                            "%3 Special characters %4"
+                            "%2%4%2" )
+                        .arg( "<ul>", "</ul>", "<li>", "</li>" );
```
tomasMizera, line 840: How about saying
```suggestion
const QString msg = tr( "%1%3 Password must be at least 8 characters long and include:%4"
                        "<ul type=\"disc\">"
                        "%3 Lowercase letters (a–z)%4"
                        "%3 Uppercase letters (A–Z)%4"
                        "%3 At least one digit (0–9) or special character%4"
                        "%2%4%2" )
                    .arg( "<ul>", "</ul>", "<li>", "</li>" );
```

tomasMizera: I think this keeps the same message, +1 to merge it then :)

tomasMizera (approved):

PR #3807, opened 2025-04-10
Multi Feature Editing

uclaros:
This PR introduces the multi-editing feature.

Process can be initiated with the new _Select more_ button in the preview form.
image
More features can be selected and de-selected by taping them on the map.
image

In the edit form, fields that have different values in each selected feature are displaying a placeholder text: `* Mixed Values *` and their contents will be ignored when saving.
Modifying a `* Mixed Values *` field will remove the placeholder and store the new value to all selected features when saving.
image
image
image

### DIFF
--- a/app/multieditmanager.cpp
+++ b/app/multieditmanager.cpp
@@ -35,7 +35,7 @@ void MultiEditManager::initialize( const FeatureLayerPair &pair )
   }
 }
 
-void MultiEditManager::toggleSelection( const FeatureLayerPair &pair )
+void MultiEditManager::toggleSelect( const FeatureLayerPair &pair )
 {
   if ( !pair.isValid() ||
        ( mLayer && mLayer != pair.layer() ) )
--- a/app/multieditmanager.cpp
+++ b/app/multieditmanager.cpp
@@ -174,7 +174,7 @@ void MultiEditManager::createTemporaryLayer()
   for ( const QgsVectorLayerJoinInfo &join : joins )
   {
     // do not copy join information for auxiliary layer
-    if ( !auxLayer || ( auxLayer && auxLayer->id() != join.joinLayerId() ) )
+    if ( !auxLayer || auxLayer->id() != join.joinLayerId() )
       mTempLayer->addJoin( join );
   }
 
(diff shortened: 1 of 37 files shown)


app/multieditmanager.cpp
```diff
@@ -0,0 +1,178 @@
...
+{
+  if ( !pair.isValid() ||
+       ( mLayer && mLayer != pair.layer() ) )
+    return;
+
+  if ( mModel->count() == 0 )
+  {
+    mModel->populate( { pair } );
+    return;
+  }
```
tomasMizera, line 47: Do we need an extra handling if `mModel` is empty? Won't the code below handle this situation?

app/qml/map/MMMapController.qml
```diff
@@ -238,6 +251,13 @@ Item {
           root.nothingIdentified()
         }
       }
+      else if ( root.state === "select" )
+      {
+        let screenPoint = Qt.point( point.x, point.y )
+        let pair = identifyKit.identifyOne( screenPoint )
```
tomasMizera, line 257: How about putting the two calls at the beginning of the function so we do not need to duplicate them here?

app/qml/map/MMMapController.qml
```diff
@@ -922,6 +942,43 @@ Item {
...
   }
 
+
+  Loader {
+    id: multiEditLoader
+
+    anchors.fill: mapCanvas
+
+    // active: true
+    active: root.state === "select"
```
tomasMizera, line 952: Could we use some other term than `select` for this functionality within the code? It somehow does not immediately hit me that this is the multi-select functionality we talk about. How about `mutliselection` or `multiselect`?

e.g. I get `onMultiSelect` instead of `onSelect`

tomasMizera (commented):
Just a bunch of comments from a quick view.

Do you plan to add the placeholder text for fields with multiple values and "apply default value on update" enabled?

tomasMizera:
This looks really good!

app/multieditmanager.cpp
uclaros: I believe this can go now, it was from before I added an explicit `initialize()` method.

app/qml/map/MMMapController.qml
uclaros: It will still need to be within a check for the two `root.state`s, as we wouldn't want to spend the time identifying for the other states

tomasMizera (approved):
🎉

PR #3809, opened 2025-04-11
Fix multiline feature title breaking eliding & unintended HTML tags parsing 

### LINKED ISSUE #3760: Long feature name overflows
Long feature name overflows the determined space for it when the feature is opened.

Image

Withalion:
Fixes #3760 and also some unintended HTML parsing

| bug | old version | new version |
| --- | --- | --- |
| Multi line feature name overflows | image | image |
| Feature name with html tag in form page | image | image |
| Feature name with html tag in form preview & Feature field with html tag in form preview | image | image |
| Feature name with html tag in feature list | image | image |

### DIFF
--- a/app/qml/components/MMListDelegate.qml
+++ b/app/qml/components/MMListDelegate.qml
@@ -84,6 +84,7 @@ Item {
 
           font: __style.t3
           text: root.text
+          textFormat: Text.PlainText
 
           color: __style.nightColor
         }
--- a/app/qml/form/MMPreviewDrawer.qml
+++ b/app/qml/form/MMPreviewDrawer.qml
@@ -90,6 +90,7 @@ Item {
           Layout.preferredHeight: implicitHeight
 
           text: root.controller.title
+          textFormat: Text.PlainText
 
           font: __style.t1
           color: __style.forestColor
@@ -249,6 +250,7 @@ Item {
                 width: parent.width
 
                 text: model.Value
+                textFormat: Text.PlainText
                 font: __style.t3
                 color: __style.nightColor
 
--- a/app/qml/components/MMPageHeader.qml
+++ b/app/qml/components/MMPageHeader.qml
@@ -56,6 +56,7 @@ Rectangle {
 
     text: root.title?.toString()?.replace(/\n/g, ' ') ?? ''
     elide: Text.ElideMiddle
+    textFormat: Text.PlainText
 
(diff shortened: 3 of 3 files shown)


VitorVieiraZ (commented):
We have the use case where we set PDF links, for instance, and we could see the formatted HTML in the preview drawer, like that:

app/qml/form/MMPreviewDrawer.qml
```diff
@@ -249,6 +250,7 @@ Item {
                 width: parent.width
 
                 text: model.Value
+                textFormat: Text.PlainText
```
tomasMizera, line 253: As @VitorVieiraZ suggested, I would not change this here - I'd leave the option to pass in html, for the sake of links :)

app/qml/components/MMListDelegate.qml
```diff
@@ -84,6 +84,7 @@ Item {
 
           font: __style.t3
           text: root.text
+          textFormat: Text.PlainText
```
tomasMizera, line 87: This does not seem to have any impact now, but sounds like a good idea to do :)

app/qml/form/MMPreviewDrawer.qml
tomasMizera: Other places look good to me!

tomasMizera (changes requested):

app/qml/components/MMListDelegate.qml
Withalion: But I think it will break the links, which wouldn't be a problem if the text would be user comprehensible

tomasMizera: Do you know about any place where we use `MMListDelegate` for links? I quickly checked the code and did not find any occurrence.

tomasMizera: If we need it, we can always override the property

Withalion: If the feature field, which is set up as a link is also used as feature title, won't we use it also as a title in `MMListDelegate`?

tomasMizera (approved):

tomasMizera:
I'd say using `html` in feature title should have never been evaluated and supported. Links there do not work either as we do not handle them via `onLinkActivated` signal.

harminius (approved):
Seems ok :green_book:

PR #3810, opened 2025-04-11
Fix testHasLocalProjectChanges on Linux

VitorVieiraZ:
Introduce a delay after file operation to ensure consistent test behavior, particularly on Linux, where ocasional failures have been observed.

### DIFF
--- a/app/test/testmerginapi.cpp
+++ b/app/test/testmerginapi.cpp
@@ -2744,6 +2744,7 @@ void TestMerginApi::writeFileContent( const QString &filename, const QByteArray 
   bool ok = f.open( QIODeviceBase::WriteOnly );
   Q_ASSERT( ok );
   f.write( data );
+  f.flush();
   f.close();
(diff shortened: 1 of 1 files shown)


app/test/testmerginapi.cpp
```diff
@@ -3286,6 +3286,9 @@ void TestMerginApi::testHasLocalProjectChanges()
 
   // 4: fourth scenario => local files differs from metadata, selective sync supported
   writeFileContent( projectDir + "/test.txt", QByteArray( "modified content" ) );
+
+  QTest::qSleep( 1000 ); // small delay to ensure filesystem operations complete
```
tomasMizera, line 3290: Can we instead listen to some "finished" signal?

tomasMizera (changes requested):

tomasMizera (approved):
Ok, let's keep the debug lines for now.

PR #3811, opened 2025-04-12
Remove dep on mobile-sdk (Hello vcpkg)

PeterPetrik:
Move vcpkg build of dependancies to main mobile repo, so we are not dependent on 2 separate repos anymore. 

fixes https://github.com/MerginMaps/mobile/issues/620


tomasMizera:
For future reference: https://doc.qt.io/qt-6/build-optimized-qt.html We should be able to optimize the app size this way

Withalion:
should also fix #4000

app/inpututils.cpp
```diff
@@ -1971,20 +1971,21 @@ void InputUtils::sanitizeFileName( QString &fileName )
   fileName = fileName.trimmed();
 }
 
-void InputUtils::updateQgisFormats( const QByteArray& output )
+void InputUtils::updateQgisFormats( const QByteArray &output )
 {
 #ifdef MM_FORMATS_PATH
   const QString sourcePath( STR( MM_FORMATS_PATH ) );
-  QFile file(sourcePath);
+  QFile file( sourcePath );
```
PeterPetrik, line 1978: I think it is better to write formats to stdout and exit the program (flag `--qgis_formats`) and then in CI append this stdout to any file you want ? so you do not need to work with files at all in the input app

Withalion: I understand your concern, however we should have an up to date file with supported formats. And using this when manually updating the app version seems like good solution to me.

PR #3834, opened 2025-04-22
Minimal fix for reuse last value

### LINKED ISSUE #3784: Reusing last entered value is broken

uclaros:
Minimal fix for #3784

Fix #3784

### DIFF
--- a/app/attributes/rememberattributescontroller.h
+++ b/app/attributes/rememberattributescontroller.h
@@ -34,6 +34,8 @@ class  RememberAttributesController : public QObject
     //! Returns TRUE if remembering values is allowed
     Q_PROPERTY( bool rememberValuesAllowed READ rememberValuesAllowed WRITE setRememberValuesAllowed NOTIFY rememberValuesAllowedChanged )
 
+    Q_PROPERTY( QString activeProjectId MEMBER mActiveProjectId )
+
   public:
     RememberAttributesController( QObject *parent = nullptr );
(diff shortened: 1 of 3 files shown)


app/attributes/rememberattributescontroller.h
```diff
@@ -68,9 +72,10 @@ class  RememberAttributesController : public QObject
     };
 
     bool mRememberValuesAllowed = false;
-    //! Remembered last created feature for each layer (key)
-    QHash<QString, RememberedValues> mRememberedValues;
+    //! Remembered last created feature for each project/layer (key)
+    static QHash<QString, RememberedValues> sRememberedValues;
```
PeterPetrik, line 76: why there is need for static variable?

uclaros: A `RememberAttributesController` is instantiated with each forms' `AttributeController`, so we need the storage to persist. We agreed with a static var instead of QSettings for now.

tomasMizera (approved):

PR #3838, opened 2025-04-23
Fix qr code by moving capture session in view port

tomasMizera:
Seems like `CaptureSession` needs to be in visual children (worked with previous Qt version)


PeterPetrik (approved):

tomasMizera:
I checked iOS and it works, build fails because we already have another version approved. It will pass once merged to master :)

harminius (approved):
:camera_flash: 
works on Android

PR #3839, opened 2025-04-23
Minimal fix for photo deletion 

### LINKED ISSUE #3786: Deleted photo is stays in the project
When you take a photo while recording a feature but don't save the feature the photo is synced.

Scenario
1. Record a feature with a photo field
2. Take a photo but don't save the new feature
3. Tap the trash icon and choose to delete the photo
4. You can repeat steps 2 and 3 few times
5. Then save the feature
6. When you look at Local changes you see all the photos that should be deleted
7. When you sync all the photos are synced

VitorVieiraZ:
Save/cancel callbacks on `MMFormPage` have been renamed to match existing methods in `MMFormPhotoEditor`, `callbackOnFormSaved` and `callbackOnFormCanceled`. `onSaved()` and `onCanceled()` methods now invoke these callbacks properly. 

Fixes #3786

### DIFF
--- a/app/qml/form/MMFormPage.qml
+++ b/app/qml/form/MMFormPage.qml
@@ -386,13 +386,13 @@ Page {
         ignoreUnknownSignals: true
 
         function onSaved() {
-          if ( formEditorsLoader.item && typeof formEditorsLoader.item.callbackOnSave === "function" ) {
+          if ( formEditorsLoader.item && typeof formEditorsLoader.item.callbackOnFormSaved === "function" ) {
             formEditorsLoader.item.callbackOnFormSaved()
           }
         }
 
         function onCanceled() {
-          if ( formEditorsLoader.item && typeof formEditorsLoader.item.callbackOnCancel === "function" ) {
+          if ( formEditorsLoader.item && typeof formEditorsLoader.item.callbackOnFormCanceled === "function" ) {
             formEditorsLoader.item.callbackOnFormCanceled()
           }
         }


app/qml/form/editors/MMFormPhotoEditor.qml
```diff
@@ -160,14 +161,43 @@ MMFormPhotoViewer {
...
+        var idx = internal.imageSourcesToDelete.indexOf( originalAbsPath )
+        if ( idx !== -1 ) {
+          internal.imageSourcesToDelete.splice( idx, 1 )
+        }
+      }
+      // 2. add current value to deletion queue if it exists and it's not the original value
+      if ( root._fieldValue && root._fieldValue !== root._fieldOriginalValue ) {
+        let existingPhotoPath = __inputUtils.getAbsolutePath( root._fieldValue, internal.prefixToRelativePath )
+        internal.imageSourcesToDelete.push( existingPhotoPath )
+      }
```
tomasMizera, line 192: I still have mixed feelings about this code. I would feel much safer if it was in c++ and covered with unit tests.. I am afraid this can explode in ways we do not imagine now. This is just too much javascript/qml logic for something this critical. My suggestion would be this:
 1. we merge just the renaming fix (`callbackOnFormCanceled`)
 2. we close the other work and keep that in mind (reference) for the photo refactoring - there we will put all of this code into cpp and cover with unit tests..

tomasMizera (changes requested):
@VitorVieiraZ can I ask you to update the PR according to my comment (keep just the renaming fix)?

tomasMizera:
> @VitorVieiraZ can I ask you to update the PR according to my comment (keep just the renaming fix)?

For the issues where you add and remove multiple images and then hit save/cancel, I suggest to create a ticket with good description and label it with `photo-refactoring`

VitorVieiraZ:
> > @VitorVieiraZ can I ask you to update the PR according to my comment (keep just the renaming fix)?
> 
> For the issues where you add and remove multiple images and then hit save/cancel, I suggest to create a ticket with good description and label it with `photo-refactoring`

Created an issue at https://github.com/MerginMaps/mobile/issues/3868.

tomasMizera (approved):

harminius (changes requested):
The fix doesn't work for me. Tried this scenarios:
- add a photo while recording a feature and don't save the feature (by hitting back icon and canceling the changes)
- same as previous but delete the photo before discarting the feature
- take a photo while recording and delete the feature before saving it (trash icon button left (note to @tomasMizera if there should be this option available - with 'delete' label))
- record the feature with a photo, save it, and then delete the feature

All cases result in having the photo in pending changes and uploading it with sync :red_circle:

VitorVieiraZ:
> The fix doesn't work for me. Tried this scenarios:
> 
> * add a photo while recording a feature and don't save the feature (by hitting back icon and canceling the changes)
> * same as previous but delete the photo before discarting the feature
> * take a photo while recording and delete the feature before saving it (trash icon button left (note to @tomasMizera if there should be this option available - with 'delete' label))
> * record the feature with a photo, save it, and then delete the feature
> 
> All cases result in having the photo in pending changes and uploading it with sync 🔴

This fix will only work in the case where we click on the trash and delete the photo, being the last to be deleted the one that will actually be deleted. The other cases mentioned won't work with this fix.

harminius:
> > The fix doesn't work for me. Tried this scenarios:
> > 
> > * add a photo while recording a feature and don't save the feature (by hitting back icon and canceling the changes)
> > * same as previous but delete the photo before discarting the feature
> > * take a photo while recording and delete the feature before saving it (trash icon button left (note to @tomasMizera if there should be this option available - with 'delete' label))
> > * record the feature with a photo, save it, and then delete the feature
> > 
> > All cases result in having the photo in pending changes and uploading it with sync 🔴
> 
> This fix will only work in the case where we click on the trash and delete the photo, being the last to be deleted the one that will actually be deleted on saving (not cancelling for now). The other cases mentioned won't work with this fix.

Thanks for the clarification. The scenario you mentioned works like you described. :green_circle:

harminius (approved):

Issue #2644, opened 2023-03-23
Store map extent, map theme and layers visibility to app settings


tomasMizera:
Let's store these values in app settings so that if you open the app again, you see the project in the state where you left it previously. It should work even when you reopen a different project.

tomasMizera:
See https://odoo.lutraconsulting.co.uk/odoo/all-tickets/10868

PR #3840, opened 2025-04-23
WIP - Store map extent, layer visibility and map themes in app settings

### LINKED ISSUE #2445: Random zooming in/out when reopening a project
1. create a project in Input (had this issue only with projects created within Input) 
2. zoom in to any scale e.g. 100km 
3. close Input 
4. open again 
5. your project will be open, but the zoom level will be either maximum or totally zoomed in. 

iPhone 11; iOS: 16.1.1.

https://user-images.githubusercontent.com/94351263/205652009-a79024a6-a6be-42b4-a849-51914a632adf.MP4

VitorVieiraZ:
Resolves #2445, #2644

### DIFF
--- a/app/activeproject.cpp
+++ b/app/activeproject.cpp
@@ -613,17 +613,14 @@ QList<QgsMapLayer *> ActiveProject::getVisibleLayers() const
 void ActiveProject::restoreLayersVisibility()
 {
   // restore layer visibility
-  if ( mMapSettings )
+  QStringList savedIds = mAppSettings.visibleLayerIdsForProject( projectFullName() );
+  if ( !savedIds.isEmpty() )
   {
-    QStringList savedIds = mAppSettings.visibleLayerIdsForProject( projectFullName() );
-    if ( !savedIds.isEmpty() )
+    QgsLayerTree *root = mQgsProject->layerTreeRoot();
+    for ( QgsLayerTreeLayer *nodeLayer : root->findLayers() )
     {
-      QgsLayerTree *root = mQgsProject->layerTreeRoot();
-      for ( QgsLayerTreeLayer *nodeLayer : root->findLayers() )
-      {
-        bool ok = savedIds.contains( nodeLayer->layer()->id() );
-        nodeLayer->setItemVisibilityChecked( ok );
-      }
+      bool ok = savedIds.contains( nodeLayer->layer()->id() );
+      nodeLayer->setItemVisibilityChecked( ok );
     }
   }
 }
--- a/app/activeproject.cpp
+++ b/app/activeproject.cpp
@@ -505,7 +505,7 @@ void ActiveProject::switchLayerTreeNodeVisibility( QgsLayerTreeNode *node )
 
   node->setItemVisibilityChecked( !node->isVisible() );
 
-  updateMapTheme();
+  restoreMapTheme();
   updateActiveLayer();
   updateMapSettingsLayers();
 
@@ -612,7 +612,7 @@ QList<QgsMapLayer *> ActiveProject::getVisibleLayers() const
 
 void ActiveProject::restoreLayersVisibility()
 {
-  // restore layer visibility
+  // restore layers visibility
   QStringList savedIds = mAppSettings.visibleLayerIdsForProject( projectFullName() );
   if ( !savedIds.isEmpty() )
   {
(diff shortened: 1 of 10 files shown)


core/coreutils.h
```diff
@@ -135,6 +137,12 @@ class CoreUtils
      */
     static bool replaceValueInJson( const QString &filePath, const QString &key, const QJsonValue &value );
 
+    /**
+     * Sanitizes a string by replacing path separators with underscores
+     * Useful to set QSettings group names, avoiding unpredicted behavior caused by slashes
+     */
+    static QString sanitizePathSlashes( const QString &input );
```
Withalion, line 144: Maybe we could wait for the introduction of project ID's and there would be no necessity for sanitizing, however the QSettings would be less human-friendly.

core/localprojectsmanager.cpp
```diff
@@ -115,6 +115,9 @@ void LocalProjectsManager::removeLocalProject( const QString &projectId )
     {
       emit aboutToRemoveLocalProject( mProjects[i] );
 
+      QString sanitizedProjectFullName = CoreUtils::sanitizePathSlashes( mProjects[i].fullName() );
```
Withalion, line 118: const

app/activeproject.h
```diff
@@ -132,6 +132,21 @@ class ActiveProject: public QObject
...
+    QString mapThemeForProject() const;
+
+    //! Retrieves map theme from QSettings
+    void setMapThemeForProject();
+
+    //! Stores visible layers ids in QSettings
+    QStringList visibleLayerIdsForProject() const;
+
+    //! Restores visible layers ids from QSettings
+    void setVisibleLayerIdsForProject( const QStringList &layerIds );
```
Withalion, line 145: The comments should be the other way around :) maybe instead of `set` we could use `store` prefix for these functions

app/activeproject.cpp
```diff
@@ -416,11 +422,6 @@ void ActiveProject::updateMapTheme()
 
 void ActiveProject::setMapTheme( const QString &themeName )
 {
-  if ( mMapTheme == themeName )
-  {
-    return;
-  }
-
```
Withalion, line 423: Maybe instead of deleting this check we would do a new check before if the map theme is already stored and if not store it. Running this gives unnecessary complexity just to be sure to store the map theme

app/activeproject.cpp
```diff
@@ -506,6 +508,14 @@ void ActiveProject::switchLayerTreeNodeVisibility( QgsLayerTreeNode *node )
   updateMapTheme();
   updateActiveLayer();
   updateMapSettingsLayers();
+
+  QStringList visibleLayerIds;
+  for ( QgsMapLayer *layer : getVisibleLayers() )
```
Withalion, line 513: const

app/activeproject.cpp
```diff
@@ -602,3 +612,83 @@ QList<QgsMapLayer *> ActiveProject::getVisibleLayers() const
 
   return visibleLayers;
 }
+
+void ActiveProject::restoreLayersVisibility()
+{
+  QStringList savedIds = visibleLayerIdsForProject();
```
Withalion, line 618:
```suggestion
  const QStringList savedIds = visibleLayerIdsForProject();
```

app/activeproject.cpp
```diff
@@ -602,3 +612,83 @@ QList<QgsMapLayer *> ActiveProject::getVisibleLayers() const
...
+  if ( !savedIds.isEmpty() )
+  {
+    QgsLayerTree *root = mQgsProject->layerTreeRoot();
```
Withalion, line 621:
```suggestion
    const QgsLayerTree *root = mQgsProject->layerTreeRoot();
```

app/activeproject.cpp
```diff
@@ -602,3 +612,83 @@ QList<QgsMapLayer *> ActiveProject::getVisibleLayers() const
...
+    for ( QgsLayerTreeLayer *nodeLayer : root->findLayers() )
+    {
+      bool visible = savedIds.contains( nodeLayer->layer()->id() );
```
Withalion, line 624:
```suggestion
      const bool visible = savedIds.contains( nodeLayer->layer()->id() );
```

app/activeproject.cpp
```diff
@@ -602,3 +612,83 @@ QList<QgsMapLayer *> ActiveProject::getVisibleLayers() const
...
+      nodeLayer->setItemVisibilityChecked( visible );
+    }
+  }
+}
+
+void ActiveProject::restoreMapTheme()
+{
+  QString savedTheme = mapThemeForProject();
```
Withalion, line 632:
```suggestion
  const QString savedTheme = mapThemeForProject();
```

Withalion (commented):
Some polishing to do, but besides that looks good for now. We should think about using project ids instead of project names, so we won't introduce new way to break our app without sanitizing

PR #3857, opened 2025-04-25
Map extent stored in app settings following no map movement

### LINKED ISSUE #2445: Random zooming in/out when reopening a project

### LINKED ISSUE #2644: Store map extent, map theme and layers visibility to app settings

VitorVieiraZ:
Map extent is now stored in app settings after 2 seconds of map movement inactivity (no extentChanged). Each time the extent changes, a timer is reset, and if no further changes occur within 2 seconds, current extent is saved. When opening the app, if no extent can be retrieved from the map canvas XML, the app falls back to using last stored extent.

Fixes #2445 
Relates to #2644

### DIFF
--- a/app/activeproject.cpp
+++ b/app/activeproject.cpp
@@ -185,6 +185,11 @@ bool ActiveProject::forceLoad( const QString &filePath, bool force )
     QString role = MerginProjectMetadata::fromCachedJson( CoreUtils::getProjectMetadataPath( mLocalProject.projectDir ) ).role;
     setProjectRole( role );
 
+    if ( mMapSettings )
+    {
+      mMapSettings->setProjectId( projectFullName() );
+    }
+
     updateMapTheme();
     updateActiveLayer();
     updateMapSettingsLayers();
@@ -346,6 +351,7 @@ void ActiveProject::setMapSettings( InputMapSettings *mapSettings )
 
   mMapSettings = mapSettings;
   updateMapSettingsLayers();
+  mMapSettings->setProjectId( projectFullName() );
 
   emit mapSettingsChanged();
 }
--- a/app/map/inputmapsettings.cpp
+++ b/app/map/inputmapsettings.cpp
@@ -20,6 +20,8 @@
 #include "qgsmessagelog.h"
 #include "qgsprojectviewsettings.h"
 
+static constexpr int EXTENT_SAVE_DELAY_MS = 2000;
+
 InputMapSettings::InputMapSettings( QObject *parent )
   : QObject( parent )
 {
@@ -34,7 +36,7 @@ InputMapSettings::InputMapSettings( QObject *parent )
   // store map extent to QSettings when extent hasn't changed for 2 seconds
   mSaveExtentTimer.setSingleShot( true );
   connect( &mSaveExtentTimer, &QTimer::timeout, this, &InputMapSettings::saveExtentToSettings );
-  connect( this, &InputMapSettings::extentChanged, this, [this]() { mSaveExtentTimer.start( 2000 ); } );
+  connect( this, &InputMapSettings::extentChanged, this, [this]() { mSaveExtentTimer.start( EXTENT_SAVE_DELAY_MS ); } );
 }
 
 void InputMapSettings::setProject( QgsProject *project )
@@ -378,5 +380,11 @@ void InputMapSettings::loadSavedExtent()
   settings.endGroup();
 
   if ( !extent.isEmpty() && extent.isFinite() )
+  {
     setExtent( extent );
+  }
+  else
+  {
+    CoreUtils::log( "MapSettings", QStringLiteral( "No valid saved extent found for project %1" ).arg( mProject->baseName() ) );
+  }
 }
--- a/app/map/inputmapsettings.cpp
+++ b/app/map/inputmapsettings.cpp
@@ -14,6 +14,7 @@
 
 #include "qgis.h"
 #include "inputmapsettings.h"
+#include "coreutils.h"
 
 #include "qgsmaplayer.h"
 #include "qgsmessagelog.h"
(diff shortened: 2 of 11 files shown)


tomasMizera (changes requested):
As we agreed on the call

app/activeproject.cpp
```diff
@@ -185,6 +185,11 @@ bool ActiveProject::forceLoad( const QString &filePath, bool force )
     QString role = MerginProjectMetadata::fromCachedJson( CoreUtils::getProjectMetadataPath( mLocalProject.projectDir ) ).role;
     setProjectRole( role );
 
+    if ( mMapSettings )
+    {
+      mMapSettings->setProjectId( projectFullName() );
```
Withalion, line 190: As already mentioned in #3840 we should probably use projectIds instead

app/activeproject.cpp
```diff
@@ -346,6 +351,7 @@ void ActiveProject::setMapSettings( InputMapSettings *mapSettings )
 
   mMapSettings = mapSettings;
   updateMapSettingsLayers();
+  mMapSettings->setProjectId( projectFullName() );
```
Withalion, line 354: Same as above

app/map/inputmapsettings.cpp
```diff
@@ -29,6 +32,12 @@ InputMapSettings::InputMapSettings( QObject *parent )
   connect( this, &InputMapSettings::extentChanged, this, &InputMapSettings::visibleExtentChanged );
   connect( this, &InputMapSettings::rotationChanged, this, &InputMapSettings::visibleExtentChanged );
   connect( this, &InputMapSettings::outputSizeChanged, this, &InputMapSettings::visibleExtentChanged );
+
+  // store map extent to QSettings when extent hasn't changed
+  mSaveExtentTimer.setSingleShot( true );
+  connect( &mSaveExtentTimer, &QTimer::timeout, this, &InputMapSettings::saveExtentToSettings );
+  connect( this, &InputMapSettings::extentChanged, this, [this]() { mSaveExtentTimer.start( EXTENT_SAVE_DELAY_MS ); } );
```
Withalion, line 39:
```suggestion
  connect( this, &InputMapSettings::extentChanged, this, [this] { mSaveExtentTimer.start( EXTENT_SAVE_DELAY_MS ); } );
```

app/map/inputmapsettings.cpp
```diff
@@ -346,3 +355,46 @@ void InputMapSettings::setTemporalEnd( const QDateTime &end )
...
+}
+
+void InputMapSettings::setProjectId( const QString &projectId )
+{
+  if ( projectId == mProjectId )
+    return;
+
+  mProjectId = projectId;
+  emit projectIdChanged();
+}
```
Withalion, line 371: Another approach that I would propose is to keep the project ID in `activeproject` and the active project would request the `mapsettings` to save the current extent or the mapsettings would fetch the project ID from active project when saving the current extent

app/map/inputmapsettings.cpp
```diff
@@ -346,3 +355,46 @@ void InputMapSettings::setTemporalEnd( const QDateTime &end )
...
+
+void InputMapSettings::saveExtentToSettings()
+{
+  QSettings settings;
```
Withalion, line 375: can be moved inside if statement

app/map/inputmapsettings.cpp
```diff
@@ -346,3 +355,46 @@ void InputMapSettings::setTemporalEnd( const QDateTime &end )
...
+}
+
+void InputMapSettings::loadSavedExtent()
+{
+  if ( mProjectId.isEmpty() )
+    return;
+
+  QSettings settings;
+  settings.beginGroup( QStringLiteral( "%1/%2" ).arg( mProjectId, CoreUtils::QSETTINGS_CACHED_MAP_EXTENT_GROUP ) );
+  QgsRectangle extent = settings.value( "extent" ).value<QgsRectangle>();
```
Withalion, line 392:
```suggestion
  const QgsRectangle extent = settings.value( "extent" ).value<QgsRectangle>();
```

app/map/inputmapsettings.h
```diff
@@ -285,10 +292,29 @@ class InputMapSettings : public QObject
     //! \copydoc InputMapSettings::temporalEnd
     void setTemporalEnd( const QDateTime &end );
 
+    //! \copydoc InputMapSettings::projectId
```
Withalion, line 295:
```suggestion
    
```
this doesn't make sense, We should keep `\copydoc` only for Qgis functions wrappers

app/map/inputmapsettings.h
```diff
@@ -285,10 +292,29 @@ class InputMapSettings : public QObject
...
+    QString projectId() const;
+
+    //! \copydoc InputMapSettings::projectId
```
Withalion, line 298: same as above

app/map/inputmapsettings.h
```diff
@@ -285,10 +292,29 @@ class InputMapSettings : public QObject
...
+    /**
+     * Retrieves saved map extent from QSettings and set it as the map extent
+     */
+    void loadSavedExtent();
+
   signals:
     //! \copydoc InputMapSettings::project
     void projectChanged();
 
+    //! \copydoc InputMapSettings::projectId
```
Withalion, line 315: same as above

app/test/testmerginapi.cpp
```diff
@@ -2744,7 +2744,6 @@ void TestMerginApi::writeFileContent( const QString &filename, const QByteArray
   bool ok = f.open( QIODeviceBase::WriteOnly );
   Q_ASSERT( ok );
   f.write( data );
-  f.flush();
```
Withalion, line 2747: ?

core/coreutils.cpp
```diff
@@ -383,3 +384,11 @@ bool CoreUtils::isValidEmail( const QString &email )
...
   return regEx.match( email ).hasMatch();
 }
+
+QString CoreUtils::sanitizePathSlashes( const QString &name )
+{
+  QString sanitizedName = name;
+  sanitizedName.replace( '/', '_' );
+  sanitizedName.replace( '\\', '_' );
+  return sanitizedName;
+}
```
Withalion, line 394: Might be unnecessary if we start using real project IDs

Withalion (commented):
Some more work is necessary to get it in line with #3840 and we should decide if we want to rebase this on project IDs adoption

Withalion:
Also all the builds are failing so we need to fix that before proceeding

tomasMizera:
Let's keep this as a prototype/reference, we will revisit the decision to add this in the future

PR #3864, opened 2025-04-29
Api root should not end with slash

### LINKED ISSUE #3862: Adjacent slashes in URL in API endpoints
Some endpoints from `merginapi.h` are called with double slashes in url. Example: `https://app.merginmaps.com//config` 

Qt previously seemed to work just fine (prob. merging slashes) but the new version is more strict and sends the call with the multiple slashes. Some servers fail to respond to such calls.

tomasMizera:
Api root will not include the trailing slash(es) from now on. URLs in api calls must start with `/`

Fixes #3862


tomasMizera:
urgent, merging

Issue #1102, opened 2021-01-18
Upload button not available


saberraz:
To replicate:
1- Make sure you have downloaded your project on device
2- Delete the project from Mergin server
3- Go to Home and click on ...
4- You are offered with _Status_ and _Remove from device_

The project on Home page is not aware of the deletion on the Mergin side and hence not allowing user to Upload the files.

wonder-sk:
Related to #1008

jozef-budac:
related to 1594

Issue #1008, opened 2020-11-17
Detect orphaned projects


wonder-sk:
If some projects in the list of local projects do not exist in Mergin anymore, the app does not indicate that in any way. Also, checking project status will only say "no changes".

It would be good to show some kind of warning icon next to the project name in the list. Clicking the warning icon would tell the user that the project does not exist anymore. And maybe later we could offer an option to create a new project and upload the content to "rescue" such project.

Slightly related to #379 as the warning icon could reuse the same icon area as "sync" icon would normally do...

jozef-budac:
Related to https://github.com/lutraconsulting/input/issues/576

Issue #576, opened 2020-01-30
Handling of deleted/replaced projects


wonder-sk:
Input should be able to identify if a project has been deleted from the server - or it may have been replaced by another project with the same name. To detect this, we may need a unique project ID from the server.

See also https://github.com/lutraconsulting/qgis-mergin-plugin/issues/110

saberraz:
Also related to https://github.com/lutraconsulting/input/issues/614

saberraz:
Closing in favour of #614

wonder-sk:
Mergin v2021.6 has introduced unique project IDs and added endpoint for listing projects by ID (`/v1/project/by_uuids?uuids=id1,id2,id3`). v2021.7 also adds fetching full project info by ID (`/v1/project/by_uuid/id1`)

Hopefully that should be enough to detect if:
- a project that is available locally with the same name is actually a different project now (IDs are not matching)
- a project that is available locally has been renamed / transferred (name may be different but IDs are the same)

PR #3884, opened 2025-05-06
Use project IDs internaly to manage projects

### LINKED ISSUE #1969: Migrate to server project ids
Currently we are not using project IDs coming from server, but use our custom ids. 
It will help us to detect orphaned projects / projects from different servers with the same name and deleted projects.

> We need to figure out how to assign IDs to local projects created in Input

Related to #1008

Withalion:
Fixes #1969 

This PR adds foundational work for supporting project IDs in mobile app. Besides issue mentioned above this is preliminary work for #1008, #1102, #576, project renaming etc. 

Outline of the changes done:
- core
  - CoreUtils
    - refactor
    - `getFullProjectName()`, `extractProjectName()` has been moved from `MerginApi`
  - Project
    - `LocalProject` supports UUID type of IDs + new `generateProjectId()`
    - `MerginProject` supports UUID type of IDs + new `fullname()`
    - `LocalProjectsList` changed to HashMap `LocalProjectsDict` with ID as key
  - LocalProjectsManager
    - keeps projects in HashMap
    - delete `projectFromMerginName()`, it was used only by tests
    - new `updateProjectId()` necessary function for creating new projects on server ( server returns new ID for already existing projects, app needs to adapt to it)
    - general refactor and move from project fullnames to project IDs as parameters in functions
  - MerginApi
    - general refactor, a lot of the functions are never called from QML for example
    - `listProjectsByName()` stays until there is `v2` endpoint for IDs
    - delete `migrateProjectToMergin()` just a wrapper for `createProject()` ( i think it was used just by tests)
    - delete `projectDiffableFiles()` not used anywhere
    - move `getFullProjectName()` & `extractProjectName()` to utils
    - change `deleteProject()` to use immediate deletion instead of scheduled, we use it only internally for project management 
    - remove `topic` parameter  from `networkErrorOccured()` it was only used by tests (they use http error codes now)
    - new `projectIdChanged()` signal to trigger update of local project ID to server generated ID
    - new `getProjectDetails()` it's leaner version of `getProjectInfo()`, the response won't contain versioning history
    - `Transactions` now keep projectId as key
    - generally everything what could be, was moved to use project IDs, the synchronization workflow couldn't, so for communication with API we use project names but internally the project ID is propagated and used where possible
    - when a local project is "migrated" to mergin there is new request for new project ID generated on server that is further processed and local ID is changed to match server ID
- app
  - QML
    - just refactor to support project IDs
  - C++
    - ActiveProject
      - new `projectId()` to get project ID of active project
    - Main
      - rework to project IDs
    - ProjectsModel
      - replaced `listProjectsByName()`  by `fetchProjectsById()`
      - reworked slots to use project IDs
      - removed `porjectNames()`, it was used internally and is not necessary since we moved to IDs
    - ProjectWizard
      - refactor, most importantly moved from deprecated `QVariant::type` to `QMetaType`
    - SynchronizationManager
      - keeps synchronization processes by project ID
      - reworked to use project IDs instead of project names
  - Tests
    - mostly refactor & minor changes
    - replaced `findProjectByName()` with template function
    - moved descriptive inline test function comments to header file or written new ones

### DIFF
--- a/app/projectsmodel.cpp
+++ b/app/projectsmodel.cpp
@@ -39,6 +39,7 @@ void ProjectsModel::initializeProjectsModel()
   if ( mModelType == ProjectModelTypes::LocalProjectsModel )
   {
     connect( mBackend, &MerginApi::listProjectsByNameFinished, this, &ProjectsModel::onListProjectsByNameFinished );
+    connect( mBackend, &MerginApi::refetchBrokenProjectsFinished, this, &ProjectsModel::onRefetchBrokenProjectsFinished );
     loadLocalProjects();
   }
   else if ( mModelType != ProjectModelTypes::RecentProjectsModel )
@@ -268,11 +269,30 @@ void ProjectsModel::onListProjectsByNameFinished( const MerginProjectsList &merg
     return;
   }
 
+  // filter out projects which returned errors and refetch them by ID
+  const QStringList brokenProjects = filterBrokenProjects( merginProjects );
+  if ( !brokenProjects.isEmpty() )
+  {
+    mBackend->refetchBrokenProjects( brokenProjects );
+  }
+
   beginResetModel();
   mergeProjects( merginProjects );
   endResetModel();
 
   setModelIsLoading( false );
+}
+
+void ProjectsModel::onRefetchBrokenProjectsFinished( const MerginProjectsList &merginProjects )
+{
+  if ( merginProjects.isEmpty() )
+  {
+    return;
+  }
+
+  beginResetModel();
+  mergeProjects( merginProjects, MergeStrategy::KeepPrevious );
+  endResetModel();
 }
 
 void ProjectsModel::mergeProjects( const MerginProjectsList &merginProjects, MergeStrategy mergeStrategy )
@@ -649,6 +669,26 @@ void ProjectsModel::loadLocalProjects()
     mergeProjects( MerginProjectsList() ); // Fills model with local projects
     endResetModel();
   }
+}
+
+QStringList ProjectsModel::filterBrokenProjects( const MerginProjectsList &list )
+{
+  QStringList errorProjects;
+  for ( MerginProject project : list )
+  {
+    if ( !project.remoteError.isEmpty() )
+    {
+      const auto res = std::find_if( mProjects.begin(), mProjects.end(), [&project]( const Project & localProject )
+      {
+        return project.fullName() == localProject.fullName();
+      } );
+      if ( res != mProjects.end() )
+      {
+        errorProjects.append( res->id() );
+      }
+    }
+  }
+  return errorProjects;
 }
 
 int ProjectsModel::projectIndexFromId( const QString &projectId ) const
--- a/app/projectsmodel.cpp
+++ b/app/projectsmodel.cpp
@@ -27,44 +27,44 @@ void ProjectsModel::initializeProjectsModel()
   if ( !mSyncManager || !mBackend || !mLocalProjectsManager || mModelType == EmptyProjectsModel ) // Model is not set up properly yet
     return;
 
-  QObject::connect( mSyncManager, &SynchronizationManager::syncStarted, this, &ProjectsModel::onProjectSyncStarted );
-  QObject::connect( mSyncManager, &SynchronizationManager::syncFinished, this, &ProjectsModel::onProjectSyncFinished );
-  QObject::connect( mSyncManager, &SynchronizationManager::syncCancelled, this, &ProjectsModel::onProjectSyncCancelled );
-  QObject::connect( mSyncManager, &SynchronizationManager::syncProgressChanged, this, &ProjectsModel::onProjectSyncProgressChanged );
+  connect( mSyncManager, &SynchronizationManager::syncStarted, this, &ProjectsModel::onProjectSyncStarted );
+  connect( mSyncManager, &SynchronizationManager::syncFinished, this, &ProjectsModel::onProjectSyncFinished );
+  connect( mSyncManager, &SynchronizationManager::syncCancelled, this, &ProjectsModel::onProjectSyncCancelled );
+  connect( mSyncManager, &SynchronizationManager::syncProgressChanged, this, &ProjectsModel::onProjectSyncProgressChanged );
 
-  QObject::connect( mBackend, &MerginApi::projectDetached, this, &ProjectsModel::onProjectDetachedFromMergin );
-  QObject::connect( mBackend, &MerginApi::projectAttachedToMergin, this, &ProjectsModel::onProjectAttachedToMergin );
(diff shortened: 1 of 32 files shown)


app/projectsmodel.cpp
```diff
@@ -549,21 +564,20 @@ void ProjectsModel::onProjectDetachedFromMergin( const QString &projectFullName
...
 }
 
-void ProjectsModel::onProjectAttachedToMergin( const QString & )
+void ProjectsModel::onProjectAttachedToMergin( const QString &projectId )
 {
-  // To ensure project will be in sync with server, send listProjectByName request.
-  // In theory we could send that request only for this one project.
-  listProjectsByName();
+  // To ensure project will be in sync with server, send fetchProjectsByProjectId request.
+  fetchProjectsByProjectId( { projectId } );
```
tomasMizera, line 580: If you fetch state only of this one project, won't other local projects states get unknown (not calculated)? When requesting this API (in `fetchProjectsByProjectId`) we clear the projects model and so the resulting state of the other projects is now _likely_ unknown (I did not test this, just assumption from the code).

app/projectsmodel.h
```diff
@@ -116,17 +120,21 @@ class ProjectsModel : public QAbstractListModel
...
+     * \param projectIds List of project IDs to fetch, by default empty
+     */
+    Q_INVOKABLE void fetchProjectsByProjectId( const QStringList &projectIds = QStringList() );
 
     //! Syncs specified project - upload or update
     Q_INVOKABLE void syncProject( const QString &projectId );
 
     //! Stops running project upload or update
-    Q_INVOKABLE void stopProjectSync( const QString &projectId );
+    Q_INVOKABLE void stopProjectSync( const QString &projectId ) const;
```
tomasMizera, line 134: This one should not be `const`, it alters things (maybe not directly, but it does)!

app/projectsmodel.h
```diff
@@ -116,17 +120,21 @@ class ProjectsModel : public QAbstractListModel
...
 
     //! Forwards call to LocalProjectsManager to remove local project
-    Q_INVOKABLE void removeLocalProject( const QString &projectId );
+    Q_INVOKABLE void removeLocalProject( const QString &projectId ) const;
```
tomasMizera, line 137: Same here, I believe this one should not be `const`.

tomasMizera (commented):
Managed to go through the `app` folder so far, will continue later!

app/projectsmodel.h
Withalion: I agree with that it changes some things. We should probably have some discussion about using `const` and in what cases is it intended to use.

app/projectsmodel.h
Withalion: Same thing as mentioned above.

app/projectsmodel.cpp
Withalion: Sounds plausible from the look of it. Will test it later again when the review is finished.

tomasMizera:
From our discussion earlier today:

For sync:
image

For listing downloaded projects:
image

core/merginapi.cpp
```diff
@@ -227,17 +228,16 @@ QString MerginApi::listProjectsByName( const QStringList &projectNames )
...
-    CoreUtils::log( "list projects by name", QStringLiteral( "Too many local projects: " ) + QString::number( projectNames.count(), 'f', 0 ) );
-    const int projectsToRemoveCount = projectNames.count() - listProjectsByNameApiLimit;
-    QString msg = tr( "Please remove some projects as the app currently\nonly allows up to %1 downloaded projects." ).arg( listProjectsByNameApiLimit );
+    CoreUtils::log( "list projects by name", QStringLiteral( "Too many local projects: " ) + QString::number( static_cast<int>( projectNames.count() ), 'f', 0 ) );
+    const QString msg = tr( "Please remove some projects as the app currently\nonly allows up to %1 downloaded projects." ).arg( maxProjectRequests );
     notifyInfo( msg );
-    projectNamesToRequest.erase( projectNamesToRequest.begin() + listProjectsByNameApiLimit, projectNamesToRequest.end() );
-    Q_ASSERT( projectNamesToRequest.count() == listProjectsByNameApiLimit );
+    projectNamesToRequest.erase( projectNamesToRequest.begin() + maxProjectRequests, projectNamesToRequest.end() );
+    Q_ASSERT( projectNamesToRequest.count() == maxProjectRequests );
```
tomasMizera, line 240: I guess
```suggestion
    Q_ASSERT( projectNamesToRequest.count() <= maxProjectRequests );
```

core/merginapi.cpp
```diff
@@ -2419,81 +2535,93 @@ void MerginApi::pullInfoReplyFinished()
...
-  TransactionStatus &transaction = mTransactionalStatus[projectFullName];
+  Q_ASSERT( mTransactionalStatus.contains( projectId ) );
+  TransactionStatus &transaction = mTransactionalStatus[projectId];
   Q_ASSERT( r == transaction.replyPullProjectInfo );
 
   if ( r->error() == QNetworkReply::NoError )
   {
-    QByteArray data = r->readAll();
+    const QByteArray data = r->readAll();
     CoreUtils::log( "pull " + projectFullName, QStringLiteral( "Downloaded project info." ) );
```
tomasMizera, line 2549: We need to compare if the server project ID is the same as the local project ID - in case remote project was renamed and some other project now bears the same name. We would try to sync a completely different project

core/merginapi.cpp
```diff
@@ -2817,7 +2949,7 @@ void MerginApi::pushInfoReplyFinished()
     transaction.replyPushProjectInfo->deleteLater();
```
tomasMizera, line 2949: We need to compare if the server project ID is the same as the local project ID - in case remote project was renamed and some other project now bears the same name. We would try to sync a completely different project

core/merginapi.h
```diff
@@ -253,62 +247,70 @@ class MerginApi: public QObject
...
      * Emits syncProjectFinished at the end.
-     * \param projectNamespace Project's namespace used in request.
-     * \param projectName Project's name used in request.
+     * \param projectFullName Project's "namespace/name" full name used for requests
+     * \param projectId Project's ID used for internal processing
      * \param isInitialPush indicates if this is first push of the project (project creation)
      * \return true when sync has started, false otherwise (e.g. due to a missing authorization or invalid server)
      */
-    Q_INVOKABLE bool pushProject( const QString &projectNamespace, const QString &projectName, bool isInitialPush = false );
+    bool pushProject( const QString &projectFullName, const QString &projectId, bool isInitialPush = false );
```
tomasMizera, line 297: This and other API definitions *should* not need to accept both `projectName` and `projectId`. The ID should be enough - we can lookup the name via `mLocalProjectsManager`

tomasMizera (changes requested):

Withalion:
We'll tackle this as part of sync v2 implementation

PR #3887, opened 2025-05-09
Hotfix - version for unsupported SDKs (2025.2.3)

tomasMizera:
We need to do a hot-fix release after Qt upgrade.
What happened:
 - We upgraded Qt library without upgrading the minimal supported SDK versions, both Android and iOS
 - Previous Qt version was `6.6.3`, its minimal supported SDKs were: 
   - Android - API lvl 26 (Android 8)
   - iOS - target platform 14
 - The new Qt version is 6.8.3, with the following minimal supported SDKs:
   - Android - API lvl 28 (Android 9)
   - iOS - target platform 16
 - The result is that some devices on unsupported system versions get undefined behavior and crashes.

What we plan to do:
 - We do a release with Qt `6.6.3` (`2025.2.3`) - includes code from release `2025.1.1` and some required cherry-picked fixes
 - Afterwards, we do another release with Qt `6.8.3` (with code from `master`) - but this time with the upgraded minimal supported system versions (`2025.2.4`)
 - Unsupported devices will be able to upgrade to `2025.2.3` and this will be their last app version
 - Other, supported, devices should update to `2025.2.4` and continue normally

**This PR is created just for reference and will not be merged.**


tomasMizera:
Released now. Keeping the branch for reference.

PR #3895, opened 2025-05-12
Switch workspace on invitation accepted

### LINKED ISSUE #3847: Open the list of workspace projects after accepting invitation
On mobile, after accepting a workspace invitation, it would be helpful to automatically redirect users to the list of projects in that workspace. Not everyone realizes they need to tap the "Projects" tab to view them. This small change could make it easier for surveyors to quickly find the right project, download it, and get to work.

Suggestions:

- Automatically switch to the invited workspace (if not already happening)
- Open the Projects page of that workspace
- Consider adding a short guiding message like: "Download a project and start collecting."

This small UX tweak could make onboarding smoother for new team members, especially for guests.

Note: this should work in scenario when you accept invitation after registration (during onboarding) - not when you accept an invitation as already existing member

VitorVieiraZ:
- Added `workspaceId` to `InvitationsModel` after `/v1/workspace/invitations` endpoint began including this field with each invitation. After accepting a workspace invitation, app now switches to accepted workspace and navigates to its project list if server version is bigger than `2025.4.1` ( see https://github.com/MerginMaps/server-private/issues/2936 ). 
- Added a new method `serverVersionIsAtLeast` to check whether server version meets or exceeds a required minimum version.
- Changed icon of `addIcon` method in `NotificationModel` from `WaitingIcon` to `InfoIcon`.

Resolves #3847

### DIFF
--- a/app/qml/account/MMAccountController.qml
+++ b/app/qml/account/MMAccountController.qml
@@ -20,6 +20,8 @@ Item {
   property bool inProgress: false
   property MM.MerginInvitation invitation
 
+  signal projectsListRequested()
+
   QtObject {
     //! Data to send to postRegister endpoint
     id: postRegisterData
@@ -303,8 +305,7 @@ Item {
 
         function onProcessInvitationFinished( accepted ) {
           controller.end()
-          stackView.currentItem.state = "workspace"
-          __notificationModel.addInfo( qsTr( "Download a project and start collecting." ), MM.NotificationType.NoAction, 10 )
+          controller.projectsListRequested()
         }
       }
     }
--- a/app/qml/account/MMAccountController.qml
+++ b/app/qml/account/MMAccountController.qml
@@ -303,6 +303,8 @@ Item {
 
         function onProcessInvitationFinished( accepted ) {
           controller.end()
+          stackView.currentItem.state = "workspace"
+          __notificationModel.addInfo( qsTr( "Download a project and start collecting." ), MM.NotificationType.NoAction, 10 )
         }
       }
     }
--- a/core/merginapi.cpp
+++ b/core/merginapi.cpp
@@ -3837,7 +3837,8 @@ void MerginApi::getServerConfigReplyFinished()
       if ( !validVersion )
       {
         CoreUtils::log( QStringLiteral( "Server version" ), QStringLiteral( "Cannot parse server version" ) );
-      } else
+      }
+      else
       {
         setApiVersion( apiVersion );
       }
@@ -4091,7 +4092,7 @@ void MerginApi::processInvitationReplyFinished()
   {
     CoreUtils::log( "process invitation", QStringLiteral( "Success" ) );
 
-    // if server version is at least 2025.4.1, let‘s get workspaceId from response and switch to it
+    // if server version is at least 2025.4.1, let's get workspaceId from response and switch to it
     // emit processInvitationSuccess to navigate to project's page in qml
     if ( serverVersionIsAtLeast( 2025, 4, 1 ) && accept )
     {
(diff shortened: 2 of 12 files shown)


VitorVieiraZ:
We should wait for [server release](https://github.com/MerginMaps/server-private/milestone/75) before testing and merging this PR.

tomasMizera (changes requested):
According to what we agreed on the call :)

core/merginapi.cpp
```diff
@@ -3631,6 +3631,8 @@ void MerginApi::getServerConfigReplyFinished()
     {
       QString serverType = doc.object().value( QStringLiteral( "server_type" ) ).toString();
       QString apiVersion = doc.object().value( QStringLiteral( "version" ) ).toString();
+      setApiVersion( apiVersion );
+
       int major = -1;
       int minor = -1;
       bool validVersion = parseVersion( apiVersion, major, minor );
```
Withalion: we should first check if the version is valid and then assign it

core/merginapi.cpp
```diff
@@ -3862,6 +3864,25 @@ void MerginApi::processInvitationReplyFinished()
...
+    {
+      QByteArray data = r->readAll();
+      QJsonDocument doc = QJsonDocument::fromJson( data );
+
+      if ( doc.isObject() )
+      {
+        QJsonObject responseObj = doc.object();
+        if ( responseObj.contains( "workspace_id" ) )
+        {
+          int workspaceId = responseObj.value( "workspace_id" ).toInt();
```
```suggestion
      const QByteArray data = r->readAll();
      const QJsonDocument doc = QJsonDocument::fromJson( data );

      if ( doc.isObject() )
      {
        const QJsonObject responseObj = doc.object();
        if ( responseObj.contains( "workspace_id" ) )
        {
          const int workspaceId = responseObj.value( "workspace_id" ).toInt();
```

core/merginapi.cpp
```diff
@@ -4073,3 +4094,51 @@ void MerginApi::setNetworkManager( QNetworkAccessManager *manager )
...
+void MerginApi::setApiVersion( const QString &apiVersion )
+{
+  if ( mApiVersion != apiVersion )
+  {
+    mApiVersion = apiVersion;
+    emit apiVersionChanged( mApiVersion );
+  }
+}
+
+bool MerginApi::serverVersionIsAtLeast( int requiredMajor, int requiredMinor, int requiredPatch ) const
```
```suggestion
bool MerginApi::serverVersionIsAtLeast( const int requiredMajor, const int requiredMinor, const int requiredPatch ) const
```

core/merginapi.cpp
```diff
@@ -4073,3 +4094,51 @@ void MerginApi::setNetworkManager( QNetworkAccessManager *manager )
...
+{
+  int serverMajor = -1;
+  int serverMinor = -1;
+  bool validVersion = parseVersion( mApiVersion, serverMajor, serverMinor );
```
```suggestion
  const bool validVersion = parseVersion( mApiVersion, serverMajor, serverMinor );
```

Withalion:
When accepting the invitation, the view gets correctly moved to a correct screen. However I get offered public projects instead of the projects from that workspace and after a while the list disappears and notification about wrong URL is shown. Clicking retry fetches workspace projects.

https://github.com/user-attachments/assets/a6d50e81-0c02-4d36-b01f-6d2d2b7cb5c8

core/merginapi.h
```diff
@@ -420,6 +420,16 @@ class MerginApi: public QObject
...
 
+    /**
+    * Parse major, minor and patch version number from version string.
+    * \param version full server version string
+    * \param major parsed major number
+    * \param minor parsed minor number
+    * \param patch parsed patch number
+    * @return true when parsing was successful
+    */
+    static bool parseVersion( const QString &version, int &major, int &minor, int &patch );
```
tomasMizera, line 432: Methods like these would better fit in utils.. let's tackle that during the upcoming refactoring

tomasMizera:
As agreed on the call, let's use the logic _only_ when accepting an invitation during the onboarding flow.

app/qml/account/MMAccountController.qml
```diff
@@ -303,6 +303,8 @@ Item {
 
         function onProcessInvitationFinished( accepted ) {
           controller.end()
+          stackView.currentItem.state = "workspace"
+          __notificationModel.addInfo( qsTr( "Download a project and start collecting." ), MM.NotificationType.NoAction, 10 )
```
tomasMizera: This works, but it is quite fragile:
 - it expects that the `stackView.currentItem` is `workspaceProjectsPanelComp` (initial item), however, the stackview might have multiple other types of pages pushed to it
 - it manipulates an item pushed to the stackView, completely outside of the scope where you would expect it

I'd suggest to replace this with something like:
```qml
root.projectListRequested()
```

and then in the `MMProjectsController`:

```qml
MMAccountController {
  ...
  onProjectListRequested: now.do.the.magic
}
```

That moves it a bit closer to the projects scope :)

tomasMizera: Once this is done, I am +1 to merge the PR

Withalion: I'm not 100% sure I did what you had in mind. `controller.end()` will clear the stackView to contain only the initial item, so we can be sure about that. I moved the other logic into `MMProjectController`

tomasMizera: This is what I meant :)

PR #3897, opened 2025-05-13
Fix testHasLocalProjectChanges 

VitorVieiraZ:
In fourth scenario, add a new file instead of modifying existing one.


tomasMizera (approved):
Let's see what it does :)

PR #3902, opened 2025-05-13
SSO implementation

uclaros:
image

image


tomasMizera (commented):
This looks great! 😍 

I was wondering if there is a way to add unit tests for this. We would probably need to get some headless IDP set up on server ... 🤔

tomasMizera:
I believe we could start playing around with this @harminius! :)

app/qml/account/MMSsoPage.qml
```diff
@@ -0,0 +1,107 @@
...
+
+      width: parent.width
+
+      spacing: __style.spacing20
+
+      MMListSpacer {
+        height: __style.margin20
+      }
+
+      MMTextInput {
```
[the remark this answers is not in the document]
uclaros: If we disable it, and the user closes the browser window without authenticating, then he won't have a chance to click sign in and retry; he will need to click back or _sign in with password_, and then _log in using sso_ again.

.github/workflows/i18n.yml
```diff
@@ -21,7 +21,7 @@ jobs:
         with:
           version: ${{ env.QT_VERSION }}
           setup-python: 'false'
-          modules: 'qtsensors qtconnectivity qt5compat qtmultimedia qtpositioning qtshadertools'
+          modules: 'qtsensors qtconnectivity qt5compat qtmultimedia qtpositioning qtshadertools qtnetworkauth'
```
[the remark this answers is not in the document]
uclaros: Makes sense, I'll try it out.

core/merginapi.cpp
```diff
@@ -4042,6 +4201,90 @@ QString MerginApi::getCachedProjectRole( const QString &projectFullName ) const
   return cachedProjectMetadata.role;
 }
 
+void MerginApi::startSsoFlow( const QString &clientId )
+{
+#ifdef MOBILE_OS
+  const QString CALLBACK_URL = QStringLiteral( "https://hello.merginmaps.com/mobile/sso-redirect" );
+#else
+  constexpr int OAUTH2_LISTEN_PORT = 8082;
```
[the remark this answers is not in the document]
uclaros: Yes, I didn't like the 8082 port much either. I think it's often used.
There's also an option to _not_ restrict to a specific port, then QT picks one. But we will need to loosen the allowed callback url in boxy to not include a port, and I'm not sure if that's the best security wise.

core/merginapi.cpp
```diff
@@ -4042,6 +4201,90 @@ QString MerginApi::getCachedProjectRole( const QString &projectFullName ) const
...
+  mOauth2Flow.setAccessTokenUrl( QUrl( mApiRoot + QStringLiteral( "/v2/sso/token" ) ) );
+  mOauth2Flow.setClientIdentifier( clientId );
+  mOauth2Flow.setScope( QString() );
+
+  if ( !mOauth2ReplyHandler )
+  {
+#ifdef MOBILE_OS
+    mOauth2ReplyHandler = new QOAuthUriSchemeReplyHandler( CALLBACK_URL, &mOauth2Flow );
+#else
+    mOauth2ReplyHandler = new QOAuthHttpServerReplyHandler( QHostAddress::Null, OAUTH2_LISTEN_PORT, &mOauth2Flow );
```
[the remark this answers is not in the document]
uclaros: `LocalHost` restricts to ipv4. With `Null` ipv4 is preferred but if failed ipv6 is also tried. I'll add a comment with link to docs

core/merginapi.cpp
```diff
@@ -4042,6 +4201,90 @@ QString MerginApi::getCachedProjectRole( const QString &projectFullName ) const
...
+    mOauth2ReplyHandler->setCallbackText( msg );
+#endif
+
+    mOauth2Flow.setReplyHandler( mOauth2ReplyHandler );
+
+    connect( &mOauth2Flow, &QAbstractOAuth::authorizeWithBrowser, this, [this]( const QUrl & url )
+    {
+      CoreUtils::log( "SSO", QStringLiteral( "Opening browser to autorize: %1" ).arg( url.toString() ) );
+      QDesktopServices::openUrl( url );
+      emit ssoAuthorizeUsingBrowser();
```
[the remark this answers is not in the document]
uclaros: I did this intentionally. I consider the "Follow instructions on the web browser" message as a safety net for users that switch back to the app without logging in, so in an ideal situation the user would not see it at all.
However, it doesn't make much difference as it takes some time to open the browser and the message is shown during that time.

core/merginapi.cpp
```diff
@@ -3338,18 +3486,26 @@ MerginProjectsList MerginApi::parseProjectsFromJson( const QJsonDocument &doc )
...
-  if ( !mUserAuth->hasAuthData() ||
-       mUserAuth->authToken().isEmpty() )
+  switch ( mUserAuth->authMethod() )
   {
-    CoreUtils::log( QStringLiteral( "Auth" ), QStringLiteral( "Can not refresh token, missing credentials" ) );
-    return;
-  }
+    case MerginUserAuth::AuthMethod::SSO:
+      // refresh tokens not implemented yet
+      // TODO: what to do when asking for refresh?
```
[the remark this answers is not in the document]
uclaros: This method should not trigger user interaction, it will be called when refresh tokens are implemented just to transparently get a fresh token, same as what's happening with the password case.

If the session eventually expires and a new sso login is required, this is handled in `bool MerginApi::validateAuth()` where `authRequested()` is emitted. This triggers the login page in qml and eventually the sso page will get loaded with the email parameter filled if the stored userAuth is SSO. So we don't really need a different signal for `ssoSessionExpired`

.github/workflows/i18n.yml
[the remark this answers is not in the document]
uclaros: On second thought, I think we need the same modules across workflows to reuse the cache, no?

harminius:
My comments:
1. I had a concern that the login screen layout suggests you should create a new account when you don't have one by displaying the text "Don't you have an account?" and the button bellow it. But it is probably fine - I don't have a better idea and I hope the users should be aware of the possibility of using SSO when their company buys it.

image

2. Email format validation
2a. When signing with SSO we care only about the domain part. However, the App somehow (in a custom way?) validates email address format. When there is plus sign, it return 400. On server there is a validation from a python lib and it let's `+` go through. I don't see a reason why to block it.

image


2b. The `Sign in` button on Sign in with SSO page enables and disables not logically during typing. It works when I close the keyboard, though.

https://github.com/user-attachments/assets/e93a126b-a5cb-4e23-a40a-83743ae144ac

3. Not sure if this is SSO issue or not but I experienced it (only) here. The app displays your fullname if you have, username in parentheses and email. Sometimes it is hard to understand what is what, but you don't need them much (mainly to close your account). However, with SSO sign up workflow, it seems to me the username is wrong - it displays the email instead. 

Screenshot_20250515_143927

harminius:
We should think of real workflows and make the UX as smooth as possible. Currently signing through SSO while having an invitation is not very user friendly. The user doesn't want their own workspace. They just need to easily accept the invitation.

https://github.com/user-attachments/assets/e9b65b6a-602a-4c4e-b0c5-3e7d30667034

uclaros:
Note to self: Check opening the app with expired token and password auth. I think something is not handled well.

app/qml/account/MMSsoPage.qml
```diff
@@ -0,0 +1,107 @@
...
+        textField.inputMethodHints: Qt.ImhNoAutoUppercase | Qt.ImhEmailCharactersOnly
+        text: __merginApi.userAuth.username
+      }
+
+      MMButton {
+        width: parent.width
+
+        text: qsTr( "Sign in" )
+
+        enabled: email.text.match( "\\S+@\\S+\\.\\S+" )
```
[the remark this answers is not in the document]
Withalion: I created a new function in `CoreUtils` as we use something similar in MerginApi and `InputUtils` expose it to QML. Also some unit tests were added...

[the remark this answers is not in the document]
uclaros: This is outdated. I have actually removed the check since it can be confusing with the mobile keyboards (text appears on the screen but is not yet sent to the widget)
I've blundered it though and wrote:
```
enabled: __inputUtils.isValidEmail( email.length > 0 )
```
instead of:
```
enabled: email.length > 0
```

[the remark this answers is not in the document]
Withalion: Looks like I was able to have the check and button is responsive even on mobile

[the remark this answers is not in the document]
uclaros: That would be great, however I think the behavior can be device/keyboard app dependent. Better check with @tomasMizera

app/qml/account/MMAccountController.qml
```diff
@@ -186,6 +197,55 @@ Item {
...
+      }
+
+      onSignInClicked: function( email ) {
+        stackView.pending = true
+        __merginApi.requestSsoConnections(email)
+      }
+
+      Connections {
+        target: __merginApi
+        enabled: stackView.currentItem.objectName === "ssoPanel"
```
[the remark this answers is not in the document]
Withalion: As we are wrapping multiple pages in this component I would keep it like it is. We also do the same thing for other pages

app/variablesmanager.cpp
```diff
@@ -104,14 +102,22 @@ void VariablesManager::apiRootChanged()
...
 void VariablesManager::setUserVariables()
 {
+  // we re-set the username variables with the data in userInfo when it changes
+  // When doing SSO, we do not have that info yet when userAuth is changed
+  QgsExpressionContextUtils::setGlobalVariable( QStringLiteral( "mergin_username" ),  mMerginApi->userInfo()->username() );
+  QgsExpressionContextUtils::setGlobalVariable( QStringLiteral( "mm_username" ),  mMerginApi->userInfo()->username() );
+
   QgsExpressionContextUtils::setGlobalVariable( QStringLiteral( "mergin_user_email" ),  mMerginApi->userInfo()->email() );
   QgsExpressionContextUtils::setGlobalVariable( QStringLiteral( "mergin_full_name" ),  mMerginApi->userInfo()->name() );
+  QgsExpressionContextUtils::setGlobalVariable( QStringLiteral( "mm_username" ),  mMerginApi->userAuth()->username() );
```
Withalion: This is probably forgotten from rebase (note for me)

core/merginapi.cpp
```diff
@@ -814,12 +828,133 @@ void MerginApi::authorize( const QString &login, const QString &password )
...
+void MerginApi::requestSsoConfig()
+{
+  if ( !mApiSupportsSso )
+  {
+    CoreUtils::log( QStringLiteral( "SSO Auth" ), QStringLiteral( "Requested sso auth for server that does not support sso!" ) );
+    return;
+  }
+
+  QNetworkRequest request = getDefaultRequest( false );
+  QUrl url( mApiRoot + QStringLiteral( "/v2/sso/config" ) );
```
Withalion: const

core/merginapi.cpp
```diff
@@ -814,12 +828,133 @@ void MerginApi::authorize( const QString &login, const QString &password )
...
+  request.setUrl( url );
+
+  QNetworkReply *reply = mManager->get( request );
```
Withalion: const

core/merginapi.cpp
```diff
@@ -814,12 +828,133 @@ void MerginApi::authorize( const QString &login, const QString &password )
...
+
+void MerginApi::ssoConfigReplyFinished()
+{
+  QNetworkReply *r = qobject_cast<QNetworkReply *>( sender() );
+  Q_ASSERT( r );
+
+  if ( r->error() == QNetworkReply::NoError )
+  {
+
+    QJsonDocument doc = QJsonDocument::fromJson( r->readAll() );
```
Withalion: const

core/merginapi.cpp
```diff
@@ -814,12 +828,133 @@ void MerginApi::authorize( const QString &login, const QString &password )
...
+  }
+
+  QNetworkRequest request = getDefaultRequest( false );
+  QUrl url( mApiRoot + QStringLiteral( "/v2/sso/connections" ) );
+  QUrlQuery query;
+  query.addQueryItem( QStringLiteral( "email" ), email.toUtf8().toPercentEncoding() );
+  url.setQuery( query );
+  request.setUrl( url );
+
+  QNetworkReply *reply = mManager->get( request );
```
Withalion: const

core/merginapi.cpp
```diff
@@ -814,12 +828,133 @@ void MerginApi::authorize( const QString &login, const QString &password )
...
+
+void MerginApi::ssoConnectionsReplyFinished()
+{
+  QNetworkReply *r = qobject_cast<QNetworkReply *>( sender() );
+  Q_ASSERT( r );
+
+  if ( r->error() == QNetworkReply::NoError )
+  {
+
+    QJsonDocument doc = QJsonDocument::fromJson( r->readAll() );
```
Withalion: const

core/merginapi.cpp
```diff
@@ -814,12 +828,133 @@ void MerginApi::authorize( const QString &login, const QString &password )
...
+        startSsoFlow( clientId );
+      }
+    }
+  }
+  else
+  {
+    const QString serverMsg = extractServerErrorMsg( r->readAll() );
+    CoreUtils::log( "SSO", QStringLiteral( "FAILED - %1. %2" ).arg( r->errorString(), serverMsg ) );
+    const QVariant statusCode = r->attribute( QNetworkRequest::HttpStatusCodeAttribute );
+    const int status = statusCode.toInt();
```
Withalion: could be just one line of code

core/merginapi.cpp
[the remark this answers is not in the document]
Withalion: If I understand it correctly should we just do a background request to the SSO backend for a new token or we just keep it empty (don't support this feature yet) or we force the normal sso authentication (not recommended, but temporary solution)?

core/merginapi.cpp
```diff
@@ -4050,6 +4215,84 @@ QString MerginApi::getCachedProjectRole( const QString &projectFullName ) const
...
+  mOauth2Flow.setClientIdentifier( clientId );
+  mOauth2Flow.setScope( QString() );
+
+  if ( !mOauth2ReplyHandler )
+  {
+#ifdef MOBILE_OS
+    mOauth2ReplyHandler = new QOAuthUriSchemeReplyHandler( CALLBACK_URL, &mOauth2Flow );
+#else
+    mOauth2ReplyHandler = new QOAuthHttpServerReplyHandler( OAUTH2_LISTEN_ADDRESS, OAUTH2_LISTEN_PORT, &mOauth2Flow );
+    const QString msg = tr( "You can now close this page and return to Mergin Maps" );
```
Withalion, line 4231: Any option to also set the text on browser tab? or is that app name dependent?

core/merginapi.cpp
```diff
@@ -4050,6 +4215,84 @@ QString MerginApi::getCachedProjectRole( const QString &projectFullName ) const
...
+    mOauth2ReplyHandler->setCallbackText( msg );
+#endif
+
+    mOauth2Flow.setReplyHandler( mOauth2ReplyHandler );
+
+    connect( &mOauth2Flow, &QAbstractOAuth::authorizeWithBrowser, this, [this]( const QUrl & url )
+    {
+      CoreUtils::log( "SSO", QStringLiteral( "Opening browser to autorize: %1" ).arg( url.toString() ) );
```
Withalion: typo

core/merginapi.cpp
```diff
@@ -4050,6 +4215,84 @@ QString MerginApi::getCachedProjectRole( const QString &projectFullName ) const
...
+      QDesktopServices::openUrl( url );
+    } );
+
+    connect( &mOauth2Flow, &QAbstractOAuth::granted, this, [this]()
```
Withalion: the round brackets are redundant

core/merginapi.cpp
```diff
@@ -4081,3 +4324,17 @@ void MerginApi::setNetworkManager( QNetworkAccessManager *manager )
 
   emit networkManagerChanged();
 }
+
+bool MerginApi::apiSupportsSso() const
+{
+  return mApiSupportsSso;
+}
+
+void MerginApi::setApiSupportsSso( bool ssoSupported )
```
Withalion: const parameter

core/merginuserauth.cpp
```diff
@@ -77,13 +104,15 @@ void MerginUserAuth::loadAuthData()
...
-             const QDateTime & tokenExpiration )
+             const QDateTime & tokenExpiration,
+             int method )
   {
     mUsername = username;
     mPassword = password;
     mUserId = userId;
     mAuthToken = token.toUtf8();
     mTokenExpiration = tokenExpiration;
+    mMethod = AuthMethod( method );
```
Withalion: let's use c++ casting instead

Withalion (commented):
Majority of the comments, are notes for me. There is still some ongoing discussion, which we should resolve before merging. Leaving any further reviews to @tomasMizera

MarcelGeo:
Hey guys. I was wondering If we are able to send some additional parameters to /authorize request:

- `login_hint` -> this says thath, Hey google prefill email with login_hint (it works for OIDC connections in boxy). I think majority of customers will have SAML connection where `login_hint` does not have effect, but it could be here.

If we are not sending such additional params, It should be good to try it.

core/merginapi.cpp
```diff
@@ -4050,6 +4214,84 @@ QString MerginApi::getCachedProjectRole( const QString &projectFullName ) const
   return cachedProjectMetadata.role;
 }
 
+void MerginApi::startSsoFlow( const QString &clientId )
+{
+  CoreUtils::log( "SSO", QStringLiteral( "Starting SSO flow for clientId: %1" ).arg( clientId ) );
+  mOauth2Flow.setAuthorizationUrl( QUrl( mApiRoot + QStringLiteral( "/v2/sso/authorize" ) ) );
+  mOauth2Flow.setAccessTokenUrl( QUrl( mApiRoot + QStringLiteral( "/v2/sso/token" ) ) );
```
Withalion, line 4221: this method is scheduled for deprecation, as this is a new code we should not disregard this. It's recommended to use ` QAbstractOAuth2::setTokenUrl()`

EDIT: I see it was added in Qt 6.9, okay at least put todo here

core/merginapi.cpp
uclaros: Haven't found something in qt docs.
Maybe it is worth replicating the page that qgis generates for the plugin?
I think it includes a _close_ link too.

app/qml/components/MMDrawerDialog.qml
```diff
@@ -54,6 +54,7 @@ MMDrawer {
         id: additionalContentGroup
 
         width: parent.width
+        // TODO: fix this binding loop
```
uclaros, line 57: Is this still valid?

Withalion: should be, I wanted to leave this here as it's one of the base components which will need to be tested on multiple places

So let's leave this fix for another PR :)

core/coreutils.cpp
```diff
@@ -380,5 +380,6 @@ bool CoreUtils::replaceValueInJson( const QString &filePath, const QString &key,
 
 bool CoreUtils::isValidEmail( const QString &email )
 {
-  return !email.isEmpty() && email.contains( '@' ) && email.contains( '.' );
+  const thread_local QRegularExpression regEx( "\\S+@\\S+\\.\\S+" );
```
Withalion, line 383: doesn't this regex fail on `+` in name part?

uclaros: It shouldn't, it validates there's at least one _non-whitespace_ character on each side of the `@` and `.`
Added a test case for this

tomasMizera:
Cmake things will be fixed in different PR

PR #3912, opened 2025-05-21
Throw exception on transform error

VitorVieiraZ:
This PR adds an exception when `screenUnitsToMeters` method fails in order to prevent app crashes.

### DIFF
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -937,7 +937,8 @@ double InputUtils::screenUnitsToMeters( InputMapSettings *mapSettings, int baseL
   }
   catch ( QgsCsException &e )
   {
-    Q_UNUSED( e )
+    Q_UNUSED( e );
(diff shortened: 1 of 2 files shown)


VitorVieiraZ:
> What happens with the app if you return `0.0` here?

In this scenario, we either get https://github.com/MerginMaps/mobile/blob/9e0249300edd4b2c46dd82d12c5870647aab564f/app/position/mapposition.cpp#L146 a default of `2.0`, which might lead to a misrepresentation of the actual GPS accuracy, and the scale bar will show value 0. Added a check to return early if value <= 0.0 to prevent unpredictable mathematical operations and possible crashes: https://github.com/MerginMaps/mobile/blob/9e0249300edd4b2c46dd82d12c5870647aab564f/app/scalebarkit.cpp#L85

tomasMizera (approved):

harminius (approved):
Fix works for https://github.com/MerginMaps/mobile/issues/3914 :green_circle:

tomasMizera:
One last question before merging, doesn't this generate a lot of lines in the diagnostic log @VitorVieiraZ?

PR #3919, opened 2025-06-02
Prevent excessive zoom by growing default map extent

VitorVieiraZ:
To prevent excessively zoomed-in initial views, especially when opening local projects with sparse data (such as a single point), the map now automatically zooms out in these scenarios.

Relates to #2445


tomasMizera:
Let's tackle this one within the other PRs .. sorry!

PR #3930, opened 2025-06-13
Photo sketching support

Withalion:
This PR adds photo sketching support to the mobile app!

Closes https://github.com/MerginMaps/mobile/issues/1374
Connected to [plugin PR](https://github.com/MerginMaps/qgis-plugin/pull/710)

The sketching is available while editing a feature with photo already attached:
image

A new page will open with the photo. 7 colors are available for sketching. If you make any mistakes, there is undo and also close button for throwing away all changes. If the sketches look good "save changes" button will save them to the original image.
image

Demo:

https://github.com/user-attachments/assets/3cca843c-5d80-43ef-b252-16b1b0dafdd7

Photo sketching can be disabled in project in QGIS plugin.


app/CMakeLists.txt
```diff
@@ -88,6 +88,7 @@ set(MM_SRCS
     variablesmanager.cpp
     workspacesmodel.cpp
     workspacesproxymodel.cpp
+    photosketchingcontroller.cpp
```
tomasMizera: Best to keep things sorted alphabetically

app/activeproject.cpp
```diff
@@ -621,3 +621,13 @@ QString ActiveProject::mapAnnotationsLayerId() const
 
   return mQgsProject->readEntry( QStringLiteral( "Mergin" ), QStringLiteral( "MapAnnotations/Layer" ), QString() );
 }
+
+bool ActiveProject::photoSketchesEnabled() const
```
tomasMizera: This is really just nit picking, but the feature is called `Photo sketching`, so we would use `PhotoSketchingEnabled` and so... Photo sketches is more about the actual drawn sketches. Like with the map sketching, the layer is called "Map sketches", but the feature is "Map sketching"

app/mmstyle.h
```diff
@@ -465,6 +466,7 @@ class MMStyle: public QObject
     QUrl xTwitterIcon() const {return QUrl( "qrc:/XTwitter.svg" );}
     QUrl youtubeIcon() const {return QUrl( "qrc:/Youtube.svg" );}
     QUrl formIcon() const {return QUrl( "qrc:/Terms.svg" );}
+    QUrl drawIcon() const {return QUrl( "qrc:/MapThemes.svg" );}
```
tomasMizera: Can you duplicate the file so the icon for photo sketching won't change in case we change the map themes icon? :)

app/qml/form/components/MMFormPhotoSketchingPageDialog.qml
```diff
@@ -0,0 +1,300 @@
...
+            shape.shapePaths[shapePathsIndex].destroy()
+            shape.shapePaths.splice( shapePathsIndex, 1 )
+          }
+        }
+      }
+    }
+
+    MMComponents.MMListSpacer { height: __style.margin20 }
+
+    RowLayout {
```
tomasMizera: What if the screen is not wide enough to fit all the colors?

app/photosketchingcontroller.cpp
```diff
@@ -0,0 +1,229 @@
...
+    mCanUndo = true;
+    emit canUndoChanged();
+  }
+  else if ( mPaths.size() <= 1 && mCanUndo )
+  {
+    mCanUndo = false;
+    emit canUndoChanged();
+  }
+}
+
```
tomasMizera: I find this shuffling with `currentLine` and path sizes a bit too complicated for what it does. Happy to have a chat, would be nice to simplify things here

app/photosketchingcontroller.cpp
```diff
@@ -0,0 +1,229 @@
...
+  for ( int i = 0; i < pathCount; ++i )
+  {
+    QColor color = mPaths.at( i ).mColor;
+    QVector<QPointF> points = mPaths.at( i ).mPoints;
+
+    if ( points.isEmpty() )
+      continue;
+
+    // if you are adjusting width here don't forget to adjust it also in MMFormPhotoDrawingPageDialog shapePathComponent
+    QPen pen( color, 4 * mPhotoScaleRatio );
```
tomasMizera: I can not find a constructor for `QPen` that would accept `QColor` and `width`. Is this one implicitly creating `QBrush`? https://doc.qt.io/qt-6.8/qpen.html#QPen-5

app/qml/form/components/MMFormPhotoSketchingPageDialog.qml
```diff
@@ -0,0 +1,301 @@
...
+      bgndColor: __style.polarColor
+
+      onClicked: {
+        root.close()
+        controller.clear()
+      }
+    }
+  }
+
+  footer: RowLayout {
```
tomasMizera: You shouldn't need `RowLayout` for just one child

app/qml/form/components/MMFormPhotoSketchingPageDialog.qml
```diff
@@ -0,0 +1,301 @@
...
+      id: container
+      Layout.fillWidth: true
+      Layout.fillHeight: true
+      Layout.leftMargin: __style.pageMargins
+      Layout.rightMargin: __style.pageMargins
+
+      Shape {
+        id: shape
+        anchors.centerIn: parent
+        anchors.fill: parent
```
tomasMizera: These two are kind-of mutually exclusive

app/qml/form/components/MMFormPhotoSketchingPageDialog.qml
```diff
@@ -0,0 +1,301 @@
...
+        }
+
+        MMComponents.MMPhoto {
+          id: photo
+
+          anchors.fill: parent
+          z: -1
+
+          photoUrl: root.photoUrl
+          isLocalFile: root.photoUrl.startsWith("file://")
```
tomasMizera: This is actually a good point, we should not allow to sketch on network images, only if the image is local

app/qml/form/editors/MMFormPhotoEditor.qml
```diff
@@ -137,6 +142,12 @@ MMFormPhotoViewer {
     }
   }
 
+  PhotoSketchingController {
+    id: sketchesController
+
+    photoSource: root.photoUrl
+  }
```
tomasMizera: This object could be instantiated via `Loader` only on projects that support the sketching

tomasMizera (dismissed):
🚀 The feature looks cool

app/photosketchingcontroller.cpp
Withalion: Sure I'm open to improvements. It might become quite a Frankenstein along the way as I was fixing other issues...

app/qml/form/components/MMFormPhotoSketchingPageDialog.qml
Withalion: Currently they get squeezed, having a scrollview might not be a bad idea

app/photosketchingcontroller.cpp
Withalion: yes the color is getting constructed to `QBrush` and then the whole thing to `QPen`

app/qml/form/components/MMFormPhotoSketchingPageDialog.qml
Withalion: That's true however I was unable to replicate the padding just with anchors

app/qml/form/editors/MMFormPhotoViewer.qml
```diff
@@ -9,9 +9,13 @@
 
 import QtQuick
 
+import mm 1.0 as MM
+import MMInput
+
 import "../../components" as MMComponents
 import "../../components/private" as MMPrivateComponents
 import "../components/photo" as MMPhotoComponents
+import "../components"as MMFormComponents
```
uclaros:
```suggestion
import "../components" as MMFormComponents
```

app/CMakeLists.txt
```diff
@@ -599,5 +601,5 @@ qt_add_qml_module(
   URI MMInput
   VERSION 1.0
   QML_FILES ${MM_QML_SRCS}
-  RESOURCE_PREFIX /com.merginmaps/imports NO_GENERATE_QMLTYPES
```
uclaros: _y u no generate qmltypes?_

app/activeproject.cpp
```diff
@@ -621,3 +621,13 @@ QString ActiveProject::mapSketchesLayerId() const
...
 }
+
+bool ActiveProject::photoSketchingEnabled() const
+{
+  if ( !isProjectLoaded() )
+  {
+    return false;
+  }
+
+  return mQgsProject->readBoolEntry( QStringLiteral( "Mergin" ), QStringLiteral( "PhotoSketches/Enabled" ), false );
```
uclaros: Since you already did the rename, let's be consistent and change the setting too:
```suggestion
  return mQgsProject->readBoolEntry( QStringLiteral( "Mergin" ), QStringLiteral( "PhotoSketching/Enabled" ), false );
```
and then update the plugin PR too :)

app/qml/form/editors/MMFormPhotoEditor.qml
```diff
@@ -164,10 +184,12 @@ MMFormPhotoViewer {
       __inputUtils.removeFile( internal.imageSourceToDelete )
       internal.imageSourceToDelete = ""
     }
+    photoSketchingLoader.item.saveDrawings()
```
uclaros: Is this not:
```suggestion
    root.controller.saveDrawings()
```
?
Same below

app/qml/form/editors/MMFormPhotoViewer.qml
```diff
@@ -28,11 +32,13 @@ MMPrivateComponents.MMBaseInput {
   property bool hasCameraCapability: true
 
   property var photoComponent: photo
+  property PhotoSketchingController controller
```
uclaros: Can you change this to
```suggestion
  property PhotoSketchingController sketchingController
```
or something self-explanatory? _Controller_ is very generic

app/photosketchingcontroller.cpp
```diff
@@ -0,0 +1,204 @@
...
+#include <QUrl>
+
+#include "coreutils.h"
+
+/**
+ * ColorPath declarations
+ */
+
+
+ColorPath::ColorPath()
```
uclaros, line 26: This object looks like a "Sketch" to me.
What do you think about adding the `width` here too instead of tracking in in multiple places, and renaming it to Sketch?

app/photosketchingcontroller.cpp
```diff
@@ -0,0 +1,204 @@
...
+    emit newPathAdded( -1 );
+    if ( !mCanUndo && !mPaths.isEmpty() )
+    {
+      mCanUndo = true;
+      emit canUndoChanged();
+    }
+  }
+  else
+  {
+    mPaths[mPaths.size() - 1] = mCurrentLine;
```
uclaros:
```suggestion
    mPaths.last() = mCurrentLine;
```

app/photosketchingcontroller.cpp
```diff
@@ -0,0 +1,204 @@
...
+  mCurrentLine.mColor = newColor;
+
+  emit activeColorChanged();
+}
+
+void PhotoSketchingController::undo()
+{
+  if ( !mPaths.empty() )
+  {
+    mPaths.remove( mPaths.size() - 1 );
```
uclaros:
```suggestion
    mPaths.removeLast();
```

app/activeproject.cpp
uclaros: Also, if it defaults to `true` on the plugin, I think we should default it to `true` here too.

app/photosketchingcontroller.cpp
```diff
@@ -0,0 +1,204 @@
...
+  else if ( mPaths.empty() && mCanUndo )
+  {
+    mCanUndo = false;
+    emit canUndoChanged();
+  }
+}
+
+void PhotoSketchingController::clear()
+{
+  mPenColor = QColor::fromString( "#FFFFFF" );
```
uclaros:
```suggestion
  mPenColor = QColor( Qt::white );
```

app/photosketchingcontroller.h
```diff
@@ -0,0 +1,102 @@
...
+    void newPathAdded( int pathIndex );
+    void pathUpdated( QVector<int> pathIndexes );
+    void lastPathRemoved();
+    void pathsReset();
+
+  private:
+    std::pair<double, double> mAnnotationOffsets = std::make_pair( 0.0, 0.0 );
+    double mPhotoScaleRatio = 1.0;
+    QString mPhotoSource;
+    QColor mPenColor = QColor::fromString( "#FFFFFF" );
```
uclaros:
```suggestion
    QColor mPenColor = QColor( Qt::white );
```

app/photosketchingcontroller.h
```diff
@@ -0,0 +1,102 @@
...
+    void annotationsChanged();
+    void photoScaleRatioChanged();
+
+    void newPathAdded( int pathIndex );
+    void pathUpdated( QVector<int> pathIndexes );
+    void lastPathRemoved();
+    void pathsReset();
+
+  private:
+    std::pair<double, double> mAnnotationOffsets = std::make_pair( 0.0, 0.0 );
```
uclaros: This would be simpler to be a `QPointF`, no?
You could then also do `QPointF p = point * mScale + offset;`

Can you also add a docstring on what this offset is?

app/photosketchingcontroller.h
```diff
@@ -0,0 +1,102 @@
...
+    double mPhotoScaleRatio = 1.0;
```
uclaros: This is super nitpicking, but _scale_ is a _ratio_ by definition! :grin:

uclaros (commented):
Nice!!!

The `annotation` keyword is still used in a couple of places (logs and mAnnotationOffset iirc), it would be good to change to Sketching or similar.

app/photosketchingcontroller.h
```diff
@@ -0,0 +1,102 @@
...
+    ~PhotoSketchingController() override = default;
+
+    // starts new line, to store points in
+    Q_INVOKABLE void newSketch();
+
+    // adds new point to current line
+    Q_INVOKABLE void addPoint( QPointF newPoint );
+
+    // sets color of new line
+    Q_INVOKABLE void setActiveColor( QColor newColor );
```
uclaros: This could be const ref, right?
```suggestion
    Q_INVOKABLE void setActiveColor( const QColor &newColor );
```

app/photosketchingcontroller.h
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * Utility class, which is used in place of std::pair, because of QML limitations.
+ */
+class ColorPath
+{
+    Q_GADGET
+    Q_PROPERTY( QColor color MEMBER mColor )
+    Q_PROPERTY( QVector<QPointF> points MEMBER mPoints )
+  public:
+    ColorPath( const QColor color, const QVector<QPointF> &list ): mColor( color ), mPoints( list ) {};
```
uclaros: This could be const ref, right?
```suggestion
    ColorPath( const QColor &color, const QVector<QPointF> &list ): mColor( color ), mPoints( list ) {};
```

app/photosketchingcontroller.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+    QPainterPath path;
+    path.moveTo( points.first().x(),  points.first().y() );
+    for ( int j = 1; j < points.size(); ++j )
+    {
+      path.lineTo( points.at( j ).x(),  points.at( j ).y() );
+    }
+
+    painter.drawPath( path );
+  }
```
uclaros: At first I thought this would be more readable like:
```suggestion
  for ( const ColorPath &path : std::as_const( mPaths ) )
  {
    if ( path.mPoints.isEmpty() )
      continue;

    const QPen pen( path.mColor, 4 * mPhotoScaleRatio );
    painter.setPen( pen );
    QPainterPath painterPath;

    bool first = true;
    for ( const QPointF &point : std::as_const( path.mPoints ) )
    {
      if ( first )
      {
        painterPath.moveTo( point );
        first = false;
      }
      else
      {
        painterPath.lineTo( point );
      }
    }
    painter.drawPath( painterPath );
  }
```
but now that I see it again it's not much different!
Maybe just use the `lineTo`/`moveTo` variants accepting `QPointF` instead of `x,y`?

app/CMakeLists.txt
Withalion: This flag prohibited use of `QML_ELEMENT`, `QML_SINGLETON` in code, which create qml types. Currently we are using context properties and `qmlRegisterType` and others to replace this behaviour, I believe it's not the ideal solution and in the future I would like to move away from it. One of the reasons is the use of QML language server in development and also it looks like it's the preferred way by Qt.

Withalion: it was part of this commit, which I wanted to use as a reference for future refactors...https://github.com/MerginMaps/mobile/pull/3930/commits/b6b239a0c9aba8f52adfe85fdadf10b850a7e304

app/activeproject.cpp
Withalion: I'm not 100% sure, but good point

app/qml/form/editors/MMFormPhotoEditor.qml
Withalion: yes and I guess it's more comprehensible from your suggestion

app/photosketchingcontroller.cpp
Withalion: I think here we need to use my approach as we are assigning a new value at that position. `[]` gives assignable reference  and for me `.last()` == `.at()`, which gives only read only access

app/photosketchingcontroller.h
Withalion: You are probably right, I originally didn't want to use `QPointF` as it isn't a point. Anyway it's not relevant anymore as the QML logic changed after Tomas's review and I forgot to remove this.

app/photosketchingcontroller.cpp
Withalion: hmm...I'm okay with renaming it even though I think my name is more descriptive for what it is, but sketch is also not far from truth. However I'm against adding width to it as there is no intention yet to have variable widths for each sketch. So for now I would keep it in `PhotoSketchingController` as a constant also accessible from QML

app/photosketchingcontroller.cpp
Withalion: Resolved with using `QPointF` variants

app/photosketchingcontroller.cpp
uclaros: not necessarily, `constLast()` is the same as `at()`, `last()` is just a ref to the last element

app/photosketchingcontroller.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+  ColorPath colorPath( mPaths.at( index ) );
+
+// we recalculate stored points into the coordinates of printed picture
+  QVector<QPointF> shapePoints;
+  for ( QPointF point : colorPath.mPoints )
+  {
+    shapePoints.append( QPointF( point.x() / mPhotoScale, point.y() / mPhotoScale ) );
+  }
```
uclaros:
```suggestion
  for ( const QPointF &point : colorPath.mPoints )
  {
    shapePoints.append( point / mPhotoScale );
  }
```

app/photosketchingcontroller.h
```diff
@@ -0,0 +1,106 @@
...
+
+  public:
+    explicit PhotoSketchingController( QObject *parent = nullptr ): QObject( parent ) {};
+    ~PhotoSketchingController() override = default;
+
+    // starts new line, to store points in
+    Q_INVOKABLE void newSketch();
+
+    // adds new point to current line
+    Q_INVOKABLE void addPoint( QPointF newPoint );
```
uclaros:
```suggestion
    Q_INVOKABLE void addPoint( const QPointF &newPoint );
```

app/photosketchingcontroller.cpp
```diff
@@ -0,0 +1,208 @@
...
+void PhotoSketchingController::newSketch()
+{
+  mCurrentLine = ColorPath( mPenColor, {} );
+}
+
+void PhotoSketchingController::addPoint( const QPointF newPoint )
+{
+// we scale up the point to picture's true position
+  const QPointF realPoint = QPointF( newPoint.x() * mPhotoScale, newPoint.y() * mPhotoScale );
+  mCurrentLine.mPoints.append( realPoint );
```
uclaros:
```suggestion
void PhotoSketchingController::addPoint( const QPointF &newPoint )
{
// we scale up the point to picture's true position
  mCurrentLine.mPoints.append( newPoint * mPhotoScale );
```

app/photosketchingcontroller.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+  for ( int i = 0; i < pathCount; ++i )
+  {
+    QColor color = mPaths.at( i ).mColor;
+    QVector<QPointF> points = mPaths.at( i ).mPoints;
+
+    if ( points.isEmpty() )
+      continue;
+
+    QPen pen( color, SKETCH_WIDTH * mPhotoScale );
```
uclaros:
```suggestion
    const QColor color = mPaths.at( i ).mColor;
    const QVector<QPointF> points = mPaths.at( i ).mPoints;

    if ( points.isEmpty() )
      continue;

    const QPen pen( color, SKETCH_WIDTH * mPhotoScale );
```

Withalion:
To answer some questions from @RastoHu (in the linked issue):
1. There was no intention of having an eraser, users can undo their last sketch or discard all sketches
2. That is not something we can influence much. The landscape mode is probably more usable for tablets. There was also some intention to have zooming and panning of photo, but that is not getting implemented yet.
3. - 5. This is a new bug which was not happening to me before have to investigate further... new pictures turn 90 degrees left on save

Withalion:
The failing link test seems to give quite random output. Somehow the http requests are getting `QNetworkReply::UnknownServerError`, but it's usually just random one. The links can be opened from Android without issues.

app/qml/form/components/MMFormPhotoSketchingPageDialog.qml
```diff
@@ -0,0 +1,287 @@
...
+      size: MMButton.Sizes.Small
+      bgndColor: __style.polarColor
+      bgndColorDisabled: __style.polarColor
+      bgndColorHover: __style.mediumGreenColor
+      fontColorHover: __style.forestColor
+      iconColorHover: __style.forestColor
+      enabled: root.controller.canUndo
+      Layout.alignment: Qt.AlignVCenter | Qt.AlignLeft
+      Layout.topMargin: __style.pageMargins + __style.safeAreaTop
+      Layout.leftMargin: __style.pageMargins + __style.safeAreaLeft
```
tomasMizera: You usually want to add some white space between properties to keep similar ones grouped and make it easier to read

```suggestion
      type: MMButton.Types.Primary
      text: qsTr( "Undo" )
      iconSourceLeft: __style.undoIcon
      size: MMButton.Sizes.Small
      
      bgndColor: __style.polarColor
      bgndColorDisabled: __style.polarColor
      bgndColorHover: __style.mediumGreenColor
      fontColorHover: __style.forestColor
      iconColorHover: __style.forestColor
      
      enabled: root.controller.canUndo
      
      Layout.alignment: Qt.AlignVCenter | Qt.AlignLeft
      Layout.topMargin: __style.pageMargins + __style.safeAreaTop
      Layout.leftMargin: __style.pageMargins + __style.safeAreaLeft
```

tomasMizera: Just like you do in other places :)

app/qml/form/components/MMFormPhotoSketchingPageDialog.qml
```diff
@@ -0,0 +1,287 @@
...
+      Layout.alignment: Qt.AlignHCenter
+      Layout.preferredHeight: scrollRow.height
+      Layout.preferredWidth: scrollRow.width
+      Layout.maximumWidth: parent.width - ( 2 * __style.pageMargins + __style.safeAreaLeft + __style.safeAreaRight )
+      Layout.bottomMargin: closeButton.implicitWidth
+      Layout.leftMargin: __style.pageMargins + __style.safeAreaLeft
+      Layout.rightMargin: __style.pageMargins + __style.safeAreaRight
+
+      ScrollBar.vertical.policy: ScrollBar.AlwaysOff
+      ScrollBar.horizontal.policy: ScrollBar.AlwaysOff
```
tomasMizera, line 221: Just a note, we have `MMScrollView` in components that sets these two automatically

app/qml/form/components/MMFormPhotoSketchingPageDialog.qml
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    MMComponents.MMListSpacer { implicitHeight: __style.margin20 }
+
+    ScrollView {
+      Layout.alignment: Qt.AlignHCenter
+      Layout.preferredHeight: scrollRow.height
+      Layout.preferredWidth: scrollRow.width
+      Layout.maximumWidth: parent.width - ( 2 * __style.pageMargins + __style.safeAreaLeft + __style.safeAreaRight )
+      Layout.bottomMargin: closeButton.implicitWidth
```
tomasMizera: BottomMargin equals to button's width? 🤔

tomasMizera: According to the design, I think it should be `__style.margin20` instead

app/qml/form/components/MMFormPhotoSketchingPageDialog.qml
```diff
@@ -0,0 +1,287 @@
...
+      }
+    }
+
+    MMComponents.MMListSpacer { implicitHeight: __style.margin20 }
+
+    ScrollView {
+      Layout.alignment: Qt.AlignHCenter
+      Layout.preferredHeight: scrollRow.height
+      Layout.preferredWidth: scrollRow.width
+      Layout.maximumWidth: parent.width - ( 2 * __style.pageMargins + __style.safeAreaLeft + __style.safeAreaRight )
```
tomasMizera, line 215: It should be enough to just say `Layout.fillWidth: true` given that you already have the `leftMargin` and `rightMargin` correctly set

tomasMizera: Then you can likely remove `Layout.preferredWidth` too

app/qml/form/components/MMFormPhotoSketchingPageDialog.qml
```diff
@@ -0,0 +1,287 @@
...
+      Layout.bottomMargin: closeButton.implicitWidth
+      Layout.leftMargin: __style.pageMargins + __style.safeAreaLeft
+      Layout.rightMargin: __style.pageMargins + __style.safeAreaRight
+
+      ScrollBar.vertical.policy: ScrollBar.AlwaysOff
+      ScrollBar.horizontal.policy: ScrollBar.AlwaysOff
+
+      Row {
+        id: scrollRow
+        width: parent.width
```
tomasMizera: Hm, shouldn't the width of this row be the rendered width (all color button's width + spacing)?

app/qml/form/components/MMFormPhotoSketchingPageDialog.qml
```diff
@@ -0,0 +1,287 @@
...
+        spacing: __style.margin12
```
tomasMizera, line 225: Please double check the values with Figma

app/qml/form/components/MMFormPhotoSketchingPageDialog.qml
```diff
@@ -0,0 +1,287 @@
...
+    }
+  }
+
+  contentItem: ColumnLayout {
+
+    Item {
+      id: container
+      Layout.fillWidth: true
+      Layout.fillHeight: true
+      Layout.leftMargin: __style.pageMargins
```
tomasMizera: Shouldn't we use the safe margins here too?

app/qml/form/components/MMFormPhotoSketchingPageDialog.qml
Withalion: Thanks! I didn't know about it, however it also locks to vertical scrolling which we don't want

app/qml/form/components/MMFormPhotoSketchingPageDialog.qml
Withalion: That sounds reasonable and I tried to use it before, but that doesn't work if you want to center the color buttons to horizontal center of screen (must have for tablets).

app/qml/form/components/MMFormPhotoSketchingPageDialog.qml
Withalion: If i can count the spacing is right and I'm not sure we want to make the color buttons smaller. I tried to keep them the same size as in map sketching. The padding can be smaller...

RastoHu:
Landscape mode issue: https://github.com/MerginMaps/mobile/issues/4012

tomasMizera (approved):
🎉 Thanks for all the adjustments!
I think this one is good to go for _beta_ release.

Two things to update:
 - autotests for `PhotoSketchingController` and `clearOrientationMetadata`
 - (optional) instantiate `PhotoSketchingController` only when requested to paint on an image

tomasMizera:
We can do the updates in a separate PRs :)

RastoHu:
iOS, app build 25.8.790411:
1. when draw photo sketches with apple pencil, undo button does not work. The drawing will be removed only when a new drawing is added to the photo. By watching this video, better experience is with sound :)

https://github.com/user-attachments/assets/9ae6ef8f-bc36-4e8c-a71c-20c51a610bff

2. app always crash when editing existing record with photo and photo sketches. App crash after clicking on confirm button on photo sketches screen. (sound on)

https://github.com/user-attachments/assets/a66b6622-8db9-41ee-8855-0810d3e715b8

3. Sometimes when adding a photo sketches with apple pencil, the sketches are connected with straight lines. (sound off)

https://github.com/user-attachments/assets/2b2a0f3d-6105-4031-9434-e7adc95c41a8

RastoHu:
log: rastislav.hudak020625_input-ios-2025.6.0_2025-08-25T11:17:26.035623.log

Withalion:
@IvaKuklica is the 3rd point related to #4020? Were you trying it on the same project or a different one?

RastoHu:
> @IvaKuklica @RastoHu some more info:
> 
> 1. the renaming bug is not connected to sketching, please file a new issue so we can track it and fix it separately
> 2. I'll need more time investigating the Samsung issue as it looks more mysterious than expected

1. is not related to the photo sketches, @IvaKuklica please create a new issue for it. thx

PR #3933, opened 2025-06-16
Rename map annotations to map sketches

Withalion:
Refactor of map annotations to change it to map sketching, mirrors also changes in qgis plugin MerginMaps/qgis-plugin#714. Besides renaming also color buttons got updated to resemble more proposed design.

image


app/qml/map/MMSketchesDrawer.qml
```diff
@@ -73,36 +73,43 @@ MMDrawer {
...
-              color: modelData
-              radius: colorButton.implicitHeight / ( isSelected ? 2.5 : 2 )
-              border.width: isSelected ? 4 * __dp : 0
-              border.color: "black"
+              anchors.verticalCenter: parent.verticalCenter
+              anchors.horizontalCenter: parent.horizontalCenter
+              radius: width / 2
+              width: scrollRow.height
+              height: scrollRow.height
+              color: modelData === root.sketchingController.activeColor.toString() ? __style.transparentColor : __style.lightGreenColor
```
tomasMizera: Best to put the comparison in to a new property. Please note we should do case insensitive comparison

Withalion: I'll move it to variable.

The comparison should be case insensitive by default, `QColor` is in upper case. Even though we parse it to `String List` it should still be saved from `QColor`

app/qml/map/MMMapController.qml
```diff
@@ -179,8 +179,8 @@ Item {
         break
       }
 
-      case "annotate": {
-        root.showInfoTextMessage( qsTr( "Draw annotation" ) )
+      case "sketch": {
+        root.showInfoTextMessage( qsTr( "Draw sketches" ) )
```
tomasMizera: Can we make this banner more informative please? :)

```suggestion
        root.showInfoTextMessage( qsTr( "Select a colour and start sketching on the map. Use two fingers to move or zoom the map." ) )
```

tomasMizera (approved):

PR #3946, opened 2025-06-26
Update server migration message

tomasMizera:


tomasMizera:
Unrelated iOS CI failure (will work on master)

PR #3962, opened 2025-07-02
Allow sending diagnostic logs to host server

### LINKED ISSUE #3885: Send diagnotic logs to the configured server
Use `POST /v2/diagnostic-logs` or value from `/config` as a host for diagnostic logs.

Read more in the parent ticket.

ValentinBuira:
fix #3885

### DIFF
--- a/app/inputhelp.cpp
+++ b/app/inputhelp.cpp
@@ -22,7 +22,6 @@
 #include <QSysInfo>
 
 const QString helpRoot = QStringLiteral( "https://merginmaps.com/docs" );
-const QString reportLogUrl = QStringLiteral( "https://g4pfq226j0.execute-api.eu-west-1.amazonaws.com/mergin_client_log_submit" );
 const QString helpDeskMail = QStringLiteral( "support@merginmaps.com" );
 const QString mmWeb = QStringLiteral( "https://merginmaps.com" );
 const QString changelogRss = QStringLiteral( "https://wishlist.merginmaps.com/rss/changelog.xml" );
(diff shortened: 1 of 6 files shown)


core/merginapi.cpp
```diff
@@ -3833,6 +3835,19 @@ void MerginApi::getServerConfigReplyFinished()
         }
       }
 
+      if ( ( major >= 2024 && minor >= 4 ) && diagnosticUrl.isEmpty() )
```
tomasMizera: I think it is since `2025` :)
```suggestion
      if ( ( major >= 2025 && minor >= 4 ) && diagnosticUrl.isEmpty() )
```

CeliaBuira: Indeed thanks, fixed and added a basic test case for the app.dev server configuration

app/inputhelp.cpp
```diff
@@ -267,11 +266,11 @@ void InputHelp::submitReport()
   if ( username.isEmpty() )
     username = "unknown";
   QString params = QStringLiteral( "?app=%1&username=%2" ).arg( app ).arg( username );
-  QNetworkRequest req( reportLogUrl + params );
+  QNetworkRequest req( mMerginApi->serverDiagnosticLogsUrl() + params );
   req.setRawHeader( "User-Agent", "InputApp" );
```
Withalion, line 270: If this is not required by server I would change it to `MM-mobile` or something reflecting our new name...eventually nobody will look for Input

app/inputhelp.cpp
```diff
@@ -267,11 +266,11 @@ void InputHelp::submitReport()
...
   req.setRawHeader( "Content-Type", "text/plain" );
   QNetworkReply *reply = mManager.post( req, logArr );
-  qDebug() << "Report to " << reportLogUrl;
+  qDebug() << "Report to " << mMerginApi->serverDiagnosticLogsUrl();
```
Withalion: I would either change it to `CoreUtils::log` or remove it completely. This is useless for production...

Withalion (commented):
Looks good, just some minor changes

app/inputhelp.cpp
Withalion: I see now that is not apparent what I was talking about. I meant the user agent

tomasMizera: It is used by the server quite a lot, we can not change it in one place. That would need to be coordinated :)

Withalion: Ahh alright, let's add it to our endless TO-DO list then :)

app/inputhelp.cpp
CeliaBuira: thanks, I removed it altogether

Withalion (approved):

PR #3964, opened 2025-07-03
Cache user profile data and fix usage of userauth and userinfo

tomasMizera:
This PR introduces cache for `merginuserinfo` class data. Previously we used to cache only details in `merginuserauth`. The PR separates the usage of the two, as follows:
 - `merginuserauth` is used solely for authorisation purposes, its `username` property got renamed to `login` as it bears either username or email used to sign in
 - `merginuserinfo` now caches the username, email and full name - we use it throughout the app, e.g. in variables (even offline) or profile details; instead of data from `merginuserauth`.

Side changes:
- refactored the `merginuserinfo` and `merginapi` a bit and filed another ticket to continue the work (see https://github.com/MerginMaps/mobile/issues/3959)
- dropped `userId` as we did not use it anywhere (again, related to https://github.com/MerginMaps/mobile/issues/3959)

It should be possible to review the PR following each commit changes.

 - [ ] Let me see if we can test this caching somehow


core/merginuserinfo.cpp
```diff
@@ -29,6 +132,14 @@ void MerginUserInfo::clear()
   mWorkspaces.clear();
   mInvitations.clear();
 
+  QSettings settings;
+  settings.remove( "Input/workspaces" );
+  settings.remove( "Input/name" );
+  settings.remove( "Input/email" );
+  settings.remove( "Input/username" );
+  settings.setValue( "Input/lastUsedWorkspace", -1 );
```
Withalion: If you use `beginGroup` there is no need for the prefix

core/merginapi.cpp
```diff
@@ -42,7 +42,7 @@ const QString MerginApi::sDefaultApiRoot = QStringLiteral( "https://app.merginma
 const QSet<QString> MerginApi::sIgnoreExtensions = QSet<QString>() << "gpkg-shm" << "gpkg-wal" << "qgs~" << "qgz~" << "pyc" << "swap";
 const QSet<QString> MerginApi::sIgnoreImageExtensions = QSet<QString>() << "jpg" << "jpeg" << "png";
 const QSet<QString> MerginApi::sIgnoreFiles = QSet<QString>() << "mergin.json" << ".DS_Store";
-const int MerginApi::UPLOAD_CHUNK_SIZE = 10 * 1024 * 1024; // Should be the same as on Mergin server
+const int MerginApi::UPLOAD_CHUNK_SIZE = 10 * 1024 * 1024; // Should be the same as on the server
```
Withalion, line 45: > should be the same on server

nitpick, but current wording sounds over-engineered

also it's `constexpr`

core/merginuserinfo.cpp
```diff
@@ -13,10 +13,113 @@
...
+  }
+
+  mActiveWorkspace = newWorkspace;
+
+  QSettings settings;
+  settings.beginGroup( "Input/" );
+  settings.setValue( "lastUsedWorkspace", mActiveWorkspace );
+  settings.endGroup();
+
+  QString logMessage = QStringLiteral( "Switched to workspace '%1' with ID %2" )
```
Withalion:
```suggestion
  const QString logMessage = QStringLiteral( "Switched to workspace '%1' with ID %2" )
```

Withalion (approved):
👏

core/merginapi.cpp
tomasMizera: I am not sure I got this 🤔

Withalion:
```suggestion
constexpr int MerginApi::UPLOAD_CHUNK_SIZE = 10 * 1024 * 1024; // Should be the same on server
```

Withalion (approved):

core/merginapi.cpp
tomasMizera: When me president, they see, they see

tomasMizera: I'd rather change all of them in one go :)

core/merginapi.cpp
```diff
@@ -1244,8 +1258,7 @@ void MerginApi::saveAuthData()
   settings.setValue( "apiRoot", mApiRoot );
   settings.endGroup();
 
-  mUserAuth->saveAuthData();
-  mUserInfo->clear();
+  mUserAuth->saveData();
```
uclaros, line 1261: This was probably not the best pace to clear the user info, indeed.
But are we now actually clearing the user info when the user auth changes?
If not, we could end up with mis-matching userInfo and userAuth, no?

Issue #1974, opened 2022-03-07
Layers Display Order Error


CJHacking93:
When selecting a map theme, and then clicking record, sometimes the layer selected by default is actually the 2nd layer in the layer tree. It is also 2nd in alphabetical order (due to numbers assigned at the start of each layer name for order purposes). Obviously, the order of the layer tree is supposed to depict the order when recording a point.

Additional information can be required if requested.

Hackster93:
Hi Mergin Maps Team,

I would like to please follow up on this issue. Please see the video and screen shots below. The default layer to edit should be `5.1. (M) Pest Monitoring` but this isn't the case. This happens in several other Map Themes as well. The order and naming convention should result in it being the default layer. Is there perhaps a work-around I could implement?

https://github.com/MerginMaps/input/assets/90187049/3a0b5f1a-f667-458a-afff-726d96cfd4b3
![WhatsApp Image 2023-08-28 at 13 22 58](https://github.com/MerginMaps/input/assets/90187049/3daf8222-d74c-4bf3-9671-c3667d8797d8)
![image](https://github.com/MerginMaps/input/assets/90187049/1a5f7522-3db8-4a35-a836-d7d99fbf2cbe)
![image](https://github.com/MerginMaps/input/assets/90187049/8851ef4c-32e5-4efa-94a1-8b8434cd4c22)

Hackster93:
Hi @Tadeo98,

Thank you for your response. 

Please see the attached video with me changing between map themes. You can see this is the case in several themes that the first layer is not selected.

https://github.com/MerginMaps/input/assets/90187049/a57c96a3-0e5a-45a6-a4f1-a582b2bc0cb7

I am happy to share the project with you if this would help? The one thing I thought it could be is we have 'duplicate' layers in some of the map themes, so for example `5.2. (M) Nematode Sample` and `5.3. (M) Diseased Tree Sample` both point to the same layer `sample_monitor` with different Attribute Forms and styling. However, the `1. Layouts & List Setup` theme has no duplicate layers in it.

Tadeo98:
Hi and thank you for more details on this issue. Please proceed with sharing the project to Mergin Maps user support@merginmaps.com (have you not done this prior to this message/issue) and let me know your workspace and project name for identification.
Please follow the tutorial if in doubt:
https://merginmaps.com/docs/tutorials/working-collaboratively/#share-a-project-with-a-specific-mergin-maps-user

[...]

PR #3996, opened 2025-07-11
Respect custom rendering order from QGIS

### LINKED ISSUE #3816: Layers Order in QGIS's Layer Order panel ignored
Layers are displayed in the same order as shown in `Layers panel`, ignoring the actual rendering order defined in `Layer order panel`. This causes visualization inconsistencies when users have specifically arranged their layers using the "Layer order" functionality in QGIS.

ValentinBuira:
This PR make the app follow the layer rendering order if it's manually overwritten in QGIS in the `Layer Order` Panel

Fix #3816
And might solve #1974 but to be confirmed

### DIFF
--- a/app/activeproject.cpp
+++ b/app/activeproject.cpp
@@ -587,17 +587,13 @@ QList<QgsMapLayer *> ActiveProject::getVisibleLayers() const
 
   // Get list of all visible valid layers in the project
   QList<QgsMapLayer *> visibleLayers;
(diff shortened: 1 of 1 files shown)


Withalion:
Doesn't solve #1974, while using map themes the active layer is not the first one. Also changing to a map theme that didn't have previous active layer visible, still didn't change active layer.

Issue #3612, opened 2024-09-20
Increase diagnostic log file size


erpas:
In many cases logs shared with users are useless because relevant info was already rotated with a more recent logs.
Please increase the size of logs.

RastoHu:
mobile app - OK
plugin

PR #3998, opened 2025-07-11
Increase log size to 8mb

ValentinBuira:
Part of #3612


Withalion (approved):
It's a nitpick, but on lines 191-193 we should decide if we are using `fileSize` or `file.size()`

PR #3999, opened 2025-07-15
Haptic feedback when recording features

### LINKED ISSUE #2681: Haptic or audio feedback when recording point
Today I tried recording a few paths for the first time, using Mergin Maps Input. And it went well :) 

I hold an external GPS in one hand and the phone in the other hand, just as when I used other apps before. It did happen a few times, just like with the other apps, that when I tapped "Add" somehow it did not register my tap. It would be great with some kind of feedback that confirms that the input was registered and that the point was succesfully added. 

Perhaps a different haptic feedback / sound for each of the following: 
1) point successfully added
2) input registered, but point not added (eg due to insufficient accuracy)

Sincerely

Withalion:
Resolves #2681 

This PR adds new settings to enable haptic feedback while recording. These settings are device bound not project bound. The users can enable vibration or sound or both as haptic feedback. Then while recording a new feature the buttons will give feedback.

Demo:
https://github.com/user-attachments/assets/43e4c59b-6d56-463d-b39c-5f1c3e223ce4

Notification dialog is for only for android as users need to enable haptic feedback in system settings too and looks like this:


app/android/src/uk/co/lutraconsulting/InputActivity.java
```diff
@@ -16,6 +16,10 @@
 import android.util.Log;
 import android.os.Bundle;
 import android.os.Build;
+import android.os.VibratorManager;
```
tomasMizera, line 19: This `VibratorManager` seems to be available from SDK 31 https://developer.android.com/reference/android/os/VibratorManager

app/ios/iosutils.mm
```diff
@@ -113,3 +114,39 @@ - ( UIViewController * )documentInteractionControllerViewControllerForPreview:(
...
   return true;
 }
+
+void IosUtils::vibrate()
+{
+  if ( @available( iOS 13.0, * ) )
+  {
+    // initialize engine
+    NSError *errorEngine;
+    CHHapticEngine *engine = [[CHHapticEngine alloc] initAndReturnError:&errorEngine];
```
tomasMizera, line 103: Is it possible to get `nil` here as result?

app/qml/map/MMRecordingTools.qml
```diff
@@ -361,6 +382,21 @@ Item {
...
+      }
+      else if ( __iosUtils.isIos ) {
+        __iosUtils.vibrate()
+      }
+    }
+
+    if ( __appSettings.useHapticsSound ) {
+      hapticSound.play()
+    }
+  }
```
tomasMizera: Is it possible to do this in cpp instead?

app/qml/settings/MMSettingsPage.qml
```diff
@@ -183,6 +186,40 @@ MMPage {
...
+        title: qsTr("Enable sound ")
+        checked: __appSettings.useHapticsSound
+        visible: __appSettings.useHaptics
+
+        onClicked: __appSettings.useHapticsSound = !checked
+      }
+
+      MMSettingsComponents.MMSettingsSwitch {
+        width: parent.width
+        title: qsTr("Enable vibration")
```
tomasMizera: How about a single switch with the following options: `No haptic feedback`, `Sound`, `Vibration`, `Sound & Vibration`?

tomasMizera (commented):
🚀

app/qml/map/MMRecordingTools.qml
Withalion: can we postpone this for the refactor of `nativeUtils`, it would be nicer to do everything at one go?

app/qml/settings/MMSettingsPage.qml
Withalion: As decided on call, it will be a single drop down menu

app/qml/settings/MMSettingsPage.qml
```diff
@@ -178,9 +193,45 @@ MMPage {
...
+      MMLine {}
+
+      MMSettingsComponents.MMSettingsDropdown {
+        id: hapticDropdown
+        width: parent.width
+
+        title: qsTr("Haptic feedback type")
+        description: qsTr("Choose a type of haptic feedback while recording")
+
+        // value: () => model.get(currentIndex).text
```
tomasMizera:
```suggestion

```

app/qml/settings/MMSettingsPage.qml
```diff
@@ -35,6 +39,17 @@ MMPage {
     }
   }
 
+  ListModel {
+    id: hapticsTypeModel
+
+    Component.onCompleted: {
+      hapticsTypeModel.append({ value: AppSettings.Off, text: qsTr("Off") });
```
tomasMizera: Can we add some prefix to the enum values? As `AppSettings.Off` might relate to many things :)
How about something like
```suggestion
      hapticsTypeModel.append({ value: AppSettings.HapticsOff, text: qsTr("Off") });
```

app/qml/settings/MMSettingsPage.qml
```diff
@@ -35,6 +39,17 @@ MMPage {
     }
   }
 
+  ListModel {
+    id: hapticsTypeModel
+
+    Component.onCompleted: {
```
tomasMizera: Isn't it possible to declare entries via `ListElement` instead? They say it should be possible to assign enum values https://doc.qt.io/qt-6/qml-qtqml-models-listelement.html (might be related only to QML enums though 🤔 )


```suggestion
    ListElement{ text: qsTr('Off'); value: AppSettings.HapticsOff }
```

app/qml/settings/MMSettingsPage.qml
```diff
@@ -35,6 +39,17 @@ MMPage {
     }
   }
 
+  ListModel {
```
tomasMizera: Can we define the model down there when declaring the dropdown?

```qml
model: ListModel {
...
}
```

tomasMizera: Non-visual children should be defined either within their visible parent (e.g. model in view) or below visible children

tomasMizera (commented):
Nice refactoring 👍🏻

app/qml/settings/MMSettingsPage.qml
Withalion: looks like with the new method of exposing c++ enums to QML the previous approach is not necessary anymore

app/android/src/uk/co/lutraconsulting/InputActivity.java
```diff
@@ -271,6 +275,20 @@ public void quitGracefully()
...
+  {
+    Vibrator vib;
+    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.S)
+    {
+      vib = (Vibrator) getSystemService(getApplicationContext().VIBRATOR_SERVICE);
+    } else {
+      VibratorManager vibManager = (VibratorManager) getSystemService(getApplicationContext().VIBRATOR_MANAGER_SERVICE);
+      vib = vibManager.getDefaultVibrator();
+    }
+    vib.vibrate(VibrationEffect.createPredefined(VibrationEffect.EFFECT_CLICK),
```
tomasMizera: Can the `vib` still be `nullptr` at this point?

tomasMizera: This `vibrate` method overload seems to be available only from API 33+ (see [docs](https://developer.android.com/reference/android/os/Vibrator#vibrate(android.os.VibrationEffect,%20android.os.VibrationAttributes))). I guess for older devices we could use [this one instead](https://developer.android.com/reference/android/os/Vibrator#vibrate(android.os.VibrationEffect)).

app/qml/settings/MMSettingsPage.qml
```diff
@@ -178,9 +182,63 @@ MMPage {
...
+      }
+
+      MMLine {}
+
+      MMSettingsComponents.MMSettingsDropdown {
+        id: hapticDropdown
+        width: parent.width
+
+        title: qsTr("Haptic feedback type")
+        description: qsTr("Choose a type of haptic feedback while recording")
```
tomasMizera: How about something less technical like:

```suggestion
        title: qsTr("Touch Feedback")
        description: qsTr("The app can vibrate and play a sound when adding a feature")
```

app/qml/settings/MMSettingsPage.qml
```diff
@@ -178,9 +182,63 @@ MMPage {
...
 
-        onClicked: __appSettings.autolockPosition = !checked
+        currentIndex: AppSettings.hapticsType
+
+        model: ListModel {
```
tomasMizera: I am wondering if we should have a different model for desktop builds where only sound works, but no vibration

Withalion: no haptic feedback for desktops :)

app/android/src/uk/co/lutraconsulting/InputActivity.java
Withalion: no

app/qml/settings/MMSettingsPage.qml
```diff
@@ -7,13 +7,17 @@
  *                                                                         *
  ***************************************************************************/
 
+pragma ComponentBehavior: Bound
```
tomasMizera: https://doc.qt.io/qt-6/qtqml-documents-structure.html#componentbehavior 🤯 

Can we not define this now? I suggest we do it at once when needed. It would give us some headache when debugging later

app/hapticsmodel.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+HapticsModel::HapticsModel( QObject *parent ) : QAbstractListModel( parent )
+{
+#ifdef DESKTOP_OS
+  mHapticsTypes.append( {AppSettings::HapticsType::HapticsOff, tr( "Off" )} );
+  mHapticsTypes.append( {AppSettings::HapticsType::HapticsSound, tr( "Sound" )} );
+#else
+  mHapticsTypes.append( {AppSettings::HapticsType::HapticsOff, tr( "Off" )} );
+  mHapticsTypes.append( {AppSettings::HapticsType::HapticsVibration, tr( "Vibration" )} );
+  mHapticsTypes.append( {AppSettings::HapticsType::HapticsSound, tr( "Sound" )} );
```
tomasMizera: Let's keep the order of indexes, just in case :) 
```suggestion
  mHapticsTypes.append( {AppSettings::HapticsType::HapticsSound, tr( "Sound" )} );
  mHapticsTypes.append( {AppSettings::HapticsType::HapticsVibration, tr( "Vibration" )} );
```

app/qml/settings/MMSettingsPage.qml
```diff
@@ -178,9 +182,41 @@ MMPage {
...
+            case AppSettings.HapticsVibration:
+              hapticDropdown.value = "Vibration"
+              break
+            case AppSettings.HapticsSound:
+              hapticDropdown.value = "Sound"
+              break
+            case AppSettings.HapticsVibrationSound:
+              hapticDropdown.value = "Vibration & Sound"
+              break
+          }
```
tomasMizera: With the cpp model in place, you could do something like

```suggestion
          value = model.data( currentIndex, ValueRole)
```

app/hapticsmodel.h
tomasMizera, line 1: Nice training with the Abstract model :) In future, `QStandardItemModel` should work well for such use cases

Withalion: Oh, I thought `QAbstractListModel` is better suited for this use case, but nice to know that this exists too

app/qml/settings/MMSettingsPage.qml
Withalion: hmm...let's talk about it tomorrow

PR #4003, opened 2025-07-22
Fix photo naming

### LINKED ISSUE #3035: Check if photo names contain invalid characters
A user reported inability to download their project to desktop after it was uploaded from a mobile device.
We have found out some photo names include "\n" or 0x0a that later breaks download attempts from the service.

See the following for more details
https://gitlab.lutraconsulting.co.uk/ops/operations/-/issues/88#note_36853

### LINKED ISSUE #3709: Photos in wrong folder with wrong name(?)
In not infrequent that despite having the project set to store photos in a specific folder with a specific name

Image

photos are being saved in the project folder with a name that seems to be the one given by default by the camera app of the device

Image

anyway the path stored in the attribute table is correct, and this photos are not ones that have been added in the form.

### LINKED ISSUE #3762: Avoid using forbiden characters in photo names
When the app is renaming an image, we should double-check if the name would be accepted on the server. If not, we should sanitize the name to allow sync.

See https://github.com/MerginMaps/qgis-plugin/issues/678#issuecomment-2746300074

uclaros:
- Make sure that photos end up in the correct path, combination of _Default Path_ and _custom photo name_
- Make sure the path stored in the feature field respects the _absolute/relative to Project Path/Default Path_ setting
- Sanitize Default paths and photo filenames, replacing invalid chars with `_`
- Allow using field alias in custom photo name expression

Fix #3762
Should fix #3709 
Should fix #3035

### DIFF
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -1965,8 +1965,10 @@ QUrl InputUtils::iconFromGeometry( const Qgis::GeometryType &geometry )
 
 void InputUtils::sanitizeFileName( QString &fileName )
 {
-  const QRegularExpression illegalChars( QStringLiteral( "[\x00-\x20<>:|?*\"]" ) );
+  // regex captures ascii codes 0 to 31 and windows path forbidden characters <>:|?*"
+  const thread_local QRegularExpression illegalChars( QStringLiteral( "[\x00-\x19<>:|?*\"]" ) );
   fileName.replace( illegalChars, QStringLiteral( "_" ) );
+  fileName = fileName.trimmed();
 }
 
 bool InputUtils::rescaleImage( const QString &path, QgsProject *activeProject )
--- a/app/test/testattributecontroller.cpp
+++ b/app/test/testattributecontroller.cpp
@@ -745,7 +745,7 @@ void TestAttributeController::testPhotoRenaming()
   // 8 - photo_with_slash_in_subfolder_relative_to_default_path -> subfolder 'media/with_slash'
   //
 
-  controller.setFormValue( items.at( 2 ), QStringLiteral( "mynotes" ) );
+  controller.setFormValue( items.at( 2 ), QStringLiteral( "mynotes " ) );
 
   // These are the field values set by QML code after attaching a photo
   controller.setFormValue( items.at( 3 ), QStringLiteral( "image1.jpg" ) );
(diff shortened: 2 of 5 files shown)


app/inpututils.cpp
```diff
@@ -69,6 +69,7 @@
 
 static const QString DATE_TIME_FORMAT = QStringLiteral( "yyMMdd-hhmmss" );
 static const QString INVALID_DATETIME_STR = QStringLiteral( "Invalid datetime" );
+const thread_local QRegularExpression ILLEGAL_FILENAME_CHARS( QStringLiteral( "[\x00-\x20<>:|?*\"]" ) );
```
tomasMizera: Is this the same logic as on server?

app/inpututils.cpp
```diff
@@ -1961,6 +1964,11 @@ QUrl InputUtils::iconFromGeometry( const Qgis::GeometryType &geometry )
   }
 }
 
+void InputUtils::sanitizeFileName( QString &fileName )
+{
+  fileName.replace( ILLEGAL_FILENAME_CHARS, QStringLiteral( "_" ) );
```
tomasMizera: Can we also trim the `fileName`?

app/test/testattributecontroller.cpp
```diff
@@ -718,30 +725,182 @@ void TestAttributeController::testPhotoRenaming()
...
+  }
 
-  QVERIFY( !QFile::exists( projectDir + QStringLiteral( "/photo2.jpg" ) ) );
-  QVERIFY( QFile::exists( projectDir + QStringLiteral( "/photos/Survey.jpg" ) ) );
-  QCOMPARE( controller.featureLayerPair().feature().attribute( 4 ), QVariant( "photos/Survey.jpg" ) );
+  // TODOs:
+  //  - report that people should not use feature IDs in custom photo names!
+  //  - fix issue where email might be used for mm_username variable as this breaks the renaming!
+  //  - fix docs recommending variables with old prefix for photo names
+  //  - report misconfiguration of the use of subfolders & subfolders in the photo naming
```
tomasMizera: I think we can remove this now

app/inpututils.cpp
uclaros: On server the [pathvalidate](https://github.com/thombashi/pathvalidate) python library is used, which does an extensive validation.

I believe with this simplistic approach we're addressing 99% of common use cases

app/test/testattributecontroller.cpp
uclaros: Or move them to plugin repo :)

Is the email a real issue though? Emails should not have illegal characters for a filename!

app/inpututils.cpp
uclaros: This is slightly tricky!
Should we trim before the file extension? After? Both?
We're using this method to also sanitize the Default Path, which doesn't have an extension, but if it has a `.` we should not trim before it!

app/test/testattributecontroller.cpp
tomasMizera: - `report misconfiguration of the use of subfolders & subfolders in the photo naming` this is not relevant anymore
- `fix docs recommending variables with old prefix for photo names` this is merged in docs
- `fix issue where email might be used for mm_username variable as this breaks the renaming!` this is fixed as we now cache the email (works even  offline)
- `report that people should not use feature IDs in custom photo names!` I believe we say this in docs :)

app/test/testattributecontroller.cpp
```diff
@@ -718,30 +725,186 @@ void TestAttributeController::testPhotoRenaming()
...
+  // Attachment fields:
+  // 3 - photo
+  // 4 - photo_with_slash
+  // 5 - photo_in_subfolder_relative_to_project_path -> subfolder 'media'
+  // 6 - photo_in_subfolder_relative_to_default_path -> subfolder 'media'
+  // 7 - photo_with_slash_in_subfolder_relative_to_project_path -> subfolder 'media/with_slash'
+  // 8 - photo_with_slash_in_subfolder_relative_to_default_path -> subfolder 'media/with_slash'
+  //
+
+  controller.setFormValue( items.at( 2 ), QStringLiteral( "mynotes" ) );
```
tomasMizera: How about testing the trimming too?
```suggestion
  controller.setFormValue( items.at( 2 ), QStringLiteral( "mynotes " ) );
```

tomasMizera (approved):
Looks good! 🚀

tomasMizera:
Unrelated iOS CI failure

PR #4005, opened 2025-07-24
Fix opening app in landscape on iOS, do not rotate the screen

tomasMizera:
Found a better workaround for https://github.com/MerginMaps/mobile/issues/3975. The app did not render properly when opening it in the landscape mode on iOS devices. Therefore, we introduced a workaround to rotate the screen automatically to the portrait mode as it triggered repaint (https://github.com/MerginMaps/mobile/pull/3824)

However, the root cause turned out to be when using `QML Settings` object (we use it to store the main window's size and position so the window opens on the same position as the last time you closed the app).
See more details in the Qt bug report: https://bugreports.qt.io/browse/QTBUG-138685

The fix is to not use `QML Settings` and rely on our own implementation -> each time the main window's x/y/width/height changes, we store the values to QSettings (in cpp) and use those values when opening the window. It is still a workaround, but seems to be working just fine :) and we do not need to rotate the screen anymore 🤪

Fixes https://github.com/MerginMaps/mobile/issues/3975

### DIFF
--- a/app/appsettings.h
+++ b/app/appsettings.h
@@ -35,7 +35,7 @@ class AppSettings: public QObject
     Q_PROPERTY( double gpsAntennaHeight READ gpsAntennaHeight WRITE setGpsAntennaHeight NOTIFY gpsAntennaHeightChanged )
     Q_PROPERTY( QString ignoreMigrateVersion READ ignoreMigrateVersion WRITE setIgnoreMigrateVersion NOTIFY ignoreMigrateVersionChanged )
     Q_PROPERTY( bool autolockPosition READ autolockPosition WRITE setAutolockPosition NOTIFY autolockPositionChanged )
-    Q_PROPERTY( QVector<QVariant> windowPosition READ windowPosition WRITE setWindowPosition NOTIFY windowPositionChanged )
+    Q_PROPERTY( QList<QVariant> windowPosition READ windowPosition WRITE setWindowPosition NOTIFY windowPositionChanged )
 
   public:
     explicit AppSettings( QObject *parent = nullptr );
--- a/app/main.cpp
+++ b/app/main.cpp
@@ -713,33 +713,6 @@ int main( int argc, char *argv[] )
   engine.rootContext()->setContextProperty( "__haveBluetooth", false );
 #endif
 
-#ifdef MOBILE_OS
-  engine.rootContext()->setContextProperty( "__appwindowvisibility", QWindow::Maximized );
-  engine.rootContext()->setContextProperty( "__appwindowx", QVariant( 0 ) );
-  engine.rootContext()->setContextProperty( "__appwindowy", QVariant( 0 ) );
-  engine.rootContext()->setContextProperty( "__appwindowwidth", QVariant( 0 ) );
-  engine.rootContext()->setContextProperty( "__appwindowheight", QVariant( 0 ) );
-#else
-  engine.rootContext()->setContextProperty( "__appwindowvisibility", QWindow::Windowed );
-
-  QVariantList windowCachedPosition = as.windowPosition();
-  if ( windowCachedPosition.isEmpty() )
-  {
-    engine.rootContext()->setContextProperty( "__appwindowx", QVariant( 0 ) );
-    engine.rootContext()->setContextProperty( "__appwindowy", QVariant( 0 ) );
-    engine.rootContext()->setContextProperty( "__appwindowwidth", QVariant( 640 ) );
-    engine.rootContext()->setContextProperty( "__appwindowheight", QVariant( 1136 ) );
-  }
-  else
-  {
-    engine.rootContext()->setContextProperty( "__appwindowx", windowCachedPosition.at( 0 ) );
-    engine.rootContext()->setContextProperty( "__appwindowy", windowCachedPosition.at( 1 ) );
-    engine.rootContext()->setContextProperty( "__appwindowwidth", windowCachedPosition.at( 2 ) );
-    engine.rootContext()->setContextProperty( "__appwindowheight", windowCachedPosition.at( 3 ) );
-  }
-#endif
-  engine.rootContext()->setContextProperty( "__version", version );
-
   // Even though enabling QT's HighDPI scaling removes the need to multiply pixel values with dp,
   // there are screens that need a "little help", because system DPR has different value than the
   // one we calculated. In these scenarios we use a ratio between real (our) DPR and DPR reported by QT.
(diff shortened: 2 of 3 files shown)


app/appsettings.h
```diff
@@ -35,6 +35,7 @@ class AppSettings: public QObject
     Q_PROPERTY( double gpsAntennaHeight READ gpsAntennaHeight WRITE setGpsAntennaHeight NOTIFY gpsAntennaHeightChanged )
     Q_PROPERTY( QString ignoreMigrateVersion READ ignoreMigrateVersion WRITE setIgnoreMigrateVersion NOTIFY ignoreMigrateVersionChanged )
     Q_PROPERTY( bool autolockPosition READ autolockPosition WRITE setAutolockPosition NOTIFY autolockPositionChanged )
+    Q_PROPERTY( QVector<QVariant> windowPosition READ windowPosition WRITE setWindowPosition NOTIFY windowPositionChanged )
```
Withalion:
```suggestion
    Q_PROPERTY( QList<QVariant> windowPosition READ windowPosition WRITE setWindowPosition NOTIFY windowPositionChanged )
```
QVector is just alias and you use QList in other places

app/main.cpp
```diff
@@ -715,12 +715,28 @@ int main( int argc, char *argv[] )
...
+    engine.rootContext()->setContextProperty( "__appwindowwidth", QVariant( 640 ) );
+    engine.rootContext()->setContextProperty( "__appwindowheight", QVariant( 1136 ) );
+  }
+  else
+  {
+    engine.rootContext()->setContextProperty( "__appwindowx", windowCachedPosition.at( 0 ) );
+    engine.rootContext()->setContextProperty( "__appwindowy", windowCachedPosition.at( 1 ) );
+    engine.rootContext()->setContextProperty( "__appwindowwidth", windowCachedPosition.at( 2 ) );
+    engine.rootContext()->setContextProperty( "__appwindowheight", windowCachedPosition.at( 3 ) );
+  }
```
Withalion: I'm just against the idea of using `contextProperty`, the information is relevant only for startup. So, how about using [initial properties](https://doc.qt.io/qt-6/qqmlapplicationengine.html#setInitialProperties), it would require switching to QQmlApplicationEngine though.

app/qml/main.qml
```diff
@@ -1171,4 +1179,11 @@ ApplicationWindow {
       __style.safeAreaLeft = safeArea[3]
     }
   }
+
+  function storeWindowPosition() {
+    if ( Qt.platform.os !== "ios" && Qt.platform.os !== "android")
+    {
+      storeWindowPositionTimer.restart()
+    }
+  }
```
Withalion, line 1188: Wouldn't it be better to set some threshold (e.g 25px) instead of timer?

app/main.cpp
```diff
@@ -715,12 +715,28 @@ int main( int argc, char *argv[] )
...
-  engine.rootContext()->setContextProperty( "__appwindowwidth", 640 );
-  engine.rootContext()->setContextProperty( "__appwindowheight", 1136 );
+
+  QVariantList windowCachedPosition = as.windowPosition();
+  if ( windowCachedPosition.isEmpty() )
+  {
+    engine.rootContext()->setContextProperty( "__appwindowx", QVariant( 0 ) );
+    engine.rootContext()->setContextProperty( "__appwindowy", QVariant( 0 ) );
+    engine.rootContext()->setContextProperty( "__appwindowwidth", QVariant( 640 ) );
+    engine.rootContext()->setContextProperty( "__appwindowheight", QVariant( 1136 ) );
```
Withalion: 🪄 numbers, let's define it somewhere as `DEFAULT_WINDOW_WIDTH/HEIGHT`
also isn't this some weird size? maybe comment why we chose this

Withalion (commented):
Looks good, but will need some more work

app/appsettings.h
tomasMizera: Ah, forgot to replace that!

app/main.cpp
tomasMizera: I simply reused what we had before. Agreed, they are a bit weird, I do not think there is much reason for them. Probably just to mimic mobile screen size so you do not need to resize the window each time.

app/main.cpp
tomasMizera: I like the idea of using new APIs/flows, but in this case it would have some further consequences like updating how we handle translations 😵‍💫  https://doc.qt.io/qt-6/qqmlapplicationengine.html#details I'd keep that as a standalone task - we can review the impact

app/qml/main.qml
tomasMizera: I believe time-based interval is better here as we store the coordinates (ideally) just once :)

app/main.cpp
tomasMizera: ~~I added them as `const` values to `MMStyle` :)~~ Whops, wrong thread

app/main.cpp
tomasMizera: I added them as `const` values to `MMStyle` :)

app/main.cpp
```diff
@@ -725,10 +735,10 @@ int main( int argc, char *argv[] )
   QVariantList windowCachedPosition = as.windowPosition();
   if ( windowCachedPosition.isEmpty() )
   {
-    engine.rootContext()->setContextProperty( "__appwindowx", QVariant( 0 ) );
-    engine.rootContext()->setContextProperty( "__appwindowy", QVariant( 0 ) );
-    engine.rootContext()->setContextProperty( "__appwindowwidth", QVariant( 640 ) );
-    engine.rootContext()->setContextProperty( "__appwindowheight", QVariant( 1136 ) );
+    engine.rootContext()->setContextProperty( "__appwindowx", style->DEFAULT_WINDOW_X );
+    engine.rootContext()->setContextProperty( "__appwindowy", style->DEFAULT_WINDOW_Y );
```
Withalion: can be also set for mobile

app/mmstyle.h
```diff
@@ -540,6 +540,12 @@ class MMStyle: public QObject
     double safeAreaBottom() const { return mSafeAreaBottom; };
     double safeAreaLeft() const { return mSafeAreaLeft; };
 
+    // Default (startup) window sizes on desktop platforms
+    const int DEFAULT_WINDOW_X = 0;
+    const int DEFAULT_WINDOW_Y = 0;
+    const int DEFAULT_WINDOW_WIDTH = 640;
+    const int DEFAULT_WINDOW_HEIGHT = 1136;
```
Withalion, line 547: can be `constexpr` but it's a nitpick

app/main.cpp
tomasMizera: how about now?

app/mmstyle.h
tomasMizera: We need to commit to a specific standard and then everyone should follow it. Let's discuss that later.

Withalion (approved):

PR #4010, opened 2025-07-30
Target android 15

### LINKED ISSUE #3888: Deprecated edge-to-edge code
Google play reports deprecated setup of edge-to-edge in our app. This is the message:

Image

It seems like this will take affect once we target SDK 35 (this August).

### LINKED ISSUE #3889: Bump target SDK version to Android 15
As every year, we need to bump the target SDK, this time to `35`. See https://developer.android.com/about/versions/15/behavior-changes-15

Withalion:
fixes #3889
fixes #3888

increase android target to Android 15 and update some more dependencies

**This change requires newer android SDK!**


PeterPetrik:
I would just close this PR and add the changes to the vcpkg build (?)

Withalion:
Merging this to master first and then rebasing VCPKG sounds better to me. There are still some things left which I need to solve here.

Withalion:
The android package size is again normal. So this is ready for merge

PR #4018, opened 2025-08-06
Fix haptics

Withalion:
Fixes the bug where on mobile devices the haptics used were opposite to what users selected.


Withalion:
Testing done by @RastoHu:
iOS ✅
Samsung Galaxy tab ✅
Samsung Galaxy A53 - The haptic feedback vibrations are very weak. Sound is ok.

PR #4032, opened 2025-08-11
Fix diagnostic logs request

Withalion:
New servers are expecting auth token to assign username to logs instead of request param

### DIFF
--- a/app/inputhelp.cpp
+++ b/app/inputhelp.cpp
@@ -257,6 +257,7 @@ void InputHelp::submitReport()
   QNetworkRequest request;
   QString params;
 
+  // log storage on old AWS lambda doesn't support any user authentication thus we pass username as parameter to "sign" the log
   if ( mMerginApi->serverDiagnosticLogsUrl() == MerginApi::sDefaultReportLogUrl )
   {
     QString username = mMerginApi->userInfo()->username().toHtmlEscaped();


app/inputhelp.cpp
```diff
@@ -254,14 +254,25 @@ void InputHelp::submitReport()
...
-    username = "unknown";
-  const QString params = QStringLiteral( "?app=%1&username=%2" ).arg( app, username );
-  QNetworkRequest req( mMerginApi->serverDiagnosticLogsUrl() + params );
-  req.setRawHeader( "User-Agent", "InputApp" );
-  req.setRawHeader( "Content-Type", "text/plain" );
-  const QNetworkReply *reply = mManager.post( req, logArr );
+  QNetworkRequest request;
+  QString params;
+
+  if ( mMerginApi->serverDiagnosticLogsUrl() == MerginApi::sDefaultReportLogUrl )
```
PeterPetrik, line 261: if the distinction is based on server version, why the if-else branch is based on URL.. ? shouldn't we check server version to decide?

app/inputhelp.cpp
```diff
@@ -254,14 +254,25 @@ void InputHelp::submitReport()
   const QString log = fullLog( false );
   const QByteArray logArr = log.toUtf8();
   const QString app = QStringLiteral( "input-%1-%2" ).arg( InputUtils::appPlatform(), CoreUtils::appVersion() );
-  QString username = mMerginApi->userInfo()->username().toHtmlEscaped();
-  if ( username.isEmpty() )
-    username = "unknown";
-  const QString params = QStringLiteral( "?app=%1&username=%2" ).arg( app, username );
-  QNetworkRequest req( mMerginApi->serverDiagnosticLogsUrl() + params );
-  req.setRawHeader( "User-Agent", "InputApp" );
```
PeterPetrik, line 262: this is dropped in new change - it is not needed anymore?

app/inputhelp.cpp
Withalion: the relevant distinction here is if we are targeting our AWS lambda or server endpoint

app/inputhelp.cpp
Withalion: - server doesn't use `username` parameter anymore as it identifies users by authentication token
- user agent is set inside `getDefaultRequest()`

app/inputhelp.cpp
PeterPetrik: fine - add please small comment for future-us and feel free to merge

PR #4034, opened 2025-08-12
cancel running background feature requests on model destruction

uclaros:
Very long running feature requests would keep running in background thread even if the caller model was destroyed.

We could also increment `mNextSearchId` upon destruction to cancel pending request, but passing a `QgsFeedback` pointer is clearer.


app/layerfeaturesmodel.cpp
tomasMizera, line 116: Shall we consider the request to be canceled even if the feedback is closed?

```suggestion
canceled = canceled || req.feedback->isCanceled()
```

PR #4042, opened 2025-08-19
Set android.useAndroidX=true and android.enableJetifier=true, and align

kaustuvpokharel:
Please keep this pr open, unless somebody tries this branch and app runs in their emulator.

compile/build tools to 34.0.0. This matches CI and fixes builds by avoiding the android-35 aapt2 crash, and does automatically chooses any other android version ex: 35.0.0

### DIFF
--- a/app/CMakeLists.txt
+++ b/app/CMakeLists.txt
@@ -182,18 +182,6 @@ set(MM_HDRS
     mmstyle.h
 )
 
-set(MM_ANDROID_GRADLE_RESOURCES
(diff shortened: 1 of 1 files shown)


kaustuvpokharel:
I would want to keep this pr open, until I am sure, it does work perfectly for any device, and emulator, some of the codes are auto generated with gradle after some configuration tweaks.

kaustuvpokharel:
Works nicely, except for the cmake coding convention isn't matching.

PR #4046, opened 2025-08-22
UPDATED: Preview panel buttons layout update

kaustuvpokharel:


app/qml/CMakeLists.txt
```diff
@@ -114,6 +114,7 @@ set(MM_QML
     form/editors/MMFormTextMultilineEditor.qml
     form/editors/MMFormValueMapEditor.qml
     form/editors/MMFormValueRelationEditor.qml
+    form/editors/MMFormActionBar.qml
```
tomasMizera: Please add the component to the gallery app too

tomasMizera (changes requested):
Visually it looks nice in the app, thanks for the adjustments. I like the API of the new component 👍🏻 it also works with other layer geometry types. The body of the component is quite complicated though. Happy to chat about it to see if we can make it bit more simple.
 
Other comments:
- I can see the following error in the console when opening the preview drawer: `MMFormActionBar.qml:144: ReferenceError: menuColumn is not defined`
- The action button in the popup should be green (forest color), not black
- Also, the "3 dots" icon should be rotated ~ vertical

Image

(Figma design ⬆️ )

tomasMizera:
@kaustuvpokharel you can merge the `dev/2025.6.1` branch here. That will fix the failing iOS build.

app/qml/form/editors/MMFormActionBar.qml
```diff
@@ -0,0 +1,280 @@
...
+  property int  debounceMs: 20          // debounce while resizing- used inside Timer
+
+  // IMPORTANT: “one extra button” worthy of space before expanding to ALL
+  property bool reserveOneMore: true
+
+  readonly property var _visible: actions.filter(a => a && a.visible !== false)
+  readonly property int _count: _visible.length
+
+  readonly property var modeEnum: ({ pair: 0, all: 1 })
+  property int _mode: modeEnum.pair
```
tomasMizera: Turn this into QML Enum

tomasMizera: Like here: https://github.com/MerginMaps/mobile/blob/298ff84079e3a87bfe4f8b28f4bc14a07e4c64ee/app/qml/components/MMPage.qml#L28

app/qml/form/editors/MMFormActionBar.qml
```diff
@@ -0,0 +1,164 @@
...
+    }
+  }
+
+  Row {
+    id: allRow
+    anchors.left: parent.left
+    spacing: __style.margin12
+    // measure content only; don't bind to parent width
+    clip: true
+    height: fitsInline ? allRow.implicitHeight : 0
```
tomasMizera: Why not `visible: fitsInline` instead of setting the height to 0?

app/qml/form/editors/MMFormActionBar.qml
tomasMizera, line 1: Move the file to "app/qml/form/components", it is in "app/qml/form/editors" folder

app/qml/form/editors/MMFormActionBar.qml
```diff
@@ -0,0 +1,164 @@
...
+    parent: hidden
+    x: parent.width - width
+    y: parent.height + __style.margin12
+    width: 155 * __dp
+
+    transformOrigin: Item.TopRight
+
+    contentItem: Column {
+      spacing: 0
+      width: root.popupWidth
```
tomasMizera: This does not exist

app/qml/form/editors/MMFormActionBar.qml
```diff
@@ -0,0 +1,164 @@
...
+
+  // Popup -> everything except the first
+  MMComponents.MMPopup {
+    id: popup
+    parent: hidden
+    x: parent.width - width
+    y: parent.height + __style.margin12
+    width: 155 * __dp
+
+    transformOrigin: Item.TopRight
```
tomasMizera: Popup is not visible if you have the preview drawer of type "image"

Image

app/qml/form/editors/MMFormActionBar.qml
kaustuvpokharel: I tried for few hours could not find a better way, as we are deciding fitsInline from allRow.implicitWidth and while allRow is the thing whose visibility we have to toggle, it will "freeze" the system as there will a no base anchor to switch from or to have a stable value. 
Putting height to 0 was something I reached at while trying, because it still makes the component gone, but just not with visibility, please allow me some suggestions if something can be done for such a case..

tomasMizera (changes requested):

kaustuvpokharel:
Ready for review:
- Dynamic Layout switch between row buttons and pair view("one button" + "more button")
- pop-up doesn't get overlap with the window height; 
- "Open Form" button issue fixed if it is the only existing button incase of readonly layer, for issue #4067 [#4067](https://github.com/MerginMaps/mobile/issues/4067)

Overall features tested video attached:
https://github.com/user-attachments/assets/307da921-b180-46f0-976d-60e87b3d7b67

kaustuvpokharel:
https://github.com/user-attachments/assets/f30f8c28-1bc3-47e0-b692-eb6b39890812

PR #4048, opened 2025-08-23
Fix: Preview panel shows UTC/ISO for datetime - now matches editor (local TZ + display_format)

### LINKED ISSUE #3907: Datetime fields in preview panel are renderd in UTC and don't follow the field's display format
Geopackages store datetimes converted to UTC (https://github.com/MerginMaps/mobile/issues/3890#issuecomment-2880192507).
While in the attribute form time is converted to local timezone before displayed, in the preview panel it is not.

Image

Image

We should do the conversion to local timezone and also respect the display format set in QGIS for the field.

kaustuvpokharel:
Preview panel showed UTC/ISO datetimes while the editor showed local time using the field’s `display_format`.

**Symptom**
* In “Fields” map-tip mode the preview renders raw ISO (often with `Z`).
* The editor drawer shows the same value in **local timezone** with the project’s **`display_format`**.

**Cause**
  * Preview used `attribute(...).toString()` which:
  * Ignores the field’s configured **`display_format`**.
  * Doesn’t normalize UTC/offset values to **LocalTime**.
  * Many attributes arrive as ISO strings or UTC, so `.toString()` leaks raw UTC/ISO.

**Fix**

* Added helper: `QString formatDateForPreview(const QgsField&, const QVariant&, const QVariantMap&)`.
  * Detects `QDate` / `QTime` / `QDateTime`.
  * Parses from `QVariant` or ISO strings (**ISODateWithMs** then **ISODate**).
  * For `QDateTime`, converts to **local time** (`toLocalTime()` + `setTimeSpec(Qt::LocalTime)`).
  * Applies the editor’s **`display_format`** from `QgsEditorWidgetSetup::config()`.
* Replaced both `.toString()` call sites in `mapTipFields()` with the helper.

**Safety / Behavior**
* If parsing fails, falls back to the original `.toString()` to avoid blanks.
* Only date/time fields change; other types unchanged.

**Why this approach**
* Keeps preview and editor perfectly in sync (same format + timezone) without duplicating format logic.

Screenshot 2025-08-23 at 5 38 58 PM
Screenshot 2025-08-23 at 5 39 07 PM

Fixes #3907

### DIFF
--- a/app/test/testattributepreviewcontroller.cpp
+++ b/app/test/testattributepreviewcontroller.cpp
@@ -266,46 +266,46 @@ void TestAttributePreviewController::testFeatureTitle()
 
 void TestAttributePreviewController::testFormatDateForPreview()
 {
-    //Layer with a single datetime field
-    auto *layer = new QgsVectorLayer(
-        QStringLiteral("Point?field=dt:datetime"),
-        QStringLiteral("layer_dt"), QStringLiteral("memory"));
-    QVERIFY(layer && layer->isValid());
+  //Layer with a single datetime field
+  auto *layer = new QgsVectorLayer(
+    QStringLiteral( "Point?field=dt:datetime" ),
+    QStringLiteral( "layer_dt" ), QStringLiteral( "memory" ) );
+  QVERIFY( layer && layer->isValid() );
 
-    //Apply display_format so we can assert a deterministic string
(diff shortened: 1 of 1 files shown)


tomasMizera (changes requested):
- Check if there is a QGIS method that we could use
- Add unit tests
- Change target branch to `dev/2025.7.0`

tomasMizera (approved):

PR #4061, opened 2025-08-30
Bug/empty detail: Fixed- detail dialog only visible if errorMessage Exists, if not it will not be visible.

kaustuvpokharel:
- Added the hasDetails flag to show or hide details component.
- Set detailedText only if error message exists
- Loader is used so details block is created only when needed

### DIFF
--- a/app/qml/dialogs/MMSyncFailedDialog.qml
+++ b/app/qml/dialogs/MMSyncFailedDialog.qml
@@ -23,6 +23,7 @@ MMComponents.MMDrawerDialog {
   id: root
 
   property string detailedText
+  property bool hasDetails: false
 
   title: qsTr( "Failed to synchronise your changes" )
   imageSource: __style.syncFailedImage
@@ -30,9 +31,13 @@ MMComponents.MMDrawerDialog {
   description: qsTr( "Your changes could not be sent to the server, make sure you have a data connection and have permission to edit this project." )
   primaryButton.text: qsTr( "Ok, I understand" )
 
-  additionalContent: MMDialogComponents.MMDialogAdditionalText {
+  additionalContent: Loader {
     width: parent.width
-    text: root.detailedText
+    active: root.hasDetails //status we receive if there is no data connection, either True or False
+    sourceComponent: MMDialogComponents.MMDialogAdditionalText {
+      width: parent.width
+      text: root.detailedText
+    }
   }
 
   onPrimaryButtonClicked: {
--- a/app/qml/main.qml
+++ b/app/qml/main.qml
@@ -959,7 +959,16 @@ ApplicationWindow {
         }
         else
         {
-          syncFailedDialog.detailedText = qsTr( "Details" ) + ": " + errorMessage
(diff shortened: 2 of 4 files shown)


app/qml/dialogs/MMSyncFailedDialog.qml
```diff
@@ -23,16 +23,21 @@ MMComponents.MMDrawerDialog {
   id: root
 
   property string detailedText
+  property bool hasDetails: false
```
PeterPetrik, line 26: Do we need it as additional API flag? wondering if we can't compare agains empty detailedText string ?

kaustuvpokharel: Yeah, that is strictly not needed, but the c++ code also has a "Details:" prefix as a string which is sent to QML as 
"Details: " + errorMessage.
If not used flag, we can have a small function is QML to trim upto that hardcoded "Details:" prefix and see if the message is empty.
But using the flag is cleaner and a bit faster(negligible performance, but if you suggest not using flag we can do that too.

kaustuvpokharel: function hasBody(str) {
  if (!str) return false
  return str.replace(Details:, "").trim().length > 0
}
visible: hasBody(root.detailedText)

//this example simple function inside QML iwould do the same thing, but with flag it would be cleaner, which will also keep QML free of added JS function, but if there is any suggestions please let me know, I would update it which would be preferred.

tomasMizera, line 26:
```suggestion
  property bool hasDetails: !detailedText.isEmpty
```

app/qml/main.qml
```diff
@@ -959,7 +959,16 @@ ApplicationWindow {
...
+          //If errorMessage is empty in this condition, we would send boolean flag "hasDetails" = false
+          //to hide the detail section in MMSyncFailedDialog
+          const hasDetails = !!errorMessage && errorMessage.trim().length > 0
+
+          syncFailedDialog.detailedText = hasDetails
+            ? qsTr("Details") + ": " + errorMessage
+            : ""
+
+          syncFailedDialog.hasDetails = hasDetails
+
```
tomasMizera, line 971:
```suggestion
syncFailedDialog.detailedText = errorMessage
```

app/qml/dialogs/MMSyncFailedDialog.qml
```diff
@@ -23,16 +23,21 @@ MMComponents.MMDrawerDialog {
...
   primaryButton.text: qsTr( "Ok, I understand" )
 
-  additionalContent: MMDialogComponents.MMDialogAdditionalText {
+  additionalContent: Loader {
     width: parent.width
-    text: root.detailedText
+    active: root.hasDetails //status we receive if there is no data connection, either True or False
+    sourceComponent: MMDialogComponents.MMDialogAdditionalText {
+      width: parent.width
+      text: root.detailedText
```
tomasMizera, line 39:
```suggestion
      text: qsTr("Details") + ": " + root.detailedText
```

PR #4066, opened 2025-09-01
Make detail dialog only visible if errorMessage exists

kaustuvpokharel:
Added the hasDetails flag to show or hide details component.
Set detailedText only if error message exists
Loader is used so details block is created only when needed

### DIFF
--- a/app/qml/dialogs/MMSyncFailedDialog.qml
+++ b/app/qml/dialogs/MMSyncFailedDialog.qml
@@ -38,10 +38,6 @@ MMComponents.MMDrawerDialog {
       width: parent.width
       text: qsTr("Details") + ": " + root.detailedText
     }
-    Component.onCompleted:
-    {
-      console.log(detailedText);
-    }
   }
 
   onPrimaryButtonClicked: {
--- a/app/qml/dialogs/MMSyncFailedDialog.qml
+++ b/app/qml/dialogs/MMSyncFailedDialog.qml
@@ -23,10 +23,7 @@ MMComponents.MMDrawerDialog {
   id: root
 
   property string detailedText
-  property bool hasDetails: {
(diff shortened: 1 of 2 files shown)


app/qml/dialogs/MMSyncFailedDialog.qml
```diff
@@ -23,16 +23,28 @@ MMComponents.MMDrawerDialog {
...
-    text: root.detailedText
+    active: root.hasDetails //status we receive if there is no data connection, either True or False
+    sourceComponent: MMDialogComponents.MMDialogAdditionalText {
+      width: parent.width
+      text: qsTr("Details") + ": " + root.detailedText
+    }
+    Component.onCompleted:
+    {
+      console.log(detailedText);
+    }
```
tomasMizera: Debug statement, please remove

app/qml/dialogs/MMSyncFailedDialog.qml
```diff
@@ -23,16 +23,28 @@ MMComponents.MMDrawerDialog {
   id: root
 
   property string detailedText
+  property bool hasDetails: {
+    const t = (root.detailedText ?? "").trim()
+    return t.length > 0
+  }
```
tomasMizera: This should do the same trick, JS will convert the string to boolean, empty strings are falsy, see https://developer.mozilla.org/en-US/docs/Glossary/Falsy

```suggestion
  property bool hasDetails: root.detailedText?.trim() ?? false
```

PeterPetrik:
51 files changed? looks like bad rebase?

kaustuvpokharel:
https://github.com/user-attachments/assets/0010bb74-208a-45f5-a72a-f36654daee25

PR #4081, opened 2025-09-03
Fix temp sketch file not loading

Withalion:
fixes some more bugs found during testing:

> Photos with photo sketches disappear when scrolling in the form on a mobile device. After saving the record, the images with sketches are visible again. You can use project tester-ws-1/test_photo-sketches

> Android only: the orientation of photo change after adding photo sketches. And if you save the record and reopen it in edit mode, the orientation change again after adding new sketches - the photo with sketches rotate 90degrees clockwise. You can use project tester-ws-1/test_photo-sketches

the android bug was a side effect probably

### DIFF
--- a/app/photosketchingcontroller.cpp
+++ b/app/photosketchingcontroller.cpp
@@ -236,7 +236,7 @@ ColorPath PhotoSketchingController::getPath( const int row ) const
 QUrl PhotoSketchingController::getCurrentPhotoPath() const
 {
   const QString photoFileName = QUrl( mPhotoSource ).fileName();
-  const QString photoPath = QString( "%1/%2/%3" ).arg( QDir::temp().absolutePath(), mProjectName, photoFileName );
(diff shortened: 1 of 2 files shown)


app/photosketchingcontroller.cpp
```diff
@@ -123,7 +123,7 @@ void PhotoSketchingController::clear()
 void PhotoSketchingController::backupSketches()
 {
   const QString photoFileName = QUrl( mPhotoSource ).fileName();
-  const QString photoPath = QString( "%1/%2/%3" ).arg( QDir::temp().absolutePath(), mProjectName, photoFileName );
```
tomasMizera, line 126: Shouldn't we update https://github.com/MerginMaps/mobile/pull/4081/files#diff-dbeaddca911ca07c0885e8dc5af9844d0c3dfce4395d5939a9d6dcaefa59677cR253 in the same manner (line 253)?

Withalion: yes there are probably multiple occurrences where I used the older version

PR #4083, opened 2025-09-03
WIP: ThemeDrawer maxHeight changed; testing left for theme lists

kaustuvpokharel:
Only changed the `maxHeight: window.height ` of existing MMMapThemeDrawer, and now it fills the window at max with all the themes, and after filling entirely, we can scroll for the rest of the themes.

Tested as in attached video:

https://github.com/user-attachments/assets/b803d8da-9283-40df-a8b7-6f23dd2f4505

### DIFF
--- a/app/qml/main.qml
+++ b/app/qml/main.qml
@@ -516,7 +516,7 @@ ApplicationWindow {
   MMMapThemeDrawer {
     id: mapThemesPanel
 
-    maxHeight: ( window.height / 2 )
+    maxHeight: window.height
     width: window.width
     edge: Qt.BottomEdge
 


app/qml/main.qml
```diff
@@ -516,7 +516,7 @@ ApplicationWindow {
   MMMapThemeDrawer {
     id: mapThemesPanel
 
-    maxHeight: ( window.height / 2 )
+    maxHeight: window.height
```
tomasMizera: I think you do not even need to define it at all. It has a proper default value set here: https://github.com/MerginMaps/mobile/blob/885f00f3dc791b68e58c2000ed97303b0de0d235/app/qml/components/MMDrawer.qml#L24

PR #4093, opened 2025-09-10
Fix: Unify measurement tools with Figma

kaustuvpokharel:
Fixed:
-Explicit value assigned for MMButton-"Undo btn" to match the requirement
- padding changed with conditionals to keep the old implementation of Tertiary type while, also enabling the height changed with size changed for MMButton to align with figma design.

### DIFF
--- a/app/qml/components/MMButton.qml
+++ b/app/qml/components/MMButton.qml
@@ -165,8 +165,19 @@ Button {
   implicitHeight: root.type === MMButton.Types.Tertiary ? buttonContent.height : buttonContent.height + topPadding + bottomPadding
   implicitWidth: row.paintedChildrenWidth + 2 * ( root.size === MMButton.Sizes.Small ? __style.margin16 : __style.margin20 )
 
-  topPadding: ( root.type === MMButton.Types.Tertiary ) ? 0 : 11 * __dp
-  bottomPadding: ( root.type === MMButton.Types.Tertiary ) ? 0 : 11 * __dp
+  topPadding: {
+    if ( root.type === MMButton.Types.Tertiary )
+      return 0
+    else
+      return ( root.size === MMButton.Sizes.Small ) ? 7 * __dp : 11 * __dp
+  }
+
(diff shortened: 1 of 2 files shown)


app/qml/components/MMButton.qml
```diff
@@ -165,8 +165,19 @@ Button {
   implicitHeight: root.type === MMButton.Types.Tertiary ? buttonContent.height : buttonContent.height + topPadding + bottomPadding
   implicitWidth: row.paintedChildrenWidth + 2 * ( root.size === MMButton.Sizes.Small ? __style.margin16 : __style.margin20 )
 
-  topPadding: ( root.type === MMButton.Types.Tertiary ) ? 0 : 11 * __dp
-  bottomPadding: ( root.type === MMButton.Types.Tertiary ) ? 0 : 11 * __dp
+  topPadding: {
+    if ( root.type === MMButton.Types.Tertiary )
+      return 0
+    else
+      return ( root.size === MMButton.Sizes.Small ) ? 7 * __dp : 11 * __dp
```
Withalion: I wouldn't use ternary operator inside if-else chain

app/qml/components/MMButton.qml
```diff
@@ -165,8 +165,19 @@ Button {
...
+  }
+
+  bottomPadding: {
+    if ( root.type === MMButton.Types.Tertiary )
+      return 0
+    else
+      return ( root.size === MMButton.Sizes.Small ) ? 7 * __dp : 11 * __dp
```
Withalion: same as above

kaustuvpokharel: 7px matches the height of the button which is 40px vertical height in total for the button, as same in the figma. The text size inside the button might have affected the one px difference in padding.

@tomasMizera let me know if I must change the text pixel size for the button when type: small for MMButton

Withalion: Okay, alright. No worries then. Let's keep it as it is.

PR #4100, opened 2025-09-11
Fix missing photosketches after editing geometry

Withalion:
This fixes a  bug where photo sketches were missing after users finished editing geometry. This bug only occurred on photo form components out of view when starting editing geometry.

### DIFF
--- a/app/qml/form/editors/MMFormPhotoEditor.qml
+++ b/app/qml/form/editors/MMFormPhotoEditor.qml
@@ -93,7 +93,12 @@ MMFormPhotoViewer {
   photoUrl: internal.tempSketchedImageSource ? internal.tempSketchedImageSource : internal.resolvedImageSource
   hasCameraCapability: __androidUtils.isAndroid || __iosUtils.isIos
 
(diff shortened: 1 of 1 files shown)


app/qml/form/editors/MMFormPhotoEditor.qml
```diff
@@ -93,7 +93,12 @@ MMFormPhotoViewer {
   photoUrl: internal.tempSketchedImageSource ? internal.tempSketchedImageSource : internal.resolvedImageSource
   hasCameraCapability: __androidUtils.isAndroid || __iosUtils.isIos
 
-  on_FieldValueChanged: internal.setImageSource()
+  on_FieldValueChanged: function (){
```
tomasMizera, line 96: Just a note, it does not have to be an explicit `function` if you do not accept any arguments.

```suggestion
  on_FieldValueChanged: {
```

Withalion: Oh, yeah. I think I used this signature for better support with QML language server, but looks like there is no difference.

PR #4102, opened 2025-09-11
Remove project loading flag few seconds after project load

### LINKED ISSUE #4069: The app falls into crashes loop when crash happens after project was loaded (on iOS)
If the app crashes after a project was loaded it doesn't use the logic of "do not open the project that crashed before when starting the app".

It was discovered during handling https://github.com/MerginMaps/mobile/issues/4068 on iOS devices.

tomasMizera:
So far we had a logic that prevented auto-opening of a project that failed to load on the last app run (crashed the app during project load). There was a case when the app crashed not during project load, but directly after that when trying to render first items.

The fix is to remove the loading flag only after a few seconds after the project is loaded. This helps with crashes like in #4069.

Fixes #4069

### DIFF
--- a/app/activeproject.cpp
+++ b/app/activeproject.cpp
@@ -26,6 +26,7 @@
 #endif
 
 const QString ActiveProject::LOADING_FLAG_FILE_PATH = QString( "%1/.input_loading_project" ).arg( QStandardPaths::standardLocations( QStandardPaths::TempLocation ).first() );
+const int ActiveProject::LOADING_FLAG_FILE_EXPIRATION_MS = 5000;
 
 ActiveProject::ActiveProject( AppSettings &appSettings
                               , ActiveLayer &activeLayer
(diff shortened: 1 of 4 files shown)


app/activeproject.cpp
```diff
@@ -197,7 +198,13 @@ bool ActiveProject::forceLoad( const QString &filePath, bool force )
 
   bool foundErrorsInLoadedProject = validateProject();
 
-  flagFile.remove();
+  // Remove the loading flag file after a while, in case the app crashes not during load, but during the first renderings
+  QTimer::singleShot( LOADING_FLAG_FILE_EXPIRATION_MS, this, []()
+  {
+    QFile::remove( ActiveProject::LOADING_FLAG_FILE_PATH );
+    CoreUtils::log( QStringLiteral( "Project loading" ), QStringLiteral( "Removed project loading flag" ) );
+  } );
```
Withalion, line 206:
```suggestion
  QTimer::singleShot( LOADING_FLAG_FILE_EXPIRATION_MS, this, []
  {
    QFile::remove( LOADING_FLAG_FILE_PATH );
    CoreUtils::log( QStringLiteral( "Project loading" ), QStringLiteral( "Removed project loading flag" ) );
  } );
```

tomasMizera: Well, `LOADING_FLAG_FILE_PATH` is not captured in the lambda's context. Would this be still valid?

Withalion: Yes, however the lambda lives in the scope of class so you don't need to specify the class and just access it directly.

PR #4105, opened 2025-09-15
Fix replacing photos with renaming expression set

### LINKED ISSUE #3600: Photo naming expression fails if replacing a photo with the same filename
Replacing a photo on a field that uses a photo naming expression results in the photo not being renamed if the resulting filename is the same as the file being replaced. This can cause issues in cases where a user needs to change the photo attached to a feature where a file naming expression is being used.

To recreate the issue, you can use [this project](https://app.merginmaps.com/projects/support/photo_name_expression/tree). 

1. Create a project that has a photo attachment field in the survey layer and at least one other field that will have a human readable name with a unique constraint (in this example it is the `"name"` field).
2. In _Project>Properties>Mergin_ set an expression that will result in a file name that is unique to the feature but not necessarily unique if another photo is taken for the feature. In the example project we use: `to_date(now())||'_'||"name"||'-photo1'`
3. Sync the project from QGIS and download it in the mobile app
4. Record a new feature and attach a photo in the photo widget then save the feature
5. Go to _More>Local changes_ to check that the correct photo name has been used for the photo
6. Select the feature on the map and click 'Edit'
7. Delete the existing photo and take a new one
8. Save the feature and go to _More>Local changes_ and the new photo should not have the correct photo name

Withalion:
fixes #3600

When user is editing feature and tries to replace photo which has expression set to rename the file name to the same as previous photo this action fails. We fix this with adding `.deleted` suffix to marked photo. This suffix gets removed on cancel.

### DIFF
--- a/app/qml/form/editors/MMFormPhotoEditor.qml
+++ b/app/qml/form/editors/MMFormPhotoEditor.qml
@@ -127,7 +127,9 @@ MMFormPhotoViewer {
 
     onDeleteImage: {
       // schedule the image for deletion
-      internal.imageSourceToDelete = imageDeleteDialog.imagePath
+      const deletedPath = imageDeleteDialog.imagePath + ".deleted"
+      __inputUtils.renameFile( imageDeleteDialog.imagePath, deletedPath )
+      internal.imageSourceToDelete = deletedPath
       resetValueAndClose()
     }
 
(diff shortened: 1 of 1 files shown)


kaustuvpokharel (approved):

Withalion:
After some more thought this will require more work. Mainly the logic needs to move out of QML, which is in line with our idealogy and most importantly current solution corrupts data. If user scrolls the photo widget outside of view and scrolls back there is no recovery and the image is lost from the perspective of the app.

PR #4109, opened 2025-09-17
Fix photo sketching yet again

Withalion:
- Fixes rotation bug found during testing on samsung device
- Fixes clearing temporary sketched file when deleting file and also clears any saved sketches saved for that photo in session (when you load a photo, sketch it, remove the photo and load it again, there won't be any leftovers )


tomasMizera (changes requested):
The plan:
- catch errors on unsupported files - png, ...
- remove temp path if metadata could not be copied
- add support for png/heic (in 2025.7.0)
- show notification if sketches could not be saved

PR #4119, opened 2025-09-25
Feature/delete multiple feature

kaustuvpokharel:


app/qml/main.qml
```diff
@@ -655,13 +655,25 @@ ApplicationWindow {
...
       onSelectionFinished: {
         multiSelectPanelLoader.active = false
         map.finishMultiSelect()
       }
     }
   }
 
+  MMFormDeleteFeatureDialog {
+    id: deleteDialog
+
```
tomasMizera, line 671: Override the description here

app/multieditmanager.h
```diff
@@ -44,6 +44,7 @@ class MultiEditManager : public QObject
 
     //! Returns a FeatureLayerPair with the feature in mTempLayer
     Q_INVOKABLE FeatureLayerPair editableFeature();
+    Q_INVOKABLE FeatureLayerPair deleteFeature();
```
tomasMizera:
```suggestion
    Q_INVOKABLE FeatureLayerPair deleteSelectedFeatures();
```

kaustuvpokharel:
image

kaustuvpokharel:
https://github.com/user-attachments/assets/3e39e936-ec7e-44de-8c40-3ef491b04c77

app/qml/main.qml
kaustuvpokharel: Done

app/multieditmanager.cpp
```diff
@@ -152,6 +153,39 @@ FeatureLayerPair MultiEditManager::editableFeature()
...
+    return;
+  }
+
+  if ( !mLayer->startEditing() )
+  {
+    CoreUtils::log( QStringLiteral( "Multi Edit Manager" ), QStringLiteral( "Could not start editing on layer %1" ).arg( mLayer->name() ) );
+    return;
+  }
+
+  bool success = mLayer->deleteFeatures( fids );
```
Withalion:
```suggestion
  const bool success = mLayer->deleteFeatures( fids );
```

app/multieditmanager.cpp
```diff
@@ -152,6 +153,39 @@ FeatureLayerPair MultiEditManager::editableFeature()
   return FeatureLayerPair( oneFeature, mTempLayer.get() );
 }
 
+void MultiEditManager::deleteSelectedFeature()
+{
+  if ( !mModel || mModel->count() == 0 || !mLayer )
```
Withalion, line 158: The `model->count()` should never be zero, the user should not be able to execute this function...

app/multieditmanager.cpp
```diff
@@ -152,6 +153,39 @@ FeatureLayerPair MultiEditManager::editableFeature()
...
+    CoreUtils::log( QStringLiteral( "Multi Edit Manager" ), QStringLiteral( "Could not start editing on layer %1" ).arg( mLayer->name() ) );
+    return;
+  }
+
+  bool success = mLayer->deleteFeatures( fids );
+  if ( success )
+  {
+    mModel->populate( {} );
+    mLayer->triggerRepaint();
+    CoreUtils::log( QStringLiteral( "Multi Edit Manager" ), QStringLiteral( "Deleted %1 features from %2" ).arg( fids.size() ).arg( mLayer->name() ) );
```
Withalion, line 184:
```suggestion
    CoreUtils::log( QStringLiteral( "Multi Edit Manager" ), QStringLiteral( "Deleted %1 features from %2" ).arg( fids.size(), mLayer->name() ) );
```

tomasMizera:
> I still believe we can do better than using `(s)` in the text, maybe propagate through the signal if it's single or multi edit and then use ternary operator inline.

https://doc.qt.io/qt-6/i18n-source-translation.html#handle-plural-forms

app/multieditmanager.cpp
kaustuvpokharel: It doesn't have a multi param for a .arg(), and safest is what already is used one.

app/multieditmanager.cpp
kaustuvpokharel: I guess that's what it fulfills here, the function right on top it uses this same flags to not do anything and return if the model has 0 features.
Or am I mistaking what you are trying to say ?

Withalion: I just wanted to mention that the check is unnecessary, but you can leave it there

app/multieditmanager.cpp
Withalion: I thought it had, interesting... Anyway I don't think `QStringLiteral` here gives you any of the benefits it should, you can use plain `QString` here

Withalion (approved):

app/test/testmultieditmanager.h
```diff
@@ -0,0 +1,20 @@
...
+ *                                                                         *
+ ***************************************************************************/
+#include <QObject>
+
+#ifndef TESTMULTIEDITMANAGER_H
+#define TESTMULTIEDITMANAGER_H
+
+class TestMultiEditManager : public QObject {
+    Q_OBJECT
+  private slots:
```
tomasMizera, line 30: you are missing the required methods
```c++
  private slots:
    void init();
    void cleanup();
```

app/test/inputtests.cpp
```diff
@@ -29,6 +29,8 @@
 #include "testlayertree.h"
 #include "testactiveproject.h"
 #include "testprojectchecksumcache.h"
+#include "multieditmanager.h"
```
Withalion:
```suggestion

```

app/test/testmultieditmanager.cpp
```diff
@@ -0,0 +1,93 @@
...
+#include <memory>
+
+#include "multieditmanager.h"
+#include "featurelayerpair.h"
+
+#include "qgsvectorlayer.h"
+#include "qgsfeature.h"
+#include "qgsfields.h"
+#include "qgsgeometry.h"
+#include "qgspoint.h"
```
Withalion:
```suggestion

```

app/test/testmultieditmanager.h
```diff
@@ -0,0 +1,46 @@
...
+#ifndef TESTMULTIEDITMANAGER_H
+#define TESTMULTIEDITMANAGER_H
+
+#include <QObject>
+
+#include <QObject>
+#include <memory>
+#include <QVector>
+
+#include "qgsfeature.h"
```
Withalion:
```suggestion
#include <memory>

#include "qgsfeature.h"
```

app/test/testmultieditmanager.h
```diff
@@ -0,0 +1,46 @@
...
+
+class QgsVectorLayer;
```
Withalion: this is probably nitpick, but I would rather include QGIS classes than forward declare as they don't come from us

Withalion: looks like you can delete it altogether

app/test/testmultieditmanager.cpp
```diff
@@ -0,0 +1,93 @@
...
+*                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+#include "testmultieditmanager.h"
+
+#include <QtTest/QtTest>
```
Withalion: [Why we should not do this](https://www.kdab.com/beware-of-qt-module-wide-includes/)

app/test/testmultieditmanager.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+#include "qgsvectorlayer.h"
+#include "qgsfeature.h"
+#include "qgsfields.h"
+#include "qgsgeometry.h"
+#include "qgspoint.h"
+
+void TestMultiEditManager::init()
+{
+  mLayer.reset( new QgsVectorLayer( QStringLiteral( "point?field=id:integer" ), QStringLiteral( "mem" ), QStringLiteral( "memory" ) ) );
```
Withalion:
```suggestion
  mLayer = std::make_unique<QgsVectorLayer> ( QStringLiteral( "point?field=id:integer" ), QStringLiteral( "mem" ), QStringLiteral( "memory" ) ) );
```

app/test/testmultieditmanager.cpp
```diff
@@ -0,0 +1,93 @@
...
+#include "multieditmanager.h"
+#include "featurelayerpair.h"
+
+#include "qgsvectorlayer.h"
+#include "qgsfeature.h"
+#include "qgsfields.h"
+#include "qgsgeometry.h"
+#include "qgspoint.h"
+
+void TestMultiEditManager::init()
```
Withalion, line 11: Maybe I would move some logic out of here to `initTestCase()` so we won't destroy and recreate everything after each function.

app/qml/main.qml
```diff
@@ -654,13 +654,37 @@ ApplicationWindow {
...
+    property int countToDelete: 0
+
+    title: (countToDelete > 1) ? qsTr( "Delete features" ) : qsTr( "Delete feature" )
+    description: (countToDelete > 1) ? qsTr( "Are you sure you want to delete selected features?" ) : qsTr( "Are you sure you want to delete selected feature?" )
```
Withalion: After returning back to translations, we should use the plural form as mention by @tomasMizera 
```qml
Text {
    text: qsTr("%n message(s) saved", "", total)
}
```

app/test/testmultieditmanager.cpp
tomasMizera: Nice 💡 thanks for sharing

app/qml/main.qml
```diff
@@ -672,7 +672,10 @@ ApplicationWindow {
     property int countToDelete: 0
 
     title: (countToDelete > 1) ? qsTr( "Delete features" ) : qsTr( "Delete feature" )
-    description: (countToDelete > 1) ? qsTr( "Are you sure you want to delete selected features?" ) : qsTr( "Are you sure you want to delete selected feature?" )
+    description: (countToDelete > 1)
+        ? qsTr("Delete %n selected features?", "", countToDelete)
+        : qsTr("Delete selected feature?")
+
```
```suggestion
    description: qsTr("Delete %n selected feature(s)?", "", countToDelete)
```

app/test/testmultieditmanager.cpp
Withalion: I'm okay with keeping this as it is, but it would be nice if we would keep the layer alive and just repopulate it when necessary.

app/multieditmanager.h
```diff
@@ -45,6 +45,9 @@ class MultiEditManager : public QObject
     //! Returns a FeatureLayerPair with the feature in mTempLayer
     Q_INVOKABLE FeatureLayerPair editableFeature();
 
+    //! Deletes selected or toggled features upon called
+    Q_INVOKABLE void deleteSelectedFeature();
```
tomasMizera:
```suggestion
    Q_INVOKABLE void deleteSelectedFeatures();
```

app/multieditmanager.cpp
```diff
@@ -152,6 +153,39 @@ FeatureLayerPair MultiEditManager::editableFeature()
...
+  if ( !mModel || mModel->count() == 0 || !mLayer )
+    return;
+
+  const FeatureLayerPairs pairs = mModel->features();
+  QgsFeatureIds fids;
+  fids.reserve( pairs.count() );
+  for ( const FeatureLayerPair &pair : pairs )
+  {
+    fids.insert( pair.feature().id() );
+  }
```
tomasMizera: I think we could save a bit of memory if we read the data directly from the model instead of copying them into `pairs`. Something like this should work in theory:

```suggestion

  QgsFeatureIds fids;
  fids.reserve( mModel->rowCount() );
  for ( int i = 0; mModel->rowCount(); i++ )
  {
    fids.insert( mModel->data( index( i ), FeaturesModel::FeatureId );
  }
```

It's quite a shame that we need to extract feature IDs in the first place. I guess for future it would be better if the model only included feature Ids and requested additional attributes when needed. Something like what @uclaros started to do in https://github.com/MerginMaps/mobile/pull/4035

app/qml/gps/MMSelectionDrawer.qml
```diff
@@ -102,6 +103,21 @@ MMComponents.MMDrawer {
         iconSourceLeft: __style.editIcon
         onClicked: editSelected()
       }
+
+      MMButton {
+        text: qsTr( "Delete" )
```
tomasMizera: Let's hope screens are wide enough to incorporate both buttons without the need to scroll (in all languages 🙃)

app/qml/main.qml
```diff
@@ -654,13 +654,38 @@ ApplicationWindow {
...
     }
   }
 
+  MMFormDeleteFeatureDialog {
+    id: deleteDialog
+
+    property int countToDelete: 0
+
+    title: (countToDelete > 1) ? qsTr( "Delete features" ) : qsTr( "Delete feature" )
+    description: qsTr("Delete %n selected feature(s)?", "", countToDelete)
```
tomasMizera: Why not using the same plural format for the `title` too? 🤔 I'd expect this to work

```suggestion
    title: qsTr( "Delete feature(s)", "", countToDelete )
    description: qsTr("Delete %n selected feature(s)?", "", countToDelete)
```

gabriel-bolbotina:
I have added a scrollbar for the button text which might be larger than in English.

Future-proof in case of adding multiple functionality buttons.

Some use case scenarios are presented in [these videos](https://drive.google.com/drive/folders/1k7q011fL99yy06dmmTMcaloWHQU3tpMf?usp=share_link)

app/qml/main.qml
```diff
@@ -654,13 +654,38 @@ ApplicationWindow {
...
       }
     }
   }
 
+  MMFormDeleteFeatureDialog {
+    id: deleteDialog
+
+    property int countToDelete: 0
+
+    title: qsTr( "Delete feature(s)", "", countToDelete )
```
tomasMizera, line 674: Have you checked if the plural form works properly even without `%n`? :)

PR #4131, opened 2025-10-05
Fix ios crash not crash

### LINKED ISSUE #4124: Incorrect Crash message displayed after App restart
**Description**
After restarting the Mergin Maps app, a system message appears stating that the app has crashed. However, the app continues to function normally and does not actually crash. 

**Environment**
Staging

Mobile app version: 2025.7.0
Build (25.9.804311)
Operating system: IOS 26.0

**Steps to reproduce**
1. Launch the Mergin Maps application
2. Trigger a restart
3. Observe the message indicating a crash, even though the app remains operational. 

**Actual results**
A crash notification appears after restart, despite the app continuing to run without issues.

**Expected results**
No crash message should be displayed unless the app has actually stopped.

**Screenshots**

Withalion:
fixes #4124 

The issue was in the double deletion of `QQmlEngine` which threw exception on ios. This was caused by previous fix of qml component deletion flow on application flow. The current approach fixes both things.

### DIFF
--- a/app/main.cpp
+++ b/app/main.cpp
@@ -573,9 +573,6 @@ int main( int argc, char *argv[] )
     CoreUtils::log( QStringLiteral( "AppState" ), QStringLiteral( "Application has quit" ) );
   } );
 
-  // this fixes the error dump from C++ defined QML components, when quiting application
-  QObject::connect( &app, &QCoreApplication::aboutToQuit, &engine, &QQmlEngine::deleteLater );
-
   QObject::connect( &help, &InputHelp::submitReportSuccessful, &lambdaContext, [&notificationModel]()
   {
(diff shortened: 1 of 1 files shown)


app/main.cpp
```diff
@@ -869,6 +866,12 @@ int main( int argc, char *argv[] )
     return EXIT_FAILURE;
   }
 
+  // this fixes the error dump from C++ defined QML components, when quiting application
+  QObject::connect(&app, &QCoreApplication::aboutToQuit, [&] {
+        object->deleteLater();
```
kaustuvpokharel: Nice, How we we even find bug like these, I would have never noticed this bug. Because this does seem like a normal engine exit, because I never knew if you do deleteLater in QQMlEngine it double calls on the stack.

kaustuvpokharel (approved):

PR #4132, opened 2025-10-07
Add empty state for value map/relation

kaustuvpokharel:
Fixed:
Component was missing emptyDelegate condition, which not fixed with "no items" text as a emptyDelegateState when no features are present on the list.
Added padding to match figma design
Added drawer bottom margin

Screenshot 2025-10-07 at 11 59 31 PM

### DIFF
--- a/app/qml/components/MMListMultiselectDrawer.qml
+++ b/app/qml/components/MMListMultiselectDrawer.qml
@@ -66,7 +66,6 @@ MMDrawer {
       MMListSpacer { id: searchBarSpacer; height: __style.spacing20; visible: root.withSearch }
 
       Item {
-        id: featureList
         width: parent.width
         height: listViewComponent.count === 0 ? emptyStateDelegateLoader.height : listViewComponent.height
 
--- a/app/qml/form/editors/MMFormValueRelationEditor.qml
+++ b/app/qml/form/editors/MMFormValueRelationEditor.qml
@@ -73,14 +73,12 @@ MMFormComboboxBaseEditor {
 
       emptyStateDelegate: Item {
         width: parent.width
-        height: noItemsText.implicitHeight
-        anchors.centerIn: parent
-
+        height: noItemsText.implicitHeight + __style.margin40
+      
         MMComponents.MMText {
           id: noItemsText
           text: qsTr( "No items" )
           anchors.centerIn: parent
-          topPadding: __style.margin20
         }
       }
 
--- a/app/qml/form/editors/MMFormValueMapEditor.qml
+++ b/app/qml/form/editors/MMFormValueMapEditor.qml
@@ -87,14 +87,12 @@ MMFormComboboxBaseEditor {
 
       emptyStateDelegate: Item {
         width: parent.width
-        height: noItemsText.implicitHeight
-        anchors.centerIn: parent
-
+        height: noItemsText.implicitHeight + __style.margin40
+      
         MMComponents.MMText {
           id: noItemsText
           text: qsTr( "No items" )
           anchors.centerIn: parent
-          topPadding: __style.margin20
         }
       }
 


app/qml/components/MMListMultiselectDrawer.qml
```diff
@@ -33,11 +33,15 @@ MMDrawer {
 
   interactive: !listViewComponent.interactive
 
-  drawerBottomMargin: 0
+  drawerBottomMargin: __style.margin20 + __style.safeAreaBottom
```
tomasMizera: This is not something we want.. you would end up with an empty white line for value map and value relation editors. No matter if they have or do not have a text; see https://www.figma.com/design/d78a4ehBBgK6MyxCTuCAqg/Mobile-App?node-id=5305-35002&t=N5ShYBn2G8IIsByh-4

app/qml/components/MMListMultiselectDrawer.qml
```diff
@@ -64,21 +68,20 @@ MMDrawer {
...
-
           enabled: contentHeight > height
 
           Loader {
             id: emptyStateDelegateLoader
 
             visible: listViewComponent.count === 0
-
-            width: parent.width
+            anchors.horizontalCenter: parent.horizontalCenter
```
tomasMizera: Again, you might be breaking existing components - you should set the `emptyStateDelegate` item (the `sourceComponent` of the loader) to use the full width of the drawer and then center text within itself

app/qml/components/MMListMultiselectDrawer.qml
tomasMizera: The correct way would be to add the bottom margin only for the empty state, below `emptyStateDelegateLoader`

app/qml/components/MMListMultiselectDrawer.qml
```diff
@@ -33,11 +33,15 @@ MMDrawer {
...
+  drawerBottomMargin: __style.margin20 + __style.safeAreaBottom
 
   drawerContent: Item {
     width: parent.width
-    height: showFullScreen ? root.drawerContentAvailableHeight : contentLayout.height
+    height: {
+      if ( root.showFullScreen ) return root.drawerContentAvailableHeight
+      if ( listViewComponent.count === 0 ) return emptyStateDelegateLoader.height
+      return listViewComponent.height
+    }
```
tomasMizera: This is not a change we want, heights work just fine afaik

app/qml/form/editors/MMFormValueRelationEditor.qml
```diff
@@ -69,9 +69,15 @@ MMFormComboboxBaseEditor {
   dropdownLoader.sourceComponent: Component {
 
     MMComponents.MMListMultiselectDrawer {
-
       drawerHeader.title: root._fieldTitle
 
+      //Incase when list is empty (ex: no relation value fields)
```
tomasMizera: Misleading comment
```suggestion
```

app/qml/form/editors/MMFormValueRelationEditor.qml
```diff
@@ -69,9 +69,15 @@ MMFormComboboxBaseEditor {
...
+      emptyStateDelegate: MMComponents.MMText {
+        text: qsTr( "No items" )
+        topPadding: __style.margin20
+        bottomPadding: __style.margin20
+      }
```
tomasMizera: This is in fact the only code we need :) 

I'd suggest we wrap it into an Item to fill in the space of the drawer, something like 
```qml
    emptyStateDelegate: Item {
      width: parent.width
      height: 100

      MMComponents.MMText {
        text: "No items"
        anchors.centerIn: parent
        ...
      }
    }
```

tomasMizera: And this same thing must go to the ValueMap too..

app/qml/components/MMListMultiselectDrawer.qml
```diff
@@ -64,13 +66,13 @@ MMDrawer {
       MMListSpacer { id: searchBarSpacer; height: __style.spacing20; visible: root.withSearch }
 
       Item {
+        id: featureList
```
tomasMizera: Unneeded id

app/qml/form/editors/MMFormValueRelationEditor.qml
```diff
@@ -69,9 +69,21 @@ MMFormComboboxBaseEditor {
   dropdownLoader.sourceComponent: Component {
 
     MMComponents.MMListMultiselectDrawer {
-
       drawerHeader.title: root._fieldTitle
 
+      emptyStateDelegate: Item {
+        width: parent.width
+        height: noItemsText.implicitHeight
+        anchors.centerIn: parent
```
tomasMizera: We should not center to parent as the parent's size depends on the height of this component 😵‍💫
```suggestion

```

app/qml/components/MMListMultiselectDrawer.qml
```diff
@@ -33,7 +33,9 @@ MMDrawer {
 
   interactive: !listViewComponent.interactive
 
-  drawerBottomMargin: 0
+  drawerBottomMargin: listViewComponent.count === 0
+                      ? (__style.margin20 + __style.safeAreaBottom)
+                      : 0
```
tomasMizera, line 38:
```suggestion

```

app/qml/components/MMListMultiselectDrawer.qml
```diff
@@ -70,7 +72,6 @@ MMDrawer {
         MMScrollView {
           width: parent.width
           height: Math.min( contentHeight, root.drawerContentAvailableHeight - internal.searchBarVerticalSpace )
-
           enabled: contentHeight > height
 
           Loader {
```
tomasMizera, line 77: Add spacer below `Loader` to handle safeArea

tomasMizera (changes requested):

app/qml/components/MMListMultiselectDrawer.qml
gabriel-bolbotina: If we remove this, it no longer shows the message of No items

app/qml/components/MMListMultiselectDrawer.qml
gabriel-bolbotina: It works without it and when I added it, the message of  "no items" no longer appears. I would keep it like this

app/qml/components/MMListMultiselectDrawer.qml
tomasMizera: Still there

app/qml/form/editors/MMFormValueRelationEditor.qml
```diff
@@ -69,9 +69,21 @@ MMFormComboboxBaseEditor {
...
+
+        MMComponents.MMText {
+          id: noItemsText
+          text: qsTr( "No items" )
+          anchors.centerIn: parent
+          topPadding: __style.margin20
+        }
+      }
```
tomasMizera, line 83: Something does not feel right here.

The top `Item`'s size is defined to take the full width of its parent and height of its only children. `anchors.centerIn` won't do anything then.

The second thing is that the inner `MMText` defines `topPadding: __style.margin20` while the parent's height is set to be the same as `MMText`'s height. What this makes is that it pushes the text outside of the parent size, something like this:

Image

I guess that is a compensation for using `__style.margin20` in `drawerBottomMargin` of `MMListMultiselectDrawer.qml‎`. I'd suggest we drop this and instead make the top `Item` bigger. Something like:


```suggestion
      emptyStateDelegate: Item {
        width: parent.width
        height: noItemsText.implicitHeight + __style.margin40

        MMComponents.MMText {
          id: noItemsText
          text: qsTr( "No items" )
          anchors.centerIn: parent
        }
      }
```

tomasMizera (changes requested):

PR #4138, opened 2025-10-08
Bugfixes/2025.6.0

kaustuvpokharel:
Fixed spacing for the multiselect drawer
Fixed the button color binding with the popup status

### DIFF
--- a/app/qml/gps/MMSelectionDrawer.qml
+++ b/app/qml/gps/MMSelectionDrawer.qml
@@ -41,8 +41,6 @@ MMComponents.MMDrawer {
   closePolicy: Popup.CloseOnEscape
 
   dropShadow: true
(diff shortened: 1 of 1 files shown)


app/qml/gps/MMSelectionDrawer.qml
```diff
@@ -41,7 +41,8 @@ MMComponents.MMDrawer {
   closePolicy: Popup.CloseOnEscape
 
   dropShadow: true
-  drawerBottomMargin: __style.margin20 + __style.safeAreaBottom
+  //We do not required bottom Margin for drawer here: uncomment if truely needed
+  //drawerBottomMargin: __style.margin20 + __style.safeAreaBottom
```
PeterPetrik: if we do not need it , remove if.

PR #4145, opened 2025-10-10
scrollbar for desktop builds in feature lists

### LINKED ISSUE #4127: Add scrollbar for desktop builds in lists
Lists like on the following screenshot are missing scrollbar on desktop builds. Let's add them to all lists.

See the original report: https://merginmaps.slack.com/archives/C02DWECUEN8/p1759318081188029

kaustuvpokharel:
Case:  Missing scrollbar for the desktop version for all the ListView components.

What has be been done:
Added slim scrollbar on the right side of the lists. 
Scroll bar only activates for desktops; ex(macos, window, linux) and not any mobile builds
Scrollbar always visibile, but highlights on active more and passive on inactive.

**Tested for mutiple page which uses ListView and the work of scrollbar**

https://github.com/user-attachments/assets/9012e551-c937-40b7-a732-eedef8f9e79b

### DIFF
--- a/app/qml/components/MMListDelegate.qml
+++ b/app/qml/components/MMListDelegate.qml
@@ -24,6 +24,7 @@ Item {
   property string text
   property string secondaryText
 
+  readonly property int scrollSpace: (ListView.view && ListView.view.scrollSpace) ? ListView.view.scrollSpace : 0
   property alias leftContent: leftContentGroup.children
   property alias rightContent: rightContentGroup.children
 
@@ -40,9 +41,10 @@ Item {
 
   property real verticalSpacing: root.secondaryText ? __style.margin8 : __style.margin20
 
-  implicitWidth: ListView?.view?.width ?? 0 // in case ListView is injected as attached property (usually it is)
+  implicitWidth: ListView?.view?.width ?? 0  // in case ListView is injected as attached property (usually it is)
   implicitHeight: contentLayout.implicitHeight
   height: visible ? implicitHeight : 0.1 // hide invisible items, for some reason setting 0 does not work ¯\_(ツ)_/¯
+  width: implicitWidth - root.scrollSpace
 
   MouseArea {
     anchors.fill: contentLayout
--- a/app/qml/components/MMListDelegate.qml
+++ b/app/qml/components/MMListDelegate.qml
@@ -24,7 +24,6 @@ Item {
   property string text
   property string secondaryText
 
-  readonly property int scrollSpace: (ListView.view && ListView.view.scrollSpace) ? ListView.view.scrollSpace : 0
   property alias leftContent: leftContentGroup.children
   property alias rightContent: rightContentGroup.children
 
@@ -44,7 +43,7 @@ Item {
   implicitWidth: ListView?.view?.width ?? 0  // in case ListView is injected as attached property (usually it is)
   implicitHeight: contentLayout.implicitHeight
   height: visible ? implicitHeight : 0.1 // hide invisible items, for some reason setting 0 does not work ¯\_(ツ)_/¯
-  width: implicitWidth - ListView.view.scrollSpace
+  width: implicitWidth - ( ListView?.view?.scrollSpace ?? 0 )
 
   MouseArea {
     anchors.fill: contentLayout
--- a/app/qml/components/MMListDelegate.qml
+++ b/app/qml/components/MMListDelegate.qml
@@ -40,10 +40,9 @@ Item {
 
   property real verticalSpacing: root.secondaryText ? __style.margin8 : __style.margin20
 
-  implicitWidth: ListView?.view?.width ?? 0  // in case ListView is injected as attached property (usually it is)
+  implicitWidth: ListView?.view?.width - ListView?.view?.scrollBarWidth ?? 0  // in case ListView is injected as attached property (usually it is)
   implicitHeight: contentLayout.implicitHeight
   height: visible ? implicitHeight : 0.1 // hide invisible items, for some reason setting 0 does not work ¯\_(ツ)_/¯
-  width: implicitWidth - ( ListView?.view?.scrollSpace ?? 0 )
 
   MouseArea {
     anchors.fill: contentLayout
--- a/app/qml/components/MMListDelegate.qml
+++ b/app/qml/components/MMListDelegate.qml
@@ -44,7 +44,7 @@ Item {
   implicitWidth: ListView?.view?.width ?? 0  // in case ListView is injected as attached property (usually it is)
   implicitHeight: contentLayout.implicitHeight
   height: visible ? implicitHeight : 0.1 // hide invisible items, for some reason setting 0 does not work ¯\_(ツ)_/¯
-  width: implicitWidth - root.scrollSpace
+  width: implicitWidth - ListView.view.scrollSpace
 
   MouseArea {
     anchors.fill: contentLayout
--- a/app/qml/components/MMListView.qml
+++ b/app/qml/components/MMListView.qml
@@ -28,6 +28,7 @@ ListView {
     id: verticalScrollBar
 
     policy: isMobile ? ScrollBar.AlwaysOff : ScrollBar.AlwaysOn
-    opacity: active ? 0.7 : 0.2
+    visible: contentHeight > availableHeight ? true  : false
+    opacity: active ? 0.7 : 0.4
   }
 }
--- a/app/qml/components/MMListView.qml
+++ b/app/qml/components/MMListView.qml
@@ -21,14 +21,14 @@ ListView {
 
   cacheBuffer: 0
   readonly property bool isMobile: (Qt.platform.os === "android" || Qt.platform.os === "ios")
-  property int scrollSpace: !isMobile ? 10 : 0
-  property bool showScrollBar: true
+  readonly property int scrollBarWidth: !isMobile ? __style.margin10 : 0
+  property alias showScrollBar: verticalScrollBar.policy
 
-  ScrollBar.vertical: ScrollBar {
+  ScrollBar.vertical: ScrollBar{
     id: verticalScrollBar
 
     policy: isMobile ? ScrollBar.AlwaysOff : ScrollBar.AlwaysOn
-    visible: showScrollBar
+    //visible: !isMobile && (root.contentHeight > root.height)
     opacity: active ? 0.7 : 0.4
 
     contentItem: Rectangle {
--- a/app/qml/components/MMListView.qml
+++ b/app/qml/components/MMListView.qml
@@ -22,13 +22,14 @@ ListView {
   cacheBuffer: 0
   readonly property bool isMobile: (Qt.platform.os === "android" || Qt.platform.os === "ios")
   property int scrollSpace: !isMobile ? 10 : 0
+  property bool showScrollBar: true
 
   ScrollBar.vertical: ScrollBar {
     id: verticalScrollBar
 
-    policy: ScrollBar.AlwaysOn
-    visible: !isMobile && (root.contentHeight > root.height)
-    opacity: (pressed || root.moving) ? 0.7 : 0.4
+    policy: isMobile ? ScrollBar.AlwaysOff : ScrollBar.AlwaysOn
+    visible: showScrollBar
+    opacity: active ? 0.7 : 0.4
 
     contentItem: Rectangle {
       implicitWidth: 5
--- a/app/qml/components/MMListView.qml
+++ b/app/qml/components/MMListView.qml
@@ -20,7 +20,7 @@ ListView {
   id: root
 
   cacheBuffer: 0
-  readonly property int scrollBarWidth: __inputUtils.isMobilePlatform ?  0 : __style.margin10
+  readonly property int scrollBarWidth: __inputUtils.isMobilePlatform() ?  0 : __style.margin10
   property alias scrollBarPolicy: verticalScrollBar.policy
 
   ScrollBar.vertical: ScrollBar{
@@ -30,7 +30,7 @@ ListView {
     // - both 'policy' and 'visible' properties need to be set, with 'policy' set to AlwaysOn.
     // - the use the 'AsNeeded' policy will make the scroll bar hide when not focused or in use,
     // and only appear after the user starts scrolling.
-    policy: __inputUtils.isMobilePlatform ? ScrollBar.AlwaysOff : ScrollBar.AlwaysOn
+    policy: __inputUtils.isMobilePlatform() ? ScrollBar.AlwaysOff : ScrollBar.AlwaysOn
     visible: contentHeight > availableHeight ? true  : false
     opacity: active ? 0.7 : 0.4
     width: scrollBarWidth
--- a/app/qml/components/MMListView.qml
+++ b/app/qml/components/MMListView.qml
@@ -20,15 +20,19 @@ ListView {
   id: root
 
   cacheBuffer: 0
-  readonly property bool isMobile: (Qt.platform.os === "android" || Qt.platform.os === "ios")
-  readonly property int scrollBarWidth: !isMobile ? __style.margin10 : 0
-  property alias showScrollBar: verticalScrollBar.policy
+  readonly property int scrollBarWidth: __inputUtils.isMobilePlatform ?  0 : __style.margin10
+  property alias scrollBarPolicy: verticalScrollBar.policy
 
   ScrollBar.vertical: ScrollBar{
     id: verticalScrollBar
 
-    policy: isMobile ? ScrollBar.AlwaysOff : ScrollBar.AlwaysOn
+    // To make sure the scroll bar stays visible on desktop, even when not focused:
+    // - both 'policy' and 'visible' properties need to be set, with 'policy' set to AlwaysOn.
+    // - the use the 'AsNeeded' policy will make the scroll bar hide when not focused or in use,
+    // and only appear after the user starts scrolling.
+    policy: __inputUtils.isMobilePlatform ? ScrollBar.AlwaysOff : ScrollBar.AlwaysOn
     visible: contentHeight > availableHeight ? true  : false
     opacity: active ? 0.7 : 0.4
+    width: scrollBarWidth
   }
 }
--- a/app/qml/components/MMListView.qml
+++ b/app/qml/components/MMListView.qml
@@ -8,6 +8,7 @@
  ***************************************************************************/
 
 import QtQuick
+import QtQuick.Controls
 
 //
 // Hot-fix for hotfix https://github.com/MerginMaps/mobile/issues/3417
@@ -15,6 +16,27 @@ import QtQuick
 //
 
 ListView {
+  id: root
 
   cacheBuffer: 0
+  readonly property bool isMobile: (Qt.platform.os === "android" || Qt.platform.os === "ios")
+  property int scrollSpace: !isMobile ? 10 : 0
+
+  ScrollBar.vertical: ScrollBar {
+    id: verticalScrollBar
+
+    policy: ScrollBar.AlwaysOn
+    visible: !isMobile && (root.contentHeight > root.height)
+    opacity: (pressed || root.moving) ? 0.7 : 0.4
+
+    anchors.right: root.right
+
+    implicitHeight: root.height
+
+    contentItem: Rectangle {
+      implicitWidth: 5
+      radius: width / 2
+      color: __style.darkGreenColor
+    }
+  }
 }
--- a/app/qml/components/MMListView.qml
+++ b/app/qml/components/MMListView.qml
@@ -9,6 +9,7 @@
 
 import QtQuick
 import QtQuick.Controls
+import QtQml
 
 //
 // Hot-fix for hotfix https://github.com/MerginMaps/mobile/issues/3417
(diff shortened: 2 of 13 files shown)


app/qml/layers/MMFeaturesListPage.qml
```diff
@@ -72,6 +74,24 @@ MMComponents.MMPage {
...
 
+      ScrollBar.vertical: ScrollBar {
+        id: verticalScrollBar
+        policy: isMobile ? ScrollBar.AlwaysOff : ScrollBar.AsNeeded
+        visible: !isMobile
+        interactive: !isMobile
+        anchors.right: listView.right
+        implicitHeight: listView.height
+        opacity: active ? 1.0 : 0.0
+        Behavior on opacity { NumberAnimation { duration: 150 } }
```
Withalion:
```suggestion
        
```
From the customers feedback I would not hide the scrollbar

app/qml/layers/MMFeaturesListPage.qml
```diff
@@ -22,6 +23,7 @@ MMComponents.MMPage {
   property var selectedLayer: null
   property bool hasToolbar: false
   property bool layerIsReadOnly: selectedLayer?.readOnly ?? false
+  readonly property bool isMobile: (Qt.platform.os === "android" || Qt.platform.os === "ios")
```
Withalion: I would move this property into `ScrollBar` to minimize it's scope

app/qml/layers/MMFeaturesListPage.qml
```diff
@@ -72,6 +74,24 @@ MMComponents.MMPage {
         onClicked: root.featureClicked( model.FeaturePair )
       }
 
+      ScrollBar.vertical: ScrollBar {
+        id: verticalScrollBar
+        policy: isMobile ? ScrollBar.AlwaysOff : ScrollBar.AsNeeded
+        visible: !isMobile
+        interactive: !isMobile
+        anchors.right: listView.right
```
Withalion: isn't there API to set this in a clearer way?

app/qml/layers/MMFeaturesListPage.qml
kaustuvpokharel: But usually every scrollbar on the internet has the hide, and it activate immediately with slight scroll.
Open to change it is we need to show scroll bar all the time!

app/qml/layers/MMFeaturesListPage.qml
kaustuvpokharel: Good idea, we could do that

app/qml/layers/MMFeaturesListPage.qml
kaustuvpokharel: I mean, I don't see a better way then this here, it does seem like micro-managed properties but it best here. We could use a singleton QObject to have the logic which is also not the best case, as it will be over kill for this tiny change.
If you had a better idea, I am open to change.

app/qml/layers/MMFeaturesListPage.qml
Withalion: The thing is that the customer that raised this issue was not able to scroll and that's why they wanted the scrollbar. I would keep the scrollbar always visible and just have an effect to change colors when in use and when not.

Withalion (commented):
On second thought, it would make more sense to move the scrollbar directly into `MMListView` and/or `MMScrollView` as we want the scrollbar for all lists in the app as mentioned in #4127 . Right now it would be just for feature list.

app/qml/layers/MMFeaturesListPage.qml
Withalion: I think you misunderstood me. What I meant is why do you use both `visible` and `interactive`? Isn't one enough? And also from the documentation it looks like attaching Scrollbar to Flickable should be enough to set it's position, there is no need for anchors.

app/qml/layers/MMFeaturesListPage.qml
kaustuvpokharel: Understood, I would make the respective changes.

app/qml/components/MMListDelegate.qml
```diff
@@ -24,6 +24,7 @@ Item {
   property string text
   property string secondaryText
 
+  readonly property int scrollSpace: (ListView.view && ListView.view.scrollSpace) ? ListView.view.scrollSpace : 0
```
tomasMizera: Unused property

app/qml/components/MMListDelegate.qml
```diff
@@ -40,9 +41,10 @@ Item {
 
   property real verticalSpacing: root.secondaryText ? __style.margin8 : __style.margin20
 
-  implicitWidth: ListView?.view?.width ?? 0 // in case ListView is injected as attached property (usually it is)
+  implicitWidth: ListView?.view?.width ?? 0  // in case ListView is injected as attached property (usually it is)
   implicitHeight: contentLayout.implicitHeight
   height: visible ? implicitHeight : 0.1 // hide invisible items, for some reason setting 0 does not work ¯\_(ツ)_/¯
+  width: implicitWidth - ListView.view.scrollSpace
```
tomasMizera:
```suggestion
  width: implicitWidth - ( ListView?.view?.scrollSpace ?? 0 )
```

app/qml/form/MMPreviewDrawer.qml
```diff
@@ -198,7 +198,7 @@ Item {
           model: root.controller.fieldModel
 
           delegate: Item {
-            width: ListView.view.width
+            width: ListView.view.width - ListView.view.width
```
tomasMizera: Huh? 🤔

app/qml/layers/MMFeaturesListPage.qml
```diff
@@ -10,6 +10,7 @@
 import QtQuick
 import QtQuick.Controls
 import QtQuick.Layouts
+import QtQml
```
tomasMizera: 🤔

app/qml/project/MMProjectWizardPage.qml
```diff
@@ -70,7 +70,7 @@ MMComponents.MMPage {
         delegate: MMProjectComponents.MMProjectWizardDelegate {
           id: fieldDelegate
 
-          width: ListView.view.width
+          width: ListView.view.width - ListView.view.width
```
tomasMizera: Shouldn't we subtract `ListView.view.scrollSpace` instead?

app/qml/project/MMProjectList.qml
```diff
@@ -94,7 +94,6 @@ Item {
     delegate: MMProjectComponents.MMProjectDelegate {
       id: projectDelegate
 
-      width: ListView.view.width
```
tomasMizera, line 97: This does not sound like something we want to do

app/qml/components/MMListView.qml
```diff
@@ -8,13 +8,35 @@
...
 
   cacheBuffer: 0
+  readonly property bool isMobile: (Qt.platform.os === "android" || Qt.platform.os === "ios")
+  property int scrollSpace: !isMobile ? 10 : 0
+
+  ScrollBar.vertical: ScrollBar {
+    id: verticalScrollBar
+
+    policy: ScrollBar.AlwaysOn
+    visible: !isMobile && (root.contentHeight > root.height)
```
tomasMizera: I'd say it would be better to set the policy to `ScrollBar.AlwaysOff` for mobiles instead of creating it, but hiding later..

```suggestion
    policy: isMobile ? ScrollBar.AlwaysOff : ScrollBar.AlwaysOn
```

app/qml/components/MMListView.qml
```diff
@@ -8,13 +8,35 @@
...
+    opacity: (pressed || root.moving) ? 0.7 : 0.4
+
+    anchors.right: root.right
+
+    implicitHeight: root.height
```
tomasMizera: Do we actually need to set these up? 🤔

app/qml/components/MMListView.qml
```diff
@@ -8,13 +8,35 @@
...
   cacheBuffer: 0
+  readonly property bool isMobile: (Qt.platform.os === "android" || Qt.platform.os === "ios")
+  property int scrollSpace: !isMobile ? 10 : 0
+
+  ScrollBar.vertical: ScrollBar {
+    id: verticalScrollBar
+
+    policy: ScrollBar.AlwaysOn
+    visible: !isMobile && (root.contentHeight > root.height)
+    opacity: (pressed || root.moving) ? 0.7 : 0.4
```
tomasMizera: Shouldn't `active` be enough https://doc.qt.io/qt-6/qml-qtquick-controls-scrollbar.html#active-prop? 
```suggestion
    opacity: active ? 0.7 : 0.4
```

app/qml/components/MMListView.qml
```diff
@@ -8,13 +8,35 @@
...
+
+    anchors.right: root.right
+
+    implicitHeight: root.height
+
+    contentItem: Rectangle {
+      implicitWidth: 5
```
tomasMizera: I believe `width` is the proper prop to set here, not `implicitWidth`

app/qml/components/MMScrollView.qml
```diff
@@ -16,8 +16,29 @@ import QtQuick.Controls
...
+
+    contentItem: Rectangle {
+      implicitWidth: 5
+      radius: width / 2
+      color: __style.darkGreenColor
+    }
+  }
 
-  ScrollBar.vertical.policy: ScrollBar.AlwaysOff
-  ScrollBar.horizontal.policy: ScrollBar.AlwaysOff
```
tomasMizera, line 22: Isn't this the Qt docs-suggested way of disabling scrollbar https://doc.qt.io/qt-6/qml-qtquick-controls-scrollview.html#scroll-bars?

app/qml/components/MMScrollView.qml
```diff
@@ -16,8 +16,29 @@ import QtQuick.Controls
...
+  property int scrollSpace: !isMobile ? 10 : 0
+
+  contentWidth: availableWidth - scrollSpace // to only scroll vertically
+
+  ScrollBar.horizontal: null
+
+  ScrollBar.vertical: ScrollBar {
+    id: verticalScrollBar
+
+    policy: ScrollBar.AsNeeded
```
tomasMizera: Why do we have a different value here compared to `MMListView`?

app/qml/components/MMScrollView.qml
tomasMizera, line 1: Similar comments apply here as in `MMListView`

tomasMizera (changes requested):

app/qml/layers/MMFeaturesListPage.qml
kaustuvpokharel: Initially MMFeatureListPage directly had the scrollbar which has Qt.platform.os so I had imported it, I took it to the ListView now.

app/qml/form/MMPreviewDrawer.qml
kaustuvpokharel: My bad, I thought I had copied the ListView.view.scrollspace.

app/qml/components/MMListView.qml
kaustuvpokharel: My intent with visible has a complete different purpose, it is to only show the bar on desktop when contentHeight > height, or else there will be a “ghost” bar ,
But 
`policy: isMobile ? ScrollBar.AlwaysOff : ScrollBar.AsNeeded`
Instead of AlwaysOn, the AsNeeded automatically knows the need of scrollbar, but I intentionally kept the explicit checking to be mindful of the scroll visibility only or specially for contentHeight > height , if you want me to switch I would use As.Needed. @tomasMizera

app/qml/components/MMListDelegate.qml
```diff
@@ -40,9 +40,10 @@ Item {
 
   property real verticalSpacing: root.secondaryText ? __style.margin8 : __style.margin20
 
-  implicitWidth: ListView?.view?.width ?? 0 // in case ListView is injected as attached property (usually it is)
+  implicitWidth: ListView?.view?.width ?? 0  // in case ListView is injected as attached property (usually it is)
   implicitHeight: contentLayout.implicitHeight
   height: visible ? implicitHeight : 0.1 // hide invisible items, for some reason setting 0 does not work ¯\_(ツ)_/¯
+  width: implicitWidth - ( ListView?.view?.scrollSpace ?? 0 )
```
Withalion:
```suggestion
  implicitWidth: ListView?.view?.width - ListView?.view?.scrollSpace ?? 0  // in case ListView is injected as attached property (usually it is)
  implicitHeight: contentLayout.implicitHeight
  height: visible ? implicitHeight : 0.1 // hide invisible items, for some reason setting 0 does not work ¯\_(ツ)_/¯
```

app/qml/components/MMListView.qml
```diff
@@ -8,13 +8,33 @@
...
 // Seems like there is some issue with cache in ListView
 //
 
 ListView {
+  id: root
 
   cacheBuffer: 0
+  readonly property bool isMobile: (Qt.platform.os === "android" || Qt.platform.os === "ios")
+  property int scrollSpace: !isMobile ? 10 : 0
+  property bool showScrollBar: true
```
Withalion: - `isMobile` could be just internal property
- `scrollSpace` I would rename it to something more meaningful like `scrollBarWidth` and it should be read only, and maybe we should use `__style.margin10`
- `showScrollBar` I would change it to alias refering `verticalScrollBar.policy` and get rid of `visible` property

app/qml/components/MMListView.qml
```diff
@@ -8,13 +8,33 @@
...
+
+  ScrollBar.vertical: ScrollBar {
+    id: verticalScrollBar
+
+    policy: isMobile ? ScrollBar.AlwaysOff : ScrollBar.AlwaysOn
+    visible: showScrollBar
+    opacity: active ? 0.7 : 0.4
+
+    contentItem: Rectangle {
+      implicitWidth: 5
```
Withalion: this has not been resolved

app/qml/components/MMScrollView.qml
Withalion, line 1: The changes here should look the same as in `MMListView`

app/qml/components/MMToolbar.qml
```diff
@@ -43,6 +43,7 @@ Rectangle {
 
     MMListView {
       id: toolbar
+      showScrollBar : false
```
Withalion: this will require change based on the stuff mentioned in `MMListView`

Withalion (commented):
I have mixed feelings about these changes. 2 things I don't like:
- when we are defining delegates we should not care if the `View` has scroll bar or not. `Flickables` have `contentWidth` property, but we don't utilize it. I'm not sure if it should be calculated automatically or by us. 
- we show the scroll bars even though they are not necessary. We should never show them on mobile, show them when necessary on desktop and they should not hide.

app/qml/components/MMListView.qml
gabriel-bolbotina: How should I make isMobile an internal property?

app/qml/components/MMListView.qml
gabriel-bolbotina: If I change it to width, the scroll bar will not appear at all.

app/qml/components/MMListView.qml
Withalion: With `QtObject` component as we do in other places

Withalion (approved):
Some stuff doesn't look the best, but it would probably require a bigger refactor, which is out of scope of this PR

app/qml/components/MMScrollView.qml
```diff
@@ -6,18 +6,25 @@
...
-// Convenient class to use as a pageContent or drawerContent
-// base element to make the content scroll
+import QtQml
 
 ScrollView {
-  id: root
+    id: root
+
+    readonly property bool isMobile: (Qt.platform.os === "android"
+                                      || Qt.platform.os === "ios")
```
tomasMizera: You can use `__inputUtis.isMobilePlatform()` instead of this property :)

app/qml/components/MMListView.qml
```diff
@@ -8,13 +8,27 @@
...
 
   cacheBuffer: 0
+  readonly property bool isMobile: (Qt.platform.os === "android" || Qt.platform.os === "ios")
+  readonly property int scrollBarWidth: !isMobile ? __style.margin10 : 0
+  property alias showScrollBar: verticalScrollBar.policy
+
+  ScrollBar.vertical: ScrollBar{
+    id: verticalScrollBar
+
+    policy: isMobile ? ScrollBar.AlwaysOff : ScrollBar.AlwaysOn
```
tomasMizera: Same here, let's use `__inputUtils.isMobilePlatform()` instead

app/qml/components/MMListView.qml
```diff
@@ -8,13 +8,27 @@
...
 // Hot-fix for hotfix https://github.com/MerginMaps/mobile/issues/3417
 // Seems like there is some issue with cache in ListView
 //
 
 ListView {
+  id: root
 
   cacheBuffer: 0
+  readonly property bool isMobile: (Qt.platform.os === "android" || Qt.platform.os === "ios")
+  readonly property int scrollBarWidth: !isMobile ? __style.margin10 : 0
```
tomasMizera: How come we say the scrollbar width is `__style.margin10`, but never use it to set the scrollbar's width? There is no `width: xyz` in `verticalScrollBar`

app/qml/components/MMListView.qml
```diff
@@ -8,13 +8,27 @@
...
+  property alias showScrollBar: verticalScrollBar.policy
+
+  ScrollBar.vertical: ScrollBar{
+    id: verticalScrollBar
+
+    policy: isMobile ? ScrollBar.AlwaysOff : ScrollBar.AlwaysOn
+    visible: contentHeight > availableHeight ? true  : false
```
tomasMizera, line 34: Can you document here why is it better to bind the `visible` prop instead of relying on Qt's `ScrollBar.AsNeeded`?

app/qml/components/MMListView.qml
```diff
@@ -8,13 +8,27 @@
...
 // Seems like there is some issue with cache in ListView
 //
 
 ListView {
+  id: root
 
   cacheBuffer: 0
+  readonly property bool isMobile: (Qt.platform.os === "android" || Qt.platform.os === "ios")
+  readonly property int scrollBarWidth: !isMobile ? __style.margin10 : 0
+  property alias showScrollBar: verticalScrollBar.policy
```
tomasMizera: The name `showScrollBar` implies that the property is `boolean` while it really is an enum value. I suggest something like `scrollBarPolicy` instead.

tomasMizera (changes requested):

app/qml/components/MMScrollView.qml
```diff
@@ -6,18 +6,27 @@
...
-  ScrollBar.vertical.policy: ScrollBar.AlwaysOff
-  ScrollBar.horizontal.policy: ScrollBar.AlwaysOff
+    // To make sure the scroll bar stays visible on desktop, even when not focused:
+    // - both 'policy' and 'visible' properties need to be set, with 'policy' set to AlwaysOn.
+    // - the use the 'AsNeeded' policy will make the scroll bar hide when not focused or in use,
+    // and only appear after the user starts scrolling.
+    ScrollBar.vertical.policy: __inputUtils.isMobilePlatform ? ScrollBar.AlwaysOff : ScrollBar.AlwaysOn
+    ScrollBar.vertical.visible: contentHeight > availableHeight ? true  : false
+    ScrollBar.vertical.opacity: active ? 0.7 : 0.4
+    ScrollBar.vertical.width: scrollBarWidth
```
tomasMizera: Just a tip, in QML you can use this syntax:
```suggestion
    ScrollBar.vertical {
      policy: __inputUtils.isMobilePlatform ? ScrollBar.AlwaysOff : ScrollBar.AlwaysOn
      visible: contentHeight > availableHeight ? true  : false
      opacity: active ? 0.7 : 0.4
      width: scrollBarWidth
    }
```

app/qml/components/MMScrollView.qml
```diff
@@ -6,18 +6,27 @@
...
-// Convenient class to use as a pageContent or drawerContent
-// base element to make the content scroll
+import QtQml
 
 ScrollView {
-  id: root
+    id: root
+
+    readonly property int scrollBarWidth: __inputUtils.isMobilePlatform ? 0 : __style.margin10 
+    property bool scrollBarPolicy:  root.ScrollBar.vertical.policy
```
tomasMizera: I'd rename this the same way as in `MMListView`.

app/qml/components/MMListView.qml
```diff
@@ -8,13 +8,31 @@
...
 //
 // Hot-fix for hotfix https://github.com/MerginMaps/mobile/issues/3417
 // Seems like there is some issue with cache in ListView
 //
 
 ListView {
+  id: root
 
   cacheBuffer: 0
+  readonly property int scrollBarWidth: __inputUtils.isMobilePlatform ?  0 : __style.margin10
```
tomasMizera: `isMobilePlatform` is a method, not a property ..

tomasMizera (changes requested):

tomasMizera (changes requested):
This is not an ideal solution to set width to `width: ListView.view.width - ListView.view.scrollBarWidth` all around the place. Let's instead:
1. Check if Qt has some automatic way to handle scrollbars, or
2. Introduce a property ~ `delegateWidth` to `MMListView` that can be used and it will automatically calculate with the scrollbar width

Withalion:
We'll keep this as a prototype for now.

PR #4146, opened 2025-10-10
Fixed the space or rebound issue

kaustuvpokharel:
Case: In featureList whenever flickable is scrolled fast or scrolled out of bound; it rebounds and parks the feature below the "add feature" button awkwardly. 

FIx: delete the cachedBuffer altogether. In addition, when scrolling up really fast, the list used to go out of bounds, to fix this I added a threshold for the verticalOvershoot and made the list come back to the first element.

https://github.com/user-attachments/assets/88c712e5-a135-4ea7-928f-c427386d01ec

And also tested on MacOS, with the project which was referred in this [ticket](https://github.com/MerginMaps/mobile/issues/3417) : 

https://github.com/user-attachments/assets/a7266add-3e7a-4a96-bf72-8a56fb7163ed

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -2,7 +2,7 @@
     "description": "<p>Mergin Maps mobile app is a QGIS powered app for Android and iOS devices.</p>", 
     "license": "GPLv3", 
     "title": "Mergin Maps mobile app", 
-    "version": "2025.8.0",
+    "version": "2026.1.0",
     "upload_type": "software", 
     "publication_date": "2022-02-24", 
     "creators": [
@@ -39,7 +39,7 @@
     "related_identifiers": [
         {
             "scheme": "url", 
-            "identifier": "https://github.com/MerginMaps/mobile/tree/2025.8.0", 
+            "identifier": "https://github.com/MerginMaps/mobile/tree/2026.1.0", 
             "relation": "isSupplementTo"
         }, 
         {
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,4 +1,4 @@
-cff-version: 2025.8.0
+cff-version: 2026.1.0
 message: "If you use this software, please cite it as below."
 authors:
 - family-names: "Martin"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,8 +5,8 @@ cmake_minimum_required(VERSION 3.22)
 # ########################################################################################
 
 # Note: To update version use script/update_all_versions.bash
-set(MM_VERSION_MAJOR "2025")
-set(MM_VERSION_MINOR "8")
+set(MM_VERSION_MAJOR "2026")
+set(MM_VERSION_MINOR "1")
 set(MM_VERSION_PATCH "0")
 
 if (VCPKG_TARGET_TRIPLET MATCHES ".*ios.*")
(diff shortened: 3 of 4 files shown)


app/qml/layers/MMFeaturesListPage.qml
```diff
@@ -50,6 +50,12 @@ MMComponents.MMPage {
 
       width: parent.width
 
+      highlightMoveDuration: 120
+      cacheBuffer: height
```
Withalion: I wouldn't touch `cacheBuffer` it will break other things, look inside `MMListView` to find more info.

app/qml/layers/MMFeaturesListPage.qml
```diff
@@ -50,6 +50,12 @@ MMComponents.MMPage {
 
       width: parent.width
 
+      highlightMoveDuration: 120
```
Withalion: But we are not using highlight, what is the point of this property?

app/qml/layers/MMFeaturesListPage.qml
```diff
@@ -50,6 +50,12 @@ MMComponents.MMPage {
...
+      cacheBuffer: height
+      footerPositioning: ListView.InlineFooter
+      bottomMargin: spacer.height
```
Withalion: so now we have bottom margin and footer too? We should probably just use footer instead

app/qml/layers/MMFeaturesListPage.qml
kaustuvpokharel: I had used highlight: property for debugging the real cause, and missed the highlightMoveDuration.

app/qml/layers/MMFeaturesListPage.qml
kaustuvpokharel: Well, I believe it wouldn't break anything else, because as you scroll, new delegates need to be created and old ones destroyed technically in the list and this is the main cause for small stutters, and rebound issue. 
And the 0 value in MMListView is only stopping the preloading, and it needs to be changed when list stutters because, it's only normal to stutter if the cacheBuffer for list is explicitly set 0. 
Preloading 10-15 pixel before hand is no harm. Plus, it only changes for MMFeaturesListPage, not for whole MMListView so I am hoping this is the best approach for this specific case.

tomasMizera: > Well, I believe it wouldn't break anything else...

Hi @kaustuvpokharel, this is why we set it to `0`: https://github.com/MerginMaps/mobile/pull/3435 .. chances are that the underlying QT bug is already fixed. If not, this would start to crash on macos/ios. Please try testing it out with >100 features in the list.

gabriel-bolbotina: I have tried testing with 100+ entries and it seems that the issue is fixed with Cali's solution. [Link to video](https://drive.google.com/file/d/1usTgPjCZ6mGCPrS75YHUCTguqpXwolO0/view?usp=share_link)

app/qml/layers/MMFeaturesListPage.qml
```diff
@@ -74,6 +77,7 @@ MMComponents.MMPage {
       }
 
       footer: MMComponents.MMListSpacer {
+        id: spacer
```
Withalion:
```suggestion

```
Let's remove it if we are not using it

Withalion (approved):

tomasMizera (approved):
This needs a proper testing on all platforms, with complicated forms and a special test for the "flicking up" fix

PR #4171, opened 2025-10-20
Created Reusable NotificationBox component

kaustuvpokharel:
Addition of new UI component - NotificationBox Componnet

- Added new MMNotificationBox.qml component
- With respected UI design with apis control for reuse

**updated image**

Capture d’écran 2025-11-27 à 15 43 17

Unknown Information: 
What's the case of the notification box or button on hover, for Warning and Error type

### DIFF
--- a/gallery/qml/Main.qml
+++ b/gallery/qml/Main.qml
@@ -85,11 +85,11 @@ ApplicationWindow {
 
   Action {
     id: switchOrientationAction
-    onTriggered: Qt.callLater(() => {
+    onTriggered: {
      let s = window.width
      window.width = window.height
      window.height = s
-    })
+    }
   }
 
   header: ToolBar {
@@ -159,10 +159,11 @@ ApplicationWindow {
         onClicked: {
           window.currentPageSource = model.source
           listView.currentIndex = index
-          const url = __isMobile
-                    ? "qrc:/qml/pages/" + model.source
-                    : "file://" + _qmlWrapperPath + model.source
-          stackView.replace(url)
+          if (__isMobile)
+            stackView.push("qrc:/qml/pages/" + model.source)
+          else
+            stackView.push("file://" + _qmlWrapperPath + model.source)
+          stackView.pop()
           drawer.close()
         }
       }
@@ -265,11 +266,10 @@ ApplicationWindow {
   StackView {
     id: stackView
     anchors.fill: parent
+
     initialItem: Loader {
       id: mainLoader
-      source: (__isMobile
-               ? "qrc:/qml/pages/" + currentPageSource
-               : "file://" + _qmlWrapperPath + currentPageSource)
+      source: (__isMobile ? "qrc:/qml/pages/" : ("file://" + _qmlWrapperPath)) + currentPageSource
       scale: 1.0
     }
   }


gallery/qml/Main.qml
```diff
@@ -159,11 +159,10 @@ ApplicationWindow {
         onClicked: {
           window.currentPageSource = model.source
           listView.currentIndex = index
-          if (__isMobile)
-            stackView.push("qrc:/qml/pages/" + model.source)
-          else
-            stackView.push("file://" + _qmlWrapperPath + model.source)
-          stackView.pop()
+          const url = __isMobile
```
kaustuvpokharel: I thought this is a better approach then pushing and popping, because this solves the screen freezing issue

app/qml/components/MMNotificationBox.qml
```diff
@@ -0,0 +1,114 @@
...
+  property alias title: titleText.text
+  property alias description: descriptionText.text
+  property alias buttonText: actionButton.text
+
+  enum Types { Warning, Error }
+  property int type: MMNotificationBox.Types.Warning
+
+  property color bgndColor: {
+    if ( type === MMNotificationBox.Types.Warning ) return __style.sandColor
+    if ( type === MMNotificationBox.Types.Error ) return __style.negativeLightColor
```
Withalion: this can be just ternary operator

app/qml/components/MMNotificationBox.qml
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  property color borderColor: {
+    if ( type === MMNotificationBox.Types.Warning ) return __style.sunsetColor
+    if ( type === MMNotificationBox.Types.Error ) return __style.grapeColor
```
Withalion: same thing here

app/qml/components/MMNotificationBox.qml
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  signal buttonClicked()
+
+  property color textColor: __style.nightColor
```
Withalion: Is this something we want to change from outside? I think it should be dependent on the `MMNotificationBox.type`.

Withalion:
Please rebase this at least on 2025.7.0 and for now it should target 2025.8.0

gallery/qml/pages/MiscPage.qml
```diff
@@ -232,6 +232,28 @@ ScrollView {
...
+        }
+
+        MMNotificationBox
+        {
+          width: page.width - 64
+          type: MMNotificationBox.Error
+          title: "Your payment is not working"
+          description: "We would like to inform you that MerginMaps will undergo scheduled maintance on 1st of october 2025. As a result of this upgrade you will be able to see better performance on our servers."
+
+          buttonText: "Service Cancled"
```
Withalion:
```suggestion
          buttonText: "Service Canceled"
```

app/qml/components/MMNotificationBox.qml
kaustuvpokharel: It is constant for either notificationBox type, so I didn't create an API to configure it from outside as it doesn't need changing.

app/qml/components/MMNotificationBox.qml
kaustuvpokharel: I m open to changing it, but specifically for this case here I feel this version is a bit more readable and the Warning and Error cases are visually easy to match at one look, plus expandable for (Error, Warning, Sucess, ..)
With a ternary operator, only one condition and its fallback would be visible, which makes it slightly harder to associate each type with its color at a glance, 
• Warning -> sandColor
• Error -> negativeLightColor

that said, Im happy to switch if you still think the ternary form is preferable @Withalion

Withalion: Eh I'm easy about it, good point about expanding it. However I would switch to `switch` at that point, as I feel like that is more readable.

app/qml/components/MMNotificationBox.qml
Withalion: But you created API for it because you defined it as property on the root element. So if it's constant I would just hard code it where you reference it.

tomasMizera:
What's the state here guys @Withalion @kaustuvpokharel ?

Withalion:
Waiting for changes

kaustuvpokharel:
Sorry, I am confused @tomasMizera @Withalion , the change we are talking about is the constant api that I created ? Because I kinda kept is as such, but I would change that, apart from that I am guessing everything else is okay no ?

app/qml/components/MMNotificationBox.qml
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  property color buttonFontColor: {
+    switch(type) {
+      case MMNotificationBox.Types.Warning:
+        return __style.earthColor
+      case MMNotificationBox.Types.Error:
+        return __style.roseColor
+    }
+  }
```
Withalion: We should not expose these properties as they are set internal depending on the `type`. We did something similar in `MMdrawerDialog` for example with the help of `QtObject`.

app/qml/components/MMNotificationBox.qml
```diff
@@ -0,0 +1,135 @@
...
+
+  signal buttonClicked()
+
+  color: bgndColor
+  border.width: 1 * __dp
+  border.color: borderColor
+  implicitHeight: dynamicContentGroup.implicitHeight + 2 * __style.margin20
+  radius: __style.radius12
```
Withalion: I think Figma defines different radius, check if it looks good

app/qml/components/MMNotificationBox.qml
```diff
@@ -0,0 +1,135 @@
...
+
+    width: parent.width - 2 * __style.margin20
+    height: implicitHeight
+
+    x: __style.margin20
+    y: __style.margin20
+
+    spacing: __style.spacing12
+
+    Column {
```
Withalion: Why are we using both `ColumnLayout` and `Column`? One should be enough.

app/qml/components/MMNotificationBox.qml
```diff
@@ -0,0 +1,135 @@
...
+  radius: __style.radius12
+
+  ColumnLayout {
+    id: dynamicContentGroup
+
+    width: parent.width - 2 * __style.margin20
+    height: implicitHeight
+
+    x: __style.margin20
+    y: __style.margin20
```
Withalion: This smells

gallery/qml/pages/MiscPage.qml
Withalion: Still not fixed

gallery/qml.qrc
```diff
@@ -155,5 +155,6 @@
         <file>../app/qml/dialogs/MMDiscardGeometryChangesDialog.qml</file>
         <file>../app/qml/dialogs/MMProviderRemoveReceiverDialog.qml</file>
         <file>../app/qml/dialogs/components/MMDialogAdditionalText.qml</file>
+        <file>../app/qml/components/MMNotificationBox.qml</file>
```
Withalion: put it back with friends :)

app/qml/components/MMNotificationBox.qml
```diff
@@ -0,0 +1,135 @@
...
+
+    spacing: __style.spacing12
+
+    Column {
+      width: parent.width
+      height: implicitHeight
```
Withalion: If you decide to keep `ColumnLayout` you should not define these in children components

app/qml/components/MMNotificationBox.qml
```diff
@@ -0,0 +1,135 @@
...
+      id: actionButton
+
+      Layout.alignment: Qt.AlignLeft
+
+      fontColor: buttonFontColor
+      bgndColor: buttonBgndColor
+
+      // reverse the color properties when the button is hovered
+      fontColorHover: buttonBgndColor
+      bgndColorHover: buttonFontColor
```
Withalion:
```suggestion
      fontColor: root.buttonFontColor
      bgndColor: root.buttonBgndColor

      // reverse the color properties when the button is hovered
      fontColorHover: root.buttonBgndColor
      bgndColorHover: root.buttonFontColor
```
I recommend using full qualifiers, maybe the qualifier will be different after refactor to "private" properties

app/qml/components/MMNotificationBox.qml
```diff
@@ -0,0 +1,132 @@
...
+  color: internal.bgndColor
+  border.width: 1 * __dp
+  border.color: internal.borderColor
+  implicitHeight: dynamicContentGroup.implicitHeight + __style.margin20
+  radius: __style.radius8
+
+  Column {
+    id: dynamicContentGroup
+
+    width: parent.width - 2 * __style.margin20
```
Withalion, line 35: couldn't we just do `padding: __style.margin20` ?

app/qml/components/MMNotificationBox.qml
```diff
@@ -0,0 +1,132 @@
...
+    height: implicitHeight
```
Withalion:
```suggestion
   
```
isn't this the default behaviour? Why do we need to force it?

app/qml/components/MMNotificationBox.qml
```diff
@@ -0,0 +1,132 @@
...
+    topPadding: __style.margin20
+    leftPadding: __style.margin20
+
+    spacing: __style.spacing12
+
+    Text {
+        id: titleText
+
+        width: parent.width
+        height: implicitHeight
```
Withalion: same as above

app/qml/components/MMNotificationBox.qml
```diff
@@ -0,0 +1,132 @@
...
+        lineHeightMode: Text.FixedHeight
+
+        verticalAlignment: Text.AlignVCenter
+      }
+
+      Text {
+        id: descriptionText
+
+        width: parent.width
+        height: text ? implicitHeight : 0
```
Withalion: hmm this also doesn't look right, I don't thing this component will be used without description. However if it will be then have a look what's the behaviour when it doesn't get any text (is there some extra space left?), so maybe not playing around with this will be enough and another thing I would rather set the component invisible than change height to 0

app/qml/components/MMNotificationBox.qml
gabriel-bolbotina: It will add padding to the bottom, so it will look like this: 
Capture d’écran 2025-11-27 à 15 40 09.

PR #4176, opened 2025-10-22
Geoid separation support for external providers

tomasMizera:
Just a quick demo for geoid separation

Resolves https://github.com/MerginMaps/mobile/issues/2725


wonder-sk (commented):
Looks good - and testing with external GNSS confirms it works.

One thing worth fixing: with internal GNSS, the GPS info panel shows geoid separation being "0.00 m" which could be confusing - it would be better to report it as N/A. Or maybe we could add calculation of geoid separation for internal GNSS while at this:

```python
ct=QgsCoordinateTransform(
    QgsCoordinateReferenceSystem("EPSG:4979"),  # WGS84 + ellipsoid height
    QgsCoordinateReferenceSystem("EPSG:9707"),  # WGS84 + EGM96 geoid height
    QgsCoordinateTransformContext())  # TODO: use project's context
# this is ellipsoid height of 200 m converted to geoid height - at lat 49°, lon 18°
print(ct.transform(QgsVector3D(18,49,200)).z())
```

(we just need to verify we have egm96 geoid shipped with PROJ - on my system it is in file `egm96_15.gtx` - cca 4 MB)

Withalion:
Let's keep this PR as it is and I'll create a new one to resolve the issue with internal providers. It seems better to me as we will be building on top of the work done here and also in #4210.

PR #4183, opened 2025-10-28
fixed minor plural translation not working

kaustuvpokharel:
Changed the string to use rules instead of arg which doesn't properly translate


app/qml/map/MMSplittingTools.qml
```diff
@@ -111,6 +111,7 @@ Item {
           else
           {
             __notificationModel.addWarning( qsTr( "You need to add at least 2 points." ) )
+            //__notificationModel.addWarning( __inputUtils.invalidGeometryWarning( mapTool.activeLayer ) )
```
tomasMizera: remove line

tomasMizera (changes requested):
Please check other places for the plural form strings

PR #4199, opened 2025-11-03
Set default QGIS_QUICK_DATA_PATH value

tomasMizera:
This PR changes cmake configuration to define default value of `QGIS_QUICK_DATA_PATH` variable. Each platform is unique:

 - `Android` ignores the value and sets its own value at runtime
 - `iOS` and `Windows` need the current `Input` value
 - `Lnx` and `Macos` need path to the build output (where we copy `proj` data)

Fixes https://github.com/MerginMaps/mobile/issues/3503

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -106,6 +106,8 @@ if (IOS)
   # after release on iOS devices
   set(QGIS_QUICK_DATA_PATH_DEFAULT "INPUT")
 elseif (ANDROID)
+  # Android sets this up automatically at runtime
+  set(QGIS_QUICK_DATA_PATH_DEFAULT "")
 elseif (WIN)
   set(QGIS_QUICK_DATA_PATH_DEFAULT "INPUT")
 elseif (LNX OR MACOS)


CMakeLists.txt
```diff
@@ -101,6 +101,17 @@ else ()
   endif ()
 endif ()
 
+if (IOS)
+  # Important, if QGIS_QUICK_DATA_PATH value is changed, old projects may not be found after release on iOS devices
+  set(QGIS_QUICK_DATA_PATH_DEFAULT "INPUT")
+elseif (ANDROID) 
+  set(QGIS_QUICK_DATA_PATH_DEFAULT "INPUT")
```
Withalion: I wouldn't touch it on Android at all, the value is completely ignored and then at runtime it's set to something meaningful

CMakeLists.txt
```diff
@@ -402,6 +413,8 @@ file(
   PATTERN "*.cmake" EXCLUDE
   PATTERN "vcpkg*" EXCLUDE
 )
+
+message(STATUS "QGIS_QUICK_DATA_PATH set to ${QGIS_QUICK_DATA_PATH}")
```
Withalion, line 421: it gets also logged during runtime, which might be more meaningful as we do some more changes at that point

tomasMizera: I'd prefer keeping it here so that I can validate my configuration is correct 👍

tomasMizera: (even without building&running it)

CMakeLists.txt
tomasMizera: Well, it needs some value otherwise we'd need to if/def it other places 🙃

PR #4202, opened 2025-11-04
Clean cmake

Withalion:
Ideas:
- remove leftovers from VCPKG migration
- remove commented out workarounds
- remove android compiler flags as they are used by default on our version of NDK
- reduce the amount of libraries we are linking explicitly, they are still linked but now transitively as required by other targets

### DIFF
--- a/app/CMakeLists.txt
+++ b/app/CMakeLists.txt
@@ -525,7 +525,7 @@ target_link_libraries(
 )
 
 if (WIN)
-  target_link_libraries(MerginMaps PUBLIC GEOS::geos_c)
+  target_link_libraries(MerginMaps PUBLIC GEOS::geos_c Qt6::Core5Compat)
 endif ()
 
 if (IOS OR MACOS)


app/CMakeLists.txt
```diff
@@ -524,6 +524,10 @@ target_link_libraries(
          meshoptimizer::meshoptimizer
 )
 
+if (WIN)
+  target_link_libraries(MerginMaps PUBLIC GEOS::geos_c)
```
tomasMizera: Isn't `GEOS` already linked https://github.com/MerginMaps/mobile/blob/70d4e10027a696a007697a78c9ce9b4f9b4c7faa/app/CMakeLists.txt#L533?

app/CMakeLists.txt
```diff
@@ -525,7 +525,7 @@ target_link_libraries(
 )
 
 if (WIN)
-  target_link_libraries(MerginMaps PUBLIC GEOS::geos_c)
+  target_link_libraries(MerginMaps PUBLIC GEOS::geos_c Qt6::Core5Compat)
```
tomasMizera, line 528: Same comment as for GEOS https://github.com/MerginMaps/mobile/blob/70d4e10027a696a007697a78c9ce9b4f9b4c7faa/app/CMakeLists.txt#L468

CMakeLists.txt
```diff
@@ -355,25 +343,9 @@ if (IOS)
 endif ()
 
 if (ANDROID)
-  if (VCPKG_TARGET_TRIPLET STREQUAL "arm-android")
```
tomasMizera, line 358: This looks serious 😱

Where is this now defined?

app/CMakeLists.txt
Withalion: The thing is for OSes different than windows `GEOS` and `Qt6::Core5Compat` become transient dependencies and are linked properly. On windows we need to define them explicitly.

CMakeLists.txt
Withalion: Let's say we got inspired by other projects when transitioning to VCPKG. However these are default flags for the linker since NDK 23+ (currently we use NDK 27), so there is no need to define them explicitly. `-ljnigraphics` flag has been moved to `app/CMakeLists.txt` as it is more appropriate and it should be defined only for android.

PR #4210, opened 2025-11-11
Recalculate ellipsoid elevation to orthometric

### LINKED ISSUE #2615: Convert ellipsoid heights to orthometric (apply geoid undulation)
In some cases it may be useful for users to get automatic conversion between ellipsoid heights and orthometric heights - or in other terms, to (un-)apply geoid undulation. (Note: external GNSS return orthometric heights, and they also return geoid separation used for the height measurement to allow switching to ellipsoid height, but this information is currently not kept in Input - see #2725)

For this we would need to have access to the EGM2008 geoid model (~77MB) - https://github.com/OSGeo/PROJ-data/blob/master/us_nga/us_nga_egm08_25.tif - we could download it from CDN first time such option would be enabled.

To list transforms to from ellipsoidal heights to orthometric heights (EPSG:4979 would be actually more correct CRS as it is 3D, EPSG:4326 is 2D):
```
projinfo -s EPSG:4326 -t EPSG:4326+3855
```
This returns `WGS 84 to EGM2008 height (1)` with this definition:
```
+proj=pipeline
  +step +proj=axisswap +order=2,1
  +step +proj=unitconvert +xy_in=deg +xy_out=rad
  +step +inv +proj=vgridshift +grids=us_nga_egm08_25.tif +multiplier=1
  +step +proj=unitconvert +xy_in=rad +xy_out=deg
  +step +proj=axisswap +order=2,1
```

Then converting lat=50 deg north, lon=20 deg east, alt=100 m above ellipsoid gives ~60 m orthometric height which seems correct:
```
$ export PROJ_NETWORK=ON    # to allow PROJ fetch the geoid raster data from network
$ echo 50 20 100 0 | cct "WGS 84 to EGM2008 height (1)"
 50.0000000000   20.0000000000       60.2805        0.0000
```

Workaround for now with ogr2ogr (with points_ellipsoid.gpkg being input file):
```
$ PROJ_NETWORK=ON ogr2ogr -s_srs EPSG:4326 -t_srs EPSG:4326+3855 points_ortho.gpkg points_ellipsoid.gpkg
```

### LINKED ISSUE #4165: Android 15+ returns geoid altitude while older Androids show ellipsoid altitude
When using internal GPS of the phone/tablet on Android, ellipsoid altitudes are reported (there's a ticket #2615 to convert those to geoid altitudes so that it's consistent with external GNSS).

BUT with Android 15+ there's a bug in Qt and the mobile app reports geoid altitudes instead 🤯  (the older Android versions should still return ellipsoid altitudes). There is a bug fix in Qt (https://github.com/qt/qtpositioning/commit/80a0ad58ed14dfbe56cc03ac8c8b93a087852971) but not available in a released version of Qt as of today.

Workaround: switch to "Internal (fused)" GPS receiver - that will consistently report ellipsoid altitudes.

More details:
- Qt 6.8+ can optionally provide geoid altitudes on Android 14+ (using "useMslAltitude" parameter - off by default) - https://doc.qt.io/qt-6/position-plugin-android.html - see https://github.com/qt/qtpositioning/commit/0f3efd9b280ba3151352e9e3bfa67b4c22e0a0f2
- In Android 15+ the geoid altitude seems to be always present in location, and Qt incorrectly thinks that the "useMslAltitude" has been turned on - see https://bugreports.qt.io/browse/QTBUG-106049

Withalion:
fixes #4165 
fixes #2615 

This PR adds EGM96_15 geoid model internally to recalculate ellipsoid heights returned by position providers to orthometric heights by default. Also, the information of which model was used is exposed to users in GPS information panel. There are some necessary fixes to Qt Positioning to force `InternalPositioningProvider` to return ellipsoid height on both android and iOS.
image
Besides this, some minor refactoring was done to prepare the rest of the codebase to support user supplied geoid models.

### DIFF
--- a/app/appsettings.cpp
+++ b/app/appsettings.cpp
@@ -10,7 +10,6 @@
 #include "appsettings.h"
 #include "coreutils.h"
 
-#include <qcoreapplication_platform.h>
 #include <QSettings>
 #include <QFileInfo>
 
@@ -199,7 +198,7 @@ void AppSettings::setActivePositionProviderId( const QString &id )
     return;
 
   mActivePositionProviderId = id;
-  setValue( QStringLiteral( "activePositionProviderId" ), mActivePositionProviderId );
+  setValue( QStringLiteral( "activePositionProviderId" ), id );
   emit activePositionProviderIdChanged( mActivePositionProviderId );
 }
 
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -476,7 +476,7 @@ QString InputUtils::geometryLengthAsString( const QgsGeometry &geometry )
 {
   QgsDistanceArea distanceArea;
   distanceArea.setEllipsoid( QStringLiteral( "WGS84" ) );
-  distanceArea.setSourceCrs( PositionKit::positionCRS(), QgsCoordinateTransformContext() );
+  distanceArea.setSourceCrs( PositionKit::positionCrs2D(), QgsCoordinateTransformContext() );
 
   qreal length = distanceArea.measureLength( geometry );
 
@@ -879,43 +879,6 @@ QgsPoint InputUtils::transformPoint( const QgsCoordinateReferenceSystem &srcCrs,
   // QGIS would convert them to a valid (0, 0) points
   if ( srcPoint.isEmpty() )
   {
-    return QgsPoint();
-  }
-
-  try
-  {
-    QgsCoordinateTransform ct( srcCrs, destCrs, context );
-    if ( ct.isValid() )
-    {
-      if ( !ct.isShortCircuited() )
-      {
-        const QgsPointXY transformed = ct.transform( srcPoint.x(), srcPoint.y() );
-        const QgsPoint pt( transformed.x(), transformed.y(), srcPoint.z(), srcPoint.m() );
-        return pt;
-      }
-      else
-      {
-        return srcPoint;
-      }
-    }
-  }
-  catch ( QgsCsException &cse )
-  {
-    Q_UNUSED( cse )
-  }
-
-  return QgsPoint();
-}
-
-QgsPoint InputUtils::transformPoint3D( const QgsCoordinateReferenceSystem &srcCrs,
-                                       const QgsCoordinateReferenceSystem &destCrs,
-                                       const QgsCoordinateTransformContext &context,
-                                       const QgsPoint &srcPoint )
-{
-  // we do not want to transform empty points,
-  // QGIS would convert them to a valid (0, 0) points
-  if ( srcPoint.isEmpty() )
-  {
     return {};
   }
 
@@ -991,7 +954,7 @@ QgsPoint InputUtils::mapPointToGps( QPointF mapPosition, InputMapSettings *mapSe
 
   const QgsPointXY transformedXY = transformPoint(
                                      mapSettings->destinationCrs(),
-                                     PositionKit::positionCRS(),
+                                     PositionKit::positionCrs2D(),
                                      QgsCoordinateTransformContext(),
                                      positionMapCrs
                                    );
@@ -1749,7 +1712,7 @@ qreal InputUtils::distanceBetweenGpsAndFeature( QgsPoint gpsPosition, const Feat
 
   // Transform gps position to map CRS
   QgsPointXY transformedPosition = transformPoint(
-                                     PositionKit::positionCRS(),
+                                     PositionKit::positionCrs3D(),
                                      mapSettings->destinationCrs(),
                                      mapSettings->transformContext(),
                                      gpsPosition
@@ -1797,7 +1760,7 @@ qreal InputUtils::angleBetweenGpsAndFeature( QgsPoint gpsPoint, const FeatureLay
 
   // Transform gps position to map CRS
   QgsPointXY transformedPosition = transformPoint(
-                                     PositionKit::positionCRS(),
+                                     PositionKit::positionCrs3D(),
                                      mapSettings->destinationCrs(),
                                      mapSettings->transformContext(),
                                      gpsPoint
--- a/app/inpututils.h
+++ b/app/inpututils.h
@@ -299,18 +299,8 @@ class InputUtils: public QObject
     /**
       * Transforms point between different crs
       * Return empty QgsPoint if the transformation could not be applied or srcPoint is empty
-      * \note This function only does 2D crs transformation, use \a transformPoint3D for 3D transformation!
       */
     Q_INVOKABLE static QgsPoint transformPoint( const QgsCoordinateReferenceSystem &srcCrs,
-        const QgsCoordinateReferenceSystem &destCrs,
-        const QgsCoordinateTransformContext &context,
-        const QgsPoint &srcPoint );
-
-    /**
-      * Transforms point between different crs
-      * Return empty QgsPoint if the transformation could not be applied or srcPoint is empty
-      */
-    Q_INVOKABLE static QgsPoint transformPoint3D( const QgsCoordinateReferenceSystem &srcCrs,
         const QgsCoordinateReferenceSystem &destCrs,
         const QgsCoordinateTransformContext &context,
         const QgsPoint &srcPoint );
--- a/app/position/positionkit.cpp
+++ b/app/position/positionkit.cpp
@@ -33,9 +33,19 @@ PositionKit::PositionKit( QObject *parent )
 {
 }
 
-QgsCoordinateReferenceSystem PositionKit::positionCRS()
+QgsCoordinateReferenceSystem PositionKit::positionCrs3D()
 {
   return QgsCoordinateReferenceSystem::fromEpsgId( 9707 );
+}
+
+QgsCoordinateReferenceSystem PositionKit::positionCrs2D()
+{
+  return QgsCoordinateReferenceSystem::fromEpsgId( 4326 );
+}
+
+QgsCoordinateReferenceSystem PositionKit::positionCrs3DEllipsoidHeight()
+{
+  return QgsCoordinateReferenceSystem::fromEpsgId( 4979 );
 }
 
 void PositionKit::startUpdates()
--- a/app/position/providers/androidpositionprovider.cpp
+++ b/app/position/providers/androidpositionprovider.cpp
@@ -56,10 +56,10 @@ void jniOnPositionUpdated( JNIEnv *env, jclass clazz, jint instanceId, jobject l
     const jdouble value = location.callMethod<jdouble>( "getAltitude" );
     if ( !qFuzzyIsNull( value ) )
     {
-      // transform the altitude from WGS84 ellipsoid to specified geoid model
-      const QgsPoint geoidPosition = InputUtils::transformPoint3D(
-                                       QgsCoordinateReferenceSystem::fromEpsgId( 4979 ),
-                                       PositionKit::positionCRS(),
+      // transform the altitude from EPSG:4979 (WGS84 (EPSG:4326) + ellipsoidal height) to specified geoid model
+      const QgsPoint geoidPosition = InputUtils::transformPoint(
+                                       PositionKit::positionCrs3DEllipsoidHeight(),
+                                       PositionKit::positionCrs3D(),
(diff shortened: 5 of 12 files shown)


gabriel-bolbotina:
@tomasMizera I think you could give it a look as well.

app/appsettings.cpp
```diff
@@ -10,6 +10,7 @@
 #include "appsettings.h"
 #include "coreutils.h"
 
+#include <qcoreapplication_platform.h>
```
wonder-sk: the change here and below in this file could be reverted for smaller diff

app/inpututils.cpp
```diff
@@ -476,7 +476,7 @@ QString InputUtils::geometryLengthAsString( const QgsGeometry &geometry )
 {
   QgsDistanceArea distanceArea;
   distanceArea.setEllipsoid( QStringLiteral( "WGS84" ) );
-  distanceArea.setSourceCrs( QgsCoordinateReferenceSystem::fromEpsgId( 4326 ), QgsCoordinateTransformContext() );
+  distanceArea.setSourceCrs( PositionKit::positionCRS(), QgsCoordinateTransformContext() );
```
wonder-sk: I am not sure if this is the right thing to do - the distance measurement only uses 2D coordinates, so using 3D coordinate system may be a bit confusing and suggest we're doing 3D measurements even we do not. I would suggest to revert this...

app/inpututils.cpp
```diff
@@ -953,11 +988,10 @@ QgsPoint InputUtils::mapPointToGps( QPointF mapPosition, InputMapSettings *mapSe
     return QgsPoint();
 
   QgsPoint positionMapCrs = mapSettings->screenToCoordinate( mapPosition );
-  QgsCoordinateReferenceSystem crsGPS = coordinateReferenceSystemFromEpsgId( 4326 );
 
   const QgsPointXY transformedXY = transformPoint(
                                      mapSettings->destinationCrs(),
-                                     crsGPS,
+                                     PositionKit::positionCRS(),
```
wonder-sk: here we only work with 2D coordinates, so using 3D CRS is misleading - let's switch back to 2D CRS...

app/inpututils.cpp
```diff
@@ -953,11 +988,10 @@ QgsPoint InputUtils::mapPointToGps( QPointF mapPosition, InputMapSettings *mapSe
...
                                      QgsCoordinateTransformContext(),
```
wonder-sk, line 958: not directly relevant to this PR, but we should not use empty QgsCoordinateTransformContext() - we should use coordinate transform context from the active project everywhere, to avoid some subtle transformation errors (e.g. when project admin picks a non-default transform). Worth checking elsewhere too!

app/inpututils.h
```diff
@@ -299,12 +299,22 @@ class InputUtils: public QObject
...
         const QgsPoint &srcPoint );
 
+    /**
+      * Transforms point between different crs
+      * Return empty QgsPoint if the transformation could not be applied or srcPoint is empty
+      */
+    Q_INVOKABLE static QgsPoint transformPoint3D( const QgsCoordinateReferenceSystem &srcCrs,
+        const QgsCoordinateReferenceSystem &destCrs,
+        const QgsCoordinateTransformContext &context,
+        const QgsPoint &srcPoint );
```
wonder-sk: Do we actually need a new function? why not change existing transformPoint() to also do 3D transformation?

app/position/positionkit.cpp
```diff
@@ -35,7 +35,7 @@ PositionKit::PositionKit( QObject *parent )
 
 QgsCoordinateReferenceSystem PositionKit::positionCRS()
 {
-  return QgsCoordinateReferenceSystem::fromEpsgId( 4326 );
+  return QgsCoordinateReferenceSystem::fromEpsgId( 9707 );
```
wonder-sk, line 38: please add a comment what EPSG:9707 means, so that we don't need to look it up every time :slightly_smiling_face:

app/position/providers/androidpositionprovider.cpp
```diff
@@ -53,14 +55,17 @@ void jniOnPositionUpdated( JNIEnv *env, jclass clazz, jint instanceId, jobject l
   {
     const jdouble value = location.callMethod<jdouble>( "getAltitude" );
     if ( !qFuzzyIsNull( value ) )
-      pos.elevation = value;
+    {
+      // transform the altitude from WGS84 ellipsoid to specified geoid model
+      const QgsPoint geoidPosition = InputUtils::transformPoint3D(
+                                       QgsCoordinateReferenceSystem::fromEpsgId( 4979 ),
```
wonder-sk: also please add a comment what EPSG:4979 stands for (i.e. WGS 84 (EPSG:4326) + ellipsoid height)

app/position/providers/internalpositionprovider.cpp
```diff
@@ -140,9 +141,15 @@ void InternalPositionProvider::parsePositionUpdate( const QGeoPositionInfo &posi
     positionDataHasChanged = true;
   }
 
-  if ( !qgsDoubleNear( position.coordinate().altitude(), mLastPosition.elevation ) )
+  // transform the altitude from WGS84 ellipsoid to specified geoid model
+  const QgsPoint geoidPosition = InputUtils::transformPoint3D(
+                                   QgsCoordinateReferenceSystem::fromEpsgId( 4979 ),
```
wonder-sk: CRS comment here as well please ... just thinking, maybe it would be useful to have the following extra static methods in PositionKits, e.g.:
- `positionCRS_2D()`  - this would return EPSG:4326
- `positionCRS_3D_EllipsoidHeight()` - this would return EPSG:4979
- `positionCRS_3D()`  - this would be renamed from positionCRS() and return EPSG:9707

In that way, it should be always clear what CRS we're dealing with, and avoid magic numbers repeated in the code...

app/position/providers/internalpositionprovider.cpp
```diff
@@ -140,9 +141,15 @@ void InternalPositionProvider::parsePositionUpdate( const QGeoPositionInfo &posi
...
+                                   PositionKit::positionCRS(),
+                                   QgsCoordinateTransformContext(),
```
wonder-sk: important - do not use empty transform context here

app/position/tracking/positiontrackingmanager.cpp
```diff
@@ -332,7 +332,7 @@ void PositionTrackingManager::setVariablesManager( VariablesManager *newVariable
 
 QgsCoordinateReferenceSystem PositionTrackingManager::crs() const
 {
-  return QgsCoordinateReferenceSystem::fromEpsgId( 4326 );
+  return PositionKit::positionCRS();
```
wonder-sk: could you please verify what CRS is used in the QGIS plugin when creating the tracking layer? If we're assigning 2D CRS to it upon creation, we should change that to 3D CRS...

app/qml/gps/MMGpsDataDrawer.qml
```diff
@@ -191,28 +191,29 @@ MMComponents.MMDrawer {
...
-              if ( !__positionKit.hasPosition || Number.isNaN( __positionKit.altitude ) ) {
+              if ( !PositionKit.hasPosition || Number.isNaN( PositionKit.altitude ) ) {
                 return qsTr( "N/A" )
               }
-              __inputUtils.formatNumber( __positionKit.altitude, 2 ) + " m"
+              __inputUtils.formatNumber( PositionKit.altitude, 2 ) + " m"
             }
 
             alignmentRight: Positioner.index % 2 === 1
+            desc: qsTr("Final value calculated with EGM96_15 geoid model.")
```
wonder-sk:
```suggestion
            desc: qsTr("Orthometric height, using EGM96 geoid")
```

vcpkg/ports/qtpositioning/ios_wgs84_ellipsoid.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/src/plugins/position/corelocation/qgeopositioninfosource_cl.mm b/src/plugins/position/corelocation/qgeopositioninfosource_cl.mm
+index 95d51a86..7529df9d 100644
+--- a/src/plugins/position/corelocation/qgeopositioninfosource_cl.mm
++++ b/src/plugins/position/corelocation/qgeopositioninfosource_cl.mm
+@@ -51,7 +51,7 @@
+     // Construct position info from location data
+     QGeoPositionInfo location(QGeoCoordinate(newLocation.coordinate.latitude,
+                                              newLocation.coordinate.longitude,
+-                                             newLocation.altitude),
++                                             newLocation.ellipsoidalAltitude),
```
wonder-sk, line 10: is this safe? I have read `ellipsoidalAltitude` is only available in iOS 15+ ... do we support older versions? if we do, what happens there?

Also, what happens with external GNSS with mock location - is the ellipsoidal altitude populated correctly?

CMakeLists.txt
```diff
@@ -402,6 +402,9 @@ file(
   PATTERN "*.cmake" EXCLUDE
   PATTERN "vcpkg*" EXCLUDE
 )
+file(COPY ${CMAKE_CURRENT_SOURCE_DIR}/app/assets/us_nga_egm96_15.tif
+     DESTINATION ${ASSETS_DIR_PATH}/app/android/assets/qgis-data/proj
+)
```
wonder-sk, line 407: just wondering, are we shipping any other assets from PROJ with the app? I expect we should be shipping at least `proj.db` (~9mb sqlite file), as that's the whole database of projections, and there are typically also some more files (e.g. see `/usr/share/proj` directory on your PC). If we do ship those other files, there's a good chance we also ship `egm96_15.gtx` file, which is equivalent of `us_nga_egm96_15.tif`, just in different format, so if that's the case, maybe we don't need to add this TIF file...

app/inpututils.cpp
Withalion: I guess this is worth to do as a separate PR and go through the whole code base. I'll create a new issue for this and try to tackle it.

app/position/positionkit.cpp
Withalion: There's a comment in header :)

app/inpututils.cpp
Withalion: see  #4221

vcpkg/ports/qtpositioning/ios_wgs84_ellipsoid.patch
Withalion: - We support only iOS 16+, so there's no need to check for lower version.
- I'll have to check the external GNSS behaviour, haven't thought much about that.

CMakeLists.txt
Withalion: I've already check and we don't ship any transformation files with our version of proj. The vcpkg version of proj is much more bare bones compared to normal install. We ship these files:
- `CH`
- `copyright`
- `deformation_model.schema.json`
- `GL27`
- `ITRF2000`
- `ITRF2008`
- `ITRF2014`
- `ITRF2020`
- `nad.lst`
- `nad27`
- `nad83`
- `other.extra`
- `proj.db`
- `proj.ini`
- `projjson.schema.json`
- `triangulation.schema.json`
- `usage`
- `world`

wonder-sk: great, thanks for checking!

app/position/tracking/positiontrackingmanager.cpp
Withalion: Currently plugin sets the CRS as `EPSG:4326` which is WGS84 2D CRS. I'll create an issue to change it from `EPSG:4326` to 3D CRS.

Withalion: see MerginMaps/qgis-plugin#838

vcpkg/ports/qtpositioning/ios_wgs84_ellipsoid.patch
Withalion: @wonder-sk there is not an easy way to debug that without access to the device, I'll ask our testing team to try it. However, I would postpone this issue for a bit and the whole bundle of enhancements will include also fix for this.

PR #4211, opened 2025-11-11
Validate polygon split geometry before committing changes

### LINKED ISSUE #4187: Prevent Split polygon cancellation with early validation instead of post-completion error
**Current Behavior:**
When splitting a polygon, if the split line is drawn inside the polygon without crossing it, users can complete the drawing and press "Done". Only then does the "We could not split the feature" error dialog appear, which cancels the entire split operation and disrupts the workflow.

**Problem:**
This delayed validation forces users to restart the entire split operation from scratch, which is frustrating and time-consuming, especially when working with complex geometries splits.

**Proposed Enhancement:**
Implement real-time validation that:

1. Detects when the split line does not properly cross the polygon boundary when hitting "Done".
2. Displays a warning message immediately and do nothing if split invalid or disables "Done" button until the split line is valid
3. Prevents operation from being totally cancelled 

Screenshots:
|Invalid Split | Workflow disrupt |
|--------|--------|
||  |

VitorVieiraZ:
- Changed `commitSplit()` return type from bool to int to support specific error codes: returns 0 on success, 1 when split line doesn't properly cross feature boundary, and 2 for other split operation failures
- Added `canCommitSplit()` private method to validate if the split line intersects the feature geometry and if start and end points are outside the feature boundary
- Message for previous `MMSplittingFailedDialog` made more generic for unsuccessful splits

https://github.com/user-attachments/assets/f08250e2-7cf9-43ba-aeab-e25eb689ec55

Resolves #4187

### DIFF
--- a/app/maptools/splittingmaptool.h
+++ b/app/maptools/splittingmaptool.h
@@ -25,6 +25,15 @@ class SplittingMapTool : public AbstractMapTool
     Q_PROPERTY( FeatureLayerPair featureToSplit READ featureToSplit WRITE setFeatureToSplit NOTIFY featureToSplitChanged )
 
   public:
+
+    enum SplitResult
+    {
+      Success = 0,
+      InvalidSplit = 1, // Split line does not properly cross the feature boundary
+      Failed = 2
+    };
+    Q_ENUM( SplitResult )
+
     explicit SplittingMapTool( QObject *parent = nullptr );
     virtual ~SplittingMapTool() override;
 
@@ -43,13 +52,8 @@ class SplittingMapTool : public AbstractMapTool
     //! Returns true if the captured geometry has enought points for the specified layer
     Q_INVOKABLE bool hasValidGeometry() const;
 
-    /**
-     * Splits the feature with recorded geometry and commits the changes to the layer
-     * Returns 0 if splitting was successful
-     * Returns 1 if the split line does not properly cross the feature boundary
-     * Returns 2 if the split operation failed
-     */
-    Q_INVOKABLE int commitSplit() const;
+    //! Splits the feature with recorded geometry and commits the changes to the layer
+    Q_INVOKABLE SplittingMapTool::SplitResult commitSplit() const;
 
     // Getters/setters
     const QgsGeometry &recordedGeometry() const;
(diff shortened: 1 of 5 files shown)


app/maptools/splittingmaptool.h
Withalion, line 1: `canCommitSplit()` feels just like utility function, which should not be exposed to QML. I would propose to keep it private and call it in the beggining of `commitSplit()`. From `commitSplit()` we could return `int` instead of bool to differ between errors, `canCommitSplit()` should return non canceling error as you defined and error from QGIS splitting could cancel splitting.

app/qml/map/MMSplittingTools.qml
Withalion, line 1: Check comment mentioned at `splittingmaptool.h`

Withalion (commented):
Besides the other comments it works nicely!

app/maptools/splittingmaptool.h
```diff
@@ -64,6 +66,12 @@ class SplittingMapTool : public AbstractMapTool
     void rebuildGeometry();
 
   private:
+    /**
+     * Validates if the recorded geometry can successfully split the feature
+     * Returns true if the split line properly crosses the feature boundary
+     */
+    bool canCommitSplit() const;
```
Withalion: maybe naming it `isValidSplit` would be better

app/maptools/splittingmaptool.h
```diff
@@ -45,9 +45,11 @@ class SplittingMapTool : public AbstractMapTool
 
     /**
      * Splits the feature with recorded geometry and commits the changes to the layer
-     * Returns true if splitting was successful, false otherwise
+     * Returns 0 if splitting was successful
+     * Returns 1 if the split line does not properly cross the feature boundary
+     * Returns 2 if the split operation failed
```
Withalion: Using Enum is starting to look like a good idea. I'll leave it up to you

app/test/testmaptools.h
```diff
@@ -26,7 +26,10 @@ class TestMapTools : public QObject
     void cleanup();
 
     void testSnapping();
+
     void testSplitting();
+    void testCanCommitSplit();
```
Withalion: I would also rename this to `testValidSplit`

app/qml/dialogs/MMSplittingFailedDialog.qml
Withalion, line 1: And this gets shown when the `commitSplit()` returns 2 right?

VitorVieiraZ: Exactly, thus a more general message

PR #4216, opened 2025-11-15
Geoid separation support for internal providers

### LINKED ISSUE #2725: Get geoid separation (undulation) from external GNSS
In some cases it may be useful to extract geoid separation (the difference between the ellisoid and orthometric height) which GNSS applied to the originally calculated ellipsoid height. This could be a new variable available to users - and it could be also reported in the GPS info page.

This information is already available in QgsGpsInformation::elevation_diff (since QGIS 3.18)

Withalion:
fixes #2725 

This PR is a continuation from #4176 and also includes the changes introduced by #4210. In short we add the geoid separation calculation for internal, android and simulated providers.

For external providers we also recalculate orthometric elevation to use our model.

### DIFF
--- a/app/position/providers/androidpositionprovider.cpp
+++ b/app/position/providers/androidpositionprovider.cpp
@@ -53,18 +53,18 @@ void jniOnPositionUpdated( JNIEnv *env, jclass clazz, jint instanceId, jobject l
 
   if ( location.callMethod<jboolean>( "hasAltitude" ) )
   {
-    const jdouble value = location.callMethod<jdouble>( "getAltitude" );
-    if ( !qFuzzyIsNull( value ) )
+    const jdouble ellipsoidHeight = location.callMethod<jdouble>( "getAltitude" );
+    if ( !qFuzzyIsNull( ellipsoidHeight ) )
     {
       // transform the altitude from EPSG:4979 (WGS84 (EPSG:4326) + ellipsoidal height) to specified geoid model
       const QgsPoint geoidPosition = InputUtils::transformPoint(
                                        PositionKit::positionCrs3DEllipsoidHeight(),
                                        PositionKit::positionCrs3D(),
                                        QgsProject::instance()->transformContext(),
-      {longitude, latitude, value} );
+      {longitude, latitude, ellipsoidHeight} );
       pos.elevation = geoidPosition.z();
 
-      const double geoidSeparation = value - geoidPosition.z();
+      const double geoidSeparation = ellipsoidHeight - geoidPosition.z();
       pos.elevation_diff = geoidSeparation;
     }
   }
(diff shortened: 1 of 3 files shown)


app/position/positionkit.cpp
```diff
@@ -40,7 +40,7 @@ QgsCoordinateReferenceSystem PositionKit::positionCrs3D()
 
 QgsCoordinateReferenceSystem PositionKit::positionCrs2D()
 {
-  return QgsCoordinateReferenceSystem::fromEpsgId( 4326 );
+  return QgsCoordinateReferenceSystem::fromEpsgId( 9707 );
```
wonder-sk: 🤯  ... looks like there are a couple of things to fix after rebase 🙂

Withalion: oops, yeah

Withalion:
@wonder-sk it should be cleaned up now 👍🏻

app/position/providers/androidpositionprovider.cpp
wonder-sk: it may be nice to rename `value` to e.g. `ellipsoidHeight` or `ellipsoidAltitude` so it's easier to read the code

app/position/providers/internalpositionprovider.cpp
```diff
@@ -153,6 +153,13 @@ void InternalPositionProvider::parsePositionUpdate( const QGeoPositionInfo &posi
     positionDataHasChanged = true;
   }
 
+  const double geoidSeparation = position.coordinate().altitude() - geoidPosition.z();
```
wonder-sk: I would suggest adding some more comments:

```suggestion
  // QGeoCoordinate::altitude() docs claim that it is above the sea level (i.e. geoid) altitude,
  // but that's not really true in our case:
  // - on Android - it is MSL altitude only if "useMslAltitude" parameter is passed to the Android
  //   Qt positioning plugin, which we don't do - see https://doc.qt.io/qt-6/position-plugin-android.html
  // - on iOS - it would return MSL altitude, but we have a custom patch in vcpkg to return
  //   ellipsoid altitude (so we do not rely of geoid model of unknown quality/resolution)
  const double ellipsoidAltitude = position.coordinate().altitude();

  const double geoidSeparation = ellipsoidAltitude - geoidPosition.z();
```

app/position/providers/simulatedpositionprovider.cpp
```diff
@@ -84,16 +88,22 @@ void SimulatedPositionProvider::generateRadiusPosition()
   position.latitude = latitude;
   position.longitude = longitude;
 
-  double altitude = ( *mGenerator )() % 40 + 20; // rand altitude <20,55>m and lost (0)
-  if ( altitude <= 55 )
+  double altitude = ( *mGenerator )() % 40 + 80; // rand altitude <80,120>m and lost (0)
+  if ( altitude <= 120 )
```
wonder-sk: is there any point of this "if"? it seems that now it will be always true if I calculate it correctly. Not sure if the code previously made much sense (altitude=0 meaning no altitude?)

Withalion: Looks like I misunderstood the code a bit too. Hopefully now it makes more sense

PR #4229, opened 2025-11-21
Better paths sanitization

tomasMizera:
This PR adds some enhancements to the sanitization logic:
 - whitespace in individual folders/files within paths is now trimmed, e.g. `/ my folder / my file.ext` ~> `/my folder/my file.ext`
 - whitespace before file extension is now trimmed, e.g. `/my folder/my file .ext` ~> `/my folder/my file.ext`

These scenarios were possible because of the custom photo name formats. It then caused problems when syncing to desktop builds or server did not allows such an upload at all. 

These two enhancements are used in:
 - `projectwizard.cpp` -> sanitize project name so it does not include a whitespace at the end
 - `attributecontroller.cpp` -> when renaming photos

Fixes https://github.com/MerginMaps/mobile/issues/3976
Fixes https://github.com/MerginMaps/mobile/issues/4022
~~Fixes (likely) https://github.com/MerginMaps/mobile/issues/4157 - needs to be tested (cc @RastoHu @IvaKuklica)~~ nope
See also https://github.com/MerginMaps/qgis-plugin/issues/693

Qt handles path separators automatically, but would be good to test this out on Windows, just to be sure.

### DIFF
--- a/app/inpututils.h
+++ b/app/inpututils.h
@@ -637,8 +637,8 @@ class InputUtils: public QObject
      * Splits path into components and sanitizes each component using sanitizeFileName().
      */
     static void sanitizePath( QString &path );
-    
-    /**    
+
+    /**
      * Returns the indexes of fields that are referenced in an expression
      */
     Q_INVOKABLE static QSet<int> referencedAttributeIndexes( QgsVectorLayer *layer, const QString &expression );


app/inpututils.cpp
```diff
@@ -1969,6 +1969,59 @@ void InputUtils::sanitizeFileName( QString &fileName )
   const thread_local QRegularExpression illegalChars( QStringLiteral( "[\x00-\x19<>:|?*\"]" ) );
   fileName.replace( illegalChars, QStringLiteral( "_" ) );
   fileName = fileName.trimmed();
+
+  // Trim whitespace immediately before the final extension, e.g. "name  .jpg" -> "name.jpg"
+  int lastDot = fileName.lastIndexOf( QChar( '.' ) );
```
Withalion:
```suggestion
  const int lastDot = fileName.lastIndexOf( QChar( '.' ) );
```
also we recast from long long to int

app/inpututils.cpp
```diff
@@ -1969,6 +1969,59 @@ void InputUtils::sanitizeFileName( QString &fileName )
...
+  if ( lastDot > 0 )
+  {
+    QString base = fileName.first( lastDot );
+    QString ext = fileName.sliced( lastDot );
+    base = base.trimmed();
+    fileName = base + ext;
```
Withalion:
```suggestion
    const QString base = fileName.first( lastDot ).trimmed();
    const QString ext = fileName.sliced( lastDot );
    fileName = base + ext;
```

app/inpututils.cpp
```diff
@@ -1969,6 +1969,59 @@ void InputUtils::sanitizeFileName( QString &fileName )
...
+  }
+}
+
+void InputUtils::sanitizePath( QString &path )
+{
+  const bool pathStartsWithFileURL = path.startsWith( "file://" );
```
Withalion, line 1985: can be replaced with -> https://doc.qt.io/qt-6/qurl.html#isLocalFile

app/inpututils.cpp
```diff
@@ -1969,6 +1969,59 @@ void InputUtils::sanitizeFileName( QString &fileName )
...
+
+  if ( pathStartsWithFileURL )
+  {
+    // remove file:// prefix before sanitization
+    path.remove( 0, 7 );
```
Withalion, line 1990: Qt has also function for this -> https://doc.qt.io/qt-6/qurl.html#toLocalFile

app/inpututils.cpp
```diff
@@ -1969,6 +1969,59 @@ void InputUtils::sanitizeFileName( QString &fileName )
...
+  }
+
+  const bool pathStartsWithSlash = path.startsWith( '/' );
+
+  QStringList parts = path.split( '/', Qt::SkipEmptyParts );
```
Withalion:
```suggestion
  const QStringList parts = path.split( '/', Qt::SkipEmptyParts );
```

app/inpututils.cpp
```diff
@@ -1969,6 +1969,59 @@ void InputUtils::sanitizeFileName( QString &fileName )
...
+  if ( pathStartsWithSlash )
+  {
+    // restore leading slash
+    sanitizedPath = '/' + sanitizedPath;
+  }
+
+  if ( pathStartsWithFileURL )
+  {
+    // restore file:// prefix
+    sanitizedPath = "file://" + sanitizedPath;
```
Withalion, line 2020: again let's use Qt's function -> https://doc.qt.io/qt-6/qurl.html#fromLocalFile

app/inpututils.cpp
tomasMizera: Qt seems to have problems with schemes like `file://relative_path`; it works well with `file://absolute_path` though.

Withalion: that is unfortunate :( the code would be more explicit as we are just dealing with `QUrl` schemes

tomasMizera:
Hi @RastoHu @IvaKuklica, here are the builds to test:

🍎 iOS: 25.11.828011
🤖 Android: [733911](https://play.google.com/apps/test/RQrgmETMGFo/ahAO29uNQ3cmpt9kVRSfcPynO_LUlAjvl3eDCKAyq3kuUUMeaykMHXYO-lCV564je7eHAh_cf3c6dBHkbrltJz_8MH)
🪟 Win: Exe from artefacts [here](https://github.com/MerginMaps/mobile/actions/runs/19640387316?pr=4229)

Issue #4221, opened 2025-11-18
Empty QgsCoordinateTransformContext in function calls


Withalion:
**Description**
Currently we are calling QGIS coordinate transform functions with new/empty `QgsCoordinateTransformContext`, which is frowned upon by QGIS itself. It might introduce some issue when users setup custom transforms in QGIS as mentioned by @wonder-sk in [here](https://github.com/MerginMaps/mobile/pull/4210#discussion_r2533822516)

We should use current `QgsCoordinateTransformContext` from active project.

PR #4234, opened 2025-11-24
GNSS enhancements bundle

Withalion:
This PR includes multiple improvements to GNSS data handling, elevation handling and some more improvements.

1. #4176 Adds geoid separation value to GPS information panel. The value is gathered from external GNSS devices. This value can be used to recalculate ellipsoid elevation to geoid elevation or vice versa. Also adds `geoid_separation` variable.
2. #4210 Adds transformation of ellipsoidal elevation to geoid elevation with the default usage of EGM96 geoid model. This transformation is applied for internal provider (both iOS and android), fused provider on android, position tracking. 
photo showing GPS panel in Mergin Maps, with dialog open, which specifies that EGM96 model is used to calculate elevation

3. #4228 Sometimes we didn't use `QgsCoordinateTransformContext` from the project, which would fail transformations if users used custom transformation.
4. #4216 Introduced slight change for external providers. We don't trust the models used in GNSSes, the received data is tranformed to ellipsoid elevation and then again to geoid elevation using our model. Geoid separation is calculated by us. Besides this, the internal, android and simulated providers also return geoid separation data.
5. Mock location detection - when using android position provider (fused in application) we detect if the position is mocked and we show this to users.
GPS panel in Mergin Maps showing 'Mocked position provider' info

6. #4238 In some cases using a more precise local geoid model is necessary for users. This PR added this support if a specific geoid model is specified for the project. The geoid model has to be specified in qgis plugin and it's grid shift file has to be packaged with the project to work correctly. Users will see that the project is using custom geoid model in GPS info panel.

|Position inside of geoid area | Position outside of geoid area |
|-----|------|
|photo showing elevation and elevation separation info when using local geoid model | photo showing uknown elevation and elevation separation info when using local geoid model |

### Current state:
- new form variable `geoid_separation` => should be used as `@position_geoid_separation`
- new form variable `elevation_ellipsoid` => should be used as `@position_elevation_ellipsoid`
- `altitude` form variable changed to `elevation` => should be used as `@position_elevation`
- bluetooth (external GNSS) provider
  - we show raw value from GNSS device by default (can be unset in plugin)  
  - other option is to show geoid (orthometric) elevation using custom model if setup (in plugin)
  - the provided data gets transformed to ellipsoid elevation and then using our model to geoid elevation again
  - we show geoid separation (undulation) when elevation transformation is turned on
- internal provider
  - iOS
    - we show geoid (orthometric) elevation now using EGM96 geoid model (by default, bundled with app) or using custom model if setup (in plugin)
    - we also show geoid separation (undulation)
    - we detect mock position and show it to user
    - for mocked position we either directly show the geoid elevation from GNSS or transform it using the selected model if setup (in plugin)
  - Android
    - we show geoid (orthometric) elevation now using EGM96 geoid model (by default, bundled with app) or using custom model if setup (in plugin)
    - we also show geoid separation (undulation)
  - Windows (other desktops)
    - we show the raw value from provider (it should be geoid elevation (applies only to Windows))
- fused provider
  - became default provider on Android 
  - we show geoid (orthometric) elevation now using EGM96 geoid model (by default, bundled with app) or using custom model if setup (in plugin)
  - we also show geoid separation (undulation)
  - we detect mock position and show it to user
  - for mocked position we either directly show the value from GNSS (default behaviour) or transform it using the selected model if setup (in plugin)
- simulated provider
  - we show geoid (orthometric) elevation now using EGM96 geoid model (by default, bundled with app) or using custom model if setup (in plugin)
  - we also show geoid separation (undulation)

Besides all of these, the altitude value in GPS info panel should show the elevation received from provider minus the antenna height. The same should apply for all elevation variables.

A tip for users setting up external GNSS receivers is to select fused provider on Android and internal provider on iOS, as these providers will show if mocked location is detected.

Some further tips discovered in testing:
 - if you are using mock on Android and you want to transform the elevation with specific geoid model, make sure the app is receiving ellipsoid elevation from receiver
 - if your receiver (app) is propagating msl elevation don't enable transformations (you'll get double transformation)

@wonder-sk soft ping if you want to have another look

### DIFF
--- a/app/test/testposition.cpp
+++ b/app/test/testposition.cpp
@@ -607,7 +607,7 @@ void TestPosition::testPositionTransformerBluetoothPosition()
   QCOMPARE( newPosition.elevation_diff, 40 );
 
   // transform with pass through disabled and elevation separation
-  newPosition = positionTransformer.processAndroidPosition( geoPosition );
+  newPosition = positionTransformer.processBluetoothPosition( geoPosition );
 
   QVERIFY( qgsDoubleNear( newPosition.latitude, 48.10305 ) );
   QVERIFY( qgsDoubleNear( newPosition.longitude, 17.1064 ) );


Withalion:
@IvaKuklica for testing #4165 #2615 #2725 #4221 feel free to use builds here:
### 2025.8.0 
- iOS: 25.11.827911
- Android: 
  - [arm-v8](https://play.google.com/apps/test/RQrgmETMGFo/ahAO29uNSmqx_Y1xlpDldJ50rIQ91uYi4uB9C3fiXW-qyeVeClLadhHP9YtG4s0giOgrfsABrfGaY-sUpyiVvGBhc7) - 733851
  - [arm-v7](https://play.google.com/apps/test/RQrgmETMGFo/ahAO29uNSLFLDAUyZHSaAYYTh-FezgzZbTtI6xFa3LdHM-BAH3KBdk60FKxC0tMBpVEZhZjUGTqe_Vtov4TVPfUOID) - 733811

app/position/providers/bluetoothpositionprovider.cpp
```diff
@@ -204,11 +204,23 @@ void BluetoothPositionProvider::positionUpdateReceived()
...
-    emit positionChanged( GeoPosition::fromQgsGpsInformation( data ) );
+    const QByteArray rawNmea = mSocket->readAll();
+    const QString nmea( rawNmea );
+
+    const QgsGpsInformation data = mNmeaParser.parseNmeaString( nmea );
+    GeoPosition positionData = GeoPosition::fromQgsGpsInformation( data );
+
+    // The geoid models used in GNSS devices can be often times unreliable, thus we apply the transformations ourselves
+    // GNSS supplied orthometric elevation -> ellipsoid elevation -> orthometric elevation based on our model
+    const double ellipsoidElevation = positionData.elevation + positionData.elevation_diff;
```
tomasMizera: What if `elevation` or `elevation_diff` will, for whatever reason, not be reported? Do you think it makes sense to try to handle this scenario? I can imagine we set a default value of `GeoPosition::elevation` and `GeoPosition::elevation_diff` to `std::numeric_limits< double >::quiet_NaN();` and then compare with `!std::isnan`. If any of the two (or both) would be invalid, we skip the transformation and report error log.

app/position/providers/bluetoothpositionprovider.cpp
```diff
@@ -204,11 +204,23 @@ void BluetoothPositionProvider::positionUpdateReceived()
...
+    const QgsPoint geoidPosition = InputUtils::transformPoint(
+                                     PositionKit::positionCrs3DEllipsoidHeight(),
+                                     PositionKit::positionCrs3D(),
+                                     QgsProject::instance()->transformContext(),
```
tomasMizera: I know we previously tried to avoid using `QgsProject::instance()` calls wherever possible. Not sure what is the current state in QGIS, maybe @wonder-sk can comment. 

If we wanted to avoid using it; I'd suggest we report the data from position providers in whatever CRS and let providers also report the source CRS; later in `position kit` - who I think has access to the active QGIS project - we could handle the transformation.

tomasMizera: The same comment applies to multiple places

Withalion: The conclusion of our discussion was that moving away from `QgsProject::instance()` is QGIS core long term goal, but nothing we should be concerned about for now

app/variablesmanager.cpp
tomasMizera, line 1: @Withalion I am thinking that it would be better to subtract the antenna height in the `PositionKit` class instead. That way it will propagate everywhere without us needing to manually subtract it on multiple places.

I am afraid we'd forget to subtract it somewhere.. for example `@position_coordinate` would currently return the original elevation and not the subtracted one.

Withalion: It should be changed now!

app/position/positionkit.cpp
```diff
@@ -148,7 +194,11 @@ AbstractPositionProvider *PositionKit::constructActiveProvider( AppSettings *app
   {
     if ( InputUtils::isMobilePlatform() )
     {
+#ifdef ANDROID
+      return constructProvider( QStringLiteral( "internal" ), QStringLiteral( "android_fused" ) );
```
tomasMizera, line 186: Let's check if fused is available on the device

Withalion: We already check for that in `constructProvider` the fallback for that is the internal provider

app/map/inputcoordinatetransformer.cpp
```diff
@@ -13,12 +13,14 @@
...
 #include "inputcoordinatetransformer.h"
+
+#include "positionkit.h"
 #include "qgslogger.h"
 
 InputCoordinateTransformer::InputCoordinateTransformer( QObject *parent )
   : QObject( parent )
 {
-  mCoordinateTransform.setSourceCrs( QgsCoordinateReferenceSystem::fromEpsgId( 4326 ) );
+  mCoordinateTransform.setSourceCrs( PositionKit::positionCrs3D() );
```
tomasMizera: TODO: Do we need to have a dedicated `positionCrs`? Why not only `positionCrs()`?

app/position/providers/androidpositionprovider.cpp
```diff
@@ -49,17 +51,54 @@ void jniOnPositionUpdated( JNIEnv *env, jclass clazz, jint instanceId, jobject l
...
+    if ( !qFuzzyIsNull( altitude ) )
+    {
+      bool positionOutsideGeoidModelArea = false;
+      bool valueRead = false;
+      const bool isVerticalCRSPassedThrough = QVariant( QgsProject::instance()->readEntry( QStringLiteral( "Mergin" ), QStringLiteral( "VerticalCRSPassThrough" ), QVariant( true ).toString(), &valueRead ) ).toBool();
+      // transform the altitude from EPSG:4979 (WGS84 (EPSG:4326) + ellipsoidal height) to specified geoid model
+      // (by default EPSG:9707 (WGS84 + EGM96))
+      // we do the transformation only in case the position is not mocked, and it's ellipsoidal altitude
+      // the second variant is when the position is mocked, the altitude is ellipsoidal plus pass through is enabled
+      if ( !isMock || ( isMock && valueRead && !isVerticalCRSPassedThrough ) )
```
tomasMizera: This should mean the same

```suggestion
      if ( !isMock || ( valueRead && !isVerticalCRSPassedThrough ) )
```

app/position/providers/androidpositionprovider.cpp
```diff
@@ -49,17 +51,54 @@ void jniOnPositionUpdated( JNIEnv *env, jclass clazz, jint instanceId, jobject l
...
 
-  // TODO: we are getting ellipsoid elevation here. From API level 34 (Android 14),
-  // there is AltitudeConverter() class in Java that can be used to add MSL altitude
-  // to Location object. How to deal with this correctly? (we could also convert
-  // to MSL (orthometric) altitude ourselves if we add geoid model to our APK
+  if ( location.callMethod<jboolean>( "hasAltitude" ) )
+  {
+    const jdouble altitude = location.callMethod<jdouble>( "getAltitude" );
+    if ( !qFuzzyIsNull( altitude ) )
+    {
```
tomasMizera: TODO: shouldn't we move the whole transformation logic outside of providers so they do not depend on QgsProject and PositionKit?

app/position/providers/bluetoothpositionprovider.cpp
```diff
@@ -204,11 +207,33 @@ void BluetoothPositionProvider::positionUpdateReceived()
...
+    const QByteArray rawNmea = mSocket->readAll();
+    const QString nmea( rawNmea );
 
-    QgsGpsInformation data = mNmeaParser.parseNmeaString( nmea );
+    const QgsGpsInformation data = mNmeaParser.parseNmeaString( nmea );
+    GeoPosition positionData = GeoPosition::fromQgsGpsInformation( data );
 
-    emit positionChanged( GeoPosition::fromQgsGpsInformation( data ) );
+    bool valueRead = false;
+    const bool isVerticalCRSPassedThrough = QVariant( QgsProject::instance()->readEntry( QStringLiteral( "Mergin" ), QStringLiteral( "VerticalCRSPassThrough" ), QVariant( true ).toString(), &valueRead ) ).toBool();
```
tomasMizera: This is a more generic comment. I am thinking about moving the transformation logic outside of individual position providers to dedicated transformator methods/classes. This has several benefits: 
 - no dependency on `QgsProject::instance` and `PositionKit` inside position providers
 - easier to unit-test the complex transformation logic with all the edge cases

Additionally, let's make sure we do not read from `QgsProject` on each position update (Imagine we can get several updates in a second).

app/position/providers/bluetoothpositionprovider.cpp
```diff
@@ -204,11 +207,33 @@ void BluetoothPositionProvider::positionUpdateReceived()
...
+    // if the user sets custom vertical crs we apply our transformation if not we propagate the value from GNSS device
+    if ( valueRead && !isVerticalCRSPassedThrough )
+    {
+      // The geoid models used in GNSS devices can be often times unreliable, thus we apply the transformations ourselves
+      // GNSS supplied orthometric elevation -> ellipsoid elevation -> orthometric elevation based on our model
+      const double ellipsoidElevation = positionData.elevation + positionData.elevation_diff;
```
tomasMizera: I would suggest to add a check if `elevation` or `elevation_diff` are null and if so, skip the transformation. Chances are that while receivers are booting up, they might not report these values.

app/position/providers/internalpositionprovider.cpp
```diff
@@ -128,69 +134,138 @@ void InternalPositionProvider::parsePositionUpdate( const QGeoPositionInfo &posi
...
+  bool positionOutsideGeoidModelArea = false;
+#ifdef Q_OS_IOS
+  // on ios we can get both ellipsoid and geoid altitude, depending on what is available we transform the altitude or not
+  // we also check if the user set vertical CRS pass through in plugin, which prohibits any transformation
+  bool valueRead = false;
+  const bool isVerticalCRSPassedThrough = QVariant( QgsProject::instance()->readEntry( QStringLiteral( "Mergin" ), QStringLiteral( "VerticalCRSPassThrough" ), QVariant( true ).toString(), &valueRead ) ).toBool();
+  const bool isEllipsoidalAltitude = localPosition.attribute( QGeoPositionInfo::VerticalSpeed );
+  localPosition.removeAttribute( QGeoPositionInfo::VerticalSpeed );
+  const bool isMockedLocation = localPosition.attribute( QGeoPositionInfo::MagneticVariation );
+  mLastPosition.isMock = isMockedLocation;
```
tomasMizera: What is `mLastPosition`? Shouldn't it be `localPosition`? 🤔

app/position/providers/internalpositionprovider.cpp
```diff
@@ -128,69 +134,138 @@ void InternalPositionProvider::parsePositionUpdate( const QGeoPositionInfo &posi
...
+  localPosition.removeAttribute( QGeoPositionInfo::MagneticVariation );
+  QgsPoint geoidPosition;
+  // transform the altitude from EPSG:4979 (WGS84 (EPSG:4326) + ellipsoidal height) to specified geoid model
+  // (by default EPSG:9707 (WGS84 + EGM96))
+  // we do the transformation only in case the position is not mocked, and it's ellipsoidal altitude
+  // the second variant is when the position is mocked, the altitude is ellipsoidal plus pass through is enabled
```
tomasMizera: Isn't it the other way around?
```suggestion
  // the second variant is when the position is mocked, the altitude is ellipsoidal plus pass through is not enabled
```

app/position/providers/internalpositionprovider.cpp
```diff
@@ -128,69 +134,138 @@ void InternalPositionProvider::parsePositionUpdate( const QGeoPositionInfo &posi
...
     positionDataHasChanged = true;
   }
 
-  if ( !qgsDoubleNear( position.coordinate().altitude(), mLastPosition.elevation ) )
+  bool positionOutsideGeoidModelArea = false;
+#ifdef Q_OS_IOS
+  // on ios we can get both ellipsoid and geoid altitude, depending on what is available we transform the altitude or not
+  // we also check if the user set vertical CRS pass through in plugin, which prohibits any transformation
+  bool valueRead = false;
+  const bool isVerticalCRSPassedThrough = QVariant( QgsProject::instance()->readEntry( QStringLiteral( "Mergin" ), QStringLiteral( "VerticalCRSPassThrough" ), QVariant( true ).toString(), &valueRead ) ).toBool();
```
tomasMizera: How about this name? I find the pass through variant a bit too hard to get my head around
```suggestion
  const bool isVerticalCRSPassedThrough = QVariant( QgsProject::instance()->readEntry( QStringLiteral( "Mergin" ), QStringLiteral( "SkipElevationTransformation" ), QVariant( true ).toString(), &valueRead ) ).toBool();
```

app/position/providers/internalpositionprovider.cpp
```diff
@@ -128,69 +134,138 @@ void InternalPositionProvider::parsePositionUpdate( const QGeoPositionInfo &posi
...
+  QgsPoint geoidPosition;
+  // transform the altitude from EPSG:4979 (WGS84 (EPSG:4326) + ellipsoidal height) to specified geoid model
+  // (by default EPSG:9707 (WGS84 + EGM96))
+  // we do the transformation only in case the position is not mocked, and it's ellipsoidal altitude
+  // the second variant is when the position is mocked, the altitude is ellipsoidal plus pass through is enabled
+  if ( ( !isMockedLocation && isEllipsoidalAltitude ) || ( isMockedLocation && isEllipsoidalAltitude && valueRead && !isVerticalCRSPassedThrough ) )
   {
-    mLastPosition.elevation = position.coordinate().altitude();
-    positionDataHasChanged = true;
+    // transform the altitude from EPSG:4979 (WGS84 (EPSG:4326) + ellipsoidal height) to
```
tomasMizera: to? :)

app/position/providers/internalpositionprovider.cpp
tomasMizera: Some other options "UsesCustomVerticalCRS", "UsesDefaultElevationTransformations", "IsCustomVerticalCRSSet" 🤔

app/position/providers/internalpositionprovider.cpp
```diff
@@ -128,69 +134,138 @@ void InternalPositionProvider::parsePositionUpdate( const QGeoPositionInfo &posi
...
+  localPosition.removeAttribute( QGeoPositionInfo::VerticalSpeed );
+  const bool isMockedLocation = localPosition.attribute( QGeoPositionInfo::MagneticVariation );
+  mLastPosition.isMock = isMockedLocation;
+  localPosition.removeAttribute( QGeoPositionInfo::MagneticVariation );
+  QgsPoint geoidPosition;
+  // transform the altitude from EPSG:4979 (WGS84 (EPSG:4326) + ellipsoidal height) to specified geoid model
+  // (by default EPSG:9707 (WGS84 + EGM96))
+  // we do the transformation only in case the position is not mocked, and it's ellipsoidal altitude
+  // the second variant is when the position is mocked, the altitude is ellipsoidal plus pass through is enabled
+  if ( ( !isMockedLocation && isEllipsoidalAltitude ) || ( isMockedLocation && isEllipsoidalAltitude && valueRead && !isVerticalCRSPassedThrough ) )
```
tomasMizera: Just a tip for better readability - sometimes I like to divide complicated conditions into multiple smaller booleans and then combine the booleans in the final condition. Something like this:

```c++
bool internalProviderUsesEllipsoid = !isMockedLocation && isEllipsoidalAltitude;
bool mockedProviderUsesEllipsoid = isMockedLocation && isEllipsoidalAltitude && valueRead

if ( internalProviderUsesEllipsoid || ( mockedProviderUsesEllipsoid && !isVerticalCRSPassedThrough ) ) ...
```

app/position/providers/internalpositionprovider.cpp
```diff
@@ -128,69 +134,138 @@ void InternalPositionProvider::parsePositionUpdate( const QGeoPositionInfo &posi
...
+  if ( !qgsDoubleNear( localPosition.coordinate().longitude(), mLastPosition.longitude ) )
   {
-    mLastPosition.longitude = position.coordinate().longitude();
+    mLastPosition.longitude = localPosition.coordinate().longitude();
     positionDataHasChanged = true;
   }
 
-  if ( !qgsDoubleNear( position.coordinate().altitude(), mLastPosition.elevation ) )
+  bool positionOutsideGeoidModelArea = false;
+#ifdef Q_OS_IOS
```
tomasMizera, line 150: For the sake of better readability, this whole section would benefit from having some whitespaces (newlines) here and there to separate logic blocks

app/position/providers/internalpositionprovider.cpp
```diff
@@ -119,7 +121,11 @@ void InternalPositionProvider::parsePositionUpdate( const QGeoPositionInfo &posi
   // emit connected signal here to know that the connection is OK
   setState( tr( "Connected" ), State::Connected );
 
-  bool hasPosition = position.coordinate().isValid();
+  // we create a local copy of position because on iOS we use QGeoPositionInfo::VerticalSpeed attribute as helper value
+  // for transformation, we need to remove it afterwards
+  QGeoPositionInfo localPosition( position );
```
tomasMizera, line 126: Is this `position` used anywhere?

app/variablesmanager.cpp
```diff
@@ -72,6 +72,7 @@ QgsExpressionContextScope *VariablesManager::positionScope()
   addPositionVariable( scope, QStringLiteral( "longitude" ), position.longitude );
   addPositionVariable( scope, QStringLiteral( "latitude" ), position.latitude );
   addPositionVariable( scope, QStringLiteral( "altitude" ), position.elevation );
+  addPositionVariable( scope, QStringLiteral( "altitude_ellipsoidal" ), position.elevation - position.elevation_diff );
```
tomasMizera: If `position.elevation` is msl height (orthometric), then I think the calculation should be `position.elevation + position.elevation_diff` (https://www.researchgate.net/figure/Relationship-between-orthometric-height-ellipsoid-height-and-geoid-height_fig1_273133676)?

tomasMizera: Also, it seems that the proper term would be `ellipsoid_height`. `Altitude` implies that this is msl (geoid)

app/inpututils.cpp
```diff
@@ -954,12 +988,11 @@ QgsPoint InputUtils::mapPointToGps( QPointF mapPosition, InputMapSettings *mapSe
     return QgsPoint();
 
   QgsPoint positionMapCrs = mapSettings->screenToCoordinate( mapPosition );
-  QgsCoordinateReferenceSystem crsGPS = coordinateReferenceSystemFromEpsgId( 4326 );
 
   const QgsPointXY transformedXY = transformPoint(
                                      mapSettings->destinationCrs(),
-                                     crsGPS,
-                                     QgsCoordinateTransformContext(),
+                                     PositionKit::positionCrs2D(),
```
tomasMizera, line 968: Why 2D in this case and 3D in all other?

app/inpututils.cpp
```diff
@@ -880,32 +880,66 @@ QgsPoint InputUtils::transformPoint( const QgsCoordinateReferenceSystem &srcCrs,
...
+  }
+  catch ( QgsCsException &cse )
+  {
+    Q_UNUSED( cse )
+  }
+
+  return {};
+}
+
+QgsPoint InputUtils::transformPoint( const QgsCoordinateReferenceSystem &srcCrs,
```
tomasMizera, line 883: If we need to have both `transformPoint` methods, can we make one call the other one instead of keeping nearly identical bodies duplicated?

Something like

```c++
QgsPoint InputUtils::transformPoint( ...the 4 arguments version ... )
{
  bool fallbackHappened; // ignored
  return InputUtils::transformPoint( ...args, &fallbackHappened );
}
```

app/qml/gps/MMGpsDataDrawer.qml
```diff
@@ -62,7 +62,17 @@ MMComponents.MMDrawer {
...
+              if ( PositionKit.positionProvider ) {
+                if ( PositionKit.isMockPosition ) {
+                  qsTr( "Mocked position provider" )
+                } else {
+                  PositionKit.positionProvider.name()
+                }
+              } else {
+                qsTr( "No receiver" )
+              }
+            }
```
tomasMizera, line 71: Let's instead set the provider name directly in cpp in providers implementation so we do not add logic in QML. 

`if (position.isMock) { setName("...") }`

app/qml/gps/components/MMGpsDataText.qml
```diff
@@ -65,4 +91,13 @@ Item {
       horizontalAlignment: alignmentRight ? Text.AlignRight : Text.AlignLeft
     }
   }
+
+  MMComponents.MMPopup {
+    id: infoPopup
+    y: ( -root.height / 2 ) - __style.margin8
```
tomasMizera, line 97: Hm, wouldn't `-root/height / 2` give you a static position? What if the screen is smaller and one needs to scroll? Maybe tooltip would be a better choice here: https://doc.qt.io/qt-6/qml-qtquick-controls-tooltip.html

app/position/positionkit.cpp
```diff
@@ -205,9 +246,9 @@ void PositionKit::parsePositionUpdate( const GeoPosition &newPosition )
     hasAnythingChanged = true;
   }
 
-  if ( !qgsDoubleNear( newPosition.elevation, mPosition.elevation ) )
+  if ( !qgsDoubleNear( newPosition.elevation - antennaHeight(), mPosition.elevation ) )
```
tomasMizera, line 246: `antennaHeight()` performs `QSettings::value()` lookup on every call. AI says that it should not be a problem because Qt caches the settings values. It would be good to verify the behavior because reading twice from a file on every position update might throw in additional milliseconds.

app/position/positionkit.cpp
```diff
@@ -33,11 +33,48 @@ PositionKit::PositionKit( QObject *parent )
...
+    CoreUtils::log( QStringLiteral( "PositionKit" ), QStringLiteral( "Failed to create custom compound crs: %1" ).arg( compoundCrsError ) );
+  }
+
+  return QgsCoordinateReferenceSystem::fromEpsgId( 9707 );
+}
+
+QString PositionKit::positionCrs3DGeoidModelName() const
+{
+  bool valueRead = false;
+  const bool isVerticalCRSPassedThrough = QVariant( QgsProject::instance()->readEntry( QStringLiteral( "Mergin" ), QStringLiteral( "VerticalCRSPassThrough" ), QVariant( true ).toString(), &valueRead ) ).toBool();
```
tomasMizera: I am seeing too many `QVariants` over here :) How about using `QgsProject::readBoolEntry` instead?

app/position/positionkit.cpp
```diff
@@ -33,11 +33,48 @@ PositionKit::PositionKit( QObject *parent )
 {
 }
 
-QgsCoordinateReferenceSystem PositionKit::positionCRS()
+QgsCoordinateReferenceSystem PositionKit::positionCrs3D()
```
tomasMizera, line 37: I forgot if I commented on this previously, but if not, do even need to keep separate `positionCrs3D` and `positionCrs2D`? What if we had just `positionCrs()` that would return what the 3D returns now? Do you forsee any problems with that?

app/position/providers/internalpositionprovider.cpp
Withalion: this probably shouldn't have been there anymore as there is comment above

app/inpututils.cpp
Withalion: Let's discuss with @wonder-sk tomorrow.

app/qml/gps/MMGpsDataDrawer.qml
Withalion: stupidify 🪄

app/position/positionkit.cpp
Withalion: Are you sure about this? I checked the code and the value gets read from `AppSettings` object property. `QSettings::value()` is used only on the creation of `AppSettings`.

app/position/positionkit.cpp
Withalion: Yes there is a reason why we want to keep both. Let's discuss this with Martin.

app/position/positionkit.cpp
```diff
@@ -31,13 +31,51 @@
...
   : QObject( parent )
 {
+  refreshPositionTransformer( QgsProject::instance() );
 }
 
-QgsCoordinateReferenceSystem PositionKit::positionCRS()
+QgsCoordinateReferenceSystem PositionKit::positionCrs3D()
+{
+  bool crsExists = false;
+  const QString crsWktDef = QgsProject::instance()->readEntry( QStringLiteral( "Mergin" ), QStringLiteral( "TargetVerticalCRS" ), QString(), &crsExists );
```
tomasMizera: Let's pass in the needed stuff from `main.cpp` instead of adding dep to `QgsProject::instance()`

app/position/positionkit.h
```diff
@@ -167,11 +185,17 @@ class PositionKit : public QObject
     // stop updates when application is minimized
     void appStateChanged( Qt::ApplicationState state );
 
+    // gets triggered when active project changes
+    void refreshPositionTransformer( const QgsProject *project );
```
tomasMizera: Let's not rebuild it, just update values + make sure it refreshes after sync

app/position/positionkit.cpp
```diff
@@ -31,13 +31,45 @@
...
+QgsCoordinateReferenceSystem PositionKit::positionCrs3D()
+{
+  if ( mVerticalCrs.isValid() )
+  {
+    QString compoundCrsError{};
+    const QgsCoordinateReferenceSystem compoundCrs = QgsCoordinateReferenceSystem::createCompoundCrs( positionCrs2D(), mVerticalCrs, compoundCrsError );
+    if ( compoundCrs.isValid() && compoundCrsError.isEmpty() )
+    {
+      return compoundCrs;
+    }
```
tomasMizera: Why not doing this in `setVerticalCrs`?

app/main.cpp
```diff
@@ -544,18 +544,37 @@ int main( int argc, char *argv[] )
...
 
   // Lambda context object can be used in all lambda functions defined here,
   // it secures lambdas, so that they are destroyed when this object is destroyed to avoid crashes.
   QObject lambdaContext;
 
+  QObject::connect( &activeProject, &ActiveProject::projectReloaded, &lambdaContext, [&pk]( QgsProject * project )
+  {
+    // read and set new vertical CRS definition
+    bool crsExists = false;
+    const QString crsWktDef = project->readEntry( QStringLiteral( "Mergin" ), QStringLiteral( "TargetVerticalCRS" ), QString(), &crsExists );
```
tomasMizera, line 561: Reset verticalCrs if not set

app/position/positionkit.cpp
```diff
@@ -31,13 +31,45 @@
 PositionKit::PositionKit( QObject *parent )
   : QObject( parent )
 {
+  mPositionTransformer = std::make_unique<PositionTransformer>( positionCrs3DEllipsoidHeight(), positionCrs3D(), mSkipElevationTransformation, QgsCoordinateTransformContext() );
 }
 
-QgsCoordinateReferenceSystem PositionKit::positionCRS()
+QgsCoordinateReferenceSystem PositionKit::positionCrs3D()
+{
+  if ( mVerticalCrs.isValid() )
```
tomasMizera:
```suggestion
if ( mSkipElevationTransformation ) {
  if ( mVerticalCrs.isValid() )
```

app/map/inputcoordinatetransformer.h
tomasMizera, line 1: +1 to remove

app/test/testposition.cpp
```diff
@@ -489,3 +488,330 @@ void TestPosition::testPositionTrackingHighlight()
...
+  geoPosition.elevation_diff = 40;
+  newPosition = passThroughTransformer.processBluetoothPosition( geoPosition );
+
+  QVERIFY( qgsDoubleNear( newPosition.latitude, 48.10305 ) );
+  QVERIFY( qgsDoubleNear( newPosition.longitude, 17.1064 ) );
+  QCOMPARE( newPosition.elevation, 171.3 );
+  QCOMPARE( newPosition.elevation_diff, 40 );
+
+  // transform with pass through disabled and elevation separation
+  newPosition = positionTransformer.processAndroidPosition( geoPosition );
```
tomasMizera: Bluetooth

PR #4235, opened 2025-11-24
Fix deletion of multiple features

### LINKED ISSUE #4097: Feature to delete multiple selected features at once
Add "delete/remove" button as an option when selecting multiple features at once. The button can be in the row, next to the "edit" button.

Image

It seems like there is also no "breathing space" between the row of buttons and system buttons. Let's unify the visual with Figma design.

Figma link: https://www.figma.com/design/d78a4ehBBgK6MyxCTuCAqg/Mobile-App?node-id=4233-18310&t=KOzmmky17djKv0q1-4

gabriel-bolbotina:
Fixed issue when deleting multiple features for the second time, as described in this comment:
https://github.com/MerginMaps/mobile/issues/4097#issuecomment-3570562912

fixes #4097

### DIFF
--- a/.github/workflows/app_delivery.yml
+++ b/.github/workflows/app_delivery.yml
@@ -1,0 +1,88 @@
+name: Mergin Maps delivery
+on:
+  workflow_run:
+    types:
+      - "completed"
+    workflows:
+      - "macOS Build"
+      - "linux Build"
+      - "win64 Build"
+      - "Android Build"
+      - "iOS Build"
+
+permissions:
+  actions: read
+  pull-requests: write
+
+jobs:
+  comment_on_pr:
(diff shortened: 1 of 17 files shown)


app/multieditmanager.cpp
```diff
@@ -182,6 +182,8 @@ void MultiEditManager::deleteSelectedFeatures()
     mModel->populate( {} );
     mLayer->triggerRepaint();
     CoreUtils::log( QStringLiteral( "Multi Edit Manager" ), QStringLiteral( "Deleted %1 features from %2" ).arg( fids.size() ).arg( mLayer->name() ) );
+    // commit the changes to stop editing the layer
+    mLayer->commitChanges();
```
tomasMizera, line 184: I think it is better to first commit and then trigger repaint

PR #4245, opened 2025-12-08
Improve autosync capabilities

### LINKED ISSUE #4178: Synchronize project when the App starts
There are workflows where projects and/or data are changed overnight, so that at the start of the work day surveyors/contributors have access to an updated version of the project and/or data (in my use case the data is cleaned with some scripts and the new version pushed to MM with MM CLI tools).

In this cases is desirable that when the App opens it will automatically download the updated project, which is kind of expected anyway if the option to synch automatically is active.

Withalion:
fixes #4178 

This PR adds new triggers for autosync when allowed:
- autosync when project is opened
- autosync after 1 minute
- autosync after app is brought to foreground and last autosync is older than a minute

Besides the triggers also UX was improved. With synchronizations triggered by autosync there won't be so much "visual smog" only notification about successful sync will be shown.

### DIFF
--- a/app/activeproject.cpp
+++ b/app/activeproject.cpp
@@ -227,7 +227,7 @@ bool ActiveProject::forceLoad( const QString &filePath, bool force )
   if ( mAppSettings.autosyncAllowed() )
   {
     setAutosyncEnabled( true );
-    requestSync();
+    requestSync( SyncOptions::AutomaticRequest );
   }
 
   // in case tracking is running, we want to show the UI
--- a/app/qml/settings/MMSettingsPage.qml
+++ b/app/qml/settings/MMSettingsPage.qml
@@ -168,7 +168,7 @@ MMPage {
       MMSettingsComponents.MMSettingsSwitch {
         width: parent.width
         title: qsTr("Automatically sync changes")
-        description: qsTr("Each time you save changes, the app will sync automatically")
+        description: qsTr("Your project is synced automatically to keep you up to date")
         checked: AppSettings.autosyncAllowed
 
         onClicked: AppSettings.autosyncAllowed = !checked


app/autosynccontroller.cpp
```diff
@@ -13,12 +13,19 @@
...
+// 10 seconds
+constexpr int SYNC_CHECK_TIMEOUT = 10000;
+
 AutosyncController::AutosyncController(
   QgsProject *openedQgsProject,
   QObject *parent
 )
   : QObject( parent )
   , mQgsProject( openedQgsProject )
+// we set the current timestamp as we sync on project open
```
gabriel-bolbotina: I would change the comment to something like "set the current timestamp as we sync with the open project"

app/autosynccontroller.cpp
```diff
@@ -35,10 +42,39 @@ AutosyncController::AutosyncController(
...
+        } );
       }
     }
   }
+
+  //every 10 seconds check if last sync was more than a minute ago and sync if it's true
+  mTimer = std::make_unique<QTimer>( this );
+  connect( mTimer.get(), &QTimer::timeout, this, [&]
+  {
+    if ( QDateTime::currentDateTime() - mLastUpdateTime >= std::chrono::milliseconds( SYNC_INTERVAL ) )
```
gabriel-bolbotina, line 58: if we keep more or equal, I think you could also change the comment to more or equal to a minute ago

Withalion:
@gioman I'm happy to hear that, after some discussion with @tomasMizera we will add some more improvements to the autosync.

app/qml/main.qml
```diff
@@ -1052,7 +1052,7 @@ ApplicationWindow {
 
     function onProjectAlreadyOnLatestVersion( projectFullName )
     {
-      if ( projectFullName === __activeProject.projectFullName() )
+      if ( projectFullName === __activeProject.projectFullName() && !AppSettings.autosyncAllowed )
```
tomasMizera: Hopaaa ~ this would hide the `Up to date` message even if this was a manually-triggered sync while autosync is enabled

app/autosynccontroller.h
```diff
@@ -24,10 +26,13 @@ class AutosyncController : public QObject
...
 
+    // Set mLastUpdateTime to "now", triggered by manual sync
+    void updateLastUpdateTime();
+
   signals:
 
-    void projectChangeDetected();
-    void projectSyncRequired();
+    void projectChangeDetected( SyncOptions::RequestOrigin origin = SyncOptions::RequestOrigin::AutomaticRequest );
+    void projectSyncRequired( SyncOptions::RequestOrigin origin = SyncOptions::RequestOrigin::AutomaticRequest );
```
tomasMizera: Let's unify into one

tomasMizera: Plus, do not define default value in here

app/activeproject.cpp
```diff
@@ -227,6 +227,7 @@ bool ActiveProject::forceLoad( const QString &filePath, bool force )
   if ( mAppSettings.autosyncAllowed() )
   {
     setAutosyncEnabled( true );
+    requestSync();
```
tomasMizera: We're missing the `RequestOrigin` here

PR #4249, opened 2025-12-12
QGIS Auth db import

gabriel-bolbotina:
Ready for review.

Workflow:
- upon loading/reloading the project, clear all the configurations from the authentication database
- If there is a XML configuration file, import it using QGIS API. 
- once configuration file is imported, the protected layers should be visible
- Add a flag to check if any modifications occurred to the configuration file (even deletion)

Note: the database is removed when the app firstly starts.

### DIFF
--- a/app/activeproject.cpp
+++ b/app/activeproject.cpp
@@ -168,7 +168,7 @@ bool ActiveProject::forceLoad( const QString &filePath, bool force )
       // import the new configuration, if it exists.
       const QString projectId = MerginProjectMetadata::fromCachedJson( CoreUtils::getProjectMetadataPath( projectDir.path() ) ).projectId;
       const bool ok = mAuthManager->importAuthenticationConfigsFromXml( cfgFile.filePath(), projectId, true );
-      CoreUtils::log( "Authentication database", QStringLiteral( "QGIS auth imported: %1" ).arg( ok ? "success" : "failure" ) );
+      CoreUtils::log( "Authentication database", QStringLiteral( "QGIS auth import of %1 configuration(s) : %2" ).arg( mAuthManager->configIds().count() ).arg( ok ? "successful" : "failed" ) );
     }
 
     res = mQgsProject->read( filePath );


app/authsync.cpp
```diff
@@ -0,0 +1,50 @@
...
+  }
+}
+
+void AuthSync::importAuth()
+{
+  if ( fileExists( mAuthFile ) )
+  {
+
+    mAuthMngr->setPasswordHelperEnabled( false );
+    mAuthMngr->setMasterPassword( QStringLiteral( "merginMaps" ) );
```
tomasMizera: Shouldn't this go to `main.cpp`?

core/merginapi.cpp
```diff
@@ -3677,6 +3677,7 @@ void MerginApi::finishProjectSync( const QString &projectFullName, bool syncSucc
   ProjectDiff diff = transaction.diff;
   int newVersion = syncSuccessful ? transaction.version : -1;
 
+  // TODO check if the xml configuration has been changed
```
tomasMizera: TODO?

app/activeproject.cpp
```diff
@@ -155,6 +156,10 @@ bool ActiveProject::forceLoad( const QString &filePath, bool force )
     emit projectWillBeReloaded();
     mActiveLayer.resetActiveLayer();
 
+    // import authentication database before loading the layers
```
tomasMizera: We wanted to reset auth configs before importing new ones..

app/authsync.h
```diff
@@ -0,0 +1,37 @@
...
+
+#include "qgsproject.h"
+#include "qgsauthmanager.h"
+#include "qgsapplication.h"
+#include "qgsmaplayer.h"
+#include "qgsproviderregistry.h"
+#include "coreutils.h"
+#include "activeproject.h"
+#include "merginprojectmetadata.h"
+#include "appsettings.h"
```
tomasMizera: I think you do not need at least half of these imports

app/authsync.cpp
```diff
@@ -0,0 +1,50 @@
+#include "authsync.h"
+#include <QFile>
+#include <QRegularExpression>
+#include <QMessageBox>
+#include <QCryptographicHash>
```
tomasMizera: I think you do not need at least half of these imports

app/activeproject.cpp
gabriel-bolbotina: I think this is not necessary, since we set the overwrite parameter to true when calling importFromXml. 
image

core/merginapi.cpp
gabriel-bolbotina: I tried to only modify the xml configuration file from QGIS. Upon syncing,  it was already detected by that function, which emitted a force reload of the project. So I don't think we should specifically check for the cfg modification at all.

app/CMakeLists.txt
```diff
@@ -506,10 +508,23 @@ if (LNX)
   target_link_libraries(MerginMaps PUBLIC QGIS::Core)
 endif ()
 
+find_library(
```
tomasMizera: Would be good to add some explanation comments on why this is needed and what it does

app/activeproject.cpp
tomasMizera: I'd still prefer if we did that so that auth db contains only entries relevant to the opened project and not entries from the previously opened projects.

app/activeproject.cpp
```diff
@@ -155,6 +159,15 @@ bool ActiveProject::forceLoad( const QString &filePath, bool force )
...
+    QgsAuthManager *authMngr = QgsApplication::authManager();
+    QString projectDir = mLocalProjectsManager.projectFromProjectFilePath( filePath ).projectDir ;
+    QString authFile = QDir( projectDir ).filePath( AUTH_CONFIG_FILENAME );
+    QString projectId = MerginProjectMetadata::fromCachedJson( CoreUtils::getProjectMetadataPath( projectDir ) ).projectId;
+
+    bool ok = authMngr->importAuthenticationConfigsFromXml( authFile, projectId, true );
```
tomasMizera: Hm, shouldn't we check if the auth file exists?

app/main.cpp
```diff
@@ -155,6 +156,8 @@
 #include "qgsapplication.h"
 #include "activeproject.h"
 #include "appsettings.h"
+#include <QtPlugin>
+Q_IMPORT_PLUGIN( opensslPlugin )
```
tomasMizera: Would be good to add some explanation comments on why this is needed and what it does

app/test/testauthimport.cpp
```diff
@@ -0,0 +1,175 @@
...
+  {
+    static int argc = 1;
+    static char *argv[] = { ( char * )"test_app", nullptr };
+    new QgsApplication( argc, argv, true, "TestApp" );
+  }
+
+  // create a temporary folder
+  mAppTempDir.reset( new QTemporaryDir() );
+  QVERIFY( mAppTempDir->isValid() );
+  qputenv( "QGIS_AUTH_DB_DIR_PATH", mAppTempDir->path().toLatin1() );
```
tomasMizera: Why exactly do we need or want to change this?

app/test/testauthimport.cpp
```diff
@@ -0,0 +1,175 @@
...
+}
+
+void TestAuthImport::initTestCase()
+{
+  if ( !QgsApplication::instance() )
+  {
+    static int argc = 1;
+    static char *argv[] = { ( char * )"test_app", nullptr };
+    new QgsApplication( argc, argv, true, "TestApp" );
+  }
```
tomasMizera: You can expect that `QgsApplication::instance()` will always exist.

app/test/testauthimport.h
```diff
@@ -0,0 +1,57 @@
...
+    void test_import_fails_master_password_not_set();
+    void test_reload_after_configuration_update();
+
+  private:
+    std::unique_ptr<QTemporaryDir> mAppTempDir;
+    const char *mPass = nullptr;
+
+    void mockProjectMetadata( const QString &projectDir );
+    void setup_ExportInitialConfig( const QString &projectDir, const QString &authId, const QString &username );
+    void copyTestAuthFile( const QString &projectDir );
```
tomasMizera: There is a mixture of camel case and snake case conventions for method names 🤔

app/test/testauthimport.cpp
tomasMizera, line 1: I do not quite get what this is testing as it does not interact with `ActiveProject` at all

tomasMizera (changes requested):

core/merginapi.cpp
gabriel-bolbotina: Check this [video](https://drive.google.com/file/d/1L6L9mwHuC4Wi2A5WXwPrpajzOYyPbFSJ/view?usp=share_link).

app/activeproject.cpp
```diff
@@ -155,6 +159,23 @@ bool ActiveProject::forceLoad( const QString &filePath, bool force )
     emit projectWillBeReloaded();
     mActiveLayer.resetActiveLayer();
 
+    // path to the authenticationn configuration file
```
Withalion:
```suggestion
    // path to the authentication configuration file
```

app/activeproject.cpp
```diff
@@ -25,6 +27,8 @@
 #include "position/tracking/androidtrackingbroadcast.h"
 #endif
 
+const QString AUTH_CONFIG_FILENAME = "qgis_cfg.xml";
```
Withalion: use `QStringLiteral`

app/activeproject.cpp
```diff
@@ -155,6 +159,23 @@ bool ActiveProject::forceLoad( const QString &filePath, bool force )
     emit projectWillBeReloaded();
     mActiveLayer.resetActiveLayer();
 
+    // path to the authenticationn configuration file
+    QFileInfo fileInfo( filePath );
+    QString projectDir  = fileInfo.absolutePath();
+    QString authFile = QDir( projectDir ).filePath( AUTH_CONFIG_FILENAME );
+    QFileInfo cfgFile( authFile );
```
Withalion: this logic feels unnecessary complex
```suggestion
    const QDir projectDir = QFileInfo( filePath ).dir()
    const QFileInfo cfgFile( projectDir.filePath( AUTH_CONFIG_FILENAME ) );
```

Withalion: this will require change on line 174

app/activeproject.cpp
```diff
@@ -155,6 +159,23 @@ bool ActiveProject::forceLoad( const QString &filePath, bool force )
...
+    if ( cfgFile.exists() && cfgFile.isFile() )
+    {
+      // clear the authentication database before importing a new one, if it exists
+      QgsAuthManager *authMngr = QgsApplication::authManager();
+      authMngr->removeAllAuthenticationConfigs();
+
+      // import the new configuration, if it exists.
+      QString projectId = MerginProjectMetadata::fromCachedJson( CoreUtils::getProjectMetadataPath( projectDir ) ).projectId;
+      bool ok = authMngr->importAuthenticationConfigsFromXml( authFile, projectId, true );
```
Withalion:
```suggestion
      const QString projectId = MerginProjectMetadata::fromCachedJson( CoreUtils::getProjectMetadataPath( projectDir ) ).projectId;
      const bool ok = authMngr->importAuthenticationConfigsFromXml( authFile, projectId, true );
```

app/test/testactiveproject.cpp
```diff
@@ -191,3 +191,38 @@ void TestActiveProject::testLoadingFlagFileExpiration()
 
   QVERIFY( !flagFile.exists() );
 }
+
+void TestActiveProject::testLoadingAuthFileFromConfiguration()
+{
+  AppSettings as;
+  ActiveLayer al;
+  ActiveProject activeProject( as, al, mApi->localProjectsManager() );
+  QString projectDir = TestUtils::testDataDir() + "/project_auth_file/";
```
Withalion:
```suggestion
  QString projectDir = TestUtils::testDataDir() + QStringLiteral( "/project_auth_file/" );
```

app/activeproject.cpp
```diff
@@ -155,6 +159,21 @@ bool ActiveProject::forceLoad( const QString &filePath, bool force )
...
+    if ( cfgFile.exists() && cfgFile.isFile() )
+    {
+      // clear the authentication database before importing a new one, if it exists
+      QgsAuthManager *authMngr = QgsApplication::authManager();
+      authMngr->removeAllAuthenticationConfigs();
+
+      // import the new configuration, if it exists.
+      const QString projectId = MerginProjectMetadata::fromCachedJson( CoreUtils::getProjectMetadataPath( projectDir.path() ) ).projectId;
+      const bool ok = authMngr->importAuthenticationConfigsFromXml( cfgFile.filePath(), projectId, true );
+      CoreUtils::log( "Database authentication: ", QString( "QGIS auth imported: %1" ).arg( ok ? "true" : "false" ) );
```
Withalion:
```suggestion
      CoreUtils::log( "Database authentication: ", QStringLiteral( "QGIS auth imported: %1" ).arg( ok ? "true" : "false" ) );
```

app/main.cpp
```diff
@@ -398,6 +399,8 @@ void addQmlImportPath( QQmlEngine &engine )
 
 int main( int argc, char *argv[] )
 {
+  // for windows qca dll
+  QCoreApplication::addLibraryPath( QCoreApplication::applicationDirPath() );
```
Withalion: Is this necessary? if the dll is missing and we need to do some path magic here, then it should be in Cmake. But this just adds the application directory, which qt should already look through.

app/static_plugins.cpp
```diff
@@ -0,0 +1,3 @@
+#include <QtPlugin>
+// Required for QGIS authentication manager API on static builds (Linux/macOS/iOS)
```
Withalion: also android?

app/test/testactiveproject.h
```diff
@@ -12,6 +12,11 @@
 
 #include <QObject>
 #include <merginapi.h>
+#include <qgsapplication.h>
+#include <qgsauthmanager.h>
+
+const QString AUTH_CONFIG_FILENAME = "qgis_cfg.xml";
+const QString AUTH_CONFIG_PASSWORD = "1234";
```
Withalion:
```suggestion
const QString AUTH_CONFIG_FILENAME = QStringLiteral( "qgis_cfg.xml" );
const QString AUTH_CONFIG_PASSWORD = QStringLiteral( "1234" );
```

Withalion (commented):
The loading of QGIS auth DB didn't work on Windows. Same project worked on Android.

app/CMakeLists.txt
```diff
@@ -506,6 +510,21 @@ if (LNX)
   target_link_libraries(MerginMaps PUBLIC QGIS::Core)
 endif ()
 
+if (NOT WIN32)
+  # for macOS and linux
```
Withalion: everything else besides Windows no?

app/test/testactiveproject.cpp
```diff
@@ -191,3 +191,39 @@ void TestActiveProject::testLoadingFlagFileExpiration()
...
+void TestActiveProject::testLoadingAuthFileFromConfiguration()
+{
+  AppSettings appSettings;
+  ActiveLayer activeLayer;
+  ActiveProject activeProject( appSettings, activeLayer, mApi->localProjectsManager() );
+  QString projectDir = TestUtils::testDataDir() + QStringLiteral( "/project_auth_file/" );
+  QString projectName = QStringLiteral( "auth-test.qgz" );
+  QString authFile = QDir( projectDir ).filePath( AUTH_CONFIG_FILENAME );
+
+  QgsApplication::initQgis();
```
tomasMizera: Why do we need to explicitly `initQgis` here?

app/activeproject.cpp
```diff
@@ -155,6 +159,21 @@ bool ActiveProject::forceLoad( const QString &filePath, bool force )
...
+    if ( cfgFile.exists() && cfgFile.isFile() )
+    {
+      // clear the authentication database before importing a new one, if it exists
+      QgsAuthManager *authMngr = QgsApplication::authManager();
+      authMngr->removeAllAuthenticationConfigs();
+
+      // import the new configuration, if it exists.
+      const QString projectId = MerginProjectMetadata::fromCachedJson( CoreUtils::getProjectMetadataPath( projectDir.path() ) ).projectId;
+      const bool ok = authMngr->importAuthenticationConfigsFromXml( cfgFile.filePath(), projectId, true );
+      CoreUtils::log( "Database authentication: ", QStringLiteral( "QGIS auth imported: %1" ).arg( ok ? "true" : "false" ) );
```
tomasMizera:
```suggestion
      CoreUtils::log( "Authentication database", QStringLiteral( "QGIS auth imported: %1" ).arg( ok ? "success" : "failure" ) );
```

CMakeLists.txt
```diff
@@ -286,6 +286,7 @@ endif ()
 if (WIN)
   add_compile_definitions(_HAS_AUTO_PTR_ETC=1)
   add_compile_definitions(_USE_MATH_DEFINES)
+  add_compile_definitions(NOMINMAX)
```
tomasMizera, line 291: Please add a comment to explain why we need this

Withalion: @gabriel-bolbotina add something along the lines
> On windows the compiler instead of using standard `min` & `max` functions expands them to `min` & `max` macros in every file that includes windows header file

Withalion:
Off topic: We should try to build the app with cmake 4.0+ now, last time QCA was the thing not letting us build it

app/main.cpp
```diff
@@ -544,10 +544,12 @@ int main( int argc, char *argv[] )
...
   LayerDetailLegendImageProvider *layerDetailLegendImageProvider( new LayerDetailLegendImageProvider );
 
-  // setting up the master password for authentication database retrieval
   QgsAuthManager *authManager = QgsApplication::authManager();
+  // remove existing authentication database when opening the app
+  authManager->removeAllAuthenticationConfigs();
+  // set up the master password for authentication database retrieval
   authManager->setPasswordHelperEnabled( false );
-  authManager->setMasterPassword( QStringLiteral( "merginMaps" ) );
+  authManager->setMasterPassword( QStringLiteral( "merginMaps" ), true );
```
tomasMizera, line 552: The addition of `verify=true` does not seem to do much as we do not read the return value. Still ok though..

PR #4267, opened 2025-12-16
Build QGIS dependency as RelWithDebInfo

### LINKED ISSUE #4266: Prepare vcpkg triplet with QGIS in relwithdebinfo mode
MacOS and Linux builds; The goal is to be able to debug QGIS code when debugging the mobile app. Let's start with desktop builds.

Withalion:
fixes #4266 

Introduce debug symbols for QGIS for better debugging of various QGIS issues. QGIS can be build in `RelWithDebInfo` for all platforms. To enable this build it's necessary to set two environment variables (see `howto.md`).

Tested on Linux & Android.

P.S: This PR might look scary, most of it is just a change in formatting. Relevant parts [here](https://github.com/MerginMaps/mobile/pull/4267/changes#diff-163dfe0e402e5457ccdfbc926be8a2e129d9b3d7afb7bd9cef755d43400e7a56R232-R235).

### DIFF
--- a/vcpkg/ports/qgis/portfile.cmake
+++ b/vcpkg/ports/qgis/portfile.cmake
@@ -229,8 +229,8 @@ if (VCPKG_CROSSCOMPILING)
   )
 endif ()
 
-# For desktop builds, build QGIS as RelWithDebInfo for better debugging
-if (TARGET_TRIPLET STREQUAL "arm64-osx" OR TARGET_TRIPLET STREQUAL "x64-linux")
+if (DEFINED ENV{QGIS_DEBUG_BUILD})
+  message(STATUS "Building requested QGIS RelWithDebInfo build")
   list(APPEND QGIS_OPTIONS_RELEASE -DCMAKE_BUILD_TYPE=RelWithDebInfo)
 endif ()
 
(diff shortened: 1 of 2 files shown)


vcpkg/ports/qgis/portfile.cmake
```diff
@@ -157,6 +157,11 @@ if(VCPKG_CROSSCOMPILING)
    list(APPEND QGIS_OPTIONS -DQT_HOST_PATH_CMAKE_DIR:PATH=${CURRENT_HOST_INSTALLED_DIR}/share)
 endif()
 
+# For desktop builds, build QGIS as RelWithDebInfo for better debugging 
+if (TARGET_TRIPLET STREQUAL "arm64-osx" OR TARGET_TRIPLET STREQUAL "x64-linux")
```
tomasMizera: Can we somehow update this to be true if the _main_ build is `RelWithDebInfo`?

Something like:
`if ($CMAKE_BUILD_TYPE == RelWithDebInfo)`

Withalion: Looks like there is not enought flexibility on VCPKG side. We are unable to access this information. We can either create new "debug" triplets or build it always as `RelWithDebInfo`

tomasMizera: I see.. Is it possible to have some cmake variable propagated here? So that we say "-DQGIS_DEBUG=TRUE"

Withalion: Hmm not really as far as I understand. We can set something up in the triplet files (so creating a different triplet file is an option) or we just keep it in the portfiles.

Withalion: Okay I think some solution, which could work, on [stackoverflow](https://stackoverflow.com/a/77954891/13176531)

tomasMizera (approved):

PR #4277, opened 2025-12-30
Fix autosync for related features

### LINKED ISSUE #4087: Improve auto-sync behavior when editing parent and child features
Description:
When automatic synchronization is enabled, the app starts syncing as soon as I open a layer form and add features

It seems that the current logic triggers synchronization immediately after it detects any edits, even if the parent feature is still being edited and the child feature is not yet fully saved.

This can be a bit distracting and also feels premature. A better approach would be:

Only trigger synchronization after the parent feature has been saved, not already when a child feature is being added to it.

Use case:
When I add features to a layer with a 1-N relation, auto-sync pushes changes while I am still saving the child feature’s data.
This interrupts the workflow and feels unexpected.

Expected behavior:
Sync should only start after completing the parent save action.

https://github.com/user-attachments/assets/ab1bb7af-1b96-4615-b121-8dfd1363ce75

Withalion:
fixes #4087

Previously we auto-synced on every change to each vector layer. This approach was causing issues while the user was creating child features while recording parent feature. Now autosync will be "paused" while feature form is open and the sync will resume after all forms are closed.

### DIFF
--- a/app/qml/form/MMFormStackController.qml
+++ b/app/qml/form/MMFormStackController.qml
@@ -255,7 +255,7 @@ Item {
   StackView {
     id: formsStack
 
-    property bool saveRequested: false
+    property bool syncWhenFormCloses: false
 
     function popOneOrClose() {
       if ( formsStack.depth > 1 ) {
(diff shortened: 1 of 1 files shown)


app/qml/form/MMFormStackController.qml
```diff
@@ -255,6 +255,8 @@ Item {
   StackView {
     id: formsStack
 
+    property bool saveRequested: false
```
tomasMizera: How about
```suggestion
    property bool syncWhenFormCloses: false
```
So we understand what this is in few months time too

app/qml/form/MMFormStackController.qml
```diff
@@ -272,6 +274,16 @@ Item {
     focus: true
 
     anchors.fill: parent
+
+    onDepthChanged: {
+      if (depth === 0) {
```
tomasMizera, line 281: Can you please add some comment to explain what we do here and why?

PR #4278, opened 2026-01-08
Add more information for the SSL erros

gabriel-bolbotina:
https://github.com/user-attachments/assets/fde732b0-e7c2-40fc-a961-75a4fd21a51c

I tested it using docker nginx and an expired certificate

### DIFF
--- a/core/merginapi.cpp
+++ b/core/merginapi.cpp
@@ -113,7 +113,7 @@ MerginApi::MerginApi( LocalProjectsManager &localProjects, QObject *parent )
   } );
 
 #ifndef QT_NO_SSL
-  QObject::connect( mManager, &QNetworkAccessManager::sslErrors, this, []( const QNetworkReply *reply, const QList<QSslError> &errors )
+  QObject::connect( mManager, &QNetworkAccessManager::sslErrors, this, []( const QNetworkReply * reply, const QList<QSslError> &errors )
   {
     CoreUtils::log( QStringLiteral( "SSL error" ), QStringLiteral( "URL attempting to access: " ) + reply->url().toString() );
     for ( const auto &error : errors )


core/merginapi.cpp
Withalion, line 1: Looks like we can also connect to the same signal on the level of `QNetworkAccessManager`, see [here](https://doc.qt.io/qt-6/qnetworkaccessmanager.html#sslErrors). It would be better to attach to it in the constructor of `MerginApi` and listen to every request not just `pingMergin`. 

Also logging the error description and which reply it is attached to seems enough to me, no need to dump the whole cert info.

core/merginapi.cpp
```diff
@@ -112,6 +112,17 @@ MerginApi::MerginApi( LocalProjectsManager &localProjects, QObject *parent )
...
 
+#ifndef QT_NO_SSL
+  QObject::connect( mManager, &QNetworkAccessManager::sslErrors, this, []( QNetworkReply * reply, const QList<QSslError> &errors )
+  {
+    CoreUtils::log( "URL attempting to access:", reply->url().toString() );
+    for ( const auto &error : errors )
+    {
+      CoreUtils::log( "Error Description:", error.errorString() );
+    }
+  } );
```
```suggestion
  QObject::connect( mManager, &QNetworkAccessManager::sslErrors, this, []( const QNetworkReply *reply, const QList<QSslError> &errors )
  {
    CoreUtils::log( QStringLiteral( "SSL error" ), QStringLiteral( "URL attempting to access: " ) + reply->url().toString() );
    for ( const auto &error : errors )
    {
      CoreUtils::log( QStringLiteral( "SSL error" ), QStringLiteral( "Error Description:" ) + error.errorString() );
    }
  } );
```

PR #4291, opened 2026-01-15
Hide the mapPanelStack layers when adding a child

gabriel-bolbotina:
Ready for review 

https://github.com/user-attachments/assets/cb134b32-cf90-4210-a491-03facb0f735b

### DIFF
--- a/app/qml/main.qml
+++ b/app/qml/main.qml
@@ -463,7 +463,7 @@ ApplicationWindow {
     }
 
     function showMapStack(){
-        mapPanelsStackView.visible = false
+        mapPanelsStackView.visible = true
     }
   }
 


app/qml/main.qml
```diff
@@ -442,6 +454,30 @@ ApplicationWindow {
...
+      id: mapPanelsHideTranslation
+      y: 0
+
+      Behavior on y {
+        NumberAnimation { 
+          duration: 500
+          easing.type: Easing.OutCubic
+        }
+      }
+    }
```
Withalion: Nice hack :) but couldn't we use something similar to what we use in `MMFormStackController.hideAll()` & `MMFormStackController.reopenAll()`?

app/qml/main.qml
```diff
@@ -442,6 +454,17 @@ ApplicationWindow {
...
+
+    function hideMapStackIfNeeded() {
+      // if present in the stack, hide the other layers when editing the geometry
+      if(mapPanelsStackView.depth > 0){
+        mapPanelsStackView.visible = false
+      }
+    }
+
+    function showMapStack(){
+        mapPanelsStackView.visible = false
```
tomasMizera: `visible = false` in `showMapStack`? 🤔

PR #4310, opened 2026-02-09
Fixed windows photo attachment and refactored input functions

gabriel-bolbotina:
Ready for review.

Modified the image paths in components that used them from 'file://' to 'file:///', which works on all the platforms, both Windows and Unix based.

I refactored the input functions that deal with paths to better accommodate system based differences for handling paths (e.g., partition letters on Windows). Also, I tried to use QDir and QFileInfo as much as possible.

Tested adding, viewing and saving photos on:

- Windows
- MacOS
- iOS
- android

https://github.com/user-attachments/assets/4b5c093f-d6ce-40de-93bc-ae3874a240ae

Also ran the testUtils and testSketching on Windows as well.

### DIFF
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -1063,7 +1063,7 @@ QString InputUtils::getRelativePath( const QString &path, const QString &prefixP
   QString relativePath = prefixDir.relativeFilePath( cleanPath );
 
   // check if the path starts with ".." or is absolute (on Windows/different drives), it's not a "child"
-  if ( relativePath.startsWith( QLatin1StringView( ".." ) ) || QDir::isAbsolutePath( relativePath ) )
+  if ( relativePath.startsWith( QStringLiteral( ".." ) ) || QDir::isAbsolutePath( relativePath ) )
   {
     return {};
   }
@@ -2059,6 +2059,11 @@ void InputUtils::sanitizePath( QString &path )
   {
     path = result;
   }
+}
+
+QString InputUtils::localFileToUrl( const QString &path )
+{
+  return QUrl::fromLocalFile( path ).toString();
 }
 
 QSet<int> InputUtils::referencedAttributeIndexes( QgsVectorLayer *layer, const QString &expression )
--- a/app/inpututils.h
+++ b/app/inpututils.h
@@ -367,7 +367,7 @@ class InputUtils: public QObject
 
     /**
      * Returns relative path of the file to given prefixPath. If prefixPath does not match a path parameter,
-     * returns an empty string. If a path starts with "file:///" or "file://", this prefix is ignored.
+     * returns an empty string. If a path starts with "file://", this prefix is ignored.
      * \param path Absolute path to file
      * \param prefixPath
      */
@@ -637,6 +637,11 @@ class InputUtils: public QObject
      * Splits path into components and sanitizes each component using sanitizeNode().
      */
     static void sanitizePath( QString &path );
+
+    /**
+     * Returns the URL to a local path
+    */
+    Q_INVOKABLE static QString localFileToUrl( const QString &path );
 
     /**
      * Returns the indexes of fields that are referenced in an expression
--- a/app/qml/form/editors/MMFormGalleryEditor.qml
+++ b/app/qml/form/editors/MMFormGalleryEditor.qml
@@ -57,7 +57,7 @@ MMPrivateComponents.MMBaseInput {
         let absolutePath = model.PhotoPath
 
         if ( absolutePath !== '' && __inputUtils.fileExists( absolutePath ) ) {
-          return "file:///" + absolutePath
+          return __inputUtils.localFileToUrl(absolutePath)
         }
         return ''
       }
--- a/app/qml/form/editors/MMFormPhotoEditor.qml
+++ b/app/qml/form/editors/MMFormPhotoEditor.qml
@@ -173,10 +173,10 @@ MMFormPhotoViewer {
     target: root.sketchingController
 
     function onTempPhotoSourceChanged( newPath ){
-      if ( internal.tempSketchedImageSource === "file:///" + newPath ) {
+      if ( internal.tempSketchedImageSource === __inputUtils.localFileToUrl( newPath ) ){
         internal.tempSketchedImageSource = ""
       }
-      internal.tempSketchedImageSource = "file:///" + newPath
+      internal.tempSketchedImageSource = __inputUtils.localFileToUrl( newPath )
     }
 
     function onSketchesSavingError(){
@@ -269,7 +269,7 @@ MMFormPhotoViewer {
 
       if ( __inputUtils.fileExists( absolutePath ) ) {
         root.photoState = "valid"
-        resolvedImageSource = "file:///" + absolutePath
+        resolvedImageSource = __inputUtils.localFileToUrl( absolutePath )
         tempSketchedImageSource = ""
       }
       else if ( __inputUtils.isValidUrl( absolutePath ) ) {


app/test/testsketching.cpp
```diff
@@ -169,7 +169,11 @@ void TestSketching::testLoadBackupSketch()
   sketchingController.mPhotoSource = path;
   sketchingController.mProjectName = QStringLiteral( "/this/is/long/path/to/image/test_sketching" );
   sketchingController.prepareController();
+#ifdef Q_OS_WIN32
+  QCOMPARE( sketchingController.mPhotoSource, "file:///" + QDir::tempPath() + QStringLiteral( "/test_sketching" ) + QStringLiteral( "/MM_test_image.jpg" ) );
+#else
   QCOMPARE( sketchingController.mPhotoSource, "file://" + QDir::tempPath() + QStringLiteral( "/test_sketching" ) + QStringLiteral( "/MM_test_image.jpg" ) );
+#endif
```
Withalion: You could probably wrap the values here in `QUrl` and then compare them, it should simplify the code

app/test/testutilsfunctions.cpp
```diff
@@ -274,7 +274,7 @@ void TestUtilsFunctions::getRelativePath()
   QString relativePath2 =  mUtils->getRelativePath( path2, prefixPath );
   QCOMPARE( fileName2, relativePath2 );
 
-  QString path3 = QStringLiteral( "file://" ) + path2;
+  QString path3 = QStringLiteral( "file:///" ) + path2;
```
Withalion:
```suggestion
  QString path3 = QUrl ( path2 );
```

app/inpututils.cpp
```diff
@@ -1051,37 +1045,46 @@ QString InputUtils::resolvePath( const QString &path, const QString &homePath, c
...
+#ifdef Q_OS_WIN
+    // remove the partition letter as well
+    modPath.remove( 0, 8 );
+#else
+    modPath.remove( 0, 7 );
+#endif
+  }
+  else if ( modPath.startsWith( QLatin1String( "file://" ), Qt::CaseInsensitive ) )
+  {
+    modPath.remove( 0, 7 );
```
Withalion: Is there no way to use QUrl here instead?

app/inpututils.cpp
```diff
@@ -1051,37 +1045,46 @@ QString InputUtils::resolvePath( const QString &path, const QString &homePath, c
...
+  }
 
-    if ( prefixCanonicalPath.length() > 1 && canonicalPath.startsWith( prefixCanonicalPath ) )
-    {
-      return canonicalPath.replace( prefixCanonicalPath, QString() );
-    }
+  // check that the file is actually a child of the prefix
+  if ( relativePath.startsWith( QLatin1String( ".." ) ) || QFileInfo( relativePath ).isAbsolute() )
+  {
+    return QString();
```
Withalion:
```suggestion
    return {};
```

app/inpututils.cpp
```diff
@@ -1051,37 +1045,46 @@ QString InputUtils::resolvePath( const QString &path, const QString &homePath, c
...
+  }
+  else if ( modPath.startsWith( QLatin1String( "file://" ), Qt::CaseInsensitive ) )
+  {
+    modPath.remove( 0, 7 );
   }
 
   if ( prefixPath.isEmpty() ) return modPath;
 
+  // use QDir to calculate the relative path
+  QDir prefixDir( prefixPath );
```
Withalion:
```suggestion
  const QDir prefixDir( prefixPath );
```

app/inpututils.cpp
```diff
@@ -1051,37 +1045,46 @@ QString InputUtils::resolvePath( const QString &path, const QString &homePath, c
...
+
   // Do not use a canonical path for non-existing path
-  if ( !QFileInfo::exists( path ) )
+  if ( QFileInfo::exists( modPath ) && QFileInfo::exists( prefixPath ) )
   {
-    if ( !prefixPath.isEmpty() && modPath.startsWith( prefixPath ) )
-    {
-      return modPath.replace( prefixPath, QString() );
-    }
+    QDir canonicalPrefix( prefixDir.canonicalPath() );
```
Withalion:
```suggestion
    const QDir canonicalPrefix( prefixDir.canonicalPath() );
```

app/inpututils.cpp
```diff
@@ -1973,62 +1976,107 @@ QUrl InputUtils::iconFromGeometry( const Qgis::GeometryType &geometry )
...
   }
-
-  if ( pathStartsWithFileURL )
+  else if ( path.startsWith( QStringLiteral( "file://" ), Qt::CaseInsensitive ) )
   {
-    // restore file:// prefix
-    sanitizedPath = "file://" + sanitizedPath;
+    prefix = QStringLiteral( "file://" );
+    cleanPath = path.mid( 7 );
   }
```
Withalion: Once again won't usage of `QUrl` simplify this?

app/inpututils.cpp
```diff
@@ -1973,62 +1976,107 @@ QUrl InputUtils::iconFromGeometry( const Qgis::GeometryType &geometry )
...
+  const thread_local QRegularExpression illegalChars( QStringLiteral( "[\x00-\x1f<>:|?*\"\\\\]" ) );
 
-  // Trim whitespace immediately before the final extension, e.g. "name  .jpg" -> "name.jpg"
-  const int lastDot = fileName.lastIndexOf( QChar( '.' ) );
-  if ( lastDot > 0 )
+#ifdef Q_OS_WIN
+  const static QRegularExpression reservedNames(
+    QStringLiteral( "^(CON|PRN|AUX|NUL|COM[1-9]|LPT[1-9])$" ),
+    QRegularExpression::CaseInsensitiveOption
+  );
```
Withalion: can you write some explanation here

app/inpututils.cpp
```diff
@@ -1973,62 +1976,107 @@ QUrl InputUtils::iconFromGeometry( const Qgis::GeometryType &geometry )
...
 {
+  if ( fileName.isEmpty() )
+    return;
+
+  // unify separators to '/' to handle splitting easily
+  QString unifiedPath = fileName;
+  unifiedPath.replace( QLatin1Char( '\\' ), QLatin1Char( '/' ) );
+
+  // split into segments (folders + filename)
+  QStringList parts = unifiedPath.split( QLatin1Char( '/' ), Qt::KeepEmptyParts );
```
Withalion: if you skip them in the loop, then remove them from the split

app/inpututils.cpp
```diff
@@ -1973,62 +1976,107 @@ QUrl InputUtils::iconFromGeometry( const Qgis::GeometryType &geometry )
 
 void InputUtils::sanitizeFileName( QString &fileName )
```
Withalion: you should also change the description of the function to reflect new version

app/test/testutilsfunctions.cpp
```diff
@@ -355,7 +355,9 @@ void TestUtilsFunctions::resolveTargetDir()
   config.insert( QStringLiteral( "PropertyCollection" ), collection );
 
   QString resultDir3 = mUtils->resolveTargetDir( homePath, config, pair, QgsProject::instance() );
-  QCOMPARE( resultDir3, QStringLiteral( "%1/photos" ).arg( projectDir ) );
+  QString expectedDir = QStringLiteral( "%1/photos" ).arg( projectDir );
+  mUtils->sanitizeFileName( expectedDir );
```
Withalion: why do we need to sanitize it?

app/test/testutilsfunctions.cpp
```diff
@@ -1029,6 +1031,11 @@ void TestUtilsFunctions::testSanitizePath()
   InputUtils::sanitizePath( str );
   QCOMPARE( str, QStringLiteral( "/complex/valid/Φ!l@#äme$%^&()-_=+[]{}`~;',.ext" ) );
 
+  // unchanged with partition letter on Windows
+  str = QStringLiteral( "C:/Users/simple/valid/filename.ext" );
+  InputUtils::sanitizePath( str );
+  QCOMPARE( str, QStringLiteral( "C:/Users/simple/valid/filename.ext" ) );
```
Withalion, line 1011: can you also add case here which is both windows and has `file://` scheme prefix please

same thing below

app/inpututils.cpp
gabriel-bolbotina: I modified the function to only use QDir like this: `QUrl url( path );
  bool isUrl = url.isValid() && url.scheme() == "file";

  QString prefix;
  QString cleanPath = isUrl ? url.toLocalFile() : path;
  cleanPath = QDir::fromNativeSeparators( cleanPath );
 
  bool needsLeadingSlash = cleanPath.startsWith('/');

  QStringList parts = cleanPath.split('/' , Qt::SkipEmptyParts);
  QStringList sanitizedParts;

  for(QString part: parts)
  {
  #ifdef Q_OS_WIN
  if( part.endsWith(":"))
  {
    sanitizedParts << part;
    continue;
  }
  #endif
  sanitizeFileName ( part);
  sanitizedParts << part;
}


  // reconstruct
  QString result = sanitizedParts.join('/');

  if (needsLeadingSlash)
  {
    result.prepend('/');
  }

  if ( isUrl )
  {
    path = QUrl::fromLocalFile(result).toString();

  }
  else {
    path = QDir::toNativeSeparators(result);
  }` 
  
  and the tests would fail on Windows.
  
  Even if I replace '\' with '/' via a helper function, 2 cases seem to fail on Windows and 1 on MacOS:
  `// unchanged - url prefix
  str = QStringLiteral( "file://simple/valid/filename.ext" );
  InputUtils::sanitizePath( str );
  verifyPath( str, QStringLiteral( "file://simple/valid/filename.ext" ) );
// and 

  str = QStringLiteral( "file:///  sa ni*tized /f<i>l?n\"a:m|e .ext " );
  InputUtils::sanitizePath( str );
  verifyPath( str, QStringLiteral( "file:///sa ni_tized/f_i_l_n_a_m_e.ext" ) );

// verify path just swaps '\' with '/' on windows and qcompares them`

app/inpututils.cpp
```diff
@@ -1050,38 +1044,36 @@ QString InputUtils::resolvePath( const QString &path, const QString &homePath, c
 
 QString InputUtils::getRelativePath( const QString &path, const QString &prefixPath )
 {
-  QString modPath = path;
-  QString filePrefix( "file://" );
+  // clean the file prefix
+  QString cleanPath = path;
+  QUrl url( path );
```
Withalion:
```suggestion
  const QUrl url( path );
```

app/inpututils.cpp
```diff
@@ -1050,38 +1044,36 @@ QString InputUtils::resolvePath( const QString &path, const QString &homePath, c
...
-    QString prefixCanonicalPath = prefixDir.canonicalPath() + "/";
 
-    if ( prefixCanonicalPath.length() > 1 && canonicalPath.startsWith( prefixCanonicalPath ) )
-    {
-      return canonicalPath.replace( prefixCanonicalPath, QString() );
-    }
+  if ( relativePath == cleanPath && !cleanPath.isEmpty() )
+  {
+    return {};
   }
```
Withalion: What case is this trying to catch? The only case I can think about is the case when `prefixPath` and `path` are pointing to different drives on windows, but that should be caught by if above. In unix you have one file tree so the first condition should never happen?

app/inpututils.cpp
```diff
@@ -1971,64 +1963,137 @@ QUrl InputUtils::iconFromGeometry( const Qgis::GeometryType &geometry )
...
+  if ( path.isEmpty() ) return;
+  QString cleanPath = path;
+
+  // QUrl treats ? as a query start, # as a fragment
+  cleanPath.replace( QLatin1Char( '?' ), QLatin1Char( '_' ) );
+  cleanPath.replace( QLatin1Char( '#' ), QLatin1Char( '_' ) );
 
-  if ( pathStartsWithFileURL )
+  // parse file prefix and path
+  QUrl url = QUrl::fromUserInput( cleanPath );
```
Withalion: do we expect the files to be also from web? I thought we worked only with local files. (but anyway this can stay here as in future it would be probably nice to support that too)

app/inpututils.cpp
```diff
@@ -1971,64 +1963,137 @@ QUrl InputUtils::iconFromGeometry( const Qgis::GeometryType &geometry )
...
+  bool isUrl = path.startsWith( QLatin1String( "file:" ), Qt::CaseInsensitive );
```
Withalion: isn't this just `url.isLocalFile()`?

app/inpututils.cpp
```diff
@@ -1971,64 +1963,137 @@ QUrl InputUtils::iconFromGeometry( const Qgis::GeometryType &geometry )
...
+
+  if ( isUrl && ( url.isLocalFile() || url.hasQuery() ) )
```
Withalion: I don't like the logic here, we check:
 - if the url is file
 - if the url is local file
 - if url has query (which it shouldn't have because you removed the query character)
we should check:
 - if the url is valid
 - if it is a local file

app/inpututils.cpp
```diff
@@ -1971,64 +1963,137 @@ QUrl InputUtils::iconFromGeometry( const Qgis::GeometryType &geometry )
...
+  {
+    cleanPath = url.toLocalFile();
+  }
+  else
   {
-    // remove file:// prefix before sanitization
-    path.remove( 0, 7 );
+    cleanPath = path;
```
Withalion: hmm, preferably this branch should not execute. `QUrl::fromUserInput` should detect both absolute and relative paths and essentially do a noop. We could possibly improve the relative path detection of the mentioned function (check docs how).
And if it happens then we have some garbage input which is not a path we want to work with further

app/inpututils.cpp
```diff
@@ -1971,64 +1963,137 @@ QUrl InputUtils::iconFromGeometry( const Qgis::GeometryType &geometry )
...
-  QString sanitizedPath;
+  // split and sanitize each segment
+  QStringList parts = cleanPath.split( QLatin1Char( '/' ), Qt::SkipEmptyParts );
+  QStringList sanitizedParts;
 
   for ( int i = 0; i < parts.size(); ++i )
   {
-    QString part = parts.at( i );
-    sanitizeFileName( part );
+    QString part = parts[i];
```
Withalion:
```suggestion
    const QString& part = parts[i];
```

app/inpututils.cpp
```diff
@@ -1971,64 +1963,137 @@ QUrl InputUtils::iconFromGeometry( const Qgis::GeometryType &geometry )
...
+  // handle Absolute Paths
+  if ( cleanPath.startsWith( QLatin1Char( '/' ) ) )
   {
-    // restore leading slash
-    sanitizedPath = '/' + sanitizedPath;
+    result.prepend( QLatin1Char( '/' ) );
   }
 
-  if ( pathStartsWithFileURL )
+  // restore file prefix rotocol
```
Withalion:
```suggestion
  // restore file prefix protocol
```

app/inpututils.cpp
```diff
@@ -1971,64 +1963,137 @@ QUrl InputUtils::iconFromGeometry( const Qgis::GeometryType &geometry )
...
+
+  // reassemble
+  if ( !suffix.isEmpty() )
+  {
+    return base + QLatin1Char( '.' ) + suffix;
+  }
+  return base;
+}
+
 void InputUtils::sanitizeFileName( QString &fileName )
```
Withalion: This function feels redundant now, why not just use `sanitizePath` instead of this? or `sanitizeNode` if we only care about file name?

app/inpututils.cpp
```diff
@@ -1971,64 +1963,137 @@ QUrl InputUtils::iconFromGeometry( const Qgis::GeometryType &geometry )
...
 
   for ( int i = 0; i < parts.size(); ++i )
   {
-    QString part = parts.at( i );
-    sanitizeFileName( part );
+    QString part = parts[i];
 
-    if ( i > 0 )
+    // keep Windows Drive Letters (e.g. "C:")
+    if ( i == 0 && part.endsWith( QLatin1Char( ':' ) ) )
```
Withalion: I would also add a check for `size == 2` so we are sure this is a drive letter and not some path part

app/inpututils.cpp
Withalion, line 1: I would replace all `QLatin1Char` that you used with `QStringLiteral`

app/inpututils.cpp
```diff
@@ -1971,64 +1963,137 @@ QUrl InputUtils::iconFromGeometry( const Qgis::GeometryType &geometry )
...
+
+  // if name has multiple extensions, clear the space for the second extension
+  // e.g., .tar .gz --> .tar.gz
+  const static QRegularExpression multipleExtensions( QStringLiteral( "(\\s)(\\.[^.]+)\\s+(?=\\.)" ) );
+  cleanOutput.replace( multipleExtensions, QStringLiteral( "\\1\\2" ) );
+
+  // split base and suffix to trim whitespace correctly.
+  QFileInfo fi( cleanOutput );
+  QString base = fi.completeBaseName();
+  QString suffix = fi.suffix().trimmed();
```
Withalion, line 1977: I'm not sure if we remove the whitespace between filename and suffix. From the tests it looks like we don't do that, but we should do it as that might break some other stuff.

app/test/testutilsfunctions.cpp
Withalion, line 1: Regarding these tests, we should not change the expected test results. Rather change the functions that they fit the expected test results

app/inpututils.cpp
gabriel-bolbotina: this was used to check the File URL, so basically the prefix which is file:///C:/.../ on Windows and which appeared in the test cases

app/inpututils.cpp
gabriel-bolbotina: In the function description it says that it does not remove the whitespace. However, in the test cases it was expected to remove it, so I think we should decide on one

app/inpututils.cpp
gabriel-bolbotina: You're right about the hasQuery and I will take it out. As for the other ones, I think we should check whether the path has the file prefix, so that we can re-attach it and also if it is a local file.

app/inpututils.cpp
gabriel-bolbotina: Not exactly, because we use isUrl to check if the original path had file prefix added, so that we can put it back once the path has been sanitized. Whereas isLocalFile will return true for absolute paths, it does not take into account if it has the file prefix/ URL

app/inpututils.cpp
gabriel-bolbotina: This is redundant, I will remove it.

app/inpututils.cpp
gabriel-bolbotina: I thought of them as each having a different role: 
- sanitizePath for full paths and the whole path needs to be sanitized
- sanitizeFileName for sanitizing filenames that could have the path attached to them
- sanitizeNode for individual segments, does not take into account the path whatsoever

app/test/testutilsfunctions.cpp
gabriel-bolbotina: I agree, but there are some cases where expected test results are not consistent. I think we should align on that, for example, the testSanitizeFileName used to expect the whole path to be sanitized when the function's responsibility was to only sanitize the file name

app/inpututils.cpp
gabriel-bolbotina: I think QLatin1Char is more efficient in this case because it does not create a whole Qstring object and it is used only for a character. 
Are there any benefits to changing them to QStringLiteral in the modified functions?

app/test/testutilsfunctions.cpp
```diff
@@ -1027,7 +1003,17 @@ void TestUtilsFunctions::testSanitizePath()
   // unchanged
   str = QStringLiteral( "/complex/valid/Φ!l@#äme$%^&()-_=+[]{}`~;',.ext" );
   InputUtils::sanitizePath( str );
-  QCOMPARE( str, QStringLiteral( "/complex/valid/Φ!l@#äme$%^&()-_=+[]{}`~;',.ext" ) );
+  QCOMPARE( str, QStringLiteral( "/complex/valid/Φ!l@_äme$%^&()-_=+[]{}`~;',.ext" ) );
```
Withalion, line 1006: Could you please note down somewhere that `#` is prohibited character now

app/inpututils.cpp
Withalion: After reading the [documentation](https://doc.qt.io/qt-6/string-processing.html) it doesn't look like it's that straight forward. This [part](https://doc.qt.io/qt-6/string-processing.html#string-class-for-using-api) talks mostly about what we are discussing. However I would like to highlight one thing in the beginning:

> Generally, [QString](https://doc.qt.io/qt-6/qstring.html) can be used everywhere and it will perform fine.

app/inpututils.cpp
```diff
@@ -1050,38 +1044,31 @@ QString InputUtils::resolvePath( const QString &path, const QString &homePath, c
...
-    if ( prefixCanonicalPath.length() > 1 && canonicalPath.startsWith( prefixCanonicalPath ) )
-    {
-      return canonicalPath.replace( prefixCanonicalPath, QString() );
-    }
+  // use QDir to calculate the relative path
+  const QDir prefixDir( prefixPath );
+  QString relativePath = prefixDir.relativeFilePath( cleanPath );
+
+  // check if the path starts with ".." or is absolute (on Windows/different drives), it's not a "child"
+  if ( relativePath.startsWith( QLatin1String( ".." ) ) || QDir::isAbsolutePath( relativePath ) )
```
Withalion:
```suggestion
  if ( relativePath.startsWith( QLatin1StringView( ".." ) ) || QDir::isAbsolutePath( relativePath ) )
```
`QLatin1String` is a synonym for `QLatin1StringView`, but it's softly deprecated

app/inpututils.cpp
```diff
@@ -1971,64 +1958,106 @@ QUrl InputUtils::iconFromGeometry( const Qgis::GeometryType &geometry )
...
-    const QString base = fileName.first( lastDot ).trimmed();
-    const QString ext = fileName.sliced( lastDot );
-    fileName = base + ext;
+    const static QRegularExpression reservedNames(
+      QStringLiteral( "^(CON|PRN|AUX|NUL|COM[1-9]|LPT[1-9])$" ),
+      QRegularExpression::CaseInsensitiveOption
+    );
+    if ( reservedNames.match( base.trimmed() ).hasMatch() )
+    {
+      base = base.trimmed() + QLatin1Char( '_' );
```
Withalion: this will get changed to `QChar`

app/inpututils.cpp
```diff
@@ -1971,64 +1958,106 @@ QUrl InputUtils::iconFromGeometry( const Qgis::GeometryType &geometry )
...
+    }
   }
+#endif
+
+  // reassemble
+  if ( !suffix.isEmpty() )
+  {
+    return base + QLatin1Char( '.' ) + suffix;
```
Withalion: this will also get changed to `QChar`

app/inpututils.cpp
```diff
@@ -1971,64 +1958,106 @@ QUrl InputUtils::iconFromGeometry( const Qgis::GeometryType &geometry )
...
+  // check if the path has a file prefix
+  // QUrl treats ? as a query start, # as a fragment
+  // so we modify the cleanPath for the file prefix check
+  cleanPath.replace( QStringLiteral( "?" ), QStringLiteral( "_" ) );
+  cleanPath.replace( QStringLiteral( "#" ), QStringLiteral( "_" ) );
 
-  if ( pathStartsWithFileURL )
+  // parse file prefix and path
+  QUrl url = QUrl::fromUserInput( cleanPath );
+  bool isUrl = path.startsWith( QLatin1String( "file:" ), Qt::CaseInsensitive );
```
Withalion:
```suggestion
  const QUrl url = QUrl::fromUserInput( cleanPath );
  const bool isUrl = path.startsWith( QStringLiteral( "file:" ), Qt::CaseInsensitive );
```

Withalion (commented):
Just some small changes to strings as we discussed

PR #4317, opened 2026-02-12
Add photo card component and update photo component

gabriel-bolbotina:
I created a new component called 'Photo card' which will be used instead of MMPhoto in the Form Gallery editor for features that have multiple photos.

This component has the following configurable properties:
- image source
- text
- boolean value to show text or not (the logic could be replaced based on the text property), which, by default, is set to `false`
- height/width

To implement the blur using MultiEffects, I have modified the MMPhoto component by changing the round edged mask to use MultiEffects. 

Here are some visuals:

Capture d’écran 2026-02-12 à 15 11 35

https://github.com/user-attachments/assets/6482c8d0-d27f-4f07-b1ea-e34867eb6e2a

Also, I tested that MMPhoto behaves as before:
https://github.com/user-attachments/assets/bec5747e-c64c-475f-98e4-001456675e70

And the gallery photo page:
image

And the current default state of the app:

https://github.com/user-attachments/assets/a3baac71-e951-4245-87e6-20c671933b51


app/qml/components/MMPhotoCard.qml
```diff
@@ -0,0 +1,100 @@
...
+      id: bngImage
+      anchors.fill: parent
+      fillMode: Image.PreserveAspectCrop
+      autoTransform: true
+      smooth: true
+    }
+
+    // the footer of the card
+    // contains the text to display over the blurred part of the image
+    FastBlur {
```
Withalion: As this is a new code let's move away from deprecated modules to something new. You'll be the guinea pig, but let's try to use [MultiEffect](https://doc.qt.io/qt-6/qml-qtquick-effects-multieffect.html).

Some more info here: [Qt5Compat module](https://doc.qt.io/qt-6/qtgraphicaleffects5-index.html)

app/qml/components/MMPhotoCard.qml
```diff
@@ -0,0 +1,100 @@
...
+  id: root
+
+  property alias imageSource: bngImage.photoUrl
+  property string text: ""
+  property bool textVisible: true
+
+  signal clicked(string path)
+
+  width: 120
+  height: 120
```
Withalion: we probably want to expose `size`, which will set `width` && `height`. same as in MMPhoto, but a bit better

gallery/qml/pages/PhotoCardPage.qml
Withalion, line 1: I guess they should be showcased in Photos page. but it's true Gallery will need some fixing first for it to work

gabriel-bolbotina: That's why I created a different page, but I could merge them and make it work as well.

app/qml/components/MMPhotoCard.qml
gabriel-bolbotina: I tried to implement the MultiEffect footer, but there seems to be a problem with how it handles the bottom margins 
image
  
I tried to find some solutions https://forum.qt.io/topic/145956/rounded-image-in-qt6/6. but they did not work as previously

app/qml/components/MMPhotoCard.qml
```diff
@@ -0,0 +1,100 @@
...
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+import QtQuick
+import QtQuick.Controls
+import Qt5Compat.GraphicalEffects
```
tomasMizera: We do not use anything from `Qt5Compat`, right?

gabriel-bolbotina: Initially, I used the OpacityMask and the FastBlur. Then, Matej suggested using MultiEffects, which worked but there is one issue I didn't solve (check the first commentary)

gallery/qml/components/FormPhotoViewer.qml
Withalion, line 1: Is there a reason why we copied the whole file into gallery and didn't reuse `MMFormPhotoViewer` from the app (which was specifically made for this use case) ?

gabriel-bolbotina: Because the original MMFormViewer has the photoSketchingLoader and other dependencies were required for this to work. I think we can park the topic of testing the photo sketching in the gallery app, and just keep it a visual tool

Withalion: yeah we don't want  to test that, can we just pass `null` there and make it work?

app/qml/components/MMPhoto.qml
```diff
@@ -8,45 +8,44 @@
...
-          radius: 20 * __dp
-        }
-      }
+  layer.effect: Component {
+    MultiEffect {
+      maskEnabled: true
+      maskSource: maskRect
+      autoPaddingEnabled: false
     }
   }
```
Withalion:
```suggestion
  layer.effect: MultiEffect {
      maskEnabled: true
      maskSource: maskRect
      autoPaddingEnabled: false
}
```

app/qml/components/MMPhoto.qml
```diff
@@ -8,45 +8,44 @@
  ***************************************************************************/
```
Withalion, line 10:
```suggestion
  pragma ComponentBehavior: Bound
  
```

Withalion: Because of the `maskRect` reference in `MultiEffect`

app/qml/components/MMPhotoCard.qml
```diff
@@ -0,0 +1,103 @@
...
+import QtQuick
+import QtQuick.Effects
+
+Item {
+  id: root
+
+  property alias imageSource: bngImage.photoUrl
+  property string text: ""
+  property bool textVisible: true
+  property int size: 120
```
Withalion:
```suggestion
  property int size: 120 * __dp
```

app/qml/components/MMPhotoCard.qml
```diff
@@ -0,0 +1,103 @@
+/***************************************************************************
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+import QtQuick
```
Withalion, line 11:
```suggestion
pragma ComponentBehavior: Bound

import QtQuick
```

Withalion: same as in `MMPhoto`, you are referencing component outside of hierarchy

app/qml/components/MMPhotoCard.qml
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  MultiEffect{
+    id: blurFooter
+    height: parent.height * 0.33
+    visible: root.textVisible
+    anchors {
+      left: parent.left
+      right: parent.right
+      bottom: parent.bottom
```
Withalion:
```suggestion
    height: root.height * 0.33
    visible: root.textVisible
    anchors {
      left: root.left
      right: root.right
      bottom: root.bottom
```

app/qml/components/MMPhotoCard.qml
```diff
@@ -0,0 +1,103 @@
...
+    verticalAlignment: Text.AlignVCenter
+
+    wrapMode: Text.WordWrap
+    elide: Text.ElideRight
+  }
+
+  MMSingleClickMouseArea {
+    anchors.fill: parent
+    onSingleClicked: root.clicked(root.imageSource)
+  }
```
Withalion, line 102:
```suggestion
  MMSingleClickMouseArea {
    anchors.fill: root
    onSingleClicked: root.clicked(root.imageSource)
  }
```

app/qml/components/MMPhotoCard.qml
```diff
@@ -0,0 +1,103 @@
...
+    width: root.width
+    height: root.height
+    radius: 20 * __dp
+    visible: false
+    layer.enabled: true
+  }
+
+  MMPhoto {
+    id: bngImage
+    anchors.fill: parent
```
Withalion:
```suggestion
    anchors.fill: root
```

app/qml/components/MMPhotoCard.qml
```diff
@@ -0,0 +1,103 @@
...
+    fillMode: Image.PreserveAspectCrop
+    autoTransform: true
+    smooth: true
```
Withalion:
```suggestion
 
```
true by default

app/qml/components/MMPhotoCard.qml
```diff
@@ -0,0 +1,103 @@
...
+    radius: 20 * __dp
+    visible: false
+    layer.enabled: true
+  }
+
+  MMPhoto {
+    id: bngImage
+    anchors.fill: parent
+    fillMode: Image.PreserveAspectCrop
+    autoTransform: true
```
Withalion:
```suggestion
    
```
already set in `MMPhoto`

gallery/qml/components/FormPhotoViewer.qml
gabriel-bolbotina: I tried it, but it fails for the MMInput that is imported by MMFormPhotoViewer with this message "module "MMInput" is not installed". I added it to the gallery's CMakeLists. If you have any suggestions to install it or a workaround, let me know

Withalion: Okay yeah that won't work, we will probably need to rip apart the `photosketchingcontroller` from viewer first

Withalion: Let's keep that for future us :)

tomasMizera (approved):
Looks good 👍🏻 Thanks

PR #4338, opened 2026-02-17
Network position provider

### LINKED ISSUE #1893: Support for network position provider
Bluetooth GPS support appears to be a dead-end for iOS. Most high-end devices support network position provision instead.

Withalion:
fixes #1893 
This PR adds new network position provider support. The user is able to connect both UDP and TCP providers.

New UI:
| Provider picker | Network provider info |
|---|---|
| image | image |

| Network information drawers | Bluetooth information drawers |
| --- | --- |
| image |  |
|  image |  |
| image |  |

| Providers list |
| --- |
| image |


Withalion:
I fixed some issues found with @IvaKuklica during testing.
- Fixed cpp check failing
- Fixed whitespace trimming for new network provider names
- Fixed missing source name in GPS panel
- Reworked default provider naming
  - Default name is  `Network device` 
  - Next default name will get number attached -> `Network device (1)`
- Fixed position processing not being used
- Fixed crash on app exit

Some more stuff that I noticed and should be fixed:
- connecting drawer popping up even when the provider is already saved ( I wouldn't do that as we didn't before, only use it when adding new provider)
- We should use the new error icon when the app fails to connect to network provider 
  - Also the title text there includes "Try again later" which conflicts with what we are doing (trying to connect again automatically after some time)
- "Try again" button missing for bluetooth provider on the failed connection screen ? (not sure about this, it's in figma)
- We should have only one "External receivers section"



**NOTE FOR TESTERS**
We were modifying code for both external providers so make sure bluetooth providers survive the transition and are detected + working correctly

app/qml/components/MMIconCheckBoxHorizontal.qml
Withalion, line 1: Now the file should be in general components as it is not only used in account pages

app/qml/account/components/MMIconCheckBoxHorizontal.qml
```diff
@@ -17,40 +17,75 @@ CheckBox {
...
+
+  leftPadding: (description !== "" ? iconBgRectangle.x : 0) + iconBgRectangle.width + 30 * __dp
+  rightPadding: 20 * __dp
 
   indicator: Rectangle {
     id: iconBgRectangle
-    width: (control.small ? 24 : 40) * __dp
-    height: (control.small ? 24 : 40) * __dp
+    width: (control.small ? 24 : (control.description !== "" ? 50 : 40)) * __dp
+    height: (control.small ? 24 : (control.description !== "" ? 50 : 40)) * __dp
```
Withalion: Please don't nest ternary operators, it's unreadable

app/qml/account/components/MMIconCheckBoxHorizontal.qml
```diff
@@ -17,40 +17,75 @@ CheckBox {
   id: control
 
   property string sourceIcon: ""
+  property string description: ""
+  property bool showBorder: false
   property bool small: false
 
-  height: (control.small ? 50 : 80) * __dp
+  height: (description !== "" ? 96 : (control.small ? 50 : 80)) * __dp
```
Withalion: `control.small`  -> `small`

app/qml/account/components/MMIconCheckBoxHorizontal.qml
```diff
@@ -17,40 +17,75 @@ CheckBox {
...
+        elide: Text.ElideRight
+      }
+    }
   }
 
   background: Rectangle {
     radius: __style.radius12
-    color: control.checked ? __style.forestColor: __style.polarColor
+    color: control.checked ? __style.forestColor : __style.polarColor
+    border.color: showBorder ? ( control.checked ? __style.transparentColor : __style.mediumGreenColor ) : __style.transparentColor
```
Withalion:
```suggestion
    border.color: showBorder && !control.checked ? __style.mediumGreenColor : __style.transparentColor
```

app/qml/components/MMIconCheckBoxHorizontal.qml
```diff
@@ -17,40 +17,75 @@ CheckBox {
...
+      id: titleText
+      visible: control.description === ""
+      width: parent.width
+      height: parent.height
+      text: control.text
+      font: __style.t3
+      color: control.checked ? __style.polarColor : __style.nightColor
+      elide: Text.ElideRight
+      verticalAlignment: Text.AlignVCenter
+    }
```
Withalion, line 59: this seems unnecessary. let's just change the visibility of the description text inside column and also remove the `Item` wrapper

app/qml/account/components/MMIconCheckBoxHorizontal.qml
```diff
@@ -17,40 +17,75 @@ CheckBox {
   id: control
 
   property string sourceIcon: ""
+  property string description: ""
```
Withalion: these two can be aliases and `sourceIcon` is actually url

app/qml/gps/MMBluetoothProviderDrawer.qml
Withalion, line 1: It would be great to change the name of this file to `MMBluetoothProviderDrawer.qml` or something more descriptive

app/qml/gps/MMBluetoothProviderDrawer.qml
```diff
@@ -21,7 +21,9 @@ MMComponents.MMListDrawer {
   signal initiatedConnectionTo( string deviceAddress, string deviceName )
 
   showFullScreen: true
-  drawerHeader.title: qsTr( "Connect to bluetooth device" )
+  drawerHeader.title: qsTr( "Bluetooth Connection" )
+
+  list.height: root.drawerContentAvailableHeight
```
Withalion, line 27: What is this solving? the same thing is getting done inside `MMListDrawer`

app/qml/gps/MMAddPositionProviderDrawer.qml
Withalion:
```suggestion

```
unnecessary

app/qml/gps/MMAddPositionProviderDrawer.qml
```diff
@@ -97,22 +99,49 @@ MMComponents.MMListDrawer {
...
+      Column {
+        id: contentColumn
 
-      MMComponents.MMListSpacer { height: __style.margin40 }
+        anchors.bottom: parent.bottom
+        anchors.horizontalCenter: parent.horizontalCenter
 
-      MMComponents.MMText {
         width: parent.width
+        spacing: __style.spacing16
```
Withalion: why do we need both spacing and also spacers inside the column?

app/qml/gps/MMBluetoothProviderDrawer.qml
```diff
@@ -97,22 +99,49 @@ MMComponents.MMListDrawer {
...
       width: ListView.view.width
+      height: Math.max( contentColumn.implicitHeight,
+                        ListView.view.height - ( ListView.view.contentHeight - height ) )
 
-      spacing: 0
+      Column {
+        id: contentColumn
 
-      MMComponents.MMListSpacer { height: __style.margin40 }
+        anchors.bottom: parent.bottom
```
Withalion, line 119: shouldn't the item be as big as the content?

app/qml/gps/MMBluetoothProviderDrawer.qml
```diff
@@ -97,22 +99,49 @@ MMComponents.MMListDrawer {
...
 
+  // footer
   Component {
     id: discoveringMessageComponent
 
-    Column {
+    Item {
       width: ListView.view.width
+      height: Math.max( contentColumn.implicitHeight,
+                        ListView.view.height - ( ListView.view.contentHeight - height ) )
```
Withalion, line 114: why do we need to specify the height now? isn't the `list.footer` placing the footer at the bottom anyway? as I understand it now the footer grows to fill the available height

app/qml/gps/MMAddPositionProviderDrawer.qml
```diff
@@ -97,22 +99,49 @@ MMComponents.MMListDrawer {
...
+
+        Image {
+          anchors.horizontalCenter: parent.horizontalCenter
+
+          source: __style.mmSymbolImage
+
+          width: 32 * __dp
+          height: 32 * __dp
```
Withalion:
```suggestion
          width: __style.icon32 * __dp
          height: __style.icon32 * __dp
```

app/qml/gps/MMExternalProviderConnectionDrawer.qml
Withalion:
```suggestion

```

app/qml/gps/MMExternalProviderConnectionDrawer.qml
```diff
@@ -45,7 +69,11 @@ MMComponents.MMDrawer {
   property string descriptionText: {
     if ( rootstate.state === "working" )
     {
-      return qsTr( "You might be asked to pair your device during this process." )
+      if ( root.providerType === "bluetooth" )
+        return qsTr( "You might be asked to pair your device during this process." )
+      else if ( root.providerType === "network" )
+      if ( root.positionProvider.getIpAddress() ) return qsTr( "Connecting to" ) + " " + root.positionProvider.getIpAddress() + ". You can close this panel, the app will continue in the background."
```
Withalion: missing translation

app/qml/gps/MMExternalProviderConnectionDrawer.qml
Withalion, line 1: We should try to move away from these huge decision trees in properties to [PropertyChanges](https://doc.qt.io/qt-6/qml-qtquick-propertychanges.html) on states if possible

Withalion: Also another thing, majority if not all properties defined in this component feel unnecessary. If we start using `PropertyChanges` it would make sense to access the targets directly and not through intermediary properties, which are also exposed to parent.

Withalion (commented):
I'll review the rest tomorrow

app/qml/gps/MMGpsDataDrawer.qml
```diff
@@ -22,327 +22,344 @@ MMComponents.MMDrawer {
   property var mapSettings
 
   property bool showReceiversButton: true
+  readonly property real reservedButtonSpace: __style.spacing20 + manageButton.implicitHeight + __style.safeAreaBottom + __style.margin8
 
   signal manageReceiversClicked()
 
   drawerHeader.title: qsTr( "GPS info" )
 
+  drawerBottomMargin: 0
```
Withalion: What I don't like about this is that we are recalculating the bottom margin, which `MMDrawer` already does inside. I would expect that we would only work with the content here and let `MMDrawer` workout the margins.

app/qml/gps/MMGpsDataDrawer.qml
```diff
@@ -22,327 +22,344 @@ MMComponents.MMDrawer {
...
 
-  drawerContent: MMComponents.MMScrollView {
+  drawerContent: Item {
 
     width: parent.width
-    height: root.maxHeightHit ? root.drawerContentAvailableHeight : contentHeight
+    height: root.maxHeightHit ? root.drawerContentAvailableHeight
+                              : ( scrollView.contentHeight + reservedButtonSpace )
```
Withalion:
```suggestion
                              : ( scrollView.contentHeight + root.reservedButtonSpace )
```

app/qml/gps/MMGpsDataDrawer.qml
```diff
@@ -22,327 +22,344 @@ MMComponents.MMDrawer {
...
 
-    Column {
-      width: parent.width
+    MMComponents.MMScrollView {
+      id: scrollView
 
-      spacing: __style.spacing40
+      width: parent.width
+      height: root.maxHeightHit ? ( parent.height - reservedButtonSpace)
```
Withalion:
```suggestion
      height: root.maxHeightHit ? ( parent.height - root.reservedButtonSpace)
```

app/qml/gps/MMGpsDataDrawer.qml
```diff
@@ -22,327 +22,344 @@ MMComponents.MMDrawer {
   property var mapSettings
 
   property bool showReceiversButton: true
+  readonly property real reservedButtonSpace: __style.spacing20 + manageButton.implicitHeight + __style.safeAreaBottom + __style.margin8
```
Withalion: also we have "internal" object already so move it there if there is no other way

app/qml/gps/MMGpsDataDrawer.qml
Withalion, line 11:
```suggestion

```

app/qml/gps/MMNetworkProviderDrawer.qml
```diff
@@ -0,0 +1,131 @@
...
+      text: qsTr( "Enter the IP address and port for your receiver. You can find these in your manufacturer's app. TCP and UDP connections are supported. Optionally, add a nickname to tell your receivers apart." )
+      font: __style.p5
+      color: __style.nightColor
+
+      wrapMode: Text.Wrap
+      horizontalAlignment: Text.AlignJustify
+
+      onLinkActivated: function( link ) {
+        Qt.openUrlExternally( link )
+      }
```
Withalion: There is no link inside the text

app/qml/gps/MMNetworkProviderDrawer.qml
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    MMInputs.MMTextInput {
+      id: portInput
+
+      width: parent.width
+      textFieldBackground.color: __style.lightGreenColor
+
+      title: qsTr( "Port" )
+      placeholderText: qsTr( "1234" )
```
Withalion: translating this probably doesn't make sense

Withalion: unless we have some Romans among us :)

app/qml/gps/MMNetworkProviderDrawer.qml
```diff
@@ -0,0 +1,131 @@
...
+      text: qsTr( "Confirm" )
+
+      onClicked: {
+        const ip = ipAddressInput.text.trim()
+        const port = portInput.text.trim()
+
+        const ipv4Regex = /^((25[0-5]|2[0-4]\d|1\d{2}|[1-9]\d|\d)\.){3}(25[0-5]|2[0-4]\d|1\d{2}|[1-9]\d|\d)$/
+        const hostnameRegex = /^[a-zA-Z0-9]([a-zA-Z0-9\-\.]*[a-zA-Z0-9])?$/
+
+        if ( ip === "" ) {
```
Withalion:
```suggestion
        if ( !ip ) {
```

app/qml/gps/MMNetworkProviderDrawer.qml
```diff
@@ -0,0 +1,131 @@
...
+        }
+        else if ( !ipv4Regex.test( ip ) && !hostnameRegex.test( ip ) ) {
+          ipAddressInput.errorMsg = qsTr( "Enter a valid IP address or hostname" )
+        }
+        else {
+          ipAddressInput.errorMsg = ""
+        }
+
+        const portNum = parseInt( port )
+        if ( port === "" ) {
```
Withalion:
```suggestion
        if ( !port ) {
```

app/qml/gps/MMNetworkProviderDrawer.qml
```diff
@@ -0,0 +1,131 @@
...
+          portInput.errorMsg = qsTr( "Port is required" )
+        }
+        else if ( !/^\d+$/.test( port ) || portNum < 1 || portNum > 65535 ) {
```
Withalion: Doesn't this regex test if it's just numbers? But the previous if already tests that

app/qml/gps/MMNetworkProviderDrawer.qml
```diff
@@ -0,0 +1,131 @@
...
+          portInput.errorMsg = qsTr( "Enter a valid port (1–65535)" )
+        }
+        else {
+          portInput.errorMsg = ""
+        }
+
+        if ( ipAddressInput.errorMsg !== "" || portInput.errorMsg !== "" ) {
```
Withalion:
```suggestion
        if ( ipAddressInput.errorMsg || portInput.errorMsg ) {
```

app/qml/gps/MMPositionProviderPage.qml
Withalion, line 1: Regarding the Header of this page, we will do a small redesign :) Please move the title to the header of this page so it will look like this: 

Image

Motivation: Even though the animation of the title is cool it doesn't make sense here. I would expect majority of the users don't open this page at all and the users that do usually have very few externall devices. So, the design looks off now.

app/qml/gps/MMPositionProviderPage.qml
```diff
@@ -80,8 +76,18 @@ MMComponents.MMPage {
...
 
-        text: model.ProviderName ? model.ProviderName : qsTr( "Unknown device" )
-        secondaryText: listdelegate.isActive ? PositionKit.positionProvider.stateMessage : model.ProviderDescription
+        text: {
+          if ( model.ProviderName ) return model.ProviderName
+          return qsTr( "Unknown device" )
+        }
+        secondaryText: {
+          if ( listdelegate.isActive ) {
+            if ( model.ProviderType === "external_ip" )
```
Withalion: "Unqualified access: model is implicitly injected into this delegate. Add a required property instead."

app/qml/gps/MMPositionProviderPage.qml
```diff
@@ -80,8 +76,18 @@ MMComponents.MMPage {
...
+              return PositionKit.positionProvider.stateMessage + " - " + PositionKit.positionProvider.getIpAddress()
```
Withalion: QML error: `qrc:/com.merginmaps/imports/MMInput/gps/MMPositionProviderPage.qml:86: TypeError: Property 'getIpAddress' of object InternalPositionProvider(0x7f80e81058e0) is not a function`

app/qml/gps/MMPositionProviderPage.qml
```diff
@@ -80,8 +76,18 @@ MMComponents.MMPage {
           }
         }
 
-        text: model.ProviderName ? model.ProviderName : qsTr( "Unknown device" )
-        secondaryText: listdelegate.isActive ? PositionKit.positionProvider.stateMessage : model.ProviderDescription
+        text: {
+          if ( model.ProviderName ) return model.ProviderName
+          return qsTr( "Unknown device" )
```
Withalion: This is probably not needed anymore as we guarantee that network providers will have a name, but let's keep it as fail safe for now.

app/qml/gps/MMPositionProviderPage.qml
```diff
@@ -135,21 +141,47 @@ MMComponents.MMPage {
...
+    }
+
+    MMNetworkProviderDrawer {
+      id: networkProviderDrawer
+
+      onConfirmed: function( alias, deviceAddress ) {
+        PositionKit.positionProvider = PositionKit.constructProvider( "external_ip", deviceAddress, alias )
+        providersModel.addProvider( PositionKit.positionProvider.name(), deviceAddress, "external_ip" )
+        connectingDialogLoaderNetwork.open()
+      }
```
Withalion, line 158: this should call `root.constructProvider()` instead, so we won't duplicate the logic in every slot

app/qml/gps/MMPositionProviderPage.qml
```diff
@@ -135,21 +141,47 @@ MMComponents.MMPage {
...
     }
 
     MMComponents.MMMessage {
+      id: infoMessage
```
Withalion: never used

app/qml/gps/MMPositionProviderPage.qml
```diff
@@ -162,7 +194,7 @@ MMComponents.MMPage {
       }
 
       onRemoveProvider: {
-        if (removeDialog.providerId === "") {
+        if ( removeDialog.providerId === "" ) {
```
Withalion:
```suggestion
        if ( !removeDialog.providerId ) {
```

app/qml/gps/MMPositionProviderPage.qml
```diff
@@ -183,52 +215,73 @@ MMComponents.MMPage {
       id: bluetoothDiscoveryLoader
 
       active: false
-      sourceComponent: bluetoothDiscoveryDrawerComponent
+      source: Qt.resolvedUrl( "MMAddPositionProviderDrawer.qml" )
```
Withalion: Why did you choose this approach?

There are other solutions which make more sense:
- `Qt.createComponent`
- Wrap the element in `Component` and inline it (choose this one)

app/qml/gps/MMPositionProviderPage.qml
```diff
@@ -183,52 +215,73 @@ MMComponents.MMPage {
...
+    Connections {
+      target: bluetoothDiscoveryLoader.item
+
+      function onInitiatedConnectionTo( deviceAddress, deviceName ) {
+        PositionKit.positionProvider = PositionKit.constructProvider( "external_bt", deviceAddress, deviceName )
+        providersModel.addProvider( deviceName, deviceAddress, "external_bt" )
+        bluetoothDiscoveryLoader.item.list.model.discovering = false
+        bluetoothDiscoveryLoader.item.close()
+        connectingDialogLoader.open()
+      }
```
Withalion, line 218: same as above we should reuse `root.constructProvider()`

app/qml/gps/MMPositionProviderPage.qml
```diff
@@ -183,52 +215,73 @@ MMComponents.MMPage {
...
+      function onClosed() { bluetoothDiscoveryLoader.active = false }
     }
 
     Loader {
       id: connectingDialogLoader
 
       active: false
       asynchronous: true
-      sourceComponent: connectionToSavedProviderDialogComponent
+      source: Qt.resolvedUrl( "MMExternalProviderConnectionDrawer.qml" )
```
Withalion: same as above this approach smells

app/qml/gps/MMPositionProviderPage.qml
```diff
@@ -183,52 +215,73 @@ MMComponents.MMPage {
...
+
+      onLoaded: {
+        item.providerType = "bluetooth"
```
Withalion: I believe the `providerType` can be passed by the element activating the loader, which would mitigate the necessity for 2 loaders and 2 connection elements

app/qml/gps/MMPositionProviderPage.qml
```diff
@@ -183,52 +215,73 @@ MMComponents.MMPage {
...
+      target: connectingDialogLoaderNetwork.item
 
-      // revert position provider back to internal provider
-      onFailure: PositionKit.positionProvider = PositionKit.constructProvider( "internal", "devicegps", "" )
-
-      Component.onCompleted: open()
+      function onClosed() { connectingDialogLoaderNetwork.active = false }
+      function onFailure() { PositionKit.positionProvider = PositionKit.constructProvider( "internal", "devicegps", "" ) }
     }
   }
```
Withalion: Check the comment above how to remove this

app/qml/gps/MMPositionProviderPage.qml
```diff
@@ -239,7 +292,7 @@ MMComponents.MMPage {
...
 
   function constructProvider( type, id, name ) {
```
Withalion: As this function got called alongside `PositionKit.constructProvider` we could rename it to `activateProvider` or something similar, so it's easier to differentiate

app/qml/gps/MMProviderTypeDrawer.qml
```diff
@@ -0,0 +1,114 @@
+/***************************************************************************
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+
+import QtQuick
```
Withalion, line 11:
```suggestion
import QtQuick
import QtQml.Models
```

app/qml/gps/MMProviderTypeDrawer.qml
```diff
@@ -0,0 +1,114 @@
...
+import "../components" as MMComponents
+import "../account/components" as MMAccountComponents
+
+MMComponents.MMListDrawer {
+  id: root
+
+  property string selectedProviderType: ""
+
+  signal bluetoothSelected()
+  signal networkSelected()
```
Withalion: Let's define just one signal e.g `providerSelected( string providerType )` as we will have more types soon and it will also help with the code reduction in `MMPositionProviderPage`

app/qml/gps/MMProviderTypeDrawer.qml
```diff
@@ -0,0 +1,114 @@
...
+
+import QtQuick
+
+import "../components" as MMComponents
+import "../account/components" as MMAccountComponents
+
+MMComponents.MMListDrawer {
+  id: root
+
+  property string selectedProviderType: ""
```
Withalion: this property can be completely removed if we use the value of `type` in [currentItem](https://doc.qt.io/qt-6/qml-qtquick-listview.html#currentItem-prop)

app/qml/gps/MMProviderTypeDrawer.qml
```diff
@@ -0,0 +1,114 @@
...
+
+    MMAccountComponents.MMIconCheckBoxHorizontal {
+      id: checkbox
+
+      width: parent.width
+      showBorder: true
+      sourceIcon: model.icon
+      text: model.name
+      description: model.description
+      checked: root.selectedProviderType === model.type
```
Withalion: "Unqualified access: Set "pragma ComponentBehavior: Bound" in order to use IDs from outer components in nested components."

app/qml/gps/MMProviderTypeDrawer.qml
```diff
@@ -0,0 +1,114 @@
...
+
+      onClicked: {
+        if ( root.selectedProviderType === model.type ) {
+          root.selectedProviderType = ""
+        }
+        else {
+          root.selectedProviderType = model.type
+        }
+      }
```
Withalion: "Unqualified access: model is implicitly injected into this delegate. Add a required property instead."

It makes more sense to declare each property separately in this case

Withalion, line 87: As a part of the `selectedProviderType` refactoring we can remove this

app/qml/gps/MMProviderTypeDrawer.qml
```diff
@@ -0,0 +1,114 @@
...
+    }
+  }
+
+  list.footer: Item {
+    width: ListView.view.width
+    height: continueButton.height + __style.margin20
+
```
Withalion: It doesn't seem like we need the wrapper here

app/qml/gps/MMProviderTypeDrawer.qml
```diff
@@ -0,0 +1,114 @@
...
+    MMComponents.MMButton {
+      id: continueButton
+
+      width: parent.width
+      anchors.top: parent.top
+      anchors.topMargin: __style.margin8
+
+      text: qsTr( "Continue" )
+      enabled: root.selectedProviderType !== ""
```
Withalion: will be based on `currentIndex`

app/qml/gps/MMProviderTypeDrawer.qml
```diff
@@ -0,0 +1,114 @@
...
+
+      onClicked: {
+        root.close()
+
+        if ( root.selectedProviderType === "bluetooth" ) {
+          root.bluetoothSelected()
+        }
+        else if ( root.selectedProviderType === "network" ) {
+          root.networkSelected()
+        }
```
Withalion: change based on the comments above

gallery/qml/pages/ChecksPage.qml
```diff
@@ -94,9 +94,11 @@ ScrollView {
 
       MMAccountComponents.MMIconCheckBoxHorizontal {
```
Withalion: it would be also nice to use the border on this one or use another one with border

Withalion (commented):
The QML will need some more work, but besides that it looks nice 👍

app/qml/gps/MMBluetoothProviderDrawer.qml
gabriel-bolbotina: It is helping the height calculation to make the footer stick to the bottom of the page. Usually the footer is placed right after the last content item in MMListDrawer when not in full screen, and the figma design suggested to keep it at the bottom.

app/qml/gps/MMBluetoothProviderDrawer.qml
gabriel-bolbotina: This would place the footer right after the last item in the list. see reply from above

app/qml/gps/MMBluetoothProviderDrawer.qml
gabriel-bolbotina: Without this height calculation the footer is placed just under the list components, so it is depending on the list actual height, not the maximum one

Withalion: okay let's keep it as it is and in the future someone can refactor it into something nicer :)

app/position/providers/networkpositionprovider.cpp
```diff
@@ -0,0 +1,228 @@
...
+      emit positionChanged( GeoPosition() );
+    }
+  } );
+
+  NetworkPositionProvider::startUpdates();
+}
+
+void NetworkPositionProvider::startUpdates()
+{
+  // TODO: QHostAddress doesn't support hostname lookup (QHostInfo does)
```
tomasMizera, line 54: todo? :)

app/position/providers/networkpositionprovider.cpp
```diff
@@ -0,0 +1,228 @@
...
+      const QByteArray rawNmeaData = datagram.data();
+      const QString nmeaData( rawNmeaData );
+      const QgsGpsInformation gpsInfo = mNmeaParser.parseNmeaString( nmeaData );
+      GeoPosition transformedPosition = mPositionTransformer->processNetworkPosition( GeoPosition::fromQgsGpsInformation( gpsInfo ) );
+
+      emit positionChanged( transformedPosition );
+    }
+
+    // "connect" to peer if we are not already connecting
+    if ( mUdpSocket->state() != QAbstractSocket::ConnectedState && mUdpSocket->state() != QAbstractSocket::ConnectingState )
```
tomasMizera, line 125: We know it is not `ConnectedState` from the condition on line 100

app/position/providers/networkpositionprovider.cpp
```diff
@@ -0,0 +1,228 @@
...
+  }
+  else if ( state == QAbstractSocket::UnconnectedState )
+  {
+    const bool isUdpSocketListening = mUdpSocket->state() == QAbstractSocket::ConnectedState || mUdpSocket->state() == QAbstractSocket::BoundState || mUdpReconnectTimer.isActive();
+    if ( socket->socketType() == QAbstractSocket::TcpSocket && !isUdpSocketListening && QApplication::applicationState() == Qt::ApplicationActive )
+    {
+      setState( tr( "No connection" ), State::NoConnection );
+      startReconnectTimer();
+      // let's also invalidate current position since we no longer have connection
+      emit positionChanged( GeoPosition() );
```
tomasMizera, line 178: Spot on! This is missing in other providers too

app/position/providers/networkpositionprovider.cpp
```diff
@@ -0,0 +1,228 @@
...
+    }
+    else if ( socket->socketType() == QAbstractSocket::UdpSocket && QApplication::applicationState() == Qt::ApplicationActive )
+    {
+      setState( tr( "No connection" ), State::NoConnection );
+      startReconnectTimer();
+      // let's also invalidate current position since we no longer have connection
+      emit positionChanged( GeoPosition() );
+    }
```
tomasMizera, line 186: These two branches perform exactly the same calls

app/position/providers/networkpositionprovider.cpp
```diff
@@ -0,0 +1,228 @@
...
+
+#include "networkpositionprovider.h"
+
+#include <QApplication>
+#include <QNetworkDatagram>
+
+constexpr int ONE_SECOND_MS = 1000;
+
+NetworkPositionProvider::NetworkPositionProvider( const QString &addr, const QString &name, PositionTransformer &positionTransformer, QObject *parent )
+  : AbstractPositionProvider( addr, QStringLiteral( "external_ip" ), name, positionTransformer, parent ),
```
tomasMizera, line 18: 😜
```suggestion
  : AbstractPositionProvider( addr, u"external_ip"_s, name, positionTransformer, parent ),
```

tomasMizera (approved):
Great work!

app/position/providers/networkpositionprovider.cpp
Withalion: Probably it should be a note instead. But i can imagine in future someone will try to connect device via company url and we will fail on that.

IvaKuklica:
**External GPS Testing — Issue Summary**

Application:
Device/OS: :robot: Samsung Galaxy A53 / Android 15 
App: Mergin Maps:  2026.3.0 - 819851


_**_1. Inconsistent receiver identity between GPS receiver list and GPS info panel_**_
When a user creates a new network receiver with an IP address and port that already exists under a different name (e.g. 192.168.1xx.xx, :9001 saved as Test2), the following happens:
1. The new entry (Test gps) is not added to the receiver list
2. The app silently selects the existing matching entry (Test2) and marks it with a green circle
3. No warning or message is shown to inform the user that this configuration already exists
4. The GPS info panel shows the source as Test gps (the name the user just typed)
5. The Manage GPS receiver list and the GPS info panel report two different things sources


**2. Mismatch between acceptance criteria and displayed device name**
**Acceptance criteria:** 
When a network receiver is added with an IP address and port number but no nickname, the position provider drawer should default to displaying "Network Device".
**Actual behaviour:** 
GPS info panel → Source: displays "Network Device" ← matches acceptance criteria
Manage GPS receivers screen: displays "Unknown device" ← does not match


**3. Frozen GPS data after offline mode - No error notification on connection loss**
When the device is connected to an external GPS receiver via TCP/IP and enters offline mode (airplane mode), the app fails to detect and communicate the lost connection. The following behaviour is observed:

- _Behaviour during offline mode:_
External GPS receiver status remains "Connected" — no connection loss is indicated
GPS info panel displays frozen data from the last known position
No error or warning message is shown to inform the user that the data connection has been lost

- _Behaviour after returning to online mode:_
GPS data remains frozen — the connection is not automatically re-established
No notification or prompt is shown to the user
The only way to restore the connection is to manually switch to a different GPS receiver and switch back to the original one

_**_4. Connection flow and error messaging**_

When connecting to a network receiver, a pop-up _"Connecting to external receiver. Connecting to 192.168.1xx.x,port:9001. **You can close this panel, the app will continue in the background."**_ appears with a close (X) button. There are two distinct paths:

- Path A — User taps X (closes the pop-up):
The dialog closes silently. No error or warning message is shown. The user receives no feedback that the connection attempt failed or was interrupted. Looking at the receiver list, they may see the green circle on the selected device (see previous issue) and incorrectly assume the connection succeeded.
- Path B — User waits (~5 seconds):
The app completes the timeout and displays an error message indicating the connection could not be established, giving the user actionable feedback.

_**_5. Misleading green circle on failed connection_ (just UX/UI suggestion)**_

The selected receiver is marked with a green circle regardless of connection status, including when the connection has failed ("No connection, reconnecting..."). Users may mistake selection for a successful connection.

_**_6.  Duplicate "External receivers" sections displayed on screen**_

- **Acceptance criteria: O**nly one "External receivers" section should be displayed in the Manage GPS receivers screen.
- **Actual behaviour:** When a user connects via Bluetooth or TCP/IP, the "External receivers" section header appears multiple times on the screen (2–3 times)

**_7. Misleading error message and missing error icon on failed network connection_**
The title tells the user to "try again later" (implying they must act manually), while the body and footer confirm the app is already retrying automatically. These two messages directly contradict each other, leaving the user confused about whether they need to do anything or not.

Screenshot 2026-06-08 212211

app/position/providers/networkpositionprovider.cpp
Withalion: no

PR #4341, opened 2026-02-19
Fix constraint translations

### LINKED ISSUE #4334: Field validation texts are not translated
**Description**
Validation texts from `fieldvalidator.h`'s namespace are not translated, they are always in English. This is because the strings are static and initialized before translator is instantiated.

**Environment**
Production

**Steps to reproduce**
1. Set the app to any language that is fully translated and is not English
2. Open a form where you have a field with a set constraint (e.g. not null)

**Actual results**
1. Field will show "Field must not be empty" in English

**Expected results**
1. Field should show "Field must not be empty" translated in the target language

Withalion:
fixes #4334 
This PR moves constraint texts from separate namespace to `FieldValidator`.

| English language | Slovak language |
| ---- | ----- |
|Constraints_english |  Constraints_slovak |

The non english translations have to be added by translators on transifex first to be visible in the app.

### DIFF
--- a/app/attributes/fieldvalidator.cpp
+++ b/app/attributes/fieldvalidator.cpp
@@ -12,8 +12,8 @@
 #include "featurelayerpair.h"
 #include "mixedattributevalue.h"
 
-#include "qgsfield.h"
-#include "qgsvectorlayerutils.h"
+#include <qgsfield.h>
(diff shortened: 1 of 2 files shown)


tomasMizera (approved):
lgtm! :)

PR #4342, opened 2026-02-20
Fixed android predictive text input

gabriel-bolbotina:
Fixed android predictive text input that was waiting for confirmation, by checking if preeditText changed and if it did, then committing the input.

Tested on Android with the following keyboards:
- swiftKey
- gboard
- samsung keyboard

https://github.com/user-attachments/assets/75d27430-85e9-46a0-9552-50ff3438a615

Also tested on iOS, where it previously worked.

### DIFF
--- a/app/qml/inputs/MMSearchInput.qml
+++ b/app/qml/inputs/MMSearchInput.qml
@@ -61,7 +61,7 @@ MMPrivateComponents.MMBaseSingleLineInput {
     else
     {
       // trim the last whitespace from the user's input
-      root.searchText = root.text.replace( /\s+$/, "" )
+      root.searchText = root.text.trim()
     }
   }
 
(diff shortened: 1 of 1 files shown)


app/qml/inputs/MMSearchInput.qml
```diff
@@ -60,7 +60,8 @@ MMPrivateComponents.MMBaseSingleLineInput {
     }
     else
     {
-      root.searchText = root.text
+      // trim the last whitespace from the user's input
+      root.searchText = root.text.replace( /\s+$/, "" )
```
Withalion:
```suggestion
      root.searchText = root.text.trim()
```
Please use `trim` instead, it's more explicit than regex

app/qml/inputs/MMSearchInput.qml
```diff
@@ -70,7 +71,11 @@ MMPrivateComponents.MMBaseSingleLineInput {
     interval: root.emitInterval
     running: false
 
-    onTriggered: root.searchText = root.text
+    onTriggered: { 
+      // when using predictive text suggestions, a whitespace is sometimes added after the word
+      // trim the last whitespace from the user's input
+      root.searchText = root.text.replace( /\s+$/, "" )
```
Withalion: same as above

PR #4356, opened 2026-02-27
Foreign key child feature fix

gabriel-bolbotina:
Modified logic in attribute controller as such:
- if the child feature form has the foreign key field visible, keep the code as it was
- if the field is not visible, search for the field in the child layer (not only in the visible form fields), and pre-fill the relation

### DIFF
--- a/app/attributes/attributecontroller.cpp
+++ b/app/attributes/attributecontroller.cpp
@@ -129,10 +129,10 @@ void AttributeController::prefillRelationReferenceField()
   {
     QVariant foreignKey = mParentController->featureLayerPair().feature().attribute( fieldPair.referencedField() );
     QString referencingField = fieldPair.referencingField();
-    QgsVectorLayer *childLayer = mLinkedRelation.referencingLayer();
+    const QgsVectorLayer *childLayer = mLinkedRelation.referencingLayer();
     if ( childLayer )
     {
-      int fieldIndex = childLayer->fields().lookupField( referencingField );
+      const int fieldIndex = childLayer->fields().lookupField( referencingField );
       if ( fieldIndex != -1 )
       {
         mFeatureLayerPair.featureRef().setAttribute( fieldIndex, foreignKey );


app/attributes/attributecontroller.cpp
```diff
@@ -127,18 +127,34 @@ void AttributeController::prefillRelationReferenceField()
...
+      if ( itemData->field().name() == referencingField )
       {
-        QVariant fk = mParentController->featureLayerPair().feature().attribute( fieldPair.referencedField() );
-        setFormValue( itemData->id(), fk );
+        setFormValue( itemData->id(), foreignKey );
+        fieldFound = true;
         break;
       }
       ++formItemsIterator;
     }
```
tomasMizera: Nice catch! :) One suggestion, how about we skip (remove) this `while` loop altogether and search for the field the new way that you introduced?

app/attributes/attributecontroller.cpp
```diff
@@ -127,17 +127,16 @@ void AttributeController::prefillRelationReferenceField()
   const QList<QgsRelation::FieldPair> fieldPairs = mLinkedRelation.fieldPairs();
   for ( const QgsRelation::FieldPair &fieldPair : fieldPairs )
   {
-    QMap<QUuid, std::shared_ptr<FormItem>>::iterator formItemsIterator = mFormItems.begin();
-    while ( formItemsIterator != mFormItems.end() )
+    QVariant foreignKey = mParentController->featureLayerPair().feature().attribute( fieldPair.referencedField() );
+    QString referencingField = fieldPair.referencingField();
+    QgsVectorLayer *childLayer = mLinkedRelation.referencingLayer();
```
Withalion:
```suggestion
    const QgsVectorLayer *childLayer = mLinkedRelation.referencingLayer();
```

app/attributes/attributecontroller.cpp
```diff
@@ -127,17 +127,16 @@ void AttributeController::prefillRelationReferenceField()
...
+    if ( childLayer )
     {
-      std::shared_ptr<FormItem> itemData = formItemsIterator.value();
-      if ( itemData->field().name() == fieldPair.referencingField() )
+      int fieldIndex = childLayer->fields().lookupField( referencingField );
```
Withalion:
```suggestion
      const int fieldIndex = childLayer->fields().lookupField( referencingField );
```

Withalion (approved):

PR #4360, opened 2026-03-02
Fix position bugs

Withalion:
Fixes these minor bugs:
- Fixes elevation available outside local geoid models. 
- Adds geoid model info button for internal providers
- Fixes translation for geoid model info bubble

### DIFF
--- a/app/position/positionkit.cpp
+++ b/app/position/positionkit.cpp
@@ -51,7 +51,7 @@ QString PositionKit::positionCrs3DGeoidModelName()
   }
 
   // for internal providers return EGM96 model
-  if ( mPositionProvider->type() == QStringLiteral( "internal" ) && !mPosition.isMock)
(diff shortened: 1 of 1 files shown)


app/position/positionkit.cpp
```diff
@@ -50,6 +50,11 @@ QString PositionKit::positionCrs3DGeoidModelName()
     return mVerticalCrs.description();
   }
 
+  // for internal providers return EGM96 model
+  if ( mPositionProvider->type() == QStringLiteral( "internal" ) )
+  {
+    return QgsCoordinateReferenceSystem::fromEpsgId( 5773 ).description();
```
tomasMizera, line 56: Hm, but this is true only for Android, no?

Withalion: mostly how about we show it for ios only if it's not mocked location?

PR #4366, opened 2026-03-03
Fix minor position issues on iOS

Withalion:
Fixes:
- elevation info icon showing EGM model when using uknown
- elevation available when outside local geoid model while using mocked position

### DIFF
--- a/app/position/positionkit.cpp
+++ b/app/position/positionkit.cpp
@@ -50,7 +50,7 @@ QString PositionKit::positionCrs3DGeoidModelName()
     return mVerticalCrs.description();
   }
 
-  // for internal providers return EGM96 model
+  // for internal providers return EGM96 model if it's not mock, and we haven't got orthometric height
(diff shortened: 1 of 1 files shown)


app/test/testposition.cpp
```diff
@@ -702,27 +702,12 @@ void TestPosition::testPositionTransformerInternalIosPosition()
...
-
-
   // mocked
   geoPosition.setAttribute( QGeoPositionInfo::MagneticVariation, 1 );
   // transform with pass through disabled, position is mocked and orthometric
   newPosition = positionTransformer.processInternalIosPosition( geoPosition );
 
-  QCOMPARE( newPosition.elevation, 171.3 );
+  QCOMPARE( newPosition.elevation, 127.53574931171875 );
   QCOMPARE( newPosition.elevation_diff, std::numeric_limits<double>::quiet_NaN() );
```
gabriel-bolbotina: Shouldn't you check that the elevation_diff is 43.76 as above?

PR #4378, opened 2026-03-09
Fixed photo orientation bug

gabriel-bolbotina:
Ready for review

Solution: instead of erasing the exif data key for image orientation, it is not set to 1 (which means no rotation). I also modified the unit test to check that the value of that key is set to 1, when calling the modified function.

Tested on iOS, Android and Windows

### DIFF
--- a/app/imageutils.cpp
+++ b/app/imageutils.cpp
@@ -155,7 +155,7 @@ bool ImageUtils::clearOrientationMetadata( const QString &sourceImage )
     Exiv2::ExifData &exifData = srcImage->exifData();
     if ( !exifData.empty() )
     {
-      auto exifIt = exifData.findKey( Exiv2::ExifKey( "Exif.Image.Orientation" ) );
+      const auto exifIt = exifData.findKey( Exiv2::ExifKey( "Exif.Image.Orientation" ) );
       if ( exifIt != exifData.end() )
       {
         exifIt->setValue( "1" ); // 1 = no rotation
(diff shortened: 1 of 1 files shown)


app/imageutils.cpp
```diff
@@ -150,21 +150,43 @@ bool ImageUtils::clearOrientationMetadata( const QString &sourceImage )
...
 
     srcImage->readMetadata();
+    bool metadataModified = false;
+
     Exiv2::ExifData &exifData = srcImage->exifData();
-    if ( exifData.empty() )
+    if ( !exifData.empty() )
     {
-      return true;
+      auto exifIt = exifData.findKey( Exiv2::ExifKey( "Exif.Image.Orientation" ) );
```
Withalion:
```suggestion
      const auto exifIt = exifData.findKey( Exiv2::ExifKey( "Exif.Image.Orientation" ) );
```

app/imageutils.cpp
```diff
@@ -150,21 +150,43 @@ bool ImageUtils::clearOrientationMetadata( const QString &sourceImage )
...
+      {
+        exifIt->setValue( "1" ); // 1 = no rotation
+        metadataModified = true;
+      }
+    }
+
+    Exiv2::XmpData &xmpData = srcImage->xmpData();
+    if ( !xmpData.empty() )
+    {
+      auto xmpIt = xmpData.findKey( Exiv2::XmpKey( "Xmp.tiff.Orientation" ) );
```
Withalion:
```suggestion
      const auto xmpIt = xmpData.findKey( Exiv2::XmpKey( "Xmp.tiff.Orientation" ) );
```

app/imageutils.cpp
```diff
@@ -150,21 +150,43 @@ bool ImageUtils::clearOrientationMetadata( const QString &sourceImage )
...
-    srcImage->setExifData( exifData );
-    srcImage->writeMetadata();
+
     return true;
   }
+  catch ( const Exiv2::Error &e )
+  {
+    CoreUtils::log( "Editing EXIF", QStringLiteral( "Failed to clear orientation: " ) + QString::fromStdString( e.what() ) );
+    return false;
+  }
```
Withalion: AFAIK there should be no exception thrown besides exiv, also I would use `.arg()` instead of string concatenation

PR #4382, opened 2026-03-10
Remove Qt5Compat effects

Withalion:
This PR removes Qt5Compat effects and replaces them with `MultiEffect` or some other approach. From the crash reports looks like they trigger quite a lot of crashes for users.

Only one reference to qt5compat has been left in `MMPhoto`, which should be refactored in #4317


app/mmstyle.h
```diff
@@ -401,7 +401,7 @@ class MMStyle: public QObject
     QColor photoSketchingBlueColor() const {return QColor::fromString( "#5E9EE4" );}
     QColor photoSketchingPinkColor() const {return QColor::fromString( "#FF8F93" );}
 
-    QColor shadowColor() const {return QColor::fromString( "#66777777" );}
+    QColor shadowColor() const {return QColor::fromString( "#4D000000" );}
```
gabriel-bolbotina, line 405: Why did you change the shadow color?

app/qml/components/MMIcon.qml
```diff
@@ -19,27 +19,25 @@ Item {
...
 
   width: size
   height: size
 
-  Image {
-    id: icon
-
+  // IconImage is not part of QML's public API, so this can break on Qt version change.
+  // However, if we don't want to use shaders, this is the most straightforward way.
+  IconImage {
```
gabriel-bolbotina, line 28: Is there a way to implement this so that it would work with a Qt version change?

gabriel-bolbotina (changes requested):
Looks good overall, just a few things to clarify

app/mmstyle.h
Withalion: Because the old one didn't produce shadow at all with the new approach. For the new approach only alpha channel is important and this color looked the most similar to the old one.

app/qml/components/MMIcon.qml
Withalion: Well we can always use shaders to do that, but this **might** break. It doesn't mean it will break every Qt release.

gabriel-bolbotina (approved):

app/qml/components/MMIcon.qml
```diff
@@ -19,27 +19,25 @@ Item {
   required property url source
   property real size: __style.icon24
   property color color: __style.forestColor
+  readonly property bool colorable: source.toString().endsWith("-coloroverlay.svg")
```
tomasMizera: This does not seem to be used anywhere

app/qml/components/MMIcon.qml
```diff
@@ -19,27 +19,25 @@ Item {
...
-    id: overlay
-
-    // Use original colors for some icons, e.g. QGIS layer icons
-    visible: !root.source.toString().endsWith("-nocoloroverlay.svg")
-
-    color: root.color
-    anchors.fill: icon
-    source: icon
+  Image {
+    visible: root.source.toString().endsWith("-nocoloroverlay.svg")
```
tomasMizera: Best to cut this out somewhere as 
```qml
QtObject {
  id: internal
  property bool iconSupportsRecolor: !root.source.toString().endsWith("-nocoloroverlay.svg")
}
```


and then

```suggestion
    visible: internal.iconSupportsRecolor
```

tomasMizera (approved):

PR #4385, opened 2026-03-12
Artefacts build bot CI update

Withalion:
This PR adds google play store link as android check output, which is then picked by `app_delivery` action. See the last build table for results.

### DIFF
--- a/app/main.cpp
+++ b/app/main.cpp
@@ -6,6 +6,7 @@
  *   (at your option) any later version.                                   *
  *                                                                         *
  ***************************************************************************/
+
 #include "mmconfig.h"
 
 #include <QFontDatabase>


app/main.cpp
gabriel-bolbotina, line 1: Is this needed for this PR?

Withalion: Nah, can you comment on this change and propose reverting it while reviewing? So we can commit directly from GitHub.

app/main.cpp
```diff
@@ -6,7 +6,6 @@
  *   (at your option) any later version.                                   *
  *                                                                         *
  ***************************************************************************/
-
 #include "mmconfig.h"
```
gabriel-bolbotina, line 10:
```suggestion

#include "mmconfig.h"
```

PR #4386, opened 2026-03-12
Added notification when a project has a newer version available on the server

### LINKED ISSUE #1769: Check for a new version of a project when it gets loaded
When loading a project (when the app starts or when the project is selected), it would be nice if Input checked whether the project is at the latest version, and if not, we would show a notification.

Related to #1566

gabriel-bolbotina:
Added a function in mergin api that checks every 5 seconds if the project needs a sync and fires a notification in case it is needed

Tested to work for the currently open project and when switching from one project to another

Use case below:
https://github.com/user-attachments/assets/e0c01bec-5a55-4e22-82a4-2a50c673825b

### DIFF
--- a/app/activeproject.h
+++ b/app/activeproject.h
@@ -26,6 +26,8 @@
 #include "synchronizationoptions.h"
 #include "filter/filtercontroller.h"
 
+class MerginApi;
+
 /**
  * \brief The ActiveProject class can load a QGIS project and holds its data.
  */
@@ -49,6 +51,7 @@ class ActiveProject: public QObject
       AppSettings &appSettings
       , ActiveLayer &activeLayer
       , LocalProjectsManager &localProjectsManager
+      , MerginApi *merginApi
       , QObject *parent = nullptr );
 
     virtual ~ActiveProject();
@@ -201,6 +204,9 @@ class ActiveProject: public QObject
 
     void requestSync( SyncOptions::RequestOrigin requestOrigin = SyncOptions::RequestOrigin::ManualRequest );
 
+    //! Checks whether the currently loaded project has a newer version available on the server
+    Q_INVOKABLE void checkForProjectUpdate();
+
   private:
 
     /**
@@ -229,6 +235,7 @@ class ActiveProject: public QObject
     LocalProject mLocalProject;
 
     AppSettings &mAppSettings;
+    MerginApi *mMerginApi = nullptr;
     ActiveLayer &mActiveLayer;
     QgsAuthManager *mAuthManager = nullptr;
     LocalProjectsManager &mLocalProjectsManager;
--- a/app/activeproject.h
+++ b/app/activeproject.h
@@ -26,8 +26,6 @@
 #include "synchronizationoptions.h"
 #include "filter/filtercontroller.h"
 
-class MerginApi;
-
 /**
  * \brief The ActiveProject class can load a QGIS project and holds its data.
  */
@@ -51,7 +49,6 @@ class ActiveProject: public QObject
       AppSettings &appSettings
       , ActiveLayer &activeLayer
       , LocalProjectsManager &localProjectsManager
-      , MerginApi *merginApi
       , QObject *parent = nullptr );
 
     virtual ~ActiveProject();
@@ -179,6 +176,8 @@ class ActiveProject: public QObject
 
     void syncActiveProject( const LocalProject &project, SyncOptions::RequestOrigin requestOrigin );
 
+    void projectSyncCheckRequested( const QString &projectFullName, bool withAuth );
+
     void mapThemeChanged( const QString &mapTheme );
 
     void positionTrackingSupportedChanged();
@@ -235,7 +234,6 @@ class ActiveProject: public QObject
     LocalProject mLocalProject;
 
     AppSettings &mAppSettings;
-    MerginApi *mMerginApi = nullptr;
     ActiveLayer &mActiveLayer;
     QgsAuthManager *mAuthManager = nullptr;
     LocalProjectsManager &mLocalProjectsManager;
--- a/app/notificationmodel.h
+++ b/app/notificationmodel.h
@@ -44,7 +44,7 @@ class NotificationType
       ShowProjectIssuesAction,
       ShowSwitchWorkspaceAction,
       ShowSyncFailedDialog,
-      ShowProjectNewVersionAction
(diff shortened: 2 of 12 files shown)


app/qml/main.qml
```diff
@@ -84,6 +84,10 @@ ApplicationWindow {
 
         // Stop/Start sync animation when user goes to map
         syncButton.iconRotateAnimationRunning = ( __syncManager.hasPendingSync( __activeProject.projectFullName() ) )
+
+        if ( __activeProject.projectFullName() !== "" ) {
+          __merginApi.isProjectSyncNeeded( __activeProject.projectFullName(), true )
+        }
```
```suggestion
        if ( __activeProject.isProjectLoaded() ) {
          __merginApi.isProjectSyncNeeded( __activeProject.projectFullName(), true )
        }
```

app/qml/main.qml
```diff
@@ -1159,6 +1177,9 @@ ApplicationWindow {
 
       AppSettings.defaultProject = __activeProject.localProject.qgisProjectFilePath ?? ""
       AppSettings.activeProject = __activeProject.localProject.qgisProjectFilePath ?? ""
+      if ( __activeProject.projectFullName() !== "" ) {
+        __merginApi.isProjectSyncNeeded( __activeProject.projectFullName(), true )
+      }
```
```suggestion
      if ( __activeProject.isProjectLoaded() ) {
        __merginApi.isProjectSyncNeeded( __activeProject.projectFullName(), true )
      }
```

app/qml/main.qml
Withalion, line 1: I would move the logic from here to `Activeproject`. To be precise trigger `MerginApi::isProjectSyncNeeded()` from `ActiveProject` after project is loaded. If there is newer version emit signal which will be connected back here and just trigger the notification from QML.

app/notificationmodel.h
```diff
@@ -43,7 +43,8 @@ class NotificationType
       NoAction,
       ShowProjectIssuesAction,
       ShowSwitchWorkspaceAction,
-      ShowSyncFailedDialog
+      ShowSyncFailedDialog,
+      ShowProjectNewVersionAction
```
Withalion: Just a nitpick, but naming it `SyncProjectAction` or similar would make more sense

core/merginapi.cpp
```diff
@@ -4597,3 +4597,42 @@ bool MerginApi::serverVersionIsAtLeast( const int requiredMajor, const int requi
...
+void MerginApi::isProjectSyncNeeded( const QString &projectFullName, bool withAuth )
+{
+  if ( mPendingSyncChecks.contains( projectFullName ) )
+    return;
+
+  QNetworkReply *reply = getProjectInfo( projectFullName, withAuth );
+  if ( !reply )
+    return;
+
+  mPendingSyncChecks.insert( projectFullName );
```
Withalion: Is there any reason why we need the set? If the request hangs we won't be able to fire a new one for example and the request should be quite light and fast

core/merginapi.cpp
```diff
@@ -4597,3 +4597,42 @@ bool MerginApi::serverVersionIsAtLeast( const int requiredMajor, const int requi
...
+  }
+  r->deleteLater();
+}
+
+void MerginApi::isProjectSyncNeeded( const QString &projectFullName, bool withAuth )
+{
+  if ( mPendingSyncChecks.contains( projectFullName ) )
+    return;
+
+  QNetworkReply *reply = getProjectInfo( projectFullName, withAuth );
```
```suggestion
  const QNetworkReply *reply = getProjectInfo( projectFullName, withAuth );
```

core/merginapi.cpp
```diff
@@ -4597,3 +4597,42 @@ bool MerginApi::serverVersionIsAtLeast( const int requiredMajor, const int requi
...
+    {
+      LocalProject projectInfo = mLocalProjects.projectFromMerginName( projectFullName );
+      if ( projectInfo.isValid() && projectInfo.localVersion != -1 && projectInfo.localVersion < serverProject.version )
+      {
+        emit projectSyncRequired( projectFullName );
+      }
+    }
+  }
+  r->deleteLater();
+}
```
Withalion, line 4628:
```suggestion
void MerginApi::isProjectSyncNeededFinished()
{
  QNetworkReply *r = qobject_cast<QNetworkReply *>( sender() );
  Q_ASSERT( r );

  const QString projectFullName = r->request().attribute( static_cast<QNetworkRequest::Attribute>( AttrProjectFullName ) ).toString();
  mPendingSyncChecks.remove( projectFullName );

  if ( r->error() == QNetworkReply::NoError )
  {
    const QByteArray data = r->readAll();
    const MerginProjectMetadata serverProject = MerginProjectMetadata::fromJson( data );

    // skip if a sync is already in progress for this project
    if ( !mTransactionalStatus.contains( projectFullName ) )
    {
      const LocalProject projectInfo = mLocalProjects.projectFromMerginName( projectFullName );
      if ( projectInfo.isValid() && projectInfo.localVersion != -1 && projectInfo.localVersion < serverProject.version )
      {
        emit projectSyncRequired( projectFullName );
      }
    }
  }
  r->deleteLater();
}
```

Withalion (commented):
It will need some more work as mentioned in `main.qml`. Also I don't really like the visuals of the notification. Why is the text a link? I propose to use the same design as we want to do in filtering notifications ([Figma](https://www.figma.com/design/d78a4ehBBgK6MyxCTuCAqg/Mobile-App?node-id=6022-12729&p=f&t=jcsrH1zajRM5Pa7J-0)). It would be great to replace the close icon with "sync" button, let's discuss it.

app/activeproject.h
Withalion, line 1: Instead of passing `MerginApi` emit signal after project is reloaded and connect it to `MerginApi::isProjectSyncNeeded()`

core/merginapi.h
```diff
@@ -979,6 +984,7 @@ class MerginApi: public QObject
     MerginServerType::ServerType mServerType = MerginServerType::ServerType::OLD;
     QString mServerDiagnosticLogsUrl = MerginApi::sDefaultReportLogUrl;
 
+
```
Withalion:
```suggestion
```

Withalion (commented):
Besides the comments below, the project list shows information that user has "pending changes to synchronise" even though there are no changes done to the project locally, if there is newer version of project available on server. I don't think we want to show that info if there are no local changes.

PR #4391, opened 2026-03-16
feat(form): show value-map options as chip buttons when ≤ 4 choices

saberraz:
Value-map fields with 4 or fewer options now render all choices as inline pill/chip buttons directly on the form, eliminating the need to open a drawer for short lists (e.g. a 4-value condition rating).

Fields with 5 or more options continue to use the existing dropdown drawer (MMFormComboboxBaseEditor + MMListMultiselectDrawer) unchanged, so there is no regression for longer lists.

Implementation
• New MMFormChipEditor.qml (extends MMBaseInput):

Displays options as a horizontally-wrapping Flow of pill Rectangles.
Selected chip: grassColor (#73D19C) background.
Unselected chip: white background with forestColor border.
Read-only state: MouseArea disabled, chips still visible.
Emits editorValueChanged / rememberValueBoxClicked for MMFormPage compatibility.
• Refactored MMFormValueMapEditor.qml (now an Item wrapper):

Parses _fieldConfig into a shared ListModel in Component.onCompleted.
After parsing, sets _modelReady = true which triggers a Loader to pick the appropriate sub-editor (chip vs combobox) in the same synchronous call — no visible flicker.
Forwards editorValueChanged and rememberValueBoxClicked from the active sub-editor to MMFormPage via Connections.
• CMakeLists.txt: registers the new MMFormChipEditor.qml QML source.

Before (check the road condition):
screenshot_before_dropdown

After:
screenshot_after_chips

### DIFF
--- a/app/qml/CMakeLists.txt
+++ b/app/qml/CMakeLists.txt
@@ -101,6 +101,7 @@ set(MM_QML
     form/components/photo/MMPhotoPreview.qml
     form/components/MMFormActionBar.qml
     form/editors/MMFormCalendarEditor.qml
+    form/editors/MMFormChipEditor.qml
     form/editors/MMFormComboboxBaseEditor.qml
     form/editors/MMFormGalleryEditor.qml
     form/editors/MMFormNotAvailable.qml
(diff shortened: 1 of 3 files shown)


tomasMizera:
Let's close this as a prototype, there are too many things to change for it to be production-ready. We will keep it as a reference, thanks! :)

PR #4403, opened 2026-03-19
Modified the source CRS to be 2D instead of 3D

gabriel-bolbotina:
Ready for review.
This modification should fix the horizontal offset issue when datum transformation is used.

### DIFF
--- a/app/inpututils.cpp
+++ b/app/inpututils.cpp
@@ -900,14 +900,25 @@ QgsPoint InputUtils::transformPoint( const QgsCoordinateReferenceSystem &srcCrs,
     {
       if ( !ct.isShortCircuited() )
       {
-        const QgsVector3D transformed = ct.transform( QgsVector3D( srcPoint.x(), srcPoint.y(), srcPoint.z() ) );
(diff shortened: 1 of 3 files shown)


app/inpututils.cpp
Withalion, line 1: I still believe we don't have to do this at all. Depending on the source and target crs proj will decide if the `z` coordinate will be used in the transform

Withalion (commented):
Looks like the issue mentioned in the ticket is not really an issue, as the user was using custom transformations and they were not getting applied as we changed the CRS being used by default. However, after looking into the changed code in #4234 looks like there are multiple places where we should use different CRS. 
- [InputUtils::distanceBetweenGpsAndFeature](https://github.com/MerginMaps/mobile/blob/3181c8f89e34148b62e3e90afda5cc921d527e72/app/inpututils.cpp#L1719)
- [InputUtils::angleBetweenGpsAndFeature](https://github.com/MerginMaps/mobile/blob/3181c8f89e34148b62e3e90afda5cc921d527e72/app/inpututils.cpp#L1767)

Withalion:
Let's park it for now and first finish the discussion about implementation

PR #4425, opened 2026-03-26
Updating plugin keys for vertical transformation

### LINKED ISSUE #4485: Adjust vertical CRS transformation code according to recent QGIS-plugin changes
QGIS-Plugin code: https://github.com/MerginMaps/qgis-plugin/pull/884

Key changed from `SkipElevationTransformation` to `ElevationTransformationEnabled` (mind the boolean flip).
Let's update the codebase to use the new naming.

Prototype branch: `fix/vertical-crs-keys`

gabriel-bolbotina:
skipElevationTransformation key is now deprecated in plugin, so the code was modified to support elevationTransformationEnabled. 

If skipElevationTransformation was set to false, it means there was vertical transformation enabled, thus elevationTransformationEnabled will then be true and vice versa.

### DIFF
--- a/app/main.cpp
+++ b/app/main.cpp
@@ -572,22 +572,9 @@ int main( int argc, char *argv[] )
     pk->setVerticalCrs( crs );
 
     // read and set new elevation transformation behavior
-    // The project key was previously named "SkipElevationTransformation" (skip=true meant no transformation).
-    // It was replaced by "ElevationTransformationEnabled" with inverted logic (enabled=true means transformation is done).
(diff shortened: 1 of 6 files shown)


Withalion (commented):
Just a question, why do we want to support both keys? Plugin never got released with the support for this functionality and we just enabled it in few project for _very early_ adopters.

Besides that it looks alright.

Withalion (approved):

Withalion (approved):

tomasMizera (approved):

PR #4442, opened 2026-04-08
Added support for custom values for filtering checkboxes

### LINKED ISSUE #4441: Support custom values for filter segment controller
Modify the UI and the filter controller to support custom values for boolean segment controller. 
These custom values can be integers ( 0 and 1), strings (customised message) or simply boolean values (True and False)

gabriel-bolbotina:
Implemented support in filter controller and in MM filter bool input for custom values for checkboxes.
Right now in QGIS a checkbox can be:
- boolean, default setting which has True and False
- string, for text values and customisable ones
- integers, using 0 or 1.


tomasMizera (approved):
Image

PR #4445, opened 2026-04-09
Added initial value map model

gabriel-bolbotina:
Did some wiring for the filter controller, still needs to be fully integrated with the QML component.


app/filtercontroller.h
Withalion, line 1: @tomasMizera can you remove just this change and I'll approve

Withalion (approved):

PR #4455, opened 2026-04-16
Integrated custom check box in filtering controller

gabriel-bolbotina:
Added function to get the checkbox configuration for custom values
Modified MM filters panel to support custom label and values for segment control.

UI:

https://github.com/user-attachments/assets/66e8ae2b-d630-4531-a1de-30b849654ac9

### DIFF
--- a/app/filtercontroller.cpp
+++ b/app/filtercontroller.cpp
@@ -376,6 +376,64 @@ bool FilterController::isDateFilterDateTime( const QString &filterId )
   return false;
 }
 
+QVariantMap FilterController::getCheckboxConfiguration( const QString &filterId )
+{
+  if ( filterId.isEmpty() ) return {};
+
+  FieldFilter fieldFilter;
+  for ( const FieldFilter &filter : std::as_const( mFieldFilters ) )
+  {
+    if ( filterId == filter.filterId )
+    {
+      fieldFilter = filter;
+      break;
+    }
+  }
+
+  if ( !fieldFilter.hasFilterInfo() ) return {};
+
+  if ( fieldFilter.filterType != FieldFilter::CheckboxFilter ) return {};
+
+  const QgsVectorLayer *layer = qobject_cast<QgsVectorLayer *>( QgsProject::instance()->mapLayers().value( fieldFilter.layerId ) );
+  if ( !layer ) return {};
+
+  const int fieldIndex = layer->fields().lookupField( fieldFilter.fieldName );
+  if ( fieldIndex < 0 ) return {};
+
+  const QgsEditorWidgetSetup widgetSetup = layer->editorWidgetSetup( fieldIndex );
+  if ( QString::compare( widgetSetup.type(), QStringLiteral( "CheckBox" ), Qt::CaseInsensitive ) != 0 ) return {};
+
+  const QVariantMap config = widgetSetup.config();
+  const QString checkedState = config.value( QStringLiteral( "CheckedState" ) ).toString();
+  const QString uncheckedState = config.value( QStringLiteral( "UncheckedState" ) ).toString();
+
+  if ( checkedState.isEmpty() && uncheckedState.isEmpty() ) return {};
+
+  const QMetaType::Type fieldType = static_cast<QMetaType::Type>( layer->fields().field( fieldIndex ).type() );
+  const bool isIntField = ( fieldType == QMetaType::Int || fieldType == QMetaType::UInt ||
+                            fieldType == QMetaType::LongLong || fieldType == QMetaType::ULongLong );
+
+  QVariantMap result;
+
+  if ( !checkedState.isEmpty() )
+  {
+    result[QStringLiteral( "customLabelForTrue" )] = checkedState;
+    bool ok = false;
+    const int intVal = checkedState.toInt( &ok );
+    result[QStringLiteral( "customValueForTrue" )] = ( isIntField && ok ) ? QVariant( intVal ) : QVariant( checkedState );
+  }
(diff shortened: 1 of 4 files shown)


app/qml/filters/components/MMFilterBoolInput.qml
```diff
@@ -87,7 +87,7 @@ Column {
   QtObject {
     id: internal
 
-    property var representationForTrue: root.customValueForTrue ? root.customValueForTrue : true
-    property var representationForFalse: root.customValueForFalse ? root.customValueForFalse : false
+    property var representationForTrue: root.customValueForTrue != null ? root.customValueForTrue : true
+    property var representationForFalse: root.customValueForFalse != null ? root.customValueForFalse : false
```
Withalion: this shouldn't be necessary or was there some specific case where it didn't work?

Have a look here: [Falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) and [Truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy) values in JS

app/qml/filters/MMFiltersPanel.qml
```diff
@@ -141,6 +141,11 @@ MMComponents.MMDrawer {
               }
               else if ( filterType === FieldFilter.CheckboxFilter )
               {
+                const checkboxConfig = __activeProject.filterController.getCheckboxConfiguration( modelData.filterId )
+                if ( checkboxConfig["customLabelForTrue"] !== undefined ) props["customLabelForTrue"] = checkboxConfig["customLabelForTrue"]
+                if ( checkboxConfig["customLabelForFalse"] !== undefined ) props["customLabelForFalse"] = checkboxConfig["customLabelForFalse"]
+                if ( checkboxConfig["customValueForTrue"] !== undefined ) props["customValueForTrue"] = checkboxConfig["customValueForTrue"]
+                if ( checkboxConfig["customValueForFalse"] !== undefined ) props["customValueForFalse"] = checkboxConfig["customValueForFalse"]
```
Withalion:
```suggestion
                if ( checkboxConfig["customLabelForTrue"] ) props["customLabelForTrue"] = checkboxConfig["customLabelForTrue"]
                if ( checkboxConfig["customLabelForFalse"] ) props["customLabelForFalse"] = checkboxConfig["customLabelForFalse"]
                if ( checkboxConfig["customValueForTrue"] ) props["customValueForTrue"] = checkboxConfig["customValueForTrue"]
                if ( checkboxConfig["customValueForFalse"] ) props["customValueForFalse"] = checkboxConfig["customValueForFalse"]
```

app/filtercontroller.cpp
```diff
@@ -376,6 +376,64 @@ bool FilterController::isDateFilterDateTime( const QString &filterId )
...
+      fieldFilter = filter;
+      break;
+    }
+  }
+
+  if ( !fieldFilter.hasFilterInfo() ) return {};
+
+  if ( fieldFilter.filterType != FieldFilter::CheckboxFilter ) return {};
+
+  const QgsVectorLayer *layer = qobject_cast<QgsVectorLayer *>( QgsProject::instance()->mapLayers().value( fieldFilter.layerId ) );
```
Withalion:
```suggestion
  const QgsVectorLayer *layer = qobject_cast<QgsVectorLayer *>( QgsProject::instance()->mapLayer( fieldFilter.layerId ) );
```

app/qml/filters/MMFiltersPanel.qml
Withalion: Also we don't have to wrap it in if's here as we also check the values inside `MMFilterBoolInput`

app/filtercontroller.cpp
```diff
@@ -376,6 +376,64 @@ bool FilterController::isDateFilterDateTime( const QString &filterId )
...
+  const QVariantMap config = widgetSetup.config();
+  const QString checkedState = config.value( QStringLiteral( "CheckedState" ) ).toString();
+  const QString uncheckedState = config.value( QStringLiteral( "UncheckedState" ) ).toString();
+
+  if ( checkedState.isEmpty() && uncheckedState.isEmpty() ) return {};
+
+  const QMetaType::Type fieldType = static_cast<QMetaType::Type>( layer->fields().field( fieldIndex ).type() );
+  const bool isIntField = ( fieldType == QMetaType::Int || fieldType == QMetaType::UInt ||
+                            fieldType == QMetaType::LongLong || fieldType == QMetaType::ULongLong );
+
```
Withalion, line 415: this feels unnecessary can't we just put the value inside QVariant without recasting it in between?

gabriel-bolbotina: To support also 0 and 1 as integer custom values, we need this. However, if we plan support only string custom value checkbox, it can be simplified.

app/qml/filters/components/MMFilterBoolInput.qml
gabriel-bolbotina: The only use case is if we want to keep integer custom values or not. If we have 0 and 1 as custom options

app/filtercontroller.cpp
Withalion: Let's keep the support for numbers too

app/qml/filters/components/MMFilterBoolInput.qml
Withalion: okay let's keep it then how it was

Withalion (approved):
Just reverse the `representationForXXX` conditions to how they were before and you can merge

PR #4456, opened 2026-04-16
Removed the hold and tap behavior of the filtering map button

gabriel-bolbotina:
Renamed the filters enabled member variable to filters activated.
Changed the behaviour of the filter floating button as discussed in [this issue ](https://github.com/MerginMaps/product/issues/106)
UI:

https://github.com/user-attachments/assets/bbc43c51-6d34-43a1-86cf-57daa89f84d4

### DIFF
--- a/app/filtercontroller.cpp
+++ b/app/filtercontroller.cpp
@@ -67,28 +67,19 @@ void FilterController::loadFilterConfig( const QgsProject *project )
 {
   mFieldFilters.clear();
 
-  if ( mFilteringActivated )
-  {
-    mFilteringActivated = false;
-    emit hasFiltersActivatedChanged();
-  }
-
-  if ( mFilteringAvailable )
-  {
-    mFilteringAvailable = false;
-    emit hasFiltersAvailableChanged();
-  }
+  setFiltersActivated( false );
 
   bool valueRead = false;
   const bool filteringAvailable = project->readBoolEntry( QStringLiteral( "Mergin" ), QStringLiteral( "Filtering/Enabled" ), false, &valueRead );
+
+  mFilteringAvailable = filteringAvailable;
+  emit hasFiltersAvailableChanged();
 
   //return early if filtering is not setup
   if ( !valueRead )
   {
     return;
   }
-  mFilteringAvailable = filteringAvailable;
-  emit hasFiltersAvailableChanged();
 
   const QString filtersDef = project->readEntry( QStringLiteral( "Mergin" ), QStringLiteral( "Filtering/Filters" ) );
   QJsonParseError jsonError;
@@ -353,9 +344,9 @@ void FilterController::processFilters( const QVariantMap &newFilters )
   applyFiltersToAllLayers();
 
   bool anyActivated = false;
-  for ( const FieldFilter &filter : std::as_const( mFieldFilters ) )
+  for ( const QVariant &value : std::as_const( newFilters ) )
   {
-    if ( filter.value.isValid() && !filter.value.isNull() )
+    if ( value.isValid() && !value.isNull() )
     {
       anyActivated = true;
       break;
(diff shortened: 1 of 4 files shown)


app/qml/main.qml
```diff
@@ -361,7 +361,11 @@ ApplicationWindow {
       MMToolbarButton {
         text: qsTr("Filters")
         iconSource: __style.filterIcon
+        visible: __activeProject.filterController?.filteringAvailable ?? false
         onClicked: {
+          if ( __activeProject.filterController ) {
+            __activeProject.filterController.filteringEnabled = true
+          }
```
```suggestion
            __activeProject.filterController?.filteringEnabled = true
```

Withalion (commented):
+ other things dicussed

app/filtercontroller.cpp
```diff
@@ -350,6 +351,17 @@ void FilterController::processFilters( const QVariantMap &newFilters )
...
+  bool anyActivated = false;
+  for ( const FieldFilter &filter : std::as_const( mFieldFilters ) )
+  {
+    if ( filter.value.isValid() && !filter.value.isNull() )
+    {
+      anyActivated = true;
+      break;
+    }
+  }
+  setFiltersActivated( anyActivated );
```
Withalion, line 364: getting non empty `newFilters` should mean that some filters are active, no need to go through all of them

app/filtercontroller.cpp
```diff
@@ -66,7 +66,18 @@ void FilterController::clearAllFilters()
...
 {
   mFieldFilters.clear();
-  setFiltersEnabled( false );
+
+  if ( mFilteringActivated )
+  {
+    mFilteringActivated = false;
+    emit hasFiltersActivatedChanged();
+  }
+
```
Withalion, line 75:
```suggestion
  setFiltersActivated( false );

```

app/filtercontroller.cpp
```diff
@@ -66,7 +66,18 @@ void FilterController::clearAllFilters()
...
+  if ( mFilteringAvailable )
+  {
+    mFilteringAvailable = false;
+    emit hasFiltersAvailableChanged();
+  }
```
Withalion, line 80: I think we can remove this and move the lines 85-89 below 90-91 to get the same effect

app/filtercontroller.cpp
```diff
@@ -66,18 +66,20 @@ void FilterController::clearAllFilters()
...
+  setFiltersActivated( false );
 
   bool valueRead = false;
   const bool filteringAvailable = project->readBoolEntry( QStringLiteral( "Mergin" ), QStringLiteral( "Filtering/Enabled" ), false, &valueRead );
 
+  mFilteringAvailable = filteringAvailable;
+  emit hasFiltersAvailableChanged();
+
   //return early if filtering is not setup
   if ( !valueRead )
```
Withalion: My mistake but this should probably be:
```suggestion
  if ( !filteringAvailable )
```
as we can successfully read false value too and it doesn't really make sense to go through the rest

app/filtercontroller.cpp
```diff
@@ -350,6 +342,17 @@ void FilterController::processFilters( const QVariantMap &newFilters )
...
+  bool anyActivated = false;
+  for ( const QVariant &value : std::as_const( newFilters ) )
+  {
+    if ( value.isValid() && !value.isNull() )
+    {
+      anyActivated = true;
+      break;
+    }
+  }
+  setFiltersActivated( anyActivated );
```
Withalion, line 355: I meant to do this:

```suggestion
  setFiltersActivated( !newFilters.isEmpty );
```
The inputs should be responsible to give you valid value back and they should not return null values

gabriel-bolbotina: They don't return null values but undefined once we clear that field. 
In this case, the newFilters will not be empty and still have the key but with undefined which will trigger the floating button to appear. Unless we reset the filters via the reset button, the floating button will always appear with this approach. I would keep it as it is.

Withalion: okay sounds good

Withalion (approved):

PR #4457, opened 2026-04-17
Fixed silent crash and filter input behaviour for numbers

gabriel-bolbotina:
- Fixed silent crashes that occurred when a filter was added to the project without a layer or field in plugin, that would lead to a silent crash in the app. Now we skip these filters and I added logging messages for them. Also, I added some guards for other potential silent crashes.
- Fixed checked status for filter inputs which was broken when inputting text, numbers or dates.
- Added support for negative numbers and 0 as inputs for number range filters
- fixed the false checked status for dropdown filter types

### DIFF
--- a/app/filtercontroller.cpp
+++ b/app/filtercontroller.cpp
@@ -131,12 +131,37 @@ void FilterController::loadFilterConfig( const QgsProject *project )
       newFieldFilter.sqlExpression = filterObject.value( QStringLiteral( "sql_expression" ) ).toString();
       newFieldFilter.layerId = filterObject.value( QStringLiteral( "layer_id" ) ).toString();
 
+      // check for missing filter fields
+      QStringList missingFilterFields;
+      if ( newFieldFilter.layerId.isEmpty() ) missingFilterFields << QStringLiteral( "'layer_id'" );
+      if ( newFieldFilter.fieldName.isEmpty() ) missingFilterFields << QStringLiteral( "'field_name'" );
+      if ( newFieldFilter.sqlExpression.isEmpty() ) missingFilterFields << QStringLiteral( "'sql_expression'" );
+
+      if ( !missingFilterFields.isEmpty() )
+      {
+        CoreUtils::log( QStringLiteral( "Feature Filtering" ),
+                        QStringLiteral( "Filter '%1' is missing required filter field(s): %2. Skipping." )
+                        .arg( newFieldFilter.filterName, missingFilterFields.join( QStringLiteral( ", " ) ) ) );
+        continue;
+      }
+
+      // check if target layer exists
       const QgsVectorLayer *filterLayer = qobject_cast<QgsVectorLayer *>( project->mapLayer( newFieldFilter.layerId ) );
       if ( !filterLayer )
       {
         CoreUtils::log( QStringLiteral( "Feature Filtering" ),
-                        QStringLiteral( "Filter '%1' is not properly configured in the project. Skipping filter." )
-                        .arg( newFieldFilter.filterName ) );
+                        QStringLiteral( "Filter '%1' has no layer with ID '%2' found in project. Skipping." )
+                        .arg( newFieldFilter.filterName, newFieldFilter.layerId ) );
+        continue;
+      }
+
+      // check if target field of target layer exists
+      if ( filterLayer->fields().lookupField( newFieldFilter.fieldName ) < 0 )
+      {
+        CoreUtils::log( QStringLiteral( "Feature Filtering" ),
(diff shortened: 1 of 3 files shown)


app/filtercontroller.cpp
```diff
@@ -32,9 +32,15 @@ void FilterController::clearLayerFilters( const QString &layerId )
 {
   QgsMapLayer *layer = QgsProject::instance()->mapLayers().value( layerId );
   QgsVectorLayer *vectorLayer = qobject_cast<QgsVectorLayer *>( layer );
+  if ( !vectorLayer )
+  {
+    CoreUtils::log( QStringLiteral( "Feature Filtering" ),
+                    QStringLiteral( "Layer '%1' is not available. Cannot clear its filters." ).arg( layerId ) );
+    return;
+  }
```
Withalion, line 40: this doesn't seem necessary as this function gets the layerID passed from the selected layer by user and not from filter

app/filtercontroller.cpp
```diff
@@ -131,6 +137,15 @@ void FilterController::loadFilterConfig( const QgsProject *project )
...
       newFieldFilter.layerId = filterObject.value( QStringLiteral( "layer_id" ) ).toString();
 
+      const QgsVectorLayer *filterLayer = qobject_cast<QgsVectorLayer *>( QgsProject::instance()->mapLayer( newFieldFilter.layerId ) );
+      if ( !filterLayer )
+      {
+        CoreUtils::log( QStringLiteral( "Feature Filtering" ),
+                        QStringLiteral( "Layer '%1' for filter '%2' is not available in the project. Skipping filter." )
+                        .arg( newFieldFilter.layerId, newFieldFilter.filterName ) );
+        continue;
+      }
```
Withalion, line 147: I wouldn't do the check here, but when creating the filter expression

Withalion: Also, `QgsProject` is passed to this function

app/filtercontroller.cpp
```diff
@@ -359,6 +381,12 @@ bool FilterController::hasActiveFilterOnLayer( const QString &layerId )
     return false;
 
   const QgsVectorLayer *layer = qobject_cast<QgsVectorLayer *>( project->mapLayers().value( layerId ) );
+  if ( !layer )
+  {
+    CoreUtils::log( QStringLiteral( "Feature Filtering" ),
+                    QStringLiteral( "Layer '%1' is not available. Cannot check active filter." ).arg( layerId ) );
+    return false;
+  }
```
Withalion, line 389: This function again gets passed layerID from layer which needs to exist or the user wouldn't be able to navigate to where this function gets triggered

Withalion (commented):
In general I would prefer triggering the logging when the filter tries to get applied and fails instead of when the filter is loaded from `QgsProject`

gabriel-bolbotina:
> In general I would prefer triggering the logging when the filter tries to get applied and fails instead of when the filter is loaded from `QgsProject`

I think we need to choose between the expected behaviour:
**current one: the filter that does not have a layer or field gets dropped entirely when loading config and it is logged there**
your suggestion: the filter does not get dropped per se, it still appears in the filters drawer, but when applying filters that one is not taken into consideration.

I think that they should not appear in the drawer at all, as they don't have a target layer nor field.

app/filtercontroller.cpp
```diff
@@ -131,6 +131,15 @@ void FilterController::loadFilterConfig( const QgsProject *project )
       newFieldFilter.sqlExpression = filterObject.value( QStringLiteral( "sql_expression" ) ).toString();
       newFieldFilter.layerId = filterObject.value( QStringLiteral( "layer_id" ) ).toString();
```
Withalion, line 132: Can you also add check and log if `fieldName`, `sqlExpression`, `layerId` is missing/empty. Log which field is missing and skip the creation of that filter.

app/filtercontroller.cpp
```diff
@@ -131,6 +131,15 @@ void FilterController::loadFilterConfig( const QgsProject *project )
...
 
+      const QgsVectorLayer *filterLayer = qobject_cast<QgsVectorLayer *>( project->mapLayer( newFieldFilter.layerId ) );
+      if ( !filterLayer )
+      {
+        CoreUtils::log( QStringLiteral( "Feature Filtering" ),
+                        QStringLiteral( "Filter '%1' is not properly configured in the project. Skipping filter." )
```
Withalion: Please give it more precise log message.

Also we should check if the field exists on that layer and if not log that too and skip the creation.

Withalion (commented):
Just last bits to improve resilience and debugging

Withalion (approved):

PR #4472, opened 2026-04-24
WIP fixed safe area rendering on iOS and cmake configuration for android

gabriel-bolbotina:
Did the following:
- Removed safe area platform-specific functions, relied on how Qt already calculates it
- Used
- Formatted code

UI:

https://github.com/user-attachments/assets/a29c2403-81e0-4c6b-a402-55e6af74c384

Remaining issues:

- some elements using mmHighlight are not rendering the highlight at all or delayed. For example, there is a delayed rendering for Map sketches which are not in real time, only after finishing up the sketch and when selecting point features on the map. Lines seem to get the highlight.

TODOs:
- [ ] Fix MMHighlight to render map sketches in real time and to highlight selected features on the map. currently it does not work
- [x] Patch the vcpkg android ports to properly configure and build the project
- [ ] check if we should refactor how safe area is handled in QML components, we can keep it as it is, or maybe there is a better way to do it


app/main.cpp
```diff
@@ -914,6 +874,22 @@ int main( int argc, char *argv[] )
   if ( QQuickWindow *quickWindow = qobject_cast<QQuickWindow *>( object ) )
   {
     quickWindow->setIcon( QIcon( logoUrl ) );
+
+#if defined( ANDROID ) || defined( Q_OS_IOS )
+    // Qt 6.9+ provides QWindow::safeAreaMargins() natively for both Android and iOS.
+
+    auto applyMargins = [style, quickWindow]()
+    {
+      const QMargins m = quickWindow->safeAreaMargins();
```
tomasMizera, line 883: Do we even need to read and store these if it is automatically handled by Qt?

tomasMizera (approved):
Looking good! :)

app/main.cpp
gabriel-bolbotina: There are QML components that still use safe area values from MM style for calculating padding or heights/widths.
It is feasible to take them out completely, but I think we should do that after we have all the builds set up, because it needs comprehensive testing on all platforms. In this case, I would keep it as it is right now

Withalion:
Let's merge and we can hammer out those missing things later

PR #4484, opened 2026-04-29
Fixes for the layer legend

uclaros:
Adjust the legend text size, color and margins.

Before             |  After
:-------------------------:|:-------------------------:
Screenshot_20260429_154531|Screenshot_20260429_154456


app/mmstyle.h
```diff
@@ -732,7 +732,7 @@ class MMStyle: public QObject
     double number400() const {return 400 * mDp;}
     double number720() const {return 720 * mDp;}
 
-    double scrollVelocityAndroid() const { return 10000; }
+    static double scrollVelocityAndroid() { return 10'000; }
```
Withalion, line 741: this is probably a typo

app/layer/layerdetaildata.cpp
```diff
@@ -100,6 +102,14 @@ void LayerDetailData::setLayerTreeNode( QgsLayerTreeNode *newLayerTreeNode )
   // setup render context for legend
   QgsLegendSettings legendSettings;
 
+  QgsTextFormat textFormat = legendSettings.rstyle( QgsLegendStyle::Style::SymbolLabel ).textFormat();
+  textFormat.setSize( 14 ); // 14 is the pixel size of MMStyle::p5, used in form widgets
+  textFormat.setSizeUnit( Qgis::RenderUnit::Pixels );
+  textFormat.setColor( MMStyle::nightColor() );
+  legendSettings.rstyle( QgsLegendStyle::Style::SymbolLabel ).setTextFormat( textFormat );
+  legendSettings.rstyle( QgsLegendStyle::Style::SymbolLabel ).setMargin( 4 ); // match the top and left total margin of the legend
+  legendSettings.rstyle( QgsLegendStyle::Style::Title ).setMargin( 0 ); // Our empty Title still adds some margins, so we set those to zero
```
Withalion: Looks like `QgsLegendStyle::Style` enum doesn't exist anymore in QGIS 3.44 -> [QGIS PR](https://github.com/qgis/QGIS/pull/60652)

app/layer/layerdetaillegendimageprovider.cpp
```diff
@@ -47,31 +47,36 @@ QImage LayerDetailLegendImageProvider::requestImage( const QString &, QSize *siz
...
+  // 60 is fallback size, not sure how it came to be
+  const int width = requestedSize.isEmpty() ? static_cast<int>( 60 * dpr ) : static_cast<int>( requestedSize.width() * dpr );
+  const int height = static_cast<int>( minimumSize.height() * dpmm );
 
-  QImage legend = QImage( desiredSize.width(), minimumSize.height() * dpm, QImage::Format_ARGB32_Premultiplied );
+  QImage legend = QImage( width, height, QImage::Format_ARGB32_Premultiplied );
 
-  QPainter painter( &legend );
-  painter.setRenderHint( QPainter::Antialiasing );
+  {
```
Withalion, line 61: out of curiosity what's the reason for having this part of the code wrapped in extra curly braces?

Withalion (commented):
Let's try another rebase :)

app/layer/layerdetaildata.cpp
uclaros: fixed in b9899c7

app/mmstyle.h
uclaros: Not a typo, single quotes can be used as visual separators between digits since c++14.
For some reason I can't instantly recognize how many zeros are in`10000`, but by using thousands' separators I can instantly say `10'000` is 10k.

Withalion: Ahh okay, I didn't know that neat

app/layer/layerdetaillegendimageprovider.cpp
uclaros: This is to limit the scope of the QPainter. By making it go out of scope, the painting is finalized and not affected by setting the `dpr`.

There's a relevant comment:

>   // Tag with DPR only after painting so QPainter does not auto-scale the
>  // coordinate system by dpr (which would otherwise double-apply the factor).
>  legend.setDevicePixelRatio( dpr );

Withalion (approved):

Withalion:
Testing passed by @IvaKuklica

PR #4488, opened 2026-05-05
Fix map canvas refresh when moving app window between screens

tomasMizera:


Withalion (approved):

Withalion:
**Note for testers:** Test with the upgraded Qt

Issue #4273, opened 2025-12-19
Aggregate: Untranslated or Incorrectly Translated Strings


RastoHu:
**Description**

This task aggregates findings from testing related to untranslated or incorrectly translated text in the Mergin Maps mobile app for specific languages. Testers should add comments with the language, location of the text, screenshots, and the expected translation.

RastoHu:
### Slovak

- [ ] Login screen ->
- Enter invalid credentials
- Leave the password empty

<img src="https://github.com/user-attachments/assets/71566fd7-4df9-4f67-8c62-3bb316831224" width="250"/>

- [ ] Calendar ->
- title "date"
- Months 

<img src="https://github.com/user-attachments/assets/9d2fb8cc-cca5-4d73-a544-232861478407" width="250"/>


- [x]  Menu for adding a field ->
✅ translated

<img width="200" height="395" alt="Image" src="https://github.com/user-attachments/assets/ba575210-5599-45ea-81cb-0146f1ce6222" />

- [x]  Clear button in the value relation/map list ->
✅ translated
<img src="https://github.com/user-attachments/assets/1ab0686e-ca3f-407b-83ce-524b738ecd9b" width="250"/>

- [ ] Active filter info banner:
_Aktívne **filter** sú aplikované_

IvaKuklica:
**GERMAN**

- [ ] 1. More options menu ---> Position_tracking should be translated: 
 
> [!TIP]
Standortverfolgung or Routenverfolgung (please, verify)

<img width="200" height="392" alt="Image" src="https://github.com/user-attachments/assets/aa31390a-88dd-4aea-842d-3f5848f92c25" />


- [ ] 2. Calendar---> date should be translated:
> [!TIP]
Datum (singular)
Daten (plural)


<img width="200" height="392" alt="Image" src="https://github.com/user-attachments/assets/103c7f97-4911-4bfb-9702-b8987fea44db" />

- [x] 3. Autosync - Description of autosync is not translated:
✅ translated

<img width="200" height="392" alt="Image" src="https://github.com/user-attachments/assets/dc35e6a7-77c1-43be-bdf2-1acce23acce8" />


- [x] 4. Menu for adding a field is not translated:
✅ translated

<img width="200" height="392" alt="Image" src="https://github.com/user-attachments/assets/e95d8f84-86e9-49cf-b2f2-de340c0cf257" />


- [ ] 5. Soft/hard QGIS constraints are not translated:

[...]

PR #4463, opened 2026-04-22
Qt 6.10.3, QGIS 3.44.10 upgrade

tomasMizera:
This PR combines all necessary upgrades to our dependencies. The major upgrades include 
- Qt 6.10.3
- QGIS 3.44.10
- iOS 26

The full checklist:
 - macOS
   - [x] build
   - [x] run
   - [x] working CI (position test failing)
 - iOS
   - [x] build
   - [x] run
   - [x] working CI (upload to testflight failing)
   - [x] problems detecting safe area on iOS
 - Linux
   - [x] build
   - [x] run
   - [x] working CI (position test failing)
 - Android
   - [x] build
   - [x] run
   - [x] working CI
 - Windows
   - [x] build
   - [x] run
   - [x] working CI

 - [x] double-check all the updated patches, especially positioning ones (some are not even used now)

PR #4489, opened 2026-05-05
Add missing translations for new project page elements

Withalion:
Fixes untranslated texts in field type dropdowns in new project creation page. Issue has been mentioned in #4273
. Besides that did some refactor in nearby code.

PS: Currently my VCPKG setup is broken, so I don't guarantee it's 100% working

### DIFF
--- a/app/fieldsmodel.cpp
+++ b/app/fieldsmodel.cpp
@@ -5,9 +5,9 @@ FieldsModel::FieldsModel( QObject *parent )
   : QAbstractListModel( parent )
 {
   // fill the model with initial data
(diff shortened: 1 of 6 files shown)


Withalion:
iOS build failing on expired provisioning profile, the new profile is already set in #4463. After rebasing on this PR it should build.

Withalion:
Note for @IvaKuklica: let's test it on regression test build as it needs translations from transifex as well

IvaKuklica:
The field type dropdowns are now correctly translated — the fix is working as expected. ✅

PR #4494, opened 2026-05-14
Fixed the UI bug not showing already selected entries for the multi select filter

gabriel-bolbotina:
Solved the UI issue described in [issue](https://github.com/MerginMaps/mobile/issues/4477) by converting the string values saved in currentValue to numeric ones which are used in Value Relation with AllowMulti option turned on. In this sense, the UI will show the previously selected entries from the Multi select filter.

The other bug mentioned regarding the filter itself will be solved in the QGIS plugin by modifying the template query for Value Relation with AllowMulti on to something like this:
`"sql_expression": "(',' || TRIM(\"NAME_OF_FIELD\", '{}') || ',' ) LIKE '%,' || @@value@@ || ',%'"` (detailed [here](https://github.com/MerginMaps/qgis-plugin/issues/914))

Once the template query is modified, the existing business logic in the code already handles this type of filtering with "OR" option.


app/qml/filters/components/MMFilterDropdownValueRelationInput.qml
```diff
@@ -109,8 +109,12 @@ Column {
...
       Component.onCompleted: {
         if ( root.currentValue ) {
-          // preselect choices if any are set
-          selected = root.currentValue
+          // preselect choices if any are set; convert numeric strings to numbers
+          // so they match the model's numeric keys (fixes AllowMulti filter returning 0 results)
+          selected = root.currentValue.map( v => {
+            const n = Number( v )
+            return isFinite( n ) && String( n ) === v ? n : v
+          } )
```
Withalion:
```suggestion
          selected = root.currentValue.map( v => isFinite( v ) ? Number( v ) : v )
```
Any reason why this is not enough?

app/qml/filters/components/MMFilterDropdownValueRelationInput.qml
```diff
@@ -109,8 +109,12 @@ Column {
 
       Component.onCompleted: {
         if ( root.currentValue ) {
-          // preselect choices if any are set
-          selected = root.currentValue
+          // preselect choices if any are set; convert numeric strings to numbers
+          // so they match the model's numeric keys (fixes AllowMulti filter returning 0 results)
```
Withalion:
```suggestion
          // preselect choices if any are set; convert numeric strings to numbers
          // so they match the model's numeric keys
```

PR #4495, opened 2026-05-14
Fix single/multi select DateTime filter

### LINKED ISSUE #4492: Filters - Single Select and Multi Select do not work correctly with DateTime fields
**Description**
Single select and Multi select filters do not work on fields with DateTime data type. It is possible to set these filters in QGIS and they are displayed correctly in the mobile app (offering used values), the result of filtering is always 0 features.

Note that both Single select and Multi select filters work well with Date data type.

**Environment**
Production

v2026.2.0

**Steps to reproduce**
1. Set up a Single select or Multi select Filter in QGIS for a DateTime field.
2. In the mobile app, try filtering features by selecting values from these filters.

**Actual results**
There are 0 features that match the filter - all features from the layer disappears from the map.

**Expected results**
Features that match the criteria should be displayed on the map.

**Screenshots**

https://github.com/user-attachments/assets/c8051590-a86b-4638-add7-baa6972297d5

gabriel-bolbotina:
GeoPackage can store DateTime as UTC with or without milliseconds.
The original code assumed that the DateTime entry for Single/Multi select contained milliseconds, which is not always the case if a point is added in QGIS compared to one saved with the mobile app.

Fix: check if the value contains milliseconds and change the format when creating the filter


app/filter/filtercontroller.cpp
```diff
@@ -262,7 +262,14 @@ QString FilterController::buildFieldExpression( const FieldFilter &filter ) cons
 
         if ( value.typeId() == QMetaType::QDateTime )
           if ( isDateFilterDateTime( filter.filterId ) )
-            expressionCopy.replace( QStringLiteral( "@@value@@" ), QgsExpression::quotedString( value.toDateTime().toUTC().toString( TIMESTAMP_FORMAT ) ) );
+          {
+            // check if the DateTime contains milliseconds
+            const QDateTime utcDateTime = value.toDateTime().toUTC();
+            const QString dateTimeFormat = utcDateTime.time().msec() != 0
+                                           ? TIMESTAMP_FORMAT
+                                           : QStringLiteral( "yyyy-MM-ddTHH:mm:ssZ" );
```
Withalion: define the formats as `TIMESTAMP_WITH_MS_FORMAT` and `TIMESTAMP_NO_MS_FORMAT` for example

Withalion:
I have a bad feeling though that, if the user will use `now()` expression as default value, which saves miliseconds too and manages to get 0 miliseconds. The filter won't be able to filter that value, so we should either try both values or get the string value before it's transformed to `QDateTime`/ JS `Date`.

Withalion (approved):

tomasMizera (commented):
Can we add unit tests here?

app/filter/filtercontroller.cpp
```diff
@@ -250,6 +250,7 @@ QString FilterController::buildFieldExpression( const FieldFilter &filter ) cons
         QStringList expressions;
         QString expressionTemplate( expressionCopy );
         expressionTemplate.replace( QStringLiteral( "@@value@@" ), QStringLiteral( "NULL" ) );
+        expressionTemplate.replace( QStringLiteral( "= NULL" ), QStringLiteral( "IS NULL" ) );
```
Withalion: if there was no change in plugin then the sql query should be `column IS value`. Did something change in the meantime?

app/filter/filtercontroller.cpp
```diff
@@ -377,6 +378,7 @@ QString FilterController::buildFieldExpression( const FieldFilter &filter ) cons
         if ( QgsVariantUtils::isNull( value ) )
         {
           expressionTemplate.replace( QStringLiteral( "@@value@@" ), QStringLiteral( "NULL" ) );
+          expressionTemplate.replace( QStringLiteral( "= NULL" ), QStringLiteral( "IS NULL" ) );
```
Withalion: Same as above

app/test/testfiltercontroller.h
```diff
@@ -0,0 +1,53 @@
...
+    void testSingleSelect_dateTime_null();   // null -> NULL OR ''
+    void testSingleSelect_date();
+
+    // Multi select
+    void testMultiSelect_dateTime_nonZeroMs();
+    void testMultiSelect_dateTime_zeroMs();  // edge case: 0 ms
+    void testMultiSelect_dateTime_mixed();   // mix of 0 ms and non-zero ms values
+    void testMultiSelect_dateTime_null();    // null -> NULL OR ''
+    void testMultiSelect_dateTime_empty();   // empty list -> no subset string
+    void testMultiSelect_date();
```
Withalion: use `camelCase` or `PascalCase`, which is used in the rest of repo

app/test/testutils.h
```diff
@@ -65,6 +67,22 @@ namespace TestUtils
...
+
+  //! Appends a single feature to layer via the data provider; value is stored in the named field. Returns false if addFeatures() fails.
+  bool addFeatureToLayer( QgsVectorLayer *layer, const QString &fieldName, const QVariant &value );
+
+  //! Writes a single-filter config into the project and loads it; returns the assigned filterId
+  QString setupControllerWithFilter( FilterController *controller,
+                                     FieldFilter::FilterType filterType,
+                                     const QString &layerId,
+                                     const QString &fieldName,
+                                     const QString &sqlExpression );
```
Withalion, line 98: Instead of doing this in the code create a QGIS project, which will have filters, layers, features already setup. That should also mitigate the filter definition duplication in `testfiltercontroller.cpp`. Place the project in `test/test_data/` and load it before running the tests

Withalion (commented):
The tests look okay in general, let's just improve the approach.

app/filter/filtercontroller.cpp
gabriel-bolbotina: Reverted it

Withalion (approved):
Let's not drag it anymore, the test can be improved later with more robust testing of filtering.

PR #4496, opened 2026-05-15
Allow adding a new part to multipart geometries

### LINKED ISSUE #2244: Add part of multiline and polygon feature
Currently there is no way how to add parts for multiline and polygon feature. 
I believe, this is something we should allow.

uclaros:
New feature for adding parts to multipart features.

Works with MultiPoint, MultiLineString and MultiPolygon layers.

Peek 2026-05-15 17-37

The button is not available for singlepart layers.


app/maptools/recordingmaptool.cpp
```diff
@@ -1150,6 +1150,36 @@ void RecordingMapTool::cancelGrab()
   setActiveVertex( Vertex() );
 }
 
+void RecordingMapTool::startDigitizingNewPart()
+{
+  // if maptool is in GRAB and VIEW state, no part should be added
+  if ( mState == RecordingMapTool::View || mState == RecordingMapTool::Grab )
```
tomasMizera: @uclaros why not? :)

tomasMizera: I'm thinking `View` is a great state to add a new part

app/qml/map/MMMapController.qml
```diff
@@ -814,17 +814,22 @@ Item {
...
+
             if (internal.redrawGeometryButtonVisible) {
-              if (!internal.splitGeometryButtonVisible && !internal.streamingModeButtonVisible)
+              if (!internal.splitGeometryButtonVisible && !internal.streamingModeButtonVisible && !internal.addPartButtonVisible)
                 return "redraw"
             }
 
             if (internal.streamingModeButtonVisible) {
-              if (!internal.redrawGeometryButtonVisible && !internal.splitGeometryButtonVisible)
+              if (!internal.redrawGeometryButtonVisible && !internal.splitGeometryButtonVisible && !internal.addPartButtonVisible)
```
tomasMizera, line 832: At some point we need to refactor this into a model :)

tomasMizera (approved):
Looks good!

app/qml/map/MMMapController.qml
```diff
@@ -1323,6 +1347,24 @@ Item {
     }
   }
 
+  function toggleAddPart() {
+    addPart( internal.featurePairToEdit )
+  }
+
```
Withalion, line 1353: why do we have this wrapper method?

app/qml/map/MMMapController.qml
```diff
@@ -1323,6 +1347,24 @@ Item {
...
+  function addPart( featurepair) {
+    __activeProject.setActiveLayer( featurepair.layer )
+    root.centerToPair( featurepair )
+    root.showInfoTextMessage( qsTr( "Add new part to the geometry" ) )
+
+    internal.featurePairToEdit = featurepair
```
Withalion, line 1358: why do we reassign the same feature pair?

Withalion (commented):
Please rebase the PR on current master so the builds pass and run astyle to fix the formatting issue

app/qml/map/MMMapController.qml
uclaros: Following the pattern set by `toggleRedraw()` and `toggleSplitting()`. Looks like all three could go.

app/qml/map/MMMapController.qml
uclaros: Again, following `toggleSplitting()` / `split()` logic.

app/maptools/recordingmaptool.cpp
uclaros: The state is set to `Record` in qml before invoking this method. This is kind of a safeguard that the method doesn't get called properly.

Should I set the state in cpp and just call `startDigitizingNewPart()` from qml?

app/qml/map/MMMapController.qml
```diff
@@ -1323,6 +1347,24 @@ Item {
     }
   }
 
+  function toggleAddPart() {
+    addPart( internal.featurePairToEdit )
+  }
+
+  function addPart( featurepair) {
+    __activeProject.setActiveLayer( featurepair.layer )
+    root.centerToPair( featurepair )
```
uclaros:
```suggestion
```
I think this is a bad idea and would like to remove it (also from `split` and `redraw` logic).
User is centered to the geometry when identifying it, then if he has panned the map he is re-centered a second time when tapping `Edit geometry`.
I think it is _very_ probable that the user has already panned the map to where he wants to add the new part / split / redraw and we're just forcing him back to the original geometry location!

IvaKuklica:
@uclaros 
**Comment for Add Part - New Feature for Multipart Geometry** **- Remove button incorrectly displayed as active when entering Add Part mode**

**Actual result:**

1. When the user enters **Add Part mode** for all multipart geometry types feature via Edit → Edit Geometry → More → Add Part, **the Remove button in the bottom toolbar is displayed as active even though no points have been added yet to the new part.** 

**Expected result:**

1. The Remove button should be inactive/disabled at this stage as there are no points to remove.
2. The issue is reproducible across all multipart geometry types: MultiPoint, MultiLineString and MultiPolygon layers.


https://github.com/user-attachments/assets/42e0c9f1-9a9b-447e-9ae2-5329aa5ab0b8

PR #4506, opened 2026-05-19
WIP Fixed the initial QT-6.10.3 migration issues regarding map tools

gabriel-bolbotina:
MMHighlight did not work with Qt 6.10.3 because the onGeometryChanged binding it relied on was not fired. Qt 6.10 seems to have stricter value-type equality checks evaluated the modified QgsGeometry as unchanged.

Updated:
- geometry properties to use QVariantList for map tools.
- install md
- cmake presets template
- the code for MMHighlight and the other components

To fix issues: 3, 4, 6, 7, 8 mentioned [here](https://github.com/MerginMaps/mobile/pull/4463#issuecomment-4439879542)

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -229,7 +229,7 @@ For building ABIs see https://www.qt.io/blog/android-multi-abi-builds-are-back
    binary vcpkg cache.
 
    ```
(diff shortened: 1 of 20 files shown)


Withalion:
Closed and we will move forward with #4511

PR #4513, opened 2026-05-22
Fixed delete button

gabriel-bolbotina:
Added check if the sketching controller is available or not, which fixes:

- the broken delete button when photo sketching is not selected
- reappearing sketches when deleting a photo with sketches and re-adding it in the project.

UI changes:

https://github.com/user-attachments/assets/3b4772aa-a900-49e5-a2b7-21317ec761e1
https://github.com/user-attachments/assets/28724eb1-1c9f-4498-bcd1-f333ad2adac7

### DIFF
--- a/app/qml/form/editors/MMFormPhotoEditor.qml
+++ b/app/qml/form/editors/MMFormPhotoEditor.qml
@@ -133,8 +133,8 @@ MMFormPhotoViewer {
     onDeleteImage: {
       // schedule the image for deletion
       internal.imageSourceToDelete = imageDeleteDialog.imagePath
-      root.sketchingController.removeBackupSketches()
-      root.sketchingController.clear()
+      root.sketchingController?.removeBackupSketches()
+      root.sketchingController?.clear()
       resetValueAndClose()
     }
 


Withalion (approved):

IvaKuklica:
Testing Result:
----------------------------------
🟢 Deletion of photos with sketches  - OK
🟢 Deletion of photos with disabled Reused last entered value - OK
🔴 **Deletion of photos with enabled Reused last entered value - HIGH RISK**

**Application:**
Device/OS/APP: Samsung Galaxy A53 / Android 15 / MM app. 2026.3.0 build 818751
Device/OS/APP: IPhone 15 / iOS 26.5 / MM app. 2026.3.0 - build 26.6.912911

Descripton
-----------------------------------

- When the Reuse Last Entered Value functionality is enabled for a photo field, all subsequently created features share a reference to the same photo file. If the user deletes a photo from any of these features using the Delete option in the form, the photo file is physically removed from the project storage. 
- **As a result, all previously created features that reference the same photo file lose their photos — the images are no longer visible in their forms.**
- This behavior is not documented in the Mergin Maps documentation, specifically on the [How to Reuse Last Entered Values](https://merginmaps.com/docs/field/reuse-last-values/)
- There is no warning displayed before the deletion to inform the user that the action will affect photos in other features.
- Using the Unlink option instead of Delete does not cause this issue — Unlink removes only the reference from the current feature without deleting the physical file, leaving Feature 1 and Feature 2 unaffected.


Project & Layer
----------------------------
Project: tc04_recording
Layer: Multiphoto layer
Feature form setting: Reuse Last Entered Value enabled for photo field

Steps to Reproduce
---------------------------------
Open the project tc04_recording in the Mergin Maps mobile app.
Navigate to Settings and enable Reuse Last Entered Value.

1. `Create Feature 1: `add two photos to the Multiphoto layer, check the checkbox next to the photo field to enable reuse, and save the feature.
2. `Create Feature 2:` the photo fields are automatically pre-filled with the photos from Feature 1 (reused). Save the feature without changes.
3. `Create Feature 3`: the photo fields are automatically pre-filled. Use the Delete option on the photos in the form and save the feature.
4. Open Feature 1 and Feature 2 and observe the photo fields.

Actual Result
--------------------------------
**After deleting the photos from Feature 3, the photos are also removed from Feature 1 and Feature 2**. The photo fields in Feature 1 and Feature 2 are empty — the images are no longer visible in their forms.
This is caused by the fact that Reuse Last Entered Value copies the file path reference to the same physical photo file across all features. When the photo is deleted via the Delete option, the physical file is removed from storage, breaking all features that share the same reference.

Expected Result
--------------------------------
Deleting a photo from Feature 3 removes only the photo reference and/or file associated with Feature 3. Photos in Feature 1 and Feature 2 remain intact and are still visible in their forms.

Attachments
---------------------------------
Screen recording showing the deletion of photos from Feature 3 and the resulting empty photo fields in Feature 1 and Feature 2.


https://github.com/user-attachments/assets/85ad66a3-269c-41c1-98e1-8c33b0b76eb5

Withalion:
@IvaKuklica this is intended behavior, user can either unlink (to remove the reference from this feature) or delete the photo (remove the file from project)

RastoHu:
> @IvaKuklica this is intended behavior, user can either unlink (to remove the reference from this feature) or delete the photo (remove the file from project)

@Withalion does the user realize that reusing the last entered values for photos causes them to be deleted from previously created features? If this behavior is intended, we need to at least document it clearly or warn the user.

Withalion:
> > @IvaKuklica this is intended behavior, user can either unlink (to remove the reference from this feature) or delete the photo (remove the file from project)
> 
> @Withalion does the user realize that reusing the last entered values for photos causes them to be deleted from previously created features? If this behavior is intended, we need to at least document it clearly or warn the user.

There is no special warning besides the information mentioned in the delete/unlink photo popup. As I mentioned to @IvaKuklica in our discussion there could be some warning text that "the photo is used in x more features" for example. For now there is no plan to add it. However this same behavior is already present in production no?

IvaKuklica:
New enhancement is created under link ----> [#4534](https://github.com/MerginMaps/mobile/issues/4534)

PR #4525, opened 2026-06-04
Value relation rework & more small updates

### LINKED ISSUE #2148: Value relation field - non-conforming value not displayed in edit mode
When a field's value doesn't match a value in the value relation it is not displayed in edit mode.

I've shared a sample project to demonstrate this - https://app.merginmaps.com/projects/andrewbailey/value_relation_test

Three points with a notes attribute of 'Good', 'Bad', 'Could be better'. A value relationship is defined that only allows 'Good' or 'Bad'. When loading the project into Input the value is displayed in the info pane after selecting it from the map but in edit mode the value is not displayed in the drop-down. I appreciate 'Could be better' should not be an option to select but currently the option box is blank indicating no data is stored in that field. I've got links to screenshots of both the QGIS project and the Input app below.

Info click https://photos.app.goo.gl/gUc8xz1f3DULXgjs5
Edit https://photos.app.goo.gl/2YnC9TU1i7D52mEj6
QGIS https://photos.app.goo.gl/JuGW5YTDutXZkiL3A

QGIS displays brackets around values when not in a value relationship. If the same could happen in Input that would be great.

### LINKED ISSUE #3297: Value relation entries beyond 10k cannot be selected
On a value relation field with > 10000 entries, only the first 10k are selectable.
Entries beyond the 10k mark can be searched but once selected they are not stored.

It seems that two different models are used for populating the list / searching and for fetching the selected key with `convertToKey()`, so the second one only contains the default first 10k entries.

### LINKED ISSUE #3737: Support to order value relation widget by column
`QGIS 3.42` introduces new sorting support for value relations, see https://qgis.org/project/visual-changelogs/visualchangelog342/#feature-additional-sorting-options-on-value-relation-widget

We could support that in our FeaturesModel.

### LINKED ISSUE #4133: Blank page appears when opening a widget Value on the tab Value Relation 2.
**Description**
When opening a tab Value Relation 2 - widget Value that does not have a form attribute defined, a new window opens and the entire page appears blank. 

**Environment**
Staging

App: Mergin Maps Mobile (IOS)
App version: 2025.7.0 (build 25.9.804311)
Device/OS: IPhone 13/ IOS 26.0

**Steps to reproduce**

1. Open the project **tester-ws-1/tc05_forms_widgets**
2. Edit the saved feature 
3. Pan to the tab _Value Relation 2_
4. Click _on the widget Type_ and choose _“Deciduous”_
5. Click _on widget Name_ and select the one that has no predefined options in QGIS.
6. **Click on the Country** and observe that **a new window opens on the bottom of the page**
7. **Click on the Values** and observe that **a new window open and entire page appears blank**

**Actual results**
A new window for the widget Values opens and shows a blank page. 

**Expected results**
Only the new window for the widget Values should appear , aligned to the bottom of the page, without affecting the whole page of the display. 

**Screenshots **

### LINKED ISSUE #4255: App does not allow NULL value on Value Relation widget
**Description**

In QGIS there is an option to enable 'Allow null values' on the Value Relation widget settings. This will create an option to choose 'no selection' in the drop down menu on the form in QGIS. However, in Mergin Maps, this option is not available when using the widget. This can be problematic if the user in the field needs to change a value on an already saved feature from any given value to a null value, currently there is no way to do this in the app.

**Environment**
Production

Mergin Maps 2025.7.0
Android 16

**Steps to reproduce**

- Create a project in QGIS with:
  - a spatial layer that has a field for recording the value relation
  - a non-spatial layer with a unique key field and a value field
- Fill in the non-spatial layer with unique key values like `uuid()` and values for the options field
- Set up the spatial layer form with a Value Relation widget to record the value from the non-spatial field, make sure to enable the 'Allow null values' option

Value relation field settings in QGIS

- Save the project and upload to Mergin Maps
- Download the project in the app
- Add a new feature to the spatial layer and select the value relation widget dropdown menu, there is no option for NULL value

**Actual results**

It is not possible to select a NULL value in the Mergin Maps app

Screenshot of Value Relation widget in Mergin Maps mobile app without the option to choose null value

**Expected results**

In QGIS, the drop-down menu for the value relation includes an option to choose 'no selection'

Value relation form in QGIS which has the option to choose null value

### LINKED ISSUE #4443: Value Relation completer freezes for ~15 seconds on tap with large lookup table (6000+ records) - regression since 2025.8.0
**Description**
When tapping a Value Relation field configured with "Use completer" enabled, the app freezes for approximately 15 seconds before becoming responsive. The freeze occurs on tap, before any text has been entered into the search field. This issue was not present prior to the 2025.8.0 release and appears to be a regression introduced by the infrastructure migration in that version.

**Environment** / **Application (+ app version, build, operating system)**
Mergin Maps version: 2026.1.3
Device: Juniper Cedar CT8X2
OS: Android 10
RAM: 4GB
Lookup table record count: ~6,318 records
Value Relation widget settings: Use completer ✅, Only match from beginning of string ❌
Lookup table format: GeoPackage

**Steps to reproduce**
- Create a Value Relation widget pointing to a lookup table with ~6,300 records
- Enable "Use completer" in the widget settings
- Open the attribute form on an Android device
- Tap the Value Relation field
- App freezes for approximately 15 seconds before the completer becomes usable

**Actual results**
The app freezes for approximately 15 seconds on tap. The freeze appears to occur regardless of whether any text has been entered, suggesting the full lookup table is being loaded into memory on tap rather than on keystroke.

**Expected results**
The completer should open immediately or near-immediately on tap, loading results only as the user types, as it did prior to the 2025.8.0 release.

**Workarounds attempted**
- VACUUM run on GeoPackage 
- Column index added to lookup field

### LINKED ISSUE #4477: Filtering does not work for fields with Value relation with Multiple selections
**Description**
Single select and Multi select filters do not work on fields that use Value relations widget with the option "Allow multiple selections" (when there are multiple values stored in a field). While it is possible to set up filter for such field in QGIS and use it in the mobile app, the result of filtering is always 0 features.

**Environment**
Production

v2026.2.0, Android

**Steps to reproduce**
1. Set up a Single select or Multi select Filter in QGIS for a field that is configured with Value relation with enabled multiple selections.
2. In the mobile app, try filtering features using one or multiple values from the list.

**Actual results**
There are 0 features that match the filter - all features from the layer disappears from the map.

**Expected results**
Depends on the intented use of filters:
- If the Single and Multi select filter is not appropriate for Value relation+multiple selection, the field should not be offered when setting up the filter in QGIS
- If it should work also for mutiple selection fields, the app should display the appropriate result of the filter. Note that there are (at least) two options for multiselect logic: OR vs. AND.

### LINKED ISSUE #4497: In fields with value map widgets, NULL is accepted as valid even if the constraint "not null" is enabled. Works as expected on QGIS Desktop.
**Description**

Take a field with a value map. The value map has also an entry for the NULL value (added by clicking "add NULL value")

Image

then also enable the "not null" constraint

Image

In QGIS Desktop all works as expected, ****if a value from the list is first selected (manually or using a default value), then** the user chooses the NULL entry**, the constraint does is job:

Image

On MM App the NULL entry from the list is accepted as a valid value:

Image

If the "NULL" entry from the value map is the default value then the constraint works also in MM App, **but not if the user changes to another entry then goes back to "NULL"**.

Not even evaluating "NULL" as text in a custom expression (rather than using the normal "not null" constraint) helps workaround this problem.

2026.2.0

### LINKED ISSUE #4505: Search field disappears when typing in Single/Multi select filter widget
## Description

When typing text into the search field of a Single select or Multi select filter widget, the search input field disappears from view after a few characters are entered. The typed text is still active (results are filtered) but the input field itself is no longer visible. Tapping Backspace causes the search input to reappear.

**Environment**
- [x] PROD

Mergin Maps iOS 2026.2.0 / iOS 26.4

## Steps to reproduce

1. Open project tester-ws-1/[preview_mode_bug](https://app.dev.merginmaps.com/projects/tester-ws-1/preview_mode_bug/tree)
2. Open Filters → Multi select-POI-Country
3. Tap the search field and start typing text
4. Observe the search field after a few characters are entered

## Actual results

- After typing some text, the search input field disappears from view
- The typed text appears near/within the results list instead of in the input field
- Filtered results are still shown correctly
- Tapping Backspace causes the search input to reappear

## Expected results

- Search input field remains visible while typing
- Typed text stays inside the search field at all times

## Screenshots or logs

https://github.com/user-attachments/assets/5f4ced3f-2346-4760-b2b5-167b50ada232

---
*Conclusion*
*Created by Claude and supervised by a human test analyst.*

### LINKED ISSUE #4508: Value Relation search matches all columns instead of display column only — affects feature form and filter widgets
## Description

When searching in a Value Relation widget (both in a feature form and in a Single/Multi select filter), the search matches against **all columns** of the referenced table instead of only the configured display column. This causes unexpected results — for example, searching for "Bratislava" returns "Slovakia" because "Bratislava" appears in another column (capital city) of the `world_capitals` lookup table, even though the display column is `country`.

Affects:
- Feature form — Country field (Value Relation widget) when adding/editing a feature
- Filter page — Multi select-POI-Country (and any Single/Multi select filter backed by a Value Relation)

**Environment**
- [x] PROD

Mergin Maps iOS 2026.2.0 / iOS 26.4

## Steps to reproduce

**Feature form:**
1. Open project tester-ws-1/preview_mode_bug
2. Add a new feature in the POI layer → open the feature form
3. Tap the Country field (Value Relation widget referencing `world_capitals`, display column = `country`)
4. Type "Bratislava" in the search field

**Filter:**
1. Open Filters → Multi select-POI-Country
2. Type "Bratislava" in the search field

## Actual results

- "Slovakia" is returned as a match for "Bratislava" — the search matched a non-display column (capital city name) in the `world_capitals` table
- Other non-display columns also produce unexpected matches

## Expected results

- Search should only match against the configured display column (`country`)
- Searching "Bratislava" should return no results since no country is named "Bratislava"

## Screenshots or logs

https://github.com/user-attachments/assets/654d1ecc-492e-4f2f-b09b-26dd42fda13e

---
*Conclusion*
*Created by Claude and supervised by a human test analyst.*

tomasMizera:
Introduces a new way of handling value relation editors. Improves performance and stability.

Changes:
 - Created a utility class `ValueRelationController` that performs lookups from `key` to `value`, handles hot reload (filter expression) and conversions from/to QGIS format, e. g. `"{1,2,3}"` to `[1,2,3]`
 - Lookup on background thread
 - Reading only Key and Value attributes (compared to all attributes before), no geometry
 - Unified search UX with filters - drawer goes fullscreen if searchbar gets focus

Todos:
- [x] Test PG lookups
- [ ] Test spatial filter expressions (@IvaKuklica) - filter expression like `distance(@geometry, @current_geometry) <= 0.000135`
- [ ] Autotests finalization

Other updates:
 - Updated min required number of features to show searchbar in drawers from 5 to 8
 - Added clear button to value map and value relation editors
 - Refactored `MMDrawerHeader` and `MMListMultiselectDrawer` multiselect drawer to make their API easier to use
 - Anhanced UX for completers - searchbar is autofocused

Resolves https://github.com/MerginMaps/mobile/issues/4255
Resolves https://github.com/MerginMaps/mobile/issues/4497
Resolves https://github.com/MerginMaps/mobile/issues/4443
Resolves https://github.com/MerginMaps/mobile/issues/3297
Resolves https://github.com/MerginMaps/mobile/issues/3737
Resolves https://github.com/MerginMaps/mobile/issues/4508
Resolves https://github.com/MerginMaps/mobile/issues/4505
Resolves https://github.com/MerginMaps/mobile/issues/2148 (<- the only limitation is VR fields with filter expression)
Resolves (? needs to be tested) https://github.com/MerginMaps/mobile/issues/4504


app/qml/form/editors/MMFormValueMapEditor.qml
```diff
@@ -82,17 +81,21 @@ MMFormComboboxBaseEditor {
...
-        height: noItemsText.implicitHeight + __style.margin40
-      
-        MMComponents.MMText {
-          id: noItemsText
-          text: qsTr( "No items" )
-          anchors.centerIn: parent
+        type: MMButton.Types.Tertiary
+
+        fontColor: __style.darkGreyColor
+        fontColorHover: __style.nightColor
```
Withalion, line 94: shouldn't the text be red?

app/qml/form/editors/MMFormValueRelationEditor.qml
```diff
@@ -58,133 +58,131 @@ MMFormComboboxBaseEditor {
...
+      drawerHeader.titleFont: __style.t2
+
+      drawerHeader.topLeftItem.visible: !root._fieldValueIsNull
+      drawerHeader.topLeftItemContent: MMComponents.MMButton {
+        text: qsTr( "Clear" )
+
+        type: MMButton.Types.Tertiary
+
+        fontColor: __style.darkGreyColor
+        fontColorHover: __style.nightColor
```
Withalion, line 82: shouldn't the text be red?

app/qml/form/MMFormStackController.qml
```diff
@@ -144,7 +144,7 @@ Item {
     // https://github.com/MerginMaps/mobile/issues/2879
     for ( let i = 0; i < formsStack.depth; i++ ) {
       let form = formsStack.get( i )
-      form.featureLayerPair = __inputUtils.createFeatureLayerPair()
+      form.featureLayerPair = __inputUtils.createFeatureLayerPair
```
Withalion: We are passing reference to the function here, that doesn't seem right

app/synchronizationmanager.cpp
```diff
@@ -49,6 +52,8 @@ void SynchronizationManager::syncProject( const Project &project, SyncOptions::A
     return;
   }
 
+  CoreUtils::log( u"Sync Manager"_s, u"Requested download of project %2"_s.arg( project.mergin.projectName ) );
```
Withalion:
```suggestion
  CoreUtils::log( QStringLiteral( "Sync Manager" ), QStringLiteral( "Requested download of project %2" ).arg( project.mergin.projectName ) );
```

app/synchronizationmanager.cpp
```diff
@@ -71,6 +76,8 @@ void SynchronizationManager::syncProject( const LocalProject &project, SyncOptio
     return;
   }
 
+  CoreUtils::log( u"Sync Manager"_s, u"Requested %1 sync of project %2"_s.arg( requestOrigin == SyncOptions::ManualRequest ? "manual" : "automatic" ).arg( project.projectName ) );
```
Withalion: Same as above

app/valuerelationcontroller.cpp
```diff
@@ -0,0 +1,349 @@
...
+
+#include "qgsproject.h"
+#include "qgsfeedback.h"
+#include "qgsvaluerelationfieldformatter.h"
+#include "qgsexpressioncontextutils.h"
+#include "qgsvectorlayerfeatureiterator.h"
+
+#include <QtConcurrentRun>
+
+using namespace Qt::Literals;
```
Withalion: Same thing as above with `QStringLiteral()`

app/valuerelationcontroller.cpp
```diff
@@ -0,0 +1,349 @@
...
+QStringList ValueRelationController::qgisFormatToArray( const QVariant &qgsValue ) const
+{
+  if ( !mIsInitialized )
+  {
+    CoreUtils::log( u"Value Relation"_s, u"Attempted to convert QGIS format to array, but the class is not initialized!"_s );
+    return {};
+  }
+
+  if ( qgsValue.isNull() || qgsValue.toString().isEmpty() )
+    return QStringList();
```
Withalion:
```suggestion
    return {};
```

app/valuerelationcontroller.cpp
```diff
@@ -0,0 +1,349 @@
...
+  {
+    const QString str = qgsValue.toString().trimmed();
+
+    if ( str.startsWith( '{' ) || str.startsWith( '[' ) )
+    {
+      return QgsValueRelationFieldFormatter::valueToStringList( qgsValue );
+    }
+  }
+
+  return QStringList() << qgsValue.toString();
```
Withalion:
```suggestion
  return { qgsValue.toString() };
```

app/valuerelationcontroller.cpp
```diff
@@ -0,0 +1,349 @@
...
+    if ( keyFieldDef.isNumeric() )
+    {
+      bool ok = false;
+      const qlonglong numVal = k.toLongLong( &ok );
+      if ( ok )
+      {
+        typedKey = QVariant( numVal );
+      }
+    }
+    quotedKeys << QgsExpression::quotedValue( typedKey );
```
Withalion: So we convert string -> number -> string?

app/valuerelationcontroller.h
```diff
@@ -0,0 +1,129 @@
...
+    Q_INVOKABLE QStringList qgisFormatToArray( const QVariant &qgsValue ) const;
+
+    /**
+     * Formats a list of key strings into the QGIS wire value "{k1,k2,...}".
+     *   allowMulti=true  → "["1","2","3"]"  becomes ["1","2","3"]
+     *   allowMulti=true  → "["1"]"          becomes "1"
+     *   allowMulti=false → "1"              becomes "1"
+     * An empty list produces "".
+     */
+    Q_INVOKABLE QString arrayToQgisFormat( const QStringList &keys ) const;
```
Withalion, line 64: I still think these should be just static utility functions, not tied to the class

app/main.cpp
```diff
@@ -332,6 +333,7 @@ void initDeclarative()
   qmlRegisterType< LayerFeaturesModel >( "mm", 1, 0, "LayerFeaturesModel" );
   qmlRegisterType< RelationFeaturesModel >( "mm", 1, 0, "RelationFeaturesModel" );
   qmlRegisterType< ValueRelationFeaturesModel >( "mm", 1, 0, "ValueRelationFeaturesModel" );
+  qmlRegisterType< ValueRelationController >( "mm", 1, 0, "ValueRelationController" );
```
Withalion:
```suggestion

```
Let's do it the declarative way, by using `QML_ELEMENT`

app/valuerelationfeaturesmodel.cpp
```diff
@@ -12,6 +12,10 @@
 
 #include "qgsvaluerelationfieldformatter.h"
 #include "qgsexpressioncontextutils.h"
+#include "qgsvectorlayer.h"
+
+using namespace Qt::Literals;
```
Withalion: Same as elsewhere

Withalion (commented):
Great improvement!

IvaKuklica:
Testing progress update
----------
The following tickets have been tested so far and the results with comments are summarized below. 
Testing activities are still ongoing and additional tests will continue. 


- https://github.com/MerginMaps/mobile/issues/4504 - **Test passed** ✅ 
- https://github.com/MerginMaps/mobile/issues/4505 - T**est passed** ✅ 
- https://github.com/MerginMaps/mobile/issues/4508 -**Test passed** ✅ 

Open issue:
-----
- https://github.com/MerginMaps/mobile/issues/4497

🟠 1. **NULL entry not visible in the value map drop-down in MM App** — the option added via Add NULL value in QGIS Desktop simply doesn't appear as a selectable choice in the mobile app - **Is it correct behavior of application?** 

🔴 3. **Clear button doesn't match the Figma design**  

Screenshot: 

Screenshot 2026-06-17 210545

app/valuerelationcontroller.cpp
tomasMizera: Should be just string -> number, no?

app/qml/form/editors/MMFormValueMapEditor.qml
tomasMizera: Let's keep it like this for now

app/qml/form/editors/MMFormValueRelationEditor.qml
tomasMizera: Let's keep it like this for now

app/qml/form/MMFormStackController.qml
tomasMizera: Thanks! This one would be hard one to debug

app/valuerelationcontroller.cpp
Withalion: yes

Withalion: We could do `QgsExpression::quotedValue( value, QMetaType::Type )`

IvaKuklica:
Testing progress update
-------------

- https://github.com/MerginMaps/mobile/issues/3297 - Test passed ✅

- https://github.com/MerginMaps/mobile/issues/4443  - Test passed ✅
**Note for Android** - the app still appears to freeze briefly (2–3 seconds for project with 24K records) but then recovers and provides data.

https://github.com/user-attachments/assets/d5c6e352-61c7-4b8e-89c3-22f5c8015228

PR #4526, opened 2026-06-04
Rework cancellation logic for LayerFeaturesModel searches

uclaros:
Rework of the cancellation logic in `LayerFeaturesModel` feature requests.

Main changes:
- Each search gets its own `QgsFeedback` and `QFutureWatcher` instances.
- To cancel a request we use `QgsFeedback::cancel()` and let the provider stop the iteration.
- Pending request is cancelled when model or layer is destroyed.


app/layerfeaturesmodel.cpp
```diff
@@ -82,70 +102,79 @@ void LayerFeaturesModel::populate()
...
 }
 
 void LayerFeaturesModel::onFutureFinished()
 {
-  QFutureWatcher<QgsFeatureList> *watcher = static_cast< QFutureWatcher<QgsFeatureList> *>( sender() );
-  const QgsFeatureList features = watcher->future().result();
-  beginResetModel();
-  mFeatures.clear();
-  for ( const auto &f : features )
+  QFutureWatcher<SearchResultData> *watcher = static_cast< QFutureWatcher<SearchResultData> *>( sender() );
```
Withalion: Should be dynamic_cast instead

Withalion (approved):
let's rebase before merging

app/layerfeaturesmodel.cpp
uclaros: The sender here is guaranteed to be the future watcher.
I'd argue that actually crashing would be a better choice in case future me tries to call `onFutureFinished()` by other means!

Withalion: Alright, I don't really have a strong argument here. It was also pointed out by [clang-tidy](https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/pro-type-static-cast-downcast.html)

uclaros: If type safety is the issue, I'd suggest a0a46c05

uclaros:
I switched back to using the `QgsVectorLayerFeatureSource` for iteration instead of the dataSource feature source in 1dbadb5.
Now searching over virtual fields will work again.

IvaKuklica:
Testing Result ✅ PASS
Application:
Device/OS/APP: Samsung Galaxy A53 / Android 15 / MM 2026.3.0 - 826651
Device/OS/APP: IPhone 15 / iOS26.5 / MM 2026.3.0 - 920811

The application has the expected behavior.

PR #4544, opened 2026-06-16
Use fixed pixel distance for line Start and End handles 

### LINKED ISSUE #2280: Geometry editing handles visual problems
There are a few minor things with handles (lines that go from the first and last points in line geometry).

1. Length of handles does not respect a change of scale (when a user zooms in/out). They are updated only when the geometry is changed. See the attached video (it has an old UI)
https://user-images.githubusercontent.com/22449698/191178167-393beea9-2e56-409c-8a49-284af2bc58c8.mp4

2. There is a visual glitch - sometimes one of the handles is just a white line. I can reproduce it in this scenario:
  a. Zoom in enough so that the end of a line feature is outside of the visible canvas
  b. Select the feature
  c. Start editing geometry
  d. When you move to the first handle, it is only a white line (see image below 👇🏻)

### LINKED ISSUE #2612: Time streaming mode problem

uclaros:
Recording map tool Vertices are now re-calculated when the map scale changed.
This allows for line Start and End handles to be rendered at a fixed pixel distance from the start/end of the line.

Peek 2026-06-16 20-34

Also, lays the ground for filtering of segment middle markers based on segment pixel length (not implemented in this PR).

Also fixes the rendering of dashed MMhighlights that are constructed off-screen.

Fixes https://github.com/MerginMaps/mobile/issues/2280


Withalion (approved):
Looks okay, the test require some changes and rebase will be also necessary for builds to pass

PR #4547, opened 2026-06-17
Added 'No value' MM List delegate for filtering

gabriel-bolbotina:
For single/ multi select filter input there was no way to select 'No value' to filter out only the values that do not have anything selected for that field.
Created an MM list delegate component for the No value entry.
In this PR, I implemented it for value relation filtering.


app/qml/filters/components/MMFilterDropdownValueRelationInput.qml
```diff
@@ -109,8 +123,9 @@ Column {
         //
         // Large fids could be converted to scientific notation on their way to cpp,
         // so we convert them to string first in JS.
+        // Null values (representing "No value") are preserved as-is.
         //
-        selectedItems = selectedItems.map( x => x.toString() )
+        selectedItems = selectedItems.map( x => x !== null && x !== undefined ? x.toString() : null )
```
Withalion, line 128:
```suggestion
        selectedItems = selectedItems.map(x => x?.toString() ?? null)
```
nullish coalescence operator (??) does what you are doing here

Withalion (changes requested):
This doesn't look like a good approach. Is there any reason why don't we add the null/empty case to value relation model? We did something similar in unique values model. So, instead of injecting the no value in QML I would have some kind of flag for value relation model, which would tell it to also include null value in it's data. That would also help on the QML side, you wouldn't need to use the header hack and extra component for the delagate.

Also please add a screenshot to the PR description

Withalion:
After discussing it with @tomasMizera we decided to have it done in c++ model, feel free to subclass the value relation model and just add functionality which would handle the null value

PR #4556, opened 2026-06-24
Fix issues found during regression testing for 2026.3.0

Withalion:
Fixes:
1. When creating new project through mobile app, setting values for fields was not registered. No way to add new fields to forms. Was caused by [new behaviour](https://doc.qt.io/qt-6/qml-qtquick-listview.html#delegateModelAccess-prop) in `ListView` since Qt 6.10


tomasMizera (approved):
🤯

PR #4560, opened 2026-06-25
Fix ios photo picker to copy all image types

Withalion:
There as regression where we supported only jpg images, now we should support all types


tomasMizera (approved):
👍🏻

PR #4572, opened 2026-06-30
Feature/clear button for calendar drawers

### LINKED ISSUE #4551: Add clear button for date and date-time form editor drawer
**Context:**  Currently there is no way to set a date/dateTime field to NULL once a value has been picked. Users need a Clear button inside the calendar drawer to reset the field.

**Solution:** Add a Clear button for date and date time form editors to set the value to NULL and collapse the drawer. 

**Implementation hints:**  
- `app/qml/form/editors/MMFormCalendarEditor.qml `and  `app/qml/form/components/MMCalendarDrawer.qml` is where the changes will be made.
- Add a new signal to be fired when the clear button is tapped
- There is a signal that already supports null via its second param in MMFormCalendarEditor:  `root.editorValueChanged( root._fieldValue, true )  // isNull = true` 
- Also, you can take a look how the clear button works in MMFilterDateInput.qml lines 62-94 for filter date fields or the Clear button in Value Map/ Relation editor.

**Acceptance criteria:**
- The button appears on the top left corner of the calendar form editor
- Once pressed, the value is set to NULL and the form editor drawer collapses 
- The design of the button follows the Figma designs that can be found here: https://www.figma.com/design/d78a4ehBBgK6MyxCTuCAqg/Mobile-App?node-id=6993-24415&m=dev

xkello:
### Problem
1. There was no way to remove value/set it as null for date field in calendar drawers. Users had to reopen the drawer by creating exactly the same point on the map for example.

2. Clearing the date field is causing runtime error.

---

### Solution
1. Added a clear button in the drawer in header space. 

2. Added guards which check for `undefined` and `null`  `_fieldValue` before assigning `dateTime` in calendar drawer.

---

### Behavior
- When user opens calendar drawer, there is a **Clear** button in the top left corner, which both closes the calendar drawer and resets the value of date field when used. 
- When calendar date value is empty, the clear button is not vissible.

---

### Images for part 1
- Tall and wide screen behavior with filled date field:
image
image

<br>

- Screen behavior when date field is empty:
image

---

### Video for part 1
[Screencast From 2026-06-30 17-43-27.webm](https://github.com/user-attachments/assets/54474d2a-012a-4f2b-83ff-8a4cd851d692)


app/qml/form/components/MMCalendarDrawer.qml
```diff
@@ -22,13 +22,29 @@ MMComponents.MMDrawer {
   property alias hasDatePicker: dateTimePicker.hasDatePicker
   property alias hasTimePicker: dateTimePicker.hasTimePicker
   property bool showSeconds: false
+  property bool fieldValueIsNull: true
```
Withalion: no need to have another property just set the value to null

app/qml/form/components/MMCalendarDrawer.qml
```diff
@@ -22,13 +22,29 @@ MMComponents.MMDrawer {
...
 
   signal primaryButtonClicked
+  signal clearButtonClicked
```
Withalion, line 28: maybe at this point we can have just one signal which says user finished picking the value

app/qml/form/components/MMCalendarDrawer.qml
```diff
@@ -22,13 +22,29 @@ MMComponents.MMDrawer {
...
 
   dim: true
 
   drawerHeader.title: root.title
 
+  drawerHeader.topLeftItem.visible: !root.fieldValueIsNull
```
Withalion: visibility based on dateTime value

app/qml/form/components/MMCalendarDrawer.qml
Withalion: I see that now this was mentioned in the ticket to create it

app/qml/form/components/MMCalendarDrawer.qml
```diff
@@ -22,13 +22,29 @@ MMComponents.MMDrawer {
...
+  drawerHeader.topLeftItemContent:
+    MMComponents.MMButton {
+      text: qsTr("Clear")
+      type: MMButton.Types.Tertiary
+      fontColor: __style.darkGreyColor
+      fontColorHover: __style.nightColor
+
+      onClicked: {
```
Withalion, line 41: I would change the the value to `null` here

Withalion: Or empty `Date()`

app/qml/form/editors/MMFormCalendarEditor.qml
```diff
@@ -122,6 +127,13 @@ MMPrivateComponents.MMBaseSingleLineInput {
     }
   }
 
+  function hasValidFieldValue() {
+    return root._fieldValueIsNull
+        || root._fieldHasMixedValues
+        || root._fieldValue === undefined
+        || root._fieldValue === null
```
Withalion: For the future this can be replaced with `??` (nullish coalescence operator), it probably won't simplify it here

app/qml/form/editors/MMFormCalendarEditor.qml
```diff
@@ -96,15 +96,20 @@ MMPrivateComponents.MMBaseSingleLineInput {
...
       hasTimePicker: root.includesTime
       showSeconds: root.showSeconds
+      fieldValueIsNull: root._fieldValueIsNull
 
       onPrimaryButtonClicked: {
         root.newDateSelected( dateTime )
       }
 
+      onClearButtonClicked: {
+        root.editorValueChanged( root._fieldValue, true )
```
Withalion: currently the actual value doesn't change we just flag it as null no?

Withalion (commented):
Welcome to the jungle!

Withalion:
Also the the clear text is not horizontally aligned with the rest of header and the text color is wrong

app/qml/form/components/MMCalendarDrawer.qml
xkello: null throws error as `date property` cannot hold `null value`.
`new Date()` would work but `close()` destroys the instance right after so it would not make sense to add it as it would be cleared right after closing the window

app/qml/form/editors/MMFormCalendarEditor.qml
```diff
@@ -122,6 +127,12 @@ MMPrivateComponents.MMBaseSingleLineInput {
     }
   }
 
+  function hasValidFieldValue() {
+    return root._fieldValueIsNull
+        || root._fieldHasMixedValues
+        || (root._fieldValue ?? null) === null
```
Withalion:
```suggestion
        || root._fieldValue ?? true
```
this should be the same

app/qml/form/editors/MMFormCalendarEditor.qml
```diff
@@ -122,6 +127,12 @@ MMPrivateComponents.MMBaseSingleLineInput {
     }
   }
 
+  function hasValidFieldValue() {
```
Withalion: It would make more sense to call it invalid values

app/qml/form/editors/MMFormCalendarEditor.qml
```diff
@@ -96,15 +96,20 @@ MMPrivateComponents.MMBaseSingleLineInput {
...
       hasTimePicker: root.includesTime
       showSeconds: root.showSeconds
+      showClearButton: root._fieldValueIsNull
 
       onPrimaryButtonClicked: {
         root.newDateSelected( dateTime )
       }
 
+      onClearButtonClicked: {
+        root.editorValueChanged( undefined, true )
```
Withalion: is there a reason we can't use null? If not then let's prefer using null

app/qml/form/editors/MMFormCalendarEditor.qml
Withalion, line 1: can you as well purge the `openCalendar()` method only keep `forceActiveFocus()` and then from `openPicker()` take the contents and put everything in `openCalendar()`. Remove `openPicker()`

app/qml/form/components/MMCalendarDrawer.qml
```diff
@@ -22,12 +22,27 @@ MMComponents.MMDrawer {
...
   property bool showSeconds: false
+  property bool showClearButton: true
 
   signal primaryButtonClicked
+  signal clearButtonClicked
 
   dim: true
 
   drawerHeader.title: root.title
+  drawerHeader.topLeftItem.visible: !root.showClearButton
```
Withalion: the logic of visibility is opposite to the naming of the variable, it should match

Withalion (commented):
The clear button doesn't work for calendar drawers in filters, but that might be intentional.

Please also rebase on current master

Withalion (approved):

PR #4578, opened 2026-07-02
Fix white map sketching bugs

xkello:
### Description

This PR fixes several issues in map sketches related to color selection and eraser state:
1. When opening map sketches and drawing with the preselected white color, the saved sketch would turn black instead of white.
2. The selection highlight for the white color was displayed as black.
3. When the eraser was selected, the previously selected color still appeared selected in the UI. This made it look like clicking the eraser again would deselect it and return to the previously selected color.

https://github.com/user-attachments/assets/49393d54-462d-4df2-9347-000b1f443b6a

### Changes

- Added explicit handling of the active sketch color.
- The active color is now initialized from the first available sketch color when map settings are set.
- The color picker now emits a color change request instead of changing internal state directly.
- The sketch drawer now updates the sketching controller with the selected color and disables the eraser when a color is selected.
- null is now used as fallback when the sketching controller or active color is not available.
- Removed a few redundant QML blocks, including anchors that had no effect.

### Final look

https://github.com/user-attachments/assets/6a5b257e-69bd-4019-a98f-ae369be1f45f

### DIFF
--- a/app/qml/map/MMMapSketchesDrawer.qml
+++ b/app/qml/map/MMMapSketchesDrawer.qml
@@ -79,15 +79,17 @@ MMComponents.MMDrawer {
       
       MMComponents.MMColorPicker {
         id: colorPicker
-        colors: root.sketchingController?.availableColors() ?? __style.photoSketchingWhiteColor
+
+        colors: root.sketchingController?.availableColors() ?? null
+        activeColor: root.sketchingController?.activeColor ?? null
 
         Layout.alignment: Qt.AlignHCenter
         Layout.maximumWidth: parent.width
 
-        onActiveColorChanged: {
+        onActiveColorChangeRequested: {
           if ( root.sketchingController )
           {
-            root.sketchingController.activeColor = colorPicker.activeColor
+            root.sketchingController.activeColor = newColor
             root.sketchingController.eraserActive = false
           }
         }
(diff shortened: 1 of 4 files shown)


tomasMizera (approved):
Fix formatting and it is ready to merge :)

app/qml/map/MMMapSketchesDrawer.qml
```diff
@@ -79,15 +79,17 @@ MMComponents.MMDrawer {
       
       MMComponents.MMColorPicker {
         id: colorPicker
-        colors: root.sketchingController?.availableColors() ?? __style.photoSketchingWhiteColor
+
+        colors: root.sketchingController?.availableColors() ?? null
+        activeColor: root.sketchingController?.activeColor ?? null
```
Withalion: btw using `??` here doesn't make sense

tomasMizera: Sketching controller can be undefined and often is at startup, but maybe that `?? null` part could be skipped

xkello: Seems to be working even without the `?? null` parts. Removing it

Withalion: Yep I was just pointing to the `?? null` part, if we don't give it some specific default value on `null`/`undefined` it doesn't really makes sense to use the operator to "double guarantee" the value will be null

PR #4589, opened 2026-07-09
Update incorrect range translations

### LINKED ISSUE #4499: Fields with a "range" widget: message should be "value must be lower/greater or equal to..." and not just "lower/greater"
**Description**

Take a field with a range widget, for example

Image

min and max values are valid as are included in the range. 

On MM App the error/warning message that is shown when an entered value is out of the range seems to suggest a different thing

Image

2026.2.0

**Actual results**

**Expected results**

**Screenshots**
Preferably add images as HTML element as shown below:

xkello:
### Description

Fixes a misleading validation message shown when a user enters a number outside the allowed range.
Ticket: https://github.com/MerginMaps/mobile/issues/4499
Previously, the message told users to select a number strictly lower/higher than the range boundary, when it should say lower/higher or equal to the boundary - since the boundary value itself is valid.

### Current behavior

For an allowed range of 1–100:
Entering 101 shows: "Value must be lower than 100" - but 100 is a valid value, so the message is misleading.
Entering 0 shows: "Value must be higher than 1" - same issue in the other direction.

### What changed

Updated the validation message strings to say "less than or equal to X" / "greater than or equal to X"

Changes were made to `fieldvalidator.cpp` manually and to `input_en.ts` using script. 

### Expected behavior

- Entering 101 now shows: "Value must be less than or equal to 100"
- Entering 0 now shows: "Value must be greater than or equal to 1"

### DIFF
--- a/app/attributes/fieldvalidator.cpp
+++ b/app/attributes/fieldvalidator.cpp
@@ -19,8 +19,8 @@
 #include <QLocale>
 
 QString FieldValidator::numberInvalid() { return tr( "Value must be a number" ); };
-QString FieldValidator::numberUpperBoundReached() { return tr( "Value must be lower than %1" ); };
-QString FieldValidator::numberLowerBoundReached() { return tr( "Value must be higher than %1" ); };
+QString FieldValidator::numberUpperBoundReached() { return tr( "Value must be less than or equal to %1" ); };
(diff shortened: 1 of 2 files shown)


Withalion (approved):
Usually we don't update the translations in specific PR's, we just bulk update before release

PR #4590, opened 2026-07-10
Align mobile regex with server regex

### LINKED ISSUE #4365: Invalid character (exclamation mark) in Project title
**Description**
When creating a project whose name ends with an exclamation mark (for example MyProject!), the project is created locally but is not uploaded on the server. The user finishes the creation flow but the project never appears in the server project list.

**Environment**
Staging

App version: 2026.1.1 (build 776151)
Device/OS: Samsung Galaxy A53 5G, / Android 15
App version: 2026.1.1 (build 870211)
Device/OS: IPone 14.5 / iOS 26.1

**Steps to reproduce**

1. Open the app and create a new project.
2. Enter a project name that ends with an exclamation mark (e.g., MyProject!).
3. Upload the project to the server.
4. Observe the server project list.

**Actual results**
- Project is created locally but does not appear on the server.
- No clear error or warning is shown to the user during creation.

**Expected results**
- The application should prevent creation of a project whose name contains invalid characters (!).
- The project should not be created locally when the name is invalid.
- The application should display a clear warning message to the user indicating the project name contains invalid characters.

**Screenshots**

https://github.com/user-attachments/assets/57526fab-ff69-412a-bac4-4d342478c467

### LINKED ISSUE #4516: Unify project name rules with server
The server rules for the project name are stricter, which means the project created in the app cannot be uploaded to the cloud.

This user's project cannot be uploaded due to an apostrophe.
Image

Hint for the fix:
The relevant server check is done here:
https://github.com/MerginMaps/server/blob/master/server/mergin/sync/forms.py#L17
Most of the checks are sorted, but the regex probably slipped through.

```
def is_reserved_word(name: str) -> str | None:
    """Check if name is reserved in system"""
    reserved = r"^support$|^helpdesk$|^merginmaps$|^lutraconsulting$|^mergin$|^lutra$|^input$|^admin$|^sales$"
    if re.match(reserved, name) is not None:
        return "The provided value is invalid."
    return None


def has_valid_characters(name: str) -> str | None:
    """Check if name contains only valid characters"""
    if re.match(r"^[\w\s\-\.]+$", name) is None:
        return "Please use only alphanumeric or the following -_. characters."
    return None


def has_valid_first_character(name: str) -> str | None:
    """Check if name contains only valid characters in first position"""
    if re.match(r"^[\s.].*$", name) is not None:
        return f"Value can not start with space or dot."
    return None

from pathvalidate import (
    validate_filename,
)
def check_filename(name: str) -> str | None:
    """Check if name contains only valid characters for filename"""
    error = None
    try:
        validate_filename(name)
    except ValidationError:
        error = "The provided value is invalid."
    return error
```

xkello:
### Description
Aligns the mobile app project name validation with the server validation rules. Previously, the mobile app used a different (blacklist-based) character validation than the server (whitelist-based), which allowed users to create local projects with names containing characters the server would reject - meaning the project could never be pushed/synced because some characters were missing from the mobile regex.

Tickets: https://github.com/MerginMaps/mobile/issues/4516, https://github.com/MerginMaps/mobile/issues/4365

### Current behaviour
The mobile app only blocked a fixed set of blacklisted special characters (``@#$%^&*(){}[]\/|+=<>~?:;,`'"``etc.). Any character not explicitly listed was allowed, even if the server would reject it later.
This meant a user could successfully create a local project with an invalid name (e.g. containing `!`), work with it locally, and only discover the mismatch when sync to the server failed.

### What changed
- Removed the special-character blacklist from `reForbiddenNames` in `CoreUtils::isValidName()`
- Added a new whitelist regex, `reValidCharacters`, allowing only word characters, whitespace, hyphens, and dots (`^[\w\s\-\.]+$`) — matching the server's `has_valid_characters()` check
- `isValidName()` now requires both checks to pass: the name must not match any forbidden pattern (reserved words, OS device names, leading space/dot) and must match the valid-character whitelist

### Expected behaviour
Project names can now only contain characters the server also accepts, preventing the local-creation/server-rejection mismatch
Previously problematic characters like `'` or `!` are now handled consistently - screenshot attached showing project creation working correctly with these characters

Screenshot_20260710-082727
Screenshot_20260710-082717
Screenshot_20260710-081406


core/coreutils.cpp
```diff
@@ -271,9 +271,13 @@ bool CoreUtils::isAuthConfigFile( const QString filePath )
 
 bool CoreUtils::isValidName( const QString &name )
 {
-  static QRegularExpression reForbiddenmNames( R"([@#$%^&*\(\)\{\}\[\]\\\/\|\+=<>~\?:;,`\'\"]|^[\s^\.].*$|^CON$|^PRN$|^AUX$|^NUL$|^COM\d$|^LPT\d|^support$|^helpdesk$|^merginmaps$|^lutraconsulting$|^mergin$|^lutra$|^input$|^sales$|^admin$)", QRegularExpression::CaseInsensitiveOption );
+  static QRegularExpression reForbiddenmNames( R"(^[\s^\.].*$|^CON$|^PRN$|^AUX$|^NUL$|^COM\d$|^LPT\d|^support$|^helpdesk$|^merginmaps$|^lutraconsulting$|^mergin$|^lutra$|^input$|^sales$|^admin$)", QRegularExpression::CaseInsensitiveOption );
```
Withalion: Could you split this long regex to multiple by category as server does?

core/coreutils.cpp
```diff
@@ -271,9 +271,13 @@ bool CoreUtils::isAuthConfigFile( const QString filePath )
...
+  static QRegularExpression reValidCharacters( R"(^[\w\s\-\.]+$)" );
+
   QRegularExpressionMatch matchForbiddenNames = reForbiddenmNames.match( name );
-  return !matchForbiddenNames.hasMatch();
+  QRegularExpressionMatch matchValidCharacters = reValidCharacters.match( name );
+
+  return !matchForbiddenNames.hasMatch() && matchValidCharacters.hasMatch();
```
Withalion: We should check the length of the file name, maximum 255 dictated by `pathvalidate` lib

Withalion (commented):
Make sure tests (`TestCoreUtils::testNameValidation()`) are up to date as well

Please rebase to current master

core/coreutils.cpp
```diff
@@ -269,11 +269,36 @@ bool CoreUtils::isAuthConfigFile( const QString filePath )
...
+{
+  static QRegularExpression re( R"(^support$|^helpdesk$|^merginmaps$|^lutraconsulting$|^mergin$|^lutra$|^input$|^sales$|^admin$)", QRegularExpression::CaseInsensitiveOption );
+  return re.match( name ).hasMatch();
+}
+
+static bool hasValidCharacters( const QString &name )
+{
+  static QRegularExpression re( R"(^[\w\s\-\.]+$)" );
+  return re.match( name ).hasMatch();
+}
```
Withalion: no need to create new functions, just bundle all the regexes in `isValidName()`
also you are missing function declarations of these new functions in header file if they would stay

Withalion (commented):
What about those tests?

core/coreutils.cpp
xkello: Totally forgot there are tests. On it

Withalion (approved):

PR #4593, opened 2026-07-12
Feature/multiline description support

### LINKED ISSUE #4502: Value map/relation field: support multi-line descriptions
## Summary

Currently, long description values in value map or value relation fields are truncated in the selection dialog. There is no way to add line breaks, forcing users to shorten their descriptions to avoid cut-off text.

## Expected behaviour

Descriptions should:
- support line breaks so multi-line text wraps
- expand the list item height to accommodate longer text without truncation.

- Let's introduce a max limit of 3 lines (instead of the current 1)

## Current behaviour

Long description strings are cut off with an ellipsis (`…`) in the value map picker dialog. No workaround exists within the app — users must shorten their descriptions.

## Steps to reproduce

1. Create a value map field in a QGIS project with long description strings.
2. Open the feature form in the mobile app and tap the value map field.
3. Observe that descriptions are truncated.

xkello:
### Description
Long description values in value map or value relation fields were truncated in the selection dialog, with no way to add line breaks. This forced users to shorten their descriptions to avoid cut-off text.

**Ticket:** https://github.com/MerginMaps/mobile/issues/4502

IMG_4764
IMG_4763
IMG_4762

### What changed
In `app/qml/components/MMListDelegate.qml`, both the primary and secondary `Text` elements now support multi-line wrapping instead of single-line truncation.
`Text.Wrap` was selected as it seemed the most appropriate of the three wrap methods, since users may use `snake_case` or similar naming conventions in their descriptions.

Screenshot_20260712-194517
Screenshot_20260712-194508
Screenshot_20260712-194458

### Expected behaviour
Descriptions now support line breaks/wrapping so multi-line text displays properly
List item height expands to accommodate up to 3 lines of text without truncation
Only descriptions exceeding 3 lines get truncated with an ellipsis (...), instead of the previous 1-line limit

### DIFF
--- a/app/qml/components/MMListDelegate.qml
+++ b/app/qml/components/MMListDelegate.qml
@@ -87,6 +87,10 @@ Item {
           textFormat: Text.PlainText
 
           color: __style.nightColor
+
+          elide: Text.ElideRight
+          wrapMode: Text.Wrap
+          maximumLineCount: 3
         }
 
         MMText {
@@ -98,6 +102,10 @@ Item {
           text: root.secondaryText
 
           color: __style.nightColor
+
+          elide: Text.ElideRight
+          wrapMode: Text.Wrap
+          maximumLineCount: 3
         }
       }
 
--- a/app/qml/gps/MMSelectionDrawer.qml
+++ b/app/qml/gps/MMSelectionDrawer.qml
@@ -76,6 +76,7 @@ MMComponents.MMDrawer {
           color: __style.nightColor
 
           elide: Text.ElideRight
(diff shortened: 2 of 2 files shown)


Withalion (commented):
I'm not 100% sold on this solution as it enables 3 line text for a lot of stuff. Maybe we could expose some property, which would enable long descriptions if necessary. Then enable it for  value map/relation editors

Withalion (approved):

Withalion:
Let's see in testing if we don't like this change somewhere

PR #4591, opened 2026-07-10
Rework handling of panning to identified features

uclaros:
This PR is a rework on how the map jumps to geometries.
Right now, a double animated pan happens is some cases while the feature form preview drawer is opening.
This is one thing handled by this PR: a single animation is performed while the drawer is opening.

The main change in behavior though is the following:
- We calculate the effective map canvas size with the form drawer open. This can change from layer to layer based on the attributes shown on the preview.
- We calculate a safe extent roughly 82% of that effective map size.
- When tapping on a feature on the map, highlight the feature and apply the following logic based on the location of the identified geometry:
  - If the geom **is already within** the safe extent, then **the map canvas does not pan**.
  - If the geom (or part of it) **is outside** the safe extent **but does fit within it**, then pan the map canvas so that **_the geom_ is centered** to the safe extent.
  - If the geom (or part of it) **is outside** the safe extent and **would not fit within it**, then pan the map canvas so that **_the tapped map location_ is centered** to the safe extent.

In all of the above the map canvas scale is unchanged and the pan is animated.
- When selecting a feature from a _list of features_ instead of the map canvas, then:
  - **Animate zooming** to, until the selected feature geom **fits centered** inside the safe extent

The advantages of this logic are that:
Identifying points or small geometries near the center of the (effective) screen does not modify the extent and the user does not need to re-orient himself (google maps does the same for POIs)
Scale is always preserved while navigating the map.
Identifying large geometries does not modify the map extent too much.

Testing notes:
- need to test all Geometries that we support (Points, Lines, Polygons and their multi variants)

PR #4595, opened 2026-07-13
Use bounding box intersections to avoid geos bug

uclaros:
This is **Plan B** from #4591


Withalion:
@uclaros can you fix the tests as well. I think only `testUtils` is necessary

Withalion (approved):

PR #4600, opened 2026-07-14
Bugfix/hide private layers from dropdown

### LINKED ISSUE #4486: Hide private layers from 'Add' list
When a spatial layer is set to private, it still appears as an option in the drop-down menu when the user selects the ‘Add’ button in map view. Being able to hide layers from the drop-down options would be useful for cases when you only want a user to add features as a child to a parent feature.

Missing `bool isPrivate = layer->flags() & QgsMapLayer::LayerFlag::Private;` in `RecordingLayersProxyModel::filterAcceptsRow`

xkello:
### Description
Layers set to private in QGIS were still visible in the add feature layer dropdown, which was inconsistent with the Layers tab, where private layers are correctly hidden.

Fixes: https://github.com/MerginMaps/mobile/issues/4486

### Problem
Users could still record new features on layers marked as private. This caused issues for admins who wanted to restrict users to adding features only as children of parent features, since the private-layer restriction was effectively bypassed in the recording flow.

### What changed
1. Added a check in `recordinglayersproxymodel` - `bool isPrivate = layer->flags() & QgsMapLayer::LayerFlag::Private` - for each layer. The app now checks whether a layer is private and excludes it from the dropdown if so.
2. Added the same guard to `activeproject`'s `recordingAllowed`. Even though the dropdown itself was filtering correctly, the app still fetched all layers on init - meaning a private layer could still be auto-selected as the active layer if it was the only layer present, or if it happened to be first in the layer list.

### Behaviour
On startup, the app now filters out private layers when iterating through available layers. When the add button is tapped, private layers no longer appear in the dropdown, matching the behavior of the Layers tab. If no eligible layers remain after filtering, the app shows the existing "No editable layers found" message.

### DIFF
--- a/app/activeproject.cpp
+++ b/app/activeproject.cpp
@@ -608,14 +608,11 @@ bool ActiveProject::recordingAllowed( QgsMapLayer *layer ) const
   if ( layer->readOnly() )
     return false;
 
-  if ( layer )
-  {
-    bool isPrivate = layer->flags() & QgsMapLayer::LayerFlag::Private;
+  bool isPrivate = layer->flags().testFlag( QgsMapLayer::LayerFlag::Private );
 
-    if ( isPrivate )
-    {
-      return false;
-    }
+  if ( isPrivate )
+  {
+    return false;
   }
 
   return QgsMapLayerProxyModel::layerMatchesFilters( layer, Qgis::LayerFilter::HasGeometry | Qgis::LayerFilter::WritableLayer ) && layer->id() != positionTrackingLayerId() && layer->id() != mapSketchesLayerId();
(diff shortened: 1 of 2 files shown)


app/activeproject.cpp
```diff
@@ -608,6 +608,16 @@ bool ActiveProject::recordingAllowed( QgsMapLayer *layer ) const
   if ( layer->readOnly() )
     return false;
 
+  if ( layer )
```
Withalion: this condition is always true

Withalion (commented):
After checking the `QFlags` API I would recommend using [testFlag](https://doc.qt.io/qt-6/qflags.html#testFlag) function instead of the and operator

Please rebase on current master as well

Withalion (approved):

PR #4602, opened 2026-07-15
Fix disappearing bg and wrong padding on widnows for reset button

xkello:
### Description
The filter reset button's background was disappearing on hover. This caused inconsistent behavior between platforms, since mobile devices did not exhibit this issue.

### What changed
1. Added `bgndColorHover` and `fontColorHover` to the action button, since `type: Tertiary` had its own styling that was overriding the defaults on hover. Explicitly setting these values aligns desktop behavior with mobile.
&nbsp; image

2. Swapped the order of the `if` statements. Previously, `type: Tertiary` was checked before `size: ExtraSmall`, causing the margin to be set to `0` instead of the intended value. Since margins were handled per-platform rather than in shared code, this meant the button had correct margins on mobile but appeared as a circle instead of a pill shape on Windows.
&nbsp; image

### Behaviour
The button no longer has hover-specific styling logic, so the background no longer disappears on mouse hover. Margins are also fixed, giving the action button its intended pill shape.

image

### DIFF
--- a/app/qml/components/MMButton.qml
+++ b/app/qml/components/MMButton.qml
@@ -166,8 +166,8 @@ Button {
 
   implicitWidth: {
     let margin = __style.margin20
-    if ( root.type === MMButton.Types.Tertiary ) margin = 0
-    else if ( root.size === MMButton.Sizes.ExtraSmall ) margin = __style.margin8
+    if ( root.size === MMButton.Sizes.ExtraSmall ) margin = __style.margin8
+    else if ( root.type === MMButton.Types.Tertiary ) margin = 0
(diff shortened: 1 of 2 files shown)


Withalion (approved):
There doesn't seem to be any behaviour now for hover, but we would have to discuss the hover behaviour further as the design system clashes with the design of this button

IvaKuklica:
Testing status: 🟢 
------------

xkello:
> There doesn't seem to be any behaviour now for hover, but we would have to discuss the hover behaviour further as the design system clashes with the design of this button

Yes I agree. I just mimicked the mobile behavior, but that naturally doesn't have any hover logic.

PR #4609, opened 2026-07-18
Fix literal wildcards in text filters

### LINKED ISSUE #4517: Text filter: % and _ treated as SQL wildcards instead of literal characters
**Description**

When entering `%` or `_` into a text filter input in the mobile app, all features are returned instead of treating these characters as literals. Both `%` and `_` are SQL wildcard characters and must be escaped before being used in a filter query — they should match only features whose field value literally contains `%` or `_`, not act as wildcards returning everything.

**Environment**
[ ] Staging  [x] PROD

Mergin Maps iOS 2026.2.0, iOS 26.0

**Steps to reproduce**
1. Open project `tester-ws-1/preview_mode_bug`
2. Navigate to the filter page
3. Open a text filter (e.g. `Text-POI-notes`) → enter `%` → apply
4. Reset filter → open the same text filter → enter `_` → apply

**Actual results**
All features with non empty "notes" field are returned for both `%` and `_` — the characters are interpreted as SQL wildcards (`%` = any string, `_` = any single character).

**Expected results**
`%` and `_` should be treated as literal characters. Only features whose field value literally contains `%` or `_` should be returned. They must not act as SQL wildcards.

**Screenshots or logs**
—

*Conclusion*
*Created by Claude and supervised by a human test analyst.*

floze-the-genius:
## Summary

- escape user-entered `!`, `%`, and `_` before substituting text filter values into `LIKE` expressions
- add an explicit `ESCAPE '!'` clause so `%` and `_` are matched literally while the surrounding `%...%` still performs substring matching
- add OGR/GeoPackage regression coverage for percent, underscore, the escape character itself, and combined `%_` input

## Root cause

Text filter values were quoted for SQL string safety, but SQL `LIKE` metacharacters were left unchanged. As a result, `%` matched any sequence and `_` matched any single character instead of the characters users typed.

Closes #4517.

## Validation

- project `astyle` check on all changed C++/header files
- `git diff --check`
- SQLite verification of the generated `LIKE ... ESCAPE '!'` patterns for `%`, `_`, `!`, and `%_`
- regression test uses a temporary copy of the project's real `roads.gpkg` through the OGR provider

The full Mergin Maps native test binary was not built locally because this host does not have the repository's Qt/QGIS/vcpkg toolchain.

## AI assistance disclosure

OpenAI Codex assisted with issue triage, implementation, regression-test drafting, and validation. The contribution was reviewed against the repository policies and the generated SQL behavior was independently checked locally.

### DIFF
--- a/app/filter/filtercontroller.cpp
+++ b/app/filter/filtercontroller.cpp
@@ -234,7 +234,11 @@ QString FilterController::buildFieldExpression( const FieldFilter &filter ) cons
       QString textValue = QgsExpression::quotedString( filter.value.toList().at( 0 ).toString() );
       // remove single quotes from the beginning and end of returned string
       textValue = textValue.slice( 1, textValue.size() - 2 );
+      textValue.replace( QStringLiteral( "!" ), QStringLiteral( "!!" ) );
+      textValue.replace( QStringLiteral( "%" ), QStringLiteral( "!%" ) );
+      textValue.replace( QStringLiteral( "_" ), QStringLiteral( "!_" ) );
       expressionCopy.replace( QStringLiteral( "@@value@@" ), textValue );
+      expressionCopy.append( QStringLiteral( " ESCAPE '!'" ) );
       break;
     }
     case FieldFilter::CheckboxFilter:
--- a/app/test/testfiltercontroller.cpp
+++ b/app/test/testfiltercontroller.cpp
@@ -13,6 +13,8 @@
 
 #include <QtTest/QtTest>
 #include <QDateTime>
+#include <QFile>
+#include <QTemporaryDir>
 
 #include <qgsproject.h>
 #include <qgsvectorlayer.h>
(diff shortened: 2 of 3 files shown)


floze-the-genius:
@Withalion I exact-rebased this onto current master 4c38f419 (new head 0b9a6d07; range-diff is identical). astyle 3.4.13 leaves all changed files unchanged, git diff --check passes, and the SQLite ESCAPE matrix for %, _, !, and %_ gives the expected literal-match counts. I could not run the full Qt/QGIS native binary on this host, and the repository workflows are still push-only for forks (#4624 is the draft that addresses this), so native CI remains a maintainer-side gate. Could you review when convenient?

app/filter/filtercontroller.cpp
```diff
@@ -234,7 +234,11 @@ QString FilterController::buildFieldExpression( const FieldFilter &filter ) cons
       QString textValue = QgsExpression::quotedString( filter.value.toList().at( 0 ).toString() );
       // remove single quotes from the beginning and end of returned string
       textValue = textValue.slice( 1, textValue.size() - 2 );
+      textValue.replace( QStringLiteral( "!" ), QStringLiteral( "!!" ) );
+      textValue.replace( QStringLiteral( "%" ), QStringLiteral( "!%" ) );
+      textValue.replace( QStringLiteral( "_" ), QStringLiteral( "!_" ) );
       expressionCopy.replace( QStringLiteral( "@@value@@" ), textValue );
+      expressionCopy.append( QStringLiteral( " ESCAPE '!'" ) );
```
Withalion, line 241: Create new function in `InputUtils` which will do these steps and just call it from here. Use backslash as the escape character instead. Also are these all the wildcards you can get?

app/test/testfiltercontroller.cpp
```diff
@@ -554,11 +556,73 @@ void TestFilterController::testTextFilter()
   filterValues[filterId] = QVariantList{ QStringLiteral( "great" ) };
   mController->processFilters( filterValues );
 
-  const QString expected = QStringLiteral( "(\"Condition\" LIKE '%great%')" );
+  const QString expected = QStringLiteral( "(\"Condition\" LIKE '%great%' ESCAPE '!')" );
```
Withalion, line 559: we are not escaping anything here

app/test/testfiltercontroller.cpp
```diff
@@ -554,11 +556,73 @@ void TestFilterController::testTextFilter()
...
+
+  const QString fieldName = QStringLiteral( "Condition" );
+  QgsVectorLayer *layer = new QgsVectorLayer(
+    layerPath + QStringLiteral( "|layername=roads" ),
+    QStringLiteral( "text-filter-wildcards" ),
+    QStringLiteral( "ogr" )
+  );
+  QVERIFY( layer );
+  QVERIFY( layer->isValid() );
+  QgsProject::instance()->addMapLayer( layer );
```
Withalion, line 595: What's wrong with the approach in other tests?

app/test/testfiltercontroller.cpp
```diff
@@ -554,11 +556,73 @@ void TestFilterController::testTextFilter()
...
+{
+  QTest::addColumn<QString>( "searchValue" );
+  QTest::addColumn<QString>( "expectedPattern" );
+  QTest::addColumn<qlonglong>( "expectedCount" );
+
+  QTest::newRow( "percent" ) << QStringLiteral( "%" ) << QStringLiteral( "%!%%" ) << 2LL;
+  QTest::newRow( "underscore" ) << QStringLiteral( "_" ) << QStringLiteral( "%!_%" ) << 2LL;
+  QTest::newRow( "escape-character" ) << QStringLiteral( "!" ) << QStringLiteral( "%!!%" ) << 1LL;
+  QTest::newRow( "combined-wildcards" ) << QStringLiteral( "%_" ) << QStringLiteral( "%!%!_%" ) << 1LL;
+}
```
Withalion, line 574: don't use this, just use the expected values inline

app/test/testfiltercontroller.cpp
```diff
@@ -554,11 +556,73 @@ void TestFilterController::testTextFilter()
...
+  QgsProject::instance()->addMapLayer( layer );
+
+  const QStringList values =
+  {
+    QStringLiteral( "plain" ),
+    QStringLiteral( "percent%only" ),
+    QStringLiteral( "under_score" ),
+    QStringLiteral( "bang!mark" ),
+    QStringLiteral( "both%_chars" )
+  };
```
Withalion, line 604: use values that could match values in condition

app/test/testfiltercontroller.cpp
```diff
@@ -554,11 +556,73 @@ void TestFilterController::testTextFilter()
...
+  for ( const QString &value : values )
+  {
+    QVERIFY( TestUtils::addFeatureToLayer( layer, fieldName, value ) );
+  }
```
Withalion, line 608: don't add more features to the layer there already are some, use them

app/test/testfiltercontroller.cpp
```diff
@@ -554,11 +556,73 @@ void TestFilterController::testTextFilter()
...
+
+  const QString sql = QStringLiteral( "\"Condition\" LIKE '%@@value@@%'" );
+  const QString filterId = TestUtils::setupControllerWithFilter(
+                             mController.get(), FieldFilter::TextFilter, layer->id(), fieldName, sql );
+  QVERIFY( !filterId.isEmpty() );
+
+  QVariantMap filterValues;
+  filterValues[filterId] = QVariantList{ searchValue };
```
Withalion, line 616: Apply the filter values one by one and check the results after each one

Withalion (commented):
You mentioned several times that you are not able to build the application on your host, what host do you have? The build infra is available for major OSes. If you haven't build the app how do you know it works? Have you run your tests?

Withalion:
Hi @floze-the-genius thank you for your time. I'll close this PR now as stale. Feel free to open it when you'll have time to continue working on this.

PR #4611, opened 2026-07-20
Update filtering for recording layer dropdown

### LINKED ISSUE #4419: Ability to reorder layers on the 'Add' drop down
Having the QGIS layer order option is great in the ‘Layers’ view, however I think workflow and user experience could be improved if there was the option to reorder the layers when adding a feature. Those are still maintained alphabetically.

xkello:
### Description
The **Layers** view already lets a project choose how layers are ordered - alphabetically, or matching the order set up in QGIS (the `Mergin/SortLayersMethod/Method` project setting). The **Choose Active Layer** dropdown shown when adding a feature didn't comply to this setting and was always alphabetical, which mismatched the workflow for projects that rely on a deliberate QGIS layer order.

Fixes: https://github.com/MerginMaps/mobile/issues/4419

### What changed
- Extracted the existing sort-method lookup out of `LayerTreeSortFilterModel::onSourceModelInitialized()` into a new static helper, `LayerTreeSortFilterModel::sortMethod(QgsProject*)`, so the setting is read from one place instead of being duplicated.
- Overrode `RecordingLayersProxyModel::lessThan()` to consult that same helper:
  - `Alphabetical` -> falls back to the existing alphabetical compare (`QgsMapLayerProxyModel::lessThan`).
  - `PreserveQgisOrder` -> ranks layers by their position in `QgsProject::layerTreeRoot()->findLayerIds()`, meaning the real QGIS layer-tree order, since the dropdown's source model (`LayersModel`) doesn't natively preserve that order the way the Layers tab's source model does.

### Behaviour
- No change for projects using alphabetical sorting.
- Projects configured for QGIS order sorting now see the active-layer dropdown (used when adding a feature) match the same layer order shown in the Layers tab, instead of always being alphabetical.

<table>
  <tr>
    <th>Qgis layer order</th>
    <th>Alphabetical</th>
  </tr>
  <tr>
    <td>image</td>
    <td>image</td>
  </tr>
</table>


app/layer/layertreesortfiltermodel.cpp
Withalion, line 1: I would keep this model as it was before

app/recordinglayersproxymodel.cpp
```diff
@@ -34,6 +36,22 @@ bool RecordingLayersProxyModel::filterAcceptsRow( int source_row, const QModelIn
   return mModel->data( index, LayersModel::LayerVisible ).toBool();
 }
 
+bool RecordingLayersProxyModel::lessThan( const QModelIndex &left, const QModelIndex &right ) const
+{
+  if ( !mModel || LayerTreeSortFilterModel::sortMethod( QgsProject::instance() ) == LayerTreeSortFilterModel::Alphabetical )
```
Withalion: Instead of creating a dependency to `LayerTreeSortFilterModel` just read the project property in this class when the `LayersModel` changes.

app/recordinglayersproxymodel.cpp
```diff
@@ -34,6 +36,22 @@ bool RecordingLayersProxyModel::filterAcceptsRow( int source_row, const QModelIn
...
+    return QgsMapLayerProxyModel::lessThan( left, right );
+
+  // preserve the layer order as arranged in the QGIS project's layer tree
+  QgsMapLayer *leftLayer = mModel->layerFromIndex( left );
+  QgsMapLayer *rightLayer = mModel->layerFromIndex( right );
+
+  if ( !leftLayer || !rightLayer )
+    return QgsMapLayerProxyModel::lessThan( left, right );
+
+  const QStringList treeOrder = QgsProject::instance()->layerTreeRoot()->findLayerIds();
```
Withalion: Using `QgsLayerTree::layerOrder()` fits here more

app/recordinglayersproxymodel.h
```diff
@@ -27,9 +27,17 @@ class RecordingLayersProxyModel : public QgsMapLayerProxyModel
     Q_PROPERTY( LayersModel *model READ model WRITE setModel NOTIFY modelChanged )
 
   public:
+    enum SortMethodEnum
+    {
+      PreserveQgisOrder = 0,
+      Alphabetical,
+    };
+    Q_ENUM( SortMethodEnum );
+
```
Withalion, line 35: Doesn't it make more sense to have `Alphabetical` first, if we use it by default?

I don't think using `Q_ENUM` here is necessary, is it used somewhere else besides this class?

Withalion (approved):
Just one more thing to solve, ready for testing after rebase

Withalion (approved):

PR #4620, opened 2026-07-22
Add scrollbar for desktop devices

### LINKED ISSUE #4127: Add scrollbar for desktop builds in lists

xkello:
### Description

Adds a proper scrollbar for desktop builds. Mobile builds rely on touch/flick scrolling and never had or needed a visible scrollbar, but on desktop there was no scroll indicator at all, and once turned on in `ScrollView` it was drawn over the left side of content instead of reserving space for itself. `ListView` based lists had no scrollbar handling at all, so the same content-overlap problem applied there too, across many pages.

Fixes: https://github.com/MerginMaps/mobile/issues/4127
Original PR: https://github.com/MerginMaps/mobile/pull/4145

### Problem

The obvious first approach was to just declare `ScrollBar.vertical` directly on the shared `MMListView` component, since `ListView`/`Flickable` auto-positions an attached scrollbar correctly on its own. The problem is that a `ListView`'s scrollbar is drawn as an overlay on top of its content - `Flickable` has no concept of `padding` the way a `Control` does, so there's nothing built in to reserve space for it.

In practice, that meant every delegate, header, footer, and section-delegate across the ~17 different lists in the app would need to know a scrollbar might be there and manually subtract its width from their own sizing to avoid being drawn underneath it. That's a lot of duplicated logic to keep in sync, easy to miss in a new or existing list, and it still didn't cover every case cleanly (e.g. footers/section headers don't automatically inherit a delegate's width-adjustment logic, so each one needed the same fix repeated separately). Wrapping each list in `MMScrollView` instead moves that reservation into one real `padding` value on a `Control`, so no list, delegate, or header needs to know a scrollbar exists at all.

### What changed

- Added `rightPadding` to the shared `MMScrollView` component so the scrollbar reserves its own space instead of overlapping content, headers, footers, etc.
- Wrapped every `ListView` in the app (via the shared `MMListView` component) in `MMScrollView`, rather than handling scrollbars per-list. `ScrollView` is able to reuse a `ListView` as its own scrollable content directly, so this didn't require a rewrite of any list, just a wrapper, and now every list in the app shares the exact same scrollbar behavior and styling from one place instead of duplicating it.
- The scrollbar now only appears on desktop builds. Mobile keeps its normal touch/flick scrolling untouched - only the visible scrollbar and its reserved padding are suppressed there.

### Behaviour

**Desktop** - scrollbar visible, correctly positioned on the right edge, content no longer drawn underneath it:

image

**Desktop** - scrollbar hides and shows as needed:

[Screencast From 2026-07-22 18-38-09.webm](https://github.com/user-attachments/assets/340bc5cb-c19d-4c83-9ba4-236c9fd64a01)

**Mobile** — no scrollbar shown, touch scrolling behaves exactly as before:

Screenshot_20260722-183046
Screenshot_20260722-183117


Withalion (commented):
Okay, so I'm not against the approach you chose it sounds good. However the idea is that if you are using `MMListView` component you are not supposed to know/care that it's a `ListView` wrapped in `ScrollView` or that you need to wrap it again in `ScrollView`. The API for `MMListView` should stay the same and if necessary the component needs to deal with scrollbar internally.

xkello:
> Okay, so I'm not against the approach you chose it sounds good. However the idea is that if you are using `MMListView` component you are not supposed to know/care that it's a `ListView` wrapped in `ScrollView` or that you need to wrap it again in `ScrollView`. The API for `MMListView` should stay the same and if necessary the component needs to deal with scrollbar internally.

Fair point, didn't think of this. Reverted files to their master version and updated `MMListView`. Had to add some properties as QMLs `ScrollView` does not support those.

app/qml/components/MMListView.qml
```diff
@@ -2,22 +2,66 @@
  *                                                                         *
  *   This program is free software; you can redistribute it and/or modify  *
  *   it under the terms of the GNU General Public License as published by  *
- *   the Free Software Foundation; either version 2 of the License, or     *
```
Withalion, line 5: probably a missclick

app/qml/components/MMListView.qml
Withalion, line 1: I'm thinking about wrapping the `ScrollView` in another `Item` and expose just that, so if used people won't accidentally modify `ScrollView` property instead of expected `Listview` property

Withalion (commented):
`MMFormPage` layout is broken. In `MMFormGalleryEditor` there should be some spacing between the photos and scrollbar, currently the scrollbar overlays the photos. `MMProjectStatusPage` not every element inside the `MMListView` gets it's width updated when scrollbar appears. Scrollbar is doing some weird things in `MMChangelogPage`, but it might be due to how the content is loaded.

There is a check in `code_style.yml` called `no_stackview`, which checks if there is any occurrence of `StackView` component in our code and throws error if it finds something. Make similar checks for `ListView` and `Scrollview`, fix any issues it finds so your PR will be complaint.

The approach looks good!

app/qml/components/MMListView.qml
xkello: > I'm thinking about wrapping the `ScrollView` in another `Item` and expose just that, so if used people won't accidentally modify `ScrollView` property instead of expected `Listview` property

Yeah, you're right.

`ScrollView` already carries `contentWidth`/`spacing`/`padding` (inherited from `Pane`/`Control`), so exposing it directly on `MMListView` means someone could set one of those thinking they're touching the real `ListView`, and get silently routed to the wrapper's own copy instead. Already hit this with `spacing` - `Control.spacing` is `FINAL` so I couldn't even alias around it.

Fix: root `MMListView` in a plain `Item` instead of `ScrollView`. `Item` is the base class `Control`/`Pane`/`ScrollView` build on top of, so it doesn't carry any of that - only what we explicitly alias gets exposed.

xkello:
> `MMFormPage` layout is broken. In `MMFormGalleryEditor` there should be some spacing between the photos and scrollbar, currently the scrollbar overlays the photos. `MMProjectStatusPage` not every element inside the `MMListView` gets it's width updated when scrollbar appears. Scrollbar is doing some weird things in `MMChangelogPage`, but it might be due to how the content is loaded.
> 
> There is a check in `code_style.yml` called `no_stackview`, which checks if there is any occurrence of `StackView` component in our code and throws error if it finds something. Make similar checks for `ListView` and `Scrollview`, fix any issues it finds so your PR will be complaint.
> 
> The approach looks good!

Thanks, went through all of these.

`MMFormPage`/toolbar layout - fixed. Turned out to be `MMToolbar` leaving a stale button behind when it swaps models on state change (readOnly -> edit); deferred the rebuild with `Qt.callLater` so old delegates finish clearing first.

`MMFormGalleryEditor` scrollbar overlay - fixed, photos now reserve space for the scrollbar instead of sitting under it.

`MMProjectStatusPage` width not updating - fixed. A few items were children of a `ColumnLayout` using a plain `width:` binding instead of `Layout.fillWidth`, which `ColumnLayout` doesn't reliably honor.

`MMChangelogPage` - left as is for now. Looks like it's from `ListView` estimating content height for off-screen items with variable-length descriptions, not something specific to this PR.

Added `no_listview`/`no_scrollview` checks to `code_style.yml` alongside `no_stackview`, and fixed the four raw `ScrollView` usages they caught (`ListView` was already clean).

.github/workflows/code_style.yml
```diff
@@ -101,3 +101,59 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Check for direct ListView instantiation in QML files
+        run: |
+          # MMListView.qml is the approved wrapper — exclude it from the search.
+          # Any other QML file with a bare "ListView" is using the raw Qt component
+          # instead of the wrapper, which breaks Squish test support.
```
Withalion, line 117: doesn't break Squish, they are just not supposed to use it :)

.github/workflows/code_style.yml
```diff
@@ -101,3 +101,59 @@ jobs:
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Check for direct ScrollView instantiation in QML files
+        run: |
+          # MMScrollView.qml and MMListView.qml are the approved wrappers (the
+          # latter uses a plain ScrollView internally) — exclude both from the
+          # search. Any other QML file with a bare "ScrollView" is using the raw
+          # Qt component instead of the wrapper, which breaks Squish test support.
```
Withalion, line 145: same thing as above

app/qml/components/MMListView.qml
```diff
@@ -8,16 +8,74 @@
  ***************************************************************************/
 
 import QtQuick
+import QtQuick.Controls
 
-ListView {
+// Drop-in replacement for ListView that also shows a scrollbar on desktop.
+// Rooted in a plain Item rather than ScrollView, so only the properties
+// aliased below are exposed - ScrollView/Pane/Control have their own
+// contentWidth/spacing/padding that would otherwise shadow ListView's.
```
Withalion, line 16:
```suggestion
// Drop-in replacement for ListView that also shows a scrollbar on desktop.
```

app/qml/form/editors/MMFormGalleryEditor.qml
Withalion, line 1: Don't shrink the photo cards here. Rather make the cards keep the original size and if scrollbar is necessary increase the `ListView` height by the margin.

app/qml/gps/MMSelectionDrawer.qml
```diff
@@ -91,12 +91,11 @@ MMComponents.MMDrawer {
...
     }
 
-    ScrollView {
+    MMComponents.MMScrollView {
         width: parent.width
         height: scrollRow.height
 
         ScrollBar.vertical.policy: ScrollBar.AlwaysOff
-        ScrollBar.horizontal.policy: ScrollBar.AlwaysOff
         ScrollBar.horizontal.interactive: true
```
Withalion, line 99: - let the `ScrollView` element decide the necessity of scrollbar
- I would remove the interactive property assignment

app/qml/gps/MMStakeoutDrawer.qml
```diff
@@ -118,11 +118,10 @@ MMDrawer {
       visible: distanceState.state === "closeRange"
     }
 
-    ScrollView {
+    MMScrollView {
       id: gpsScrollView
 
       width: parent.width
-      ScrollBar.horizontal.policy: ScrollBar.AlwaysOff
       ScrollBar.vertical.policy: ScrollBar.AlwaysOff
```
Withalion, line 125: Any reason why we want to force this?

app/qml/settings/MMAboutPage.qml
```diff
@@ -20,14 +20,12 @@ MMPage {
...
   pageHeader.title: qsTr( "About" )
 
-  pageContent: ScrollView {
+  pageContent: MMScrollView {
 
     width: parent.width
     height: parent.height
 
-    contentWidth: availableWidth // to only scroll vertically
     ScrollBar.vertical.policy: ScrollBar.AlwaysOff
```
Withalion, line 28: any reason why we want to force this?

Withalion (commented):
Looks good, just some minor things now

tomasMizera:
Thanks for the attempt, but the amount of changes and the impact is too risky to merge. Let's see if we'll be able to find a good solution to this..

PR #4635, opened 2026-08-04
Fix camera capture crash on devices with lower RAM

### LINKED ISSUE #2784: Adding photo attachment from camera crashes Input
Hello!

We're finding that adding a photo from the phone's camera crashes the Android app about half the time. Adding a photo from the gallery does not cause crashing. My assistant and I are both using 2021 Moto G Powers, in case that makes a difference. We have been experiencing the issue over the course of many weeks.

Thanks!

Katherine

### LINKED ISSUE #3538: App crashing after taking sequential photos
Text field form and "Take a picture" in Relations Editors on Android causes app freeze. There are two known ways to reproduce this issue:

Example project: [support/multiple_multiple_photos](https://app.merginmaps.com/projects/support/multiple_multiple_photos/)
### Steps to reproduce issue:
_First Method:_
1. In "Survey" form, add some text to "Notes" field.
2. Add a new feature to "Other photos" field.
4. Click on "Take a picture":   
   - The camera opens to take a photo.
   - The application crashes.
   - After capturing the photo, the camera reopens immediately for another photo.
4. Clicking the button to take this additional photo leads to the app freezing on the splash screen.

_Second Method:_
1. In "Survey" form.
2. Add a new feature to "photos" field.
3. Go to "external-pk" field and click on Link icon.
4. Change the link to another one.
5. Stay in the "external-pk" field and click on it (not on the Link icon area).
6. Add a new feature to "photos" field.
7. Click on "Take a picture":   
   - The camera opens to take a photo.
   - The application crashes.

xkello:
### Description

On some devices with heavy OEM camera stacks (confirmed on Motorola, 4GB RAM), tapping "add photo" -> camera reliably crashed the app. Root cause, confirmed via multiple `adb logcat` captures: launching the OS camera app spikes system-wide memory pressure, and Android's `lowmemorykiller` kills our own backgrounded process moments later - before the user can even take the photo. It's a silent OS process kill, not a Java exception (no `FATAL EXCEPTION` in any capture).

Fixes: https://github.com/MerginMaps/mobile/issues/2784

### What changed

- `CameraActivity.java`
  - Explicit `grantUriPermission()`/`revokeUriPermission()` for the capture `FileProvider` URI - some OEM camera apps (confirmed: Motorola) don't reliably honor `FLAG_GRANT_*` flags when the URI is passed via `EXTRA_OUTPUT` rather than `setData()`.
  - `onSaveInstanceState()`/resume branch in `onCreate()` - if the process is killed and recreated by Android to redeliver a pending camera result, resume from the saved `targetPath`/`cameraFile` instead of relaunching the capture or losing track of the file.
  - Null-guard on `cameraFile` in `onActivityResult()` - fails cleanly instead of a `NullPointerException` if state wasn't restored.
- `CameraForegroundService.java` (new)
  - Started right before the camera intent, stopped as soon as a result comes back. Keeps the process in a protected priority tier (`oom_score_adj`) for the duration the OS camera is in front, making the kill significantly less likely in the first place.
- `AndroidManifest.xml`
  - Registers `CameraForegroundService` (`foregroundServiceType="shortService"`, `stopWithTask="true"`).

### Behaviour

**Before:** on affected devices, opening the camera from a form would kill the app outright - the project reloads from scratch and the in-progress edit is lost.

**After:** the process is kept in a protected priority tier for as long as the camera is open, verified directly via `adb` - `oom_score_adj` stays at `0`/`50` during the camera wait, versus the `700` tier every observed kill happened at. If the process is still killed under extreme memory pressure, `CameraActivity` now resumes and completes the capture correctly instead of crashing or silently corrupting the saved file.

**Before fix:** https://github.com/MerginMaps/mobile/issues/2784#issuecomment-3323069209
**After fix:** 

https://github.com/user-attachments/assets/389461f9-e313-4ce8-8711-c70bf0b83422

### TLDR @Withalion 

Camera crash on some devices = OS killing our process under memory pressure while the OEM camera app is open, not a code bug. Fix is two-layered: `CameraForegroundService` reduces the odds of the kill (confirmed via `oom_score_adj` measurement, not just crash/no-crash testing), and `CameraActivity`'s save/resume logic makes it non-destructive if the kill happens anyway. No new user-facing UI - the foreground service's notification is intentionally not requested/shown (`POST_NOTIFICATIONS` not requested), since the user doesn't need to know this is happening.

### DIFF
--- a/app/android/src/uk/co/lutraconsulting/CameraActivity.java
+++ b/app/android/src/uk/co/lutraconsulting/CameraActivity.java
@@ -27,8 +27,6 @@ import android.net.Uri;
 import android.app.Activity;
 import android.content.Context;
 import android.content.Intent;
-import android.content.pm.PackageManager;
-import android.content.pm.ResolveInfo;
 import android.util.Log;
 import android.provider.MediaStore;
 import android.graphics.Bitmap;
@@ -42,8 +40,6 @@ import uk.co.lutraconsulting.OrientationSensor;
 public class CameraActivity extends Activity {
     private static final String TAG = "Camera Activity";
     private static final int CAMERA_CODE = 102;
-    private static final String KEY_TARGET_PATH = "targetPath";
-    private static final String KEY_CAMERA_FILE_PATH = "cameraFilePath";
 
     private String targetPath;
     private File cameraFile;
@@ -61,15 +57,6 @@ public class CameraActivity extends Activity {
                 Context.SENSOR_SERVICE);
         orientationSensor = new OrientationSensor(mSensorManager, null);
         orientationSensor.Register(this, SensorManager.SENSOR_DELAY_NORMAL);
-
-        if (savedInstanceState != null && savedInstanceState.containsKey(KEY_CAMERA_FILE_PATH)) {
-            // Process was killed and recreated while the camera app held the foreground.
-            // The capture is already in flight -- resume instead of relaunching it.
-            targetPath = savedInstanceState.getString(KEY_TARGET_PATH);
-            cameraFile = new File(savedInstanceState.getString(KEY_CAMERA_FILE_PATH));
-            Log.d(TAG, "Resumed after process recreation, cameraFile: " + cameraFile.getAbsolutePath());
-            return;
-        }
 
         targetPath = getIntent().getExtras().getString("targetPath");
         Log.d(TAG, "targetPath: " + targetPath);
@@ -90,20 +77,6 @@ public class CameraActivity extends Activity {
                         photoFile);
 
                 takePictureIntent.putExtra(MediaStore.EXTRA_OUTPUT, photoURI);
-                takePictureIntent.setFlags(Intent.FLAG_GRANT_WRITE_URI_PERMISSION
-                        | Intent.FLAG_GRANT_READ_URI_PERMISSION);
-
-                // Explicitly grant URI permission to every app that can resolve this intent.
-                // Required because the URI is passed via EXTRA_OUTPUT rather than setData(),
-                // and some OEM camera apps (confirmed: Motorola) don't reliably honor the
-                // FLAG_GRANT_* flags in that case.
-                List<ResolveInfo> resolvedActivities = getPackageManager()
-                        .queryIntentActivities(takePictureIntent, PackageManager.MATCH_DEFAULT_ONLY);
-                for (ResolveInfo resolveInfo : resolvedActivities) {
-                    grantUriPermission(resolveInfo.activityInfo.packageName, photoURI,
-                            Intent.FLAG_GRANT_WRITE_URI_PERMISSION | Intent.FLAG_GRANT_READ_URI_PERMISSION);
-                }
-
                 takePictureIntent.putExtra("__RESULT__", "takePictureIntent__RESULT__");
                 startForegroundService(new Intent(this, CameraForegroundService.class));
                 startActivityForResult(takePictureIntent, CAMERA_CODE);
@@ -116,15 +89,6 @@ public class CameraActivity extends Activity {
         }
 
         return;
-    }
-
-    @Override
-    protected void onSaveInstanceState(Bundle outState) {
-        super.onSaveInstanceState(outState);
-        outState.putString(KEY_TARGET_PATH, targetPath);
-        if (cameraFile != null) {
-            outState.putString(KEY_CAMERA_FILE_PATH, cameraFile.getAbsolutePath());
-        }
     }
 
     private File createImageFile(String targetPath) throws IOException {
@@ -158,14 +122,6 @@ public class CameraActivity extends Activity {
         }
 
         if (requestCode == CAMERA_CODE && resultCode == Activity.RESULT_OK) {
-            if (cameraFile == null) {
-                Log.e(TAG, "cameraFile is null in onActivityResult - lost capture state.");
-                Intent resultData = getIntent();
-                resultData.putExtra("__RESULT__", "Lost photo capture state.");
-                setResult(Activity.RESULT_CANCELED, resultData);
-                finish();
-                return;
-            }
             Log.d(TAG, "tmp exists: " + cameraFile.exists());
             Log.d(TAG, "tmp path: " + cameraFile.getAbsolutePath());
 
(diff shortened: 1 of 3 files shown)


app/android/src/uk/co/lutraconsulting/CameraActivity.java
```diff
@@ -77,7 +90,22 @@ protected void onCreate(Bundle savedInstanceState) {
...
+                // FLAG_GRANT_* flags in that case.
+                List<ResolveInfo> resolvedActivities = getPackageManager()
+                        .queryIntentActivities(takePictureIntent, PackageManager.MATCH_DEFAULT_ONLY);
+                for (ResolveInfo resolveInfo : resolvedActivities) {
+                    grantUriPermission(resolveInfo.activityInfo.packageName, photoURI,
+                            Intent.FLAG_GRANT_WRITE_URI_PERMISSION | Intent.FLAG_GRANT_READ_URI_PERMISSION);
+                }
+
                 takePictureIntent.putExtra("__RESULT__", "takePictureIntent__RESULT__");
+                startForegroundService(new Intent(this, CameraForegroundService.class));
```
Withalion, line 81: Shouldn't the `CameraForegroundService` be started before this one gets created in `AndroidUtils::callCamera`?

app/android/src/uk/co/lutraconsulting/CameraActivity.java
```diff
@@ -142,6 +192,23 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
...
+        if (cameraFile != null) {
+            try {
+                Uri photoURI = FileProvider.getUriForFile(this,
+                        "uk.co.lutraconsulting.fileprovider", cameraFile);
+                revokeUriPermission(photoURI,
+                        Intent.FLAG_GRANT_WRITE_URI_PERMISSION | Intent.FLAG_GRANT_READ_URI_PERMISSION);
+            } catch (IllegalArgumentException e) {
+                // cameraFile isn't covered by file_paths.xml -- nothing was granted, nothing to revoke
+            }
+        }
```
Withalion: I would remove this as well

app/android/src/uk/co/lutraconsulting/CameraForegroundService.java
```diff
@@ -0,0 +1,73 @@
...
+import android.app.NotificationChannel;
+import android.app.NotificationManager;
+
+/**
+ * Runs in the app's default process (unlike PositionTrackingService, which deliberately
+ * runs in its own :trackingThread) for as long as CameraActivity is waiting on the external
+ * camera app, to raise this process's priority and make it much less likely to be picked by
+ * the OS's low-memory killer while the (often heavy) OEM camera stack is in the foreground.
+ * This is a mitigation, not a guarantee -- under severe enough memory pressure the process can
+ * still be killed, which is what CameraActivity's saved-instance-state resume handles.
```
Withalion:
```suggestion
 * Runs in the app's default process for as long as CameraActivity is waiting on the external
 * camera app, to raise this process's priority and make it much less likely to be killed while the camera
 * is in the foreground.
 * This is a mitigation, not a guarantee -- under severe enough memory pressure the process can
 * still be killed, which is what CameraActivity's saved-instance-state resume handles.
```

app/android/src/uk/co/lutraconsulting/CameraForegroundService.java
```diff
@@ -0,0 +1,73 @@
...
+        PendingIntent pendingIntent = PendingIntent.getActivity(this, 0, notificationIntent, PendingIntent.FLAG_IMMUTABLE);
+
+        Notification notification = new Notification.Builder(this, CHANNEL_ID)
+                .setSmallIcon(R.drawable.ic_notification)
+                .setContentTitle("Waiting for photo")
+                .setColor(getResources().getColor(R.color.grassColor))
+                .setContentIntent(pendingIntent)
+                .build();
+
+        if (Build.VERSION.SDK_INT >= 35) { // Android 15 (Vanilla Ice Cream) -- FOREGROUND_SERVICE_TYPE_SHORT_SERVICE
```
Withalion:
```suggestion
        if (Build.VERSION.SDK_INT >= 35) {
```

app/android/src/uk/co/lutraconsulting/CameraForegroundService.java
```diff
@@ -0,0 +1,73 @@
...
+            startForeground(SERVICE_ID, notification, ServiceInfo.FOREGROUND_SERVICE_TYPE_SHORT_SERVICE);
+        } else {
+            startForeground(SERVICE_ID, notification);
+        }
+
+        // Not START_STICKY: if this service alone gets killed there is nothing useful to resume --
+        // it holds no state, it only exists to keep the process's priority elevated while
+        // CameraActivity waits on startActivityForResult().
+        return START_NOT_STICKY;
```
Withalion:
```suggestion
        // if this service alone gets killed there is nothing useful to resume, it only exists to 
        // keep the process's priority elevated while CameraActivity waits on startActivityForResult().
        return START_NOT_STICKY;
```

app/android/src/uk/co/lutraconsulting/CameraActivity.java
```diff
@@ -77,7 +90,22 @@ protected void onCreate(Bundle savedInstanceState) {
...
+                // Explicitly grant URI permission to every app that can resolve this intent.
+                // Required because the URI is passed via EXTRA_OUTPUT rather than setData(),
+                // and some OEM camera apps (confirmed: Motorola) don't reliably honor the
+                // FLAG_GRANT_* flags in that case.
+                List<ResolveInfo> resolvedActivities = getPackageManager()
+                        .queryIntentActivities(takePictureIntent, PackageManager.MATCH_DEFAULT_ONLY);
+                for (ResolveInfo resolveInfo : resolvedActivities) {
+                    grantUriPermission(resolveInfo.activityInfo.packageName, photoURI,
+                            Intent.FLAG_GRANT_WRITE_URI_PERMISSION | Intent.FLAG_GRANT_READ_URI_PERMISSION);
+                }
```
Withalion: I have mixed feelings about this and it feels sloppy

Withalion (commented):
There are no changes for `AndroidManifest`. The service won't get created without it. 
Please rebase on Master as well

IvaKuklica:
Testing note
---------

Device/OS/App: Motorola, Moto G62 5 G, Android 13 - 2026.3.3/846451


**Summary of testing:**

- When the user takes a photo using the camera, the camera action is triggered twice instead of once. - **the issue still persists.** 
- After confirming the photo the second time, the app doesn't return to the form — instead, only a green screen is displayed. This looks like a crash, but the app doesn't actually crash. **- app crash is fixed, but it doesn't behave correctly.**
- Instead of returning to the form screen (where the user would make further adjustments or tap Save to save the feature), the app goes straight back to the map canvas. - **the issue still persists.** 
- As a result, the feature is not saved and doesn't appear on the map canvas (main screen). - **new finding**
- If the project has multiple map themes and the user switches themes, then takes a photo with the camera and taps Save, the app returns to the map canvas but displays the previous map theme instead of the one currently selected. - **new finding**  
`Projet: tester-ws-1/tc04_recording`


https://github.com/user-attachments/assets/bb3b47bc-fd1d-439f-9a99-51bf65483f4e

Withalion (approved):

IvaKuklica:
Retest Round 2 (after fix) - All tests passed. It's ready to merge! :shipit:
------------------

Taking a photo from the camera now works correctly:
✅ the camera action triggers once, 
✅ the app returns to the form screen after confirming the photo, 
✅ the feature saves correctly and appears on the map canvas, 
✅ the correct map theme is displayed after saving.

Withalion:
@xkello please address my comments, update code comments where necessary and I'll merge it

PR #4641, opened 2026-08-09
Add Trimble position provider

Withalion:
resolves https://github.com/MerginMaps/product/issues/121

For easier integration with Trimble external GNSS devices a new provider has been added. The setup has been essentially changed to be "one-click". Trimble Mobile Manager (TMM) companion app is required for proper setup. 

New Workflows:

https://github.com/user-attachments/assets/e2da1df5-b722-47e0-9c6d-e836f41e41c1

https://github.com/user-attachments/assets/3bbfff83-64f3-4687-af17-d5772600eb31

Testing criteria:
- It's possible to add trimble external GNSS device 
- All information gets filled in GPS info Panel
- It's possible to manage provider
- It's not possible to change antenna height in the app, user should get redirected to TMM
- Elevation transformation is supported
- Geoid info is visible for both TMM provided height and transformed height
- Custom CRS setup in TMM is supported
- App tells you TMM is missing when it's missing


IvaKuklica:
Testing Comments — Android (Trimble Mobile Manager integration)
----------------------------

- [x] 🧨 **1. GPS connection status shows "connected" when Trimble receiver is not actually connected**

- When the user is NOT connected to the Trimble external GPS device (only the Trimble Mobile Manager app is open, with no device paired), the GPS info panel and the Manage GPS receiver screen both incorrectly show the status as "connected." 
- The app is not actually receiving data from the Trimble external GPS device in this state — only TMM itself is running. This is misleading, as it suggests an active GPS connection to the external receiver when none exists.  
- **The connection status should reflect whether the external Trimble device itself is connected, not just whether TMM is open.**



- [ ]  **2. Redirect to TMM on app launch**
When the app is opened, there's a brief redirect to TMM with a loading screen before Mergin Maps is displayed.

https://github.com/user-attachments/assets/d9f1e765-f676-48d1-983d-c6c107a41273

- [x]  **3. Satellites in use/view**
The number of satellites shown for both fields (in use/in view) stays the same, since there's only one value coming from TMM to display. 

https://github.com/user-attachments/assets/7ef58ba8-59bc-41b2-9fcc-5dd8a45fbc0a

Withalion:
Issues found in testing should be fixed now:
1. After successful connection to TMM the status is "No data", after connecting to Trimble device the state should change between "Connecting", "Connected", "Disconnecting", "Disconnected". After the device is disconnected the data is not hidden.
2.  The app should not redirect to TMM right away on startup, it will wait until project is loaded. The issue was more noticeable if TMM was not running, if it was running there was no app switch.
3. Issue has been fixed correct values are shown

RastoHu:
### Testing 2
- [x] If antenna height is set in TMM, the MM app does not display its value in the bubble on map canvas.
- [x] iOS: Connecting to a Trimble device causes an infinite switching loop between the TMM and MM apps.

app/position/providers/trimblepositionprovider.cpp
```diff
@@ -0,0 +1,416 @@
...
+  { QStringLiteral( "NAD83(CSRS)v7 (EPOCH:1997)" ), QStringLiteral( "EPSG:8254" ) },
+  { QStringLiteral( "NAD83(CSRS)v7 (EPOCH:2002)" ), QStringLiteral( "EPSG:8254" ) },
+  { QStringLiteral( "NAD83(CSRS)v7 (EPOCH:2010)" ), QStringLiteral( "EPSG:8254" ) },
+  { QStringLiteral( "NAD83(CSRS)v8 (EPOCH:1997)" ), QStringLiteral( "EPSG:10413" ) },
+  { QStringLiteral( "NAD83(CSRS)v8 (EPOCH:2002)" ), QStringLiteral( "EPSG:10413" ) },
+  { QStringLiteral( "NAD83(CSRS)v8 (EPOCH:2010)" ), QStringLiteral( "EPSG:10413" ) },
+  { QStringLiteral( "NAD83(MA11) (EPOCH:2010)" ), QStringLiteral( "EPSG:6324" ) },
+  { QStringLiteral( "NAD83(PA11) (EPOCH:2010)" ), QStringLiteral( "EPSG:6321" ) },
+  { QStringLiteral( "NZGD2000" ), QStringLiteral( "EPSG:4959" ) },
+  { QStringLiteral( "OSNetv2009" ), QStringLiteral( "EPSG:4277" ) },  // TODO: create 3D CRS here
```
tomasMizera, line 79: todo?

app/position/providers/trimblepositionprovider.cpp
```diff
@@ -0,0 +1,416 @@
...
+  emit positionChanged( GeoPosition() );
+}
+
+void TrimblePositionProvider::onSocketDisconnected()
+{
+  setState( tr( "Disconnected" ), State::NoConnection );
+  emit positionChanged( GeoPosition() );
+}
+
+void TrimblePositionProvider::openAntennaHeightPage()
```
tomasMizera, line 403: Why not bothering with this native code in `trimbleregistration`?

RastoHu:
### Testing 3
- [x] iOS/Android: When antenna height (AH) is set in TMM, TMM applies the height offset, but our app subtracts the AH value again (e.g., subtracting 5m twice)
- [x] Android 9: when user disconnects GNSS in TMM, our app should display status "No data" in GPS info panel, but it still show status "Connected" - bug: https://github.com/MerginMaps/mobile/issues/4674

Enhancement: https://github.com/MerginMaps/mobile/issues/4673

Withalion:
Let's make the Android issue as enhancement for future

app/position/providers/trimblepositionprovider.cpp
Withalion: It should go completely away either way, registration is not good anyway.

app/position/providers/trimblepositionprovider.cpp
Withalion: this will need to be rethought again in #4431

PR #4274, opened 2025-12-22
Fixed color selection for map sketching

gabriel-bolbotina:
Modified the components inside the MMSKetchesDrawer.

Below there is the latest UI:

https://github.com/user-attachments/assets/ffdc880c-8a66-4ed9-a5ba-4e220a1842b2

PR #4649, opened 2026-08-12
Fix map sketching points order

### LINKED ISSUE #3932: Map sketch geometries insert vertices at `0` instead of appending
When drawing on the canvas, new vertices should be appended at the end of both the screen space geometry and the map space geometry (used for highlights).
There should not be any visible difference to the user, though it will avoid copying all vertices data on each insert.

tomasMizera:
This PR reverts the order in which we add points to map sketching lines. Previously we prepended points, which resulted in points being stored in reversed order, starting with the last. Now we append points. This feels natural and it was a missed todo in map sketching code.

The request came from the community, see https://community.merginmaps.com/p/line-vertices - this will enable easy setup of arrow symbology.

There were a couple of JS runtime exceptions in the code that I fixed along the way:

On map sketching drawer open: 
```
qrc:/com.merginmaps/imports/MMInput/map/MMMapSketchesDrawer.qml:84:9: Unable to assign [undefined] to QColor
qrc:/com.merginmaps/imports/MMInput/map/MMMapSketchesDrawer.qml:83:9: Unable to assign [undefined] to QList<QColor>
```

On color selection:
```
qt.qml.context: qrc:/com.merginmaps/imports/MMInput/map/MMMapSketchesDrawer.qml:89:9 Parameter "newColor" is not declared. Injection of parameters into signal handlers is deprecated. Use JavaScript functions with formal parameters instead
```
---
No UI/UX change. Map sketching works just like before.

### DIFF
--- a/app/mapsketchingcontroller.cpp
+++ b/app/mapsketchingcontroller.cpp
@@ -17,6 +17,7 @@
 #include "qgsvectorlayerutils.h"
 #include "qgslinestring.h"
 #include "qgsmultilinestring.h"
+#include "qgsvertexid.h"
 
 
 MapSketchingController::MapSketchingController( QObject *parent )
@@ -49,10 +50,13 @@ void MapSketchingController::updateHighlight( const QPointF &oldPoint, const QPo
     mScreenPoints = QgsGeometry( new QgsLineString( { QgsPointXY( oldPoint.x(), oldPoint.y() ) } ) );
   }
 
-  // TODO: append instead of insert to zero
-  mScreenPoints.insertVertex( newPoint.x(), newPoint.y(), 0 );
+  // Append point to the geometry
+  const QgsVertexId lastScreenVertex( 0, 0, mScreenPoints.constGet()->vertexCount() );
+  mScreenPoints.get()->insertVertex( lastScreenVertex, QgsPoint( newPoint.x(), newPoint.y() ) );
+
   const QgsPoint p1 = mMapSettings->screenToCoordinate( newPoint );
-  mHighlight.insertVertex( p1, 0 );
+  const QgsVertexId lastHighlightVertex( 0, 0, mHighlight.constGet()->vertexCount() );
+  mHighlight.get()->insertVertex( lastHighlightVertex, p1 );
 
   emit highlightGeometryChanged();
 }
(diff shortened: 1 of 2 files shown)


app/mapsketchingcontroller.cpp
```diff
@@ -49,10 +50,13 @@ void MapSketchingController::updateHighlight( const QPointF &oldPoint, const QPo
     mScreenPoints = QgsGeometry( new QgsLineString( { QgsPointXY( oldPoint.x(), oldPoint.y() ) } ) );
   }
 
-  // TODO: append instead of insert to zero
-  mScreenPoints.insertVertex( newPoint.x(), newPoint.y(), 0 );
+  // Append point to the geometry
+  const QgsVertexId lastScreenVertex( 0, 0, mScreenPoints.constGet()->vertexCount() );
+  mScreenPoints.get()->insertVertex( lastScreenVertex, QgsPoint( newPoint.x(), newPoint.y() ) );
```
Withalion, line 55: Is there a reason why we don't use `insertVertex` from `QgsGeometry` instead of `QgsAbstractGeometry`?

tomasMizera: We use this approach in other places too, I believe it is because `QgsGeometry::insertVertex` does not support appending (despite saying so in docstrings..)

Withalion: okay makes sense

Withalion (approved):

IvaKuklica:
Testing note:
------
Application (+ app version, build, operating system)
DEV/OS/APP: IPhone 15/ iOS 26/ Build number:  2026.3.3 - 26.08.940611
DEV/OS/APP:  Android: Samsung Galaxy A53 / Android 15 / 2026.3.3 - 846551

- Tested map sketching point order on a basic QGIS project by styling the sketch line layer with a Marker Line symbol (arrow marker, placement "On last vertex", rotation following line direction).

- ✅ Point order fix works as expected: after drawing a sketch, the arrow marker consistently points in the direction of the drawing motion and ends up on the last vertex of the line — matching the direction the sketch was actually drawn.

Hi @tomasMizera 
> [!WARNING]
**Summary:**
When opening map sketching and selecting different colours, sketched objects/lines display as white while editing; after closing the map sketching feature all sketched lines display as black instead of retaining the selected colours.
**Related PRs:** #4649, #4578, #4274

Screenrecord:

https://github.com/user-attachments/assets/2bd6fb8b-bc16-48e1-9120-c5cfedad0cd8

IvaKuklica:
Testing note: Update
-------
✅ The application behaves as expected. 
Verified: the issue where sketched lines/objects displayed as white while editing and then changed to black after closing map sketching (instead of retaining the selected colour) is fixed.

PR #4654, opened 2026-08-19
Remove height override causing layout break

### LINKED ISSUE #4653: Long layer name in "Choose Active Layer" list causes incorrect layout of the row below it
**Description:**

In the "Choose Active Layer" bottom sheet (shown when starting to record a feature), if a layer's name is long enough to wrap onto two lines, the row for that layer does not get properly sized/padded to fit both lines, and this pushes the layout of the following row out of alignment.
_Project: tester-ws-1 / test_sort_layers_B_

Environment:
[ x } Staging  [ x ] PROD

**Application:**
Dev/OS/App: Samsung galaxy A53 / Android 15 / MM 2026.3.3 - 847551
Dev/OS/App: IPhone 15, iOS 26 / MM 2026.3.3 - 26.08.941611

**Steps to reproduce:**
1. Add a layer with a very long name to the project (long enough to wrap to two lines in the layer list), e.g. PolygonLayer_with_loooooooooooooooooooooooooong_name_again.
2. Start recording a new feature.
3. Open the "Choose Active Layer" selector.
4. Observe the row for the long-named layer and the row immediately below it.

**Expected behavior:**

The row height should expand to accommodate both lines of the wrapped layer name, with the icon vertically centered against the full two-line label, and the divider line placed cleanly below the complete row — consistent with how single-line rows are spaced.

**Actual behavior:**

The divider line sits right under the first line of the wrapped name instead of below the full two-line label, making it look like it's separating the two lines of the same name rather than separating this row from the next one. The layer icon is top-aligned to the first line only, instead of being vertically centered against the two-line label. This makes the row below the long-named layer appear squeezed / inconsistently spaced compared to the rest of the list

**Screenshots:**

Image

xkello:
### Description
The "Choose Active Layer" drawer broke visually for layers with long, multi-line names - rows overlapped instead of expanding.

Fixes: https://github.com/MerginMaps/mobile/issues/4653

### What changed
Removed a hardcoded `height: __style.menuDrawerHeight` on the drawer's list delegate in `MMMapController.qml`, letting `MMListDelegate`'s own content-based height (which already supports wrapped multi-line text) take over.

### Behaviour
Rows in the active-layer dropdown now grow to fit multi-line layer names instead of overlapping.

image

### DIFF
--- a/app/qml/map/MMMapController.qml
+++ b/app/qml/map/MMMapController.qml
@@ -925,9 +925,6 @@ Item {
       list.delegate: MMListDelegate {
         text: model.layerName
 
-        // TODO: why we need to set hight here?
-        height: __style.menuDrawerHeight
-
         leftContent: MMIcon {
           source: model.iconSource
         }


Withalion (approved):

IvaKuklica:
Testing note:
-------
Retested after fix: All layer names in the dropdown display correctly with consistent row spacing throughout the list — no squeezed or inconsistent spacing on rows following a long-named layer

PR #4663, opened 2026-08-24
Add wayland feature to QT

### LINKED ISSUE #4276: Enable wayland plugin for Qt
When migrating to newer version of Qt (6.9.3+) we should also enable wayland plugin for qtbase dependency. This change should improve UX for linux users as current mainstream DEs on Linux already migrated to wayland. We also supported this feature in previous versions before migrating to VCPKG.

Withalion:
fixes #4276

Another try to add this plugin. Let's see if we can make this work on all relevant Linux devices.

### DIFF
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -84,7 +84,13 @@
         },
         {
           "name": "qtbase",
-          "default-features": false
+          "default-features": false,
+          "features": [
+            {
+              "name": "wayland",
+              "platform": "linux"
+            }
+          ]
         },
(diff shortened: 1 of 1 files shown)


Withalion:
@uclaros can you retest if the build passes on your setup? Check `INSTALL.md` again as there might be some system packages missing, but they should be available for you from what I've seen.

uclaros (approved):
Installing `libwayland-dev` fixed it for me

PR #4664, opened 2026-08-25
Fix antenna height value sometimes showing floating point noise

Withalion:
The antenna height setting in _Settings_ sometimes didn't round the value to 3 decimal places and instead showed floating points noise as well (e.g. `1.45000000000000002`). Apparently QML was completely okay with use supplying double to string :). Added another guard (`InputUtils::formatNumber()`) to fix it.

### DIFF
--- a/app/appsettings.cpp
+++ b/app/appsettings.cpp
@@ -353,10 +353,10 @@ void AppSettings::setGpsAntennaHeight( const double gpsAntennaHeight )
   }
   if ( mGpsAntennaHeight != height )
   {
-    mGpsAntennaHeight = gpsAntennaHeight;
+    mGpsAntennaHeight = height;
     setValue( "gpsHeight", height );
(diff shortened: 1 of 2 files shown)


Withalion:
@IvaKuklica could you please test it, I couldn't reliably replicate this issue even on previous builds

xkello:
@Withalion can't we just update the setter and getter for `GpsAntennaHeight` in `appsettings`? Then there is no need for guard in `MMSettingsPage`. And if we ever need to reuse this, we don't have to worry about decimals.

Withalion:
@xkello the thing is that the C++ code already does it, but the issue was coming from JS <-> C++ transformation between number types. Another issue is that C++ side works with double, but `InputUtils::formatNumber()` returns `QString`

xkello (approved):

PR #4619, opened 2026-07-22
Build all libraries with 16kb pages support

Withalion:
The changes done in #4463 didn't build all dependencies with support of 16kb pages required by google play. The new version of NDK does that by default.

Also we are just creating overlay port for libaec to change the source repo see https://github.com/microsoft/vcpkg/issues/51921

PR #4666, opened 2026-08-25
Update QGIS to 3.44.13

Withalion:
If we are doing extensive testing due to #4619 we might as well update QGIS to latest release. Just patch updates, nothing major.


Withalion:
we can update to [3.44.14](https://github.com/qgis/QGIS/releases/tag/final-3_44_14) now

PR #4668, opened 2026-08-25
Bugfix/fix reuse last entered value

### LINKED ISSUE #4487: "reuse last entered value" not working for attachment/photos fields
App version 2026.2.0

Description:

form with more than one filed for photos. In our case 2 (say A and B) of them are mandatory.

The user start adding a new (point) feature (id 1) and take the picture for field A, checking the "reuse last entered value" checkbox, then fill the other fields and also take photo for field B.

After finishing id=1 the user surveys another point (id=2): the picture for field A is already in place, as is re-using the one taken previously.

At this point the user syncs edits, no errors.

If someone now syncs on QGIS Desktop sees being downloaded 3 photos, as expected

Image

but now inspecting the data the picture for field A in record id=1 is missing

Image

because simply the file does not exists. Looking at record id=2 is ok

Image

I believe this is a regression, as it used to work ok.

xkello:
### Description

Fixes a bug where enabling "Use last entered value" on a photo/file field causes the previous feature's photo file to be corrupted, renamed, or orphaned when reused by a new feature - most visibly when a custom filename expression is set in QGIS, where the previous feature's file gets physically renamed and reassigned to the new feature, leaving the previous feature with a broken/missing image.

Fixes: https://github.com/MerginMaps/mobile/issues/4487

### What changed

- `AttributeController::updateOnFeatureChange()` now clones the underlying file when a photo/file field's value is reused, instead of just copying the path string. Each feature gets its own independent file from the moment the drawer opens, so it's never possible for two features to end up sharing (and therefore corrupting) the same photo.
- The clone is tracked on the `FormItem` (`isReusedValue()`, `reusedCopyPath()`) so it can be cleaned up if it's never actually needed: `AttributeController::discardReusedPhotoCopies()` deletes it if the draft feature is discarded (cancel/rollback), or if the photo is deleted or replaced before the feature is saved.
- `renamePhotos()` now evaluates the project's custom filename expression fresh against the clone (instead of skipping reused fields entirely), so a reused photo gets a correctly computed, per-feature filename rather than a generic incremented name. A static expression that collides with an existing file still gets a proper incremental suffix.
- `recalculateDefaultValues()` no longer lets a QGIS field-level Default Value expression overwrite a value that was just populated by "reuse last entered value" while the form is open.
- `InputUtils::resolveTargetDir()` now resolves a relative root path expression/Default Path against the project home path. Previously a relative path (e.g. `'photos/' + @layer_name`) was used as-is, which could resolve to the wrong location on disk and produce a broken/missing image.
- Android (`MMActivity.importImage`) and iOS (`IOSGalleryPickerDelegate`) gallery import now create the destination directory before copying the picked image, matching what capture-from-camera already did.

### Behaviour

- Reusing a photo value for a new feature clones the file, so previously saved features keep their original, valid photo path - no more orphaned attachments or broken image icons after sync.
- Deleting or unlinking a reused photo from one feature no longer affects any other feature that reused it.
- Cancelling a draft feature, or deleting/replacing a reused photo before saving, cleans up the clone instead of leaving it behind.
- Custom filename expressions are evaluated fresh for each feature's cloned photo, with correct incremental suffixing when the expression is static.
- No impact on fields without a custom filename expression, and no impact on the "unsaved changes" dialog.

### TLDR @Withalion 

"Use last entered value" on a photo field now clones the file per feature instead of sharing/renaming the original, so reused photos never corrupt or orphan another feature's image, on save or on cancel.


app/attributes/attributecontroller.cpp
```diff
@@ -612,7 +612,8 @@ void AttributeController::updateOnFeatureChange()
       const QVariant newVal = feature.attribute( fieldIndex );
       mFormItems[itemData->id()]->setOriginalValue( newVal );
       mFormItems[itemData->id()]->setRawValue( newVal ); // we need to set raw value as well, as we use it in form now
-      if ( mRememberAttributesController && isNewFeature() ) // this is a new feature
+      itemData->setReusedValue( false );
```
Withalion, line 615: I would reuse the access approach above

app/attributes/attributecontroller.cpp
```diff
@@ -612,7 +612,8 @@ void AttributeController::updateOnFeatureChange()
...
+      if ( mRememberAttributesController && isNewFeature() && newVal.toString().isEmpty() )
```
Withalion, line 616: won't the new condition somehow mess up other fieldtypes? or was it just expected until now?

app/attributes/attributecontroller.cpp
```diff
@@ -791,7 +819,8 @@ void AttributeController::recalculateDefaultValues(
 
     bool shouldApplyDefaultValue =
       !defaultDefinition.expression().isEmpty() &&
-      ( isFirstUpdateOfNewFeature || ( isFormValueChange && defaultDefinition.applyOnUpdate() ) );
+      ( isFirstUpdateOfNewFeature || ( isFormValueChange && defaultDefinition.applyOnUpdate() ) ) &&
+      !item->isReusedValue();
```
Withalion, line 823: why is this necessary, what does it solve?

app/attributes/attributecontroller.cpp
```diff
@@ -1293,6 +1324,11 @@ bool AttributeController::save()
     disconnect( mFeatureLayerPair.layer(), &QgsVectorLayer::featureAdded, this, &AttributeController::onFeatureAdded );
   }
 
+  if ( rv )
+  {
+    discardReusedPhotoCopies( false );
+  }
```
Withalion, line 1330: what's the case we want to remove photocopies on successful save?

app/attributes/attributedata.h
```diff
@@ -152,6 +152,12 @@ class FormItem
     QVariant rawValue() const;
     void setRawValue( const QVariant &rawValue );
 
+    bool isReusedValue() const;
+    void setReusedValue( bool reused );
+
+    QString reusedCopyPath() const;
+    void setReusedCopyPath( const QString &path );
```
Withalion, line 159: can't we instead of this field reuse `rawValue`?

Withalion (commented):
Looks like in some cases the photo renaming doesn't work correctly. Either it get's renamed again, when using `now()` or the name somehow gets mixed with previous value as well. Have a look at this project `Matej-test/photo-naming-expr` and POI layer.

Renaming gets triggered on empty photo fields with reuse last entered value turned on.

My comment on the issue about unlinking/deleting is addressed by #4685?

Add some unit tests for using default value on photos in `testattributecontroller.h`

app/attributes/attributecontroller.cpp
xkello: This runs right after `updateOnFeatureChange()` sets the reused value, so without this guard, a field-level Default Value expression would immediately overwrite it (or re-clobber it on every `applyOnUpdate` re-eval) silently discarding the reused/cloned value before the form even shows it. Added a comment explaining this in `attributecontroller.cpp`.

app/attributes/attributecontroller.cpp
xkello: Fields with a custom filename expression already get `reusedCopyPath()` cleared by `renamePhotos()`, which runs before commit. This only catches the leftover case: a reused copy with no rename expression that got deleted/replaced before save - otherwise it'd be an orphaned file on disk. Added a comment at the call site.

app/attributes/attributedata.h
xkello: `rawValue()` gets overwritten by every edit (rename, replace, delete via `setFormValue()`), so it can't hold the original clone's path once superseded so we'd lose it right when we need it. But there was a duplication: `setFormValue()` already clears `isReusedValue()` on any change, so I swapped the manual path-recompute for a direct `!item->isReusedValue()` check instead.

app/attributes/attributecontroller.cpp
```diff
@@ -622,8 +623,36 @@ void AttributeController::updateOnFeatureChange()
                                         );
         if ( shouldUseRememberedValue )
         {
-          mFeatureLayerPair.featureRef().setAttribute( fieldIndex, rememberedValue );
-          itemData->setRawValue( rememberedValue );
+          QVariant valueToUse = rememberedValue;
+
+          if ( itemData->editorWidgetType() == QStringLiteral( "ExternalResource" ) && !rememberedValue.toString().isEmpty() )
```
tomasMizera, line 628: Check if `valueToUse` is a file

app/attributes/attributecontroller.cpp
```diff
@@ -622,8 +623,36 @@ void AttributeController::updateOnFeatureChange()
...
+            const QString prefix = InputUtils::resolvePrefixForRelativePath( config[ QStringLiteral( "RelativeStorage" ) ].toInt(), QgsProject::instance()->homePath(), targetDir );
+            const QString src = InputUtils::getAbsolutePath( rememberedValue.toString(), prefix );
+            const QFileInfo fi( src );
+
+            static const QRegularExpression trailingCounter( QStringLiteral( "\\s\\(\\d+\\)$" ) );
+            QString baseName = fi.completeBaseName();
+            baseName.remove( trailingCounter );
+            const QString canonicalName = fi.suffix().isEmpty() ? baseName : QStringLiteral( "%1.%2" ).arg( baseName, fi.suffix() );
+
+            const QString dst = CoreUtils::findUniquePath( InputUtils::getAbsolutePath( canonicalName, targetDir ), true );
```
tomasMizera, line 642: Set file name as the default timestamp, store it in the correct folder

app/attributes/attributecontroller.cpp
```diff
@@ -622,8 +623,36 @@ void AttributeController::updateOnFeatureChange()
...
+
+            if ( InputUtils::copyFile( src, dst ) )
+            {
+              valueToUse = InputUtils::getRelativePath( dst, prefix );
+              itemData->setReusedCopyPath( dst );
+            }
+          }
+
+          mFeatureLayerPair.featureRef().setAttribute( fieldIndex, valueToUse );
```
tomasMizera, line 651: Watch out, this code is run even if `InputUtils::copy` on line `644` fails

app/attributes/attributedata.h
tomasMizera, line 159: This part of the app should not be affected
```suggestion
```

app/ios/iosviewdelegate.mm
```diff
@@ -94,6 +94,8 @@ - ( void )picker:( PHPickerViewController * )picker didFinishPicking:( NSArray<P
     BOOL writeSuccess = NO;
     if ( url && !error )
     {
+      [[NSFileManager defaultManager] createDirectoryAtPath:targetDir withIntermediateDirectories:YES attributes:nil error:nil];
```
tomasMizera, line 97: If needed, please create a separate PR

app/android/src/uk/co/lutraconsulting/MMActivity.java
```diff
@@ -201,6 +201,7 @@ public String importImage(Uri imageUri, String targetPath) {
     String fileName = getFileName( imageUri );
     File newCopyFile = new File( targetPath + "/" + fileName );
     try {
+      newCopyFile.getParentFile().mkdirs();
```
tomasMizera, line 204: If needed, please create a separate PR

core/coreutils.cpp
```diff
@@ -165,7 +165,7 @@ void CoreUtils::appendLog( const QByteArray &data, const QString &path )
   }
 }
 
-QString CoreUtils::findUniquePath( const QString &path )
+QString CoreUtils::findUniquePath( const QString &path, bool splitOnLastDot )
```
tomasMizera, line 168: We do not need the `splitOnLastDot` parameter, that is the only correct behaviour :) 

Put this in a new PR and reference https://github.com/MerginMaps/mobile/issues/4135

app/attributes/attributecontroller.h
```diff
@@ -222,6 +222,7 @@ class  AttributeController : public QObject
      */
     bool allowTabs( QgsAttributeEditorContainer *container );
 
+    void discardReusedPhotoCopies( bool force );
```
tomasMizera, line 225: Let's drop this

app/attributes/attributecontroller.cpp
tomasMizera: +1 to remove this - let's keep the logic as it was

app/attributes/attributecontroller.cpp
```diff
@@ -1576,6 +1616,27 @@ void AttributeController::onFeatureAdded( QgsFeatureId newFeatureId )
   emit featureIdChanged();
 }
 
+void AttributeController::discardReusedPhotoCopies( bool force )
```
tomasMizera, line 1619: drop

tomasMizera (changes requested):

PR #4670, opened 2026-08-25
Reset missing position data if missing

Withalion:
This fixes issue where stale/untrue position data is shown to users if the data is missing from newer position update. We will clear the data now.

### DIFF
--- a/app/position/providers/internalpositionprovider.cpp
+++ b/app/position/providers/internalpositionprovider.cpp
@@ -173,45 +173,51 @@ void InternalPositionProvider::parsePositionUpdate( const QGeoPositionInfo &posi
     positionDataHasChanged = true;
   }
 
-  const bool hasSpeedInfo = localPosition.hasAttribute( QGeoPositionInfo::GroundSpeed );
-  if ( hasSpeedInfo && !qgsDoubleNear( localPosition.attribute( QGeoPositionInfo::GroundSpeed ), mLastPosition.speed ) )
+  const double newSpeed = localPosition.hasAttribute( QGeoPositionInfo::GroundSpeed ) ?
+                          localPosition.attribute( QGeoPositionInfo::GroundSpeed ) * 3.6 : std::numeric_limits<double>::quiet_NaN(); // convert from m/s to km/h
+  if ( !qgsDoubleNear( newSpeed, mLastPosition.speed ) )
   {
-    mLastPosition.speed = localPosition.attribute( QGeoPositionInfo::GroundSpeed ) * 3.6; // convert from m/s to km/h
+    mLastPosition.speed = newSpeed;
     positionDataHasChanged = true;
   }
 
-  const bool hasVerticalSpeedInfo = localPosition.hasAttribute( QGeoPositionInfo::VerticalSpeed );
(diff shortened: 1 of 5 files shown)


app/position/providers/internalpositionprovider.cpp
```diff
@@ -173,45 +173,51 @@ void InternalPositionProvider::parsePositionUpdate( const QGeoPositionInfo &posi
...
   {
-    mLastPosition.magneticVariation = localPosition.attribute( QGeoPositionInfo::MagneticVariation );
+    mLastPosition.magneticVariation = newMagneticVariation;
     positionDataHasChanged = true;
   }
 
-  const bool hasHacc = localPosition.hasAttribute( QGeoPositionInfo::HorizontalAccuracy );
-  if ( hasHacc && !qgsDoubleNear( localPosition.attribute( QGeoPositionInfo::HorizontalAccuracy ), mLastPosition.hacc ) )
+  const double newHorizontalAccuracy = localPosition.hasAttribute( QGeoPositionInfo::HorizontalAccuracy ) ?
+                                       localPosition.attribute( QGeoPositionInfo::HorizontalAccuracy ) : std::numeric_limits<double>::quiet_NaN();;
```
xkello, line 209: The double semicolon might be a typo. Same on line `217`

app/position/providers/internalpositionprovider.cpp
```diff
@@ -173,45 +173,51 @@ void InternalPositionProvider::parsePositionUpdate( const QGeoPositionInfo &posi
     positionDataHasChanged = true;
   }
 
-  const bool hasSpeedInfo = localPosition.hasAttribute( QGeoPositionInfo::GroundSpeed );
-  if ( hasSpeedInfo && !qgsDoubleNear( localPosition.attribute( QGeoPositionInfo::GroundSpeed ), mLastPosition.speed ) )
+  const double newSpeed = localPosition.hasAttribute( QGeoPositionInfo::GroundSpeed ) ?
+                          localPosition.attribute( QGeoPositionInfo::GroundSpeed ) * 3.6 : std::numeric_limits<double>::quiet_NaN(); // convert from m/s to km/h
+  if ( !qgsDoubleNear( newSpeed, mLastPosition.speed ) )
```
xkello: If both of these will be `NaN`, thats always false no? I think `qgsDoubleNear` won't ever return true if it is comparing `NaN` values. So it goes through even when it shouldn't and forces `positionDataHasChanged` to true. The attributes stay missing though and this stays like that on every consecutive update right? Is that intended? Same applies to other `if` statements bellow

xkello, line 178: If both of these will be `NaN`, thats always false no? I think `qgsDoubleNear` won't ever return true if it is comparing `NaN` values. So it goes through even when it shouldn't and forces `positionDataHasChanged` to true. The attributes stay missing though and this stays like that on every consecutive update right? Is that intended? Same applies to other `if` statements bellow

Withalion: `qgsDoubleNear` checks for `NaN` values through `qgsNumberNear` and if both values are `NaN` the function returns true, so this doesn't get triggered

xkello (approved):

RastoHu:
@Withalion @tomasMizera can we close this ticket and the rest of issues will be fixed in other tickets or mark as testing fail?

Withalion:
No let's leave it open and split the mentioned ticket into multiple actionable tickets and then proceed here.

PR #4671, opened 2026-08-25
Fix photo naming expression returning NULL

### LINKED ISSUE #4622: Photo naming expression returns NULL value in some cases
In some cases the photo naming expression will fail due to a NULL value in a field used in the naming expression. This results in a NULL value for the photo name. For some reason, the photo renaming check doesn't catch this and replace it with a valid character so it creates file names that are prohibited like this:

Image

Since there is no way to rename the photo in the app, the user is unable to sync. This can be further complicated if the device is an iOS device because it is not possible to access the project files and the unallowed filename prevents file transfer to another device. 

### To reproduce the issue:

1. Create a project with a survey layer that contains a photo field and at least one other text entry field like `name`
2. In project settings, set a photo naming expression that includes the text input field, example: `"name" || '-' || format_date(now(), 'yyyyMMdd_HHmmss')`
3. In the mobile app, create a feature but leave the 'name' field blank and attach a photo, this will create a photo with an empty space for a filename
4. Repeat the previous step to create a second photo with no filename, the device will rename it ` (1).jpg`
5. Try to sync, you should get the error but there is no way to rename the files and complete the sync

### Workaround
It is possible to avoid this by wrapping any field in the naming expression in a coalesce() function. For example, the previous expression would look like:
`coalesce("name", 'no_name') || '-' || format_date(now(), 'yyyyMMdd_HHmmss')`

It would be helpful if the app could catch these instances and either replace the null value with something else, give the user an option to rename the file, or revert to the device's default photo name.

xkello:
### Description

When a photo naming expression references a field that's blank, QGIS's `||` concatenation returns NULL for the whole expression. This NULL wasn't caught before being used as a filename, producing invalid names (a leading space, no real name) that the server rejects on sync with no way to fix it from the app.

Fixes: https://github.com/MerginMaps/mobile/issues/4622

### What changed

In `AttributeController::renamePhotos()`, after evaluating and sanitizing the naming expression, an empty/NULL result is now detected before it's turned into a filename. In that case the rename is skipped and the file keeps its original, valid name instead.


app/attributes/attributecontroller.cpp
```diff
@@ -1715,6 +1715,13 @@ void AttributeController::renamePhotos()
 
         InputUtils::sanitizePath( newName );
 
+        if ( newName.trimmed().isEmpty() )
+        {
+          CoreUtils::log( QStringLiteral( "Photo name format" ), QStringLiteral( "Expression for %1:%2 evaluated to an empty or NULL name, keeping the original file name" ).arg( mFeatureLayerPair.layer()->name(), field.name() ) );
```
Withalion, line 1720: It would be good to log Layer, feature and then field, to help support debug this issue.

Withalion (approved):

Withalion:
@xkello can you fix the history and conflicts here? I'll merge afterwards

xkello:
> @xkello can you fix the history and conflicts here? I'll merge afterwards

@Withalion guess its done?

PR #4685, opened 2026-08-28
Add check so unlink button won't show on a new feature

### LINKED ISSUE #4542: Unlinked photo stays unrenamed
**Description**
The photo naming expression is ignored when the photo is unlinked from the unsaved feature

**Environment**
Production

Teste on 2026.2.0, Android 16

**Steps to reproduce**
1. Create a field with custom photo naming in QGIS
2. Create a new feature in mobile
3. Take a photo in the form
4. Unlink the photo before saving the feature
5. Save the feature

**Actual results**
The photo ignores custom Photo naming settings and is created in the project folder with a default name.

**Expected results**
Either remove the photo entirely on Unlink
or do not even offer to unlink unsaved photo
or follow the custom naming.

**Screenshots**

Image

xkello:
### Description
The "Unlink photo" option in the photo delete drawer was showing even when the feature hadn't been saved yet. Since an unsaved feature has no persisted attachment relation to unlink from, the option didn't make sense there.

Fixes: https://github.com/MerginMaps/mobile/issues/4542

### What changed
Exposed a new `isNewFeature` property on `AttributeController` (reusing the existing `featureLayerPairChanged` signal so it stays reactive). `MMFormPhotoEditor` passes this down as `featureState` to `MMRemovePhotoDialog`, which now hides the "Unlink photo" button when `featureState === "add"`.

### Behaviour
On a new, unsaved feature, the delete photo drawer now only shows "Delete photo". On an existing, saved feature, both "Delete photo" and "Unlink photo" still show as before.

**Desktop:**

image

image 

<br/>**Android:**

Screenshot_20260828-123943

Screenshot_20260828-124002


app/qml/dialogs/MMRemovePhotoDialog.qml
uclaros: I think this message should also be conditionally changed to "Would you like to delete the photo?" to avoid confusion

app/qml/dialogs/MMRemovePhotoDialog.qml
```diff
@@ -18,6 +18,7 @@ MMDrawerDialog {
   signal unlinkImage()
 
   property string imagePath
+  property string featureState: ""
```
Withalion:
```suggestion
  property bool offerUnlinkOption: true
```
string is not state and it opens doors to undefined behaviour

app/qml/dialogs/MMRemovePhotoDialog.qml
```diff
@@ -35,6 +36,8 @@ MMDrawerDialog {
   secondaryButton {
     text: qsTr( "Unlink photo" )
 
+    visible: root.featureState !== "add"
```
Withalion:
```suggestion
    visible: root.offerUnlinkOption
```

Withalion (commented):
And also the comment from Stefanos

app/qml/dialogs/MMRemovePhotoDialog.qml
xkello: I agree, thank you for pointing that out

Withalion (approved):
Please rebase on current master, so iOS build will pass

IvaKuklica:
Testing status:
-------------------

**Tested on:**
iPhone 15, iOS 26
Samsung Galaxy A53, Android 16
Windows application

**Result:** ✅ Success on all platforms.
Verified that the "Unlink photo" option is correctly hidden when adding a photo to a new, unsaved feature, and remains available when editing an existing, previously saved feature — per the implemented fix.

PR #4687, opened 2026-08-28
Improved telemetry by adding anonymous usage reports

gabriel-bolbotina:
This enhancement has been done to better understand our app and which features are valuable to our users, as well as continuing on improving the experience overall. It also helps prioritise development and simplifies UX/ UI design. It is an **OPT-OUT enhancement** that stores no personal data.

## Summary of changes: 
- if **enabled**, the app now sends a weekly anonymous usage snapshot to our telemetry endpoint
- all analytics logic lives in main.cpp as a static function `trySubmitUsageSnapshot` + signal connections for each feature that is analysed
- Added **OPT-OUT** button in the app settings, which removes all the analysed data
- on app install each devices receives a telemetry id, that is **not linked** to device UUID or user account
- no username, no file paths nor other personal information are being stored

Resolves: https://github.com/MerginMaps/product/issues/127


app/main.cpp
```diff
@@ -387,6 +397,153 @@ void addQmlImportPath( QQmlEngine &engine )
 #endif
 }
 
+static const QString USAGE_REPORT_ENDPOINT = QStringLiteral( "https://api.merginmaps.com/mobile/usage-statistics" );
```
PeterPetrik, line 400: Wouldn't be better for long term maintenance of the code to have usage code that is now in main.cpp have separate file and/or maybe even some object such as `TelemetryManager`?

PR #4688, opened 2026-08-28
Add condition so readOnly child cannot be added from parent

### LINKED ISSUE #4640: Relations always show the "+" button even when the child table is read-only (as defined in the QGIS project properties) or a view
**Description**

I have a project with two 1:n relations.

In the first one the child table is a postgresql view, which is read-only by definition.
In the second the child table is a postgresql table which has been set as read-only in the QGIS project properties.

In MM App no features can be added or edited directly in this tables/views or when passing from the relation widget, as expected, BUT...

when navigating the form for the parent table, when one of this relations is shown, the "+" button is always shown... and indeed by pressing it it will open the form as it would for a table that is editable. This also happens for the view.

I believe this should not show when the child is not editable... also because in QGIS Desktop this works as expected... in the meantime...

Is there a way to hide the "+" button, for example the same way as the "nogallery" trick?

2026.3.2 Android 16.

**Steps to reproduce**
See above.

**Screenshots**

Image

Image

Image

Image

"+" button should not be there

Image

by pressing it MM App allows to add a feature to table that is read only (or even a view)

Image

xkello:
### Description

The "add" button for relation-linked child features (photo gallery tiles and relation bubbles) was shown based only on the user's project role, without checking whether the underlying child layer itself was read-only. This allowed users to attempt adding features to layers that were explicitly configured as read-only in the QGIS project.

Fixes: https://github.com/MerginMaps/mobile/issues/4640

### What changed

Extended the visibility condition for the add button in `MMFormGalleryEditor.qml` and the add relation button in `MMFormRelationEditor.qml` to also check `rmodel.layer.readOnly`, so the add button is hidden whenever the child layer is marked read-only, regardless of project role.

### Behavior

When the child layer is read-only, the add button is hidden and only existing linked child features remain visible. When the child layer is editable, the add button is shown as before (subject to the existing project role check). See attached screenshots for both editors, each with a read-only and non-read-only comparison alongside the corresponding QGIS layer settings.

**Linked feature list - child layer**

image

image

<br/>**Photo gallery - child layer**

image

image


Withalion:
Looks like it doesn't work for new features correctly:

https://github.com/user-attachments/assets/399a7869-fc63-4cd0-b2e8-1d7b95c0cf8c

The survey layer can link both photo and non-spatial layer. Non-spatial layer can also link photo layer. Non-spatial layer is set to be **read only**. It's possible to open the form of new feature for non-spatial layer, but impossible to save. When opening the for for non-spatial feature there is a visible switch between gallery and relation editor.

app/qml/form/editors/MMFormGalleryEditor.qml
Withalion, line 1: we should probably add a special case here when no pictures are linked and no are available to link

app/qml/form/editors/MMFormGalleryEditor.qml
```diff
@@ -75,7 +75,7 @@ MMPrivateComponents.MMBaseInput {
       }
     }
 
-    header: __activeProject.projectRole !== "reader" ? addFeatureComponent : null
+    header: (__activeProject.projectRole !== "reader" && !rmodel?.layer.readOnly) ? addFeatureComponent : null
```
Withalion, line 78:
```suggestion
    header: (__activeProject.projectRole !== "reader" && !rmodel?.layer?.readOnly) ? addFeatureComponent : null
```

app/qml/form/editors/MMFormRelationEditor.qml
```diff
@@ -86,7 +86,7 @@ MMPrivateComponents.MMBaseInput {
         height: privates.itemHeight
         radius: 8 * __dp
         color: __style.lightGreenColor
-        visible: __activeProject.projectRole !== "reader"
+        visible: (__activeProject.projectRole !== "reader" && !rmodel?.layer.readOnly)
```
Withalion, line 89:
```suggestion
        visible: (__activeProject.projectRole !== "reader" && !rmodel?.layer?.readOnly)
```

Withalion (commented):
See my comment above

Withalion (approved):
Looks good now!

IvaKuklica:
Testing result
---------------

**Setup:** 
Project: tester-ws-1/tc05_photo_relations
**Child layers** photos_nogallery and photos_gallery are non-spatial and both set to Read-only 
**Parent layer:** Survey_points.

image


**Scenario 1 — creating a new parent feature:** ✅ Works as expected.

- When creating a new feature in Survey_points, the Add button for child features is correctly not shown for either photos_nogallery or photos_gallery — no way to add a child feature to a read-only layer. 

**Scenario 2 — editing an existing parent feature:** 🔴 

- When editing an existing Survey_points feature, the Add button is also not shown by default for photos_nogallery — consistent with Scenario 1. 
- However, clicking directly on the child feature list area activates the Add button, and a new child feature can then be created. Attempting to save this new feature fails ("saving mode was not working").

**Result: Partially fixed.** The fix correctly hides the Add button for read-only child layers when creating a new parent feature, but the check is inconsistent when editing an existing parent feature — the Add button can still be triggered via the child feature list area, letting the user start creating a child feature that then cannot be saved.


https://github.com/user-attachments/assets/03cafdc2-151b-48ac-8b03-9dde699b8373

PR #4393, opened 2026-03-17
Fixed iOS photo GPS information 

gabriel-bolbotina:
Modified the image picker logic for gallery and camera inputs.
- the camera logic use UIImagePickerController
- the gallery picker now uses PHPickerViewController instead of UIImagePickerController API. This controller preserves all the EXIF metadata from photos, thus the GPS information is kept.

Added IOSGalleryPickerDelegate, a PHPickerViewControllerDelegate implementation that retrieves the original image metadata via NSItemProvider, which preserves GPS data and calls back into the Qt layer on the main thread

Tested on iOS 26 using the clone of [this project](https://app.merginmaps.com/projects/documentation/exif-metadata/tree?_gl=1*1p36gcn*_gcl_au*MjA0OTkyNTczOC4xNzY4Mzg3NTU1*_ga*MjU4MjMwNTA3LjE3NjgzODc1NTU.*_ga_RKTVJMQ7R2*czE3NzM3NTg1NTckbzIyJGcxJHQxNzczNzU4NTc2JGo0MSRsMCRoMA..)

Edit after QA: added date exif data for the photo taken with camera

PR #4690, opened 2026-09-01
Fix Incorrect handling of Lat and Long from EXIF data for S&W coords

### LINKED ISSUE #3685: Incorrect Handling of Latitude and Longitude from EXIF Data for S and W Coordinates
When dealing with Southern or Western latitudes and longitudes, which should always be negative (e.g., Belém (-1.455833, -48.503887) in S / W), the EXIF data is not processed correctly. As a result, the latitude and longitude fields are incorrectly set as positive ( using the expression read_exif_latitude(@project_home + '/' + "photo") ).

xkello:
### Description

`read_exif_latitude()` and `read_exif_longitude()` return incorrect (always positive) coordinates on Android for photos taken in the Southern or Western hemisphere. EXIF stores the coordinate magnitude in degrees/minutes/seconds as an unsigned value, and the hemisphere is stored separately in the `GPSLatitudeRef` and `GPSLongitudeRef` tags (`N`/`S` and `E`/`W`). The Android code path converts the magnitude to decimal degrees but never reads the ref tag, so the sign is lost. iOS was already fixed for this in commit PR #4393, Gabriel Bolbotina, May 2026, which negates the value when the ref indicates South or West.

Fixes: https://github.com/MerginMaps/mobile/issues/3685

### Behavior

**Before the fix**, a photo taken in Belem, Brazil (expected around `1.455833 S, 48.503887 W`) would populate a coordinate field with `(1.455833, 48.503887)` instead of `(−1.455833, −48.503887)`, placing the point in the wrong hemisphere entirely.

**After the fix**, Android matches iOS behavior: the ref tag is checked and the resulting decimal degree value is negated when appropriate, so coordinates in the Southern and Western hemispheres come out with the correct sign.

Screenshot_20260901-092622_Fake GPS Location

Screenshot_20260901-092510

### What changed

In `app/inputexpressionfunctions.cpp`, the Android branches of `ReadExifLatitude::func` and `ReadExifLongitude::func` now also read `GPSLatitudeRef` and `GPSLongitudeRef` via `AndroidUtils::readExif`, and negate the converted coordinate when the ref is `S` or `W` respectively, mirroring the existing iOS implementation.

### DIFF
--- a/app/androidutils.cpp
+++ b/app/androidutils.cpp
@@ -260,10 +260,11 @@ bool AndroidUtils::requestMediaLocationPermission()
 {
 #ifdef ANDROID
   const double buildVersion = QSysInfo::productVersion().toDouble();
-  // ACCESS_MEDIA_LOCATION is a runtime permission without UI dialog (User do not need to click anything to grant it, it is granted automatically)
-  if ( buildVersion < ANDROID_VERSION_13 )
+  // ACCESS_MEDIA_LOCATION was introduced in Android 10 and requires explicit user consent via a runtime permission dialog
+  if ( buildVersion >= ANDROID_VERSION_10 && !checkAndAcquirePermissions( QStringLiteral( "android.permission.ACCESS_MEDIA_LOCATION" ) ) )
   {
-    return checkAndAcquirePermissions( "android.permission.ACCESS_MEDIA_LOCATION" );
+    CoreUtils::log( QStringLiteral( "AndroidUtils" ), QStringLiteral( "Media location permission denied: GPS EXIF data will not be read from gallery-picked photos" ) );
+    return false;
   }
 #endif
   return true;
@@ -281,7 +282,7 @@ void AndroidUtils::callImagePicker( const QString &targetPath, const QString &co
   mLastCode = code;
   mTargetPath = targetPath;
 
-  // request media location permission to be able to read EXIF metadata from gallery image (only necessary for android < 14)
+  // request media location permission to be able to read unredacted EXIF metadata from gallery image
   // it is not a mandatory permission, so continue even if it is rejected
   requestMediaLocationPermission();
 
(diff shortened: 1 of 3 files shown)


Withalion (commented):
the code looks okay, but it doesn't work on my android phone only direction gets supplied for photos taken via camera and nothing gets supplied for photos from gallery

Android 15 phone, you can try it on this project `Matej-test/position-data` layer `POI`

xkello:
> the code looks okay, but it doesn't work on my android phone only direction gets supplied for photos taken via camera and nothing gets supplied for photos from gallery
> 
> Android 15 phone, you can try it on this project `Matej-test/position-data` layer `POI`

This looks like a mock location problem. It seems that due to some of our code parts, the app cannot read mocked exif data. And even if it does (when the exif data was edited manually), they are read as zeros. I asked Vitor to retest my fix with real exif data. Also we will have to discuss how we want to approach the mocked data problem.

xkello:
Update. The problem was with requesting access to media on android versions 10 and above. App did not request access. So it never got green light on reading exif data of the photos when user chose a photo from gallery. 

For the camera bug its a bit tricky. User needs to wait till native device camera picks up geodata from the phone. There is no info message that this succeeded. Generally waiting a few seconds should be enough. Solution could be to pull geodata from the device with MM app.

app/androidutils.cpp
```diff
@@ -309,10 +310,6 @@ void AndroidUtils::callCamera( const QString &targetPath, const QString &code )
 
   mLastCode = code;
 
-  // request media location permission to be able to read EXIF metadata from captured image
-  // it is not a mandatory permission, so continue even if it is rejected
-  requestMediaLocationPermission();
-
```
Withalion, line 315: why don't we ask for permission here before camera is called?

xkello: Because the app doesn't need permission to media to be able to access EXIF data of a photo. All it needs is a camera permission which we ask the user to provide. Tested on Android 16 and Android 10. Both work correctly.

All this permission does is ask for a not needed access to the media on the phone.

Withalion (approved):

PR #4691, opened 2026-09-02
Disable predictive back on android 16+ devices

Withalion:
Android 16 introduced new [_predictive back_ behaviour](https://developer.android.com/about/versions/16/behavior-changes-16#predictive-back), which instead of firing back button signal to the app instead switches apps/returns back to home screen. Let's disable it for now, so users will get the expected behaviour.

### DIFF
--- a/cmake_templates/AndroidManifest.xml.in
+++ b/cmake_templates/AndroidManifest.xml.in
@@ -48,6 +48,7 @@
             android:screenOrientation="unspecified"
             android:exported="true"
             android:theme="@style/Theme.App.Starting"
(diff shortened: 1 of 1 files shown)


Withalion:
Tested by @IvaKuklica on android 16 phone, the behaviour changed back to "return" inside app

xkello (approved):

IvaKuklica:
Testing note:
------------------

✅ **Testing OK** – the original app behaviour is preserved by the fix; back gesture/back button behaves in the app the same way as before Android 16.

PR #4700, opened 2026-09-12
Push v2 core implementation

### LINKED ISSUE #4008: Add support for concurrent push (v2)
Same as in https://github.com/MerginMaps/product/issues/84, but now for mobile.

tomasMizera:
This PR adds support for the new push mechanism - parallel project sync.

Implementation comments:
 - Transferred size is now updated with every chunk upload instead of every file
 - `TransactionStatus` now keeps a dedicated `pushChanges` member with all added, updated and removed files metadata. This is important as we inform the server about project changes only at the end of the sync (previously known as ~`pushFinish`)
 - Chunks in v2 are pregenerated with dummy IDs and swapped with server IDs after upload
 - Chunks upload is still sequential

What needs to be fixed in follow-up PRs:
 - Server feature flags (v2 push enabled and v2 pull enabled) are ignored now
 - Project ID is forced to a specific one now - need support for project IDs
 - Push cancel logic
 - Error handling of `check_only`
 - Further, robust, autotests suite

Fixes #4008


Withalion (approved):
LGTM 🚀

PR #4703, opened 2026-09-16
Introduce sync loop into synchronization manager

Withalion:
This branch consists of these changes:
- Server feature flag for pull/push v2 is read and saved
- New sync loop has been established in synchronization manager
  - every sync starts with pull, except initial project push to server
  - after pull, push is done with new limits for amount of files changed, maximum media file and maximum versionable file
  - if there are any unsynced changes start from step 1 (pull)
  - if any error occurs check if it is possible to retry and do so (try 10 times, with 5secs pauses)
- new server error code parsing has been added to error handling
- `hasLocalChanges` function has been refactored 
- better QML types integration from `SyncManager`

TODO:
- ~~`SynchronizationManager` with it's utility classes should be moved to `core`~~
- `MerginApi` tests are failing

### DIFF
--- a/app/synchronizationmanager.cpp
+++ b/app/synchronizationmanager.cpp
@@ -55,17 +55,14 @@ void SynchronizationManager::syncProject( const Project &project, SyncOptions::A
   CoreUtils::log( QStringLiteral( "Sync Manager" ), QStringLiteral( "Requested download of project %2" ).arg( project.mergin.projectName ) );
 
   // project is not local yet -> we download it for the first time
-  bool syncHasStarted = mMerginApi->pullProject( project.mergin.projectNamespace, project.mergin.projectName, auth == SyncOptions::Authorized );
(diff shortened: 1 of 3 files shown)


core/synchronizationerror.cpp
```diff
@@ -27,17 +27,37 @@ SynchronizationError::ErrorType SynchronizationError::errorType( int errorCode,
...
   {
     // Project no longer exists / is on different server
     return ErrorType::ProjectNotFound;
   }
-  else if ( errorCode >= 500 )
+  else if ( httpErrorCode == 409 )
+  {
+    //choose depending on serverErrorCode
+    if ( serverErrorCode == QStringLiteral( "ProjectsLimitHit" ) )
+    {
```
tomasMizera, line 44: Let's focus primarily on the error code instead of the http code

tomasMizera (approved):

PR #4709, opened 2026-09-19
Filter generated diff by size limits

Withalion:
The generated diff for push first gets filtered with new filters (max media file size, max versionable file size, max files in transaction) and then gets saved into transaction.


core/merginapi.cpp
```diff
@@ -359,111 +361,123 @@ void MerginApi::preparePushPayload( const QString &projectFullName )
...
-  for ( const QString &filePath : std::as_const( transaction.diff.localAdded ) )
+  for ( auto filePath = localDiff.localAdded.begin(); filePath != localDiff.localAdded.end(); ++filePath )
   {
-    MerginFile file = findFile( filePath, localFiles );
+    if ( fileCounter >= MAX_UPLOAD_CHANGES )
+    {
+      CoreUtils::log( "push " + projectFullName, QStringLiteral( "Maximum amount of changed files reached, push will be split to multiple versions" ) );
+      filePath = localDiff.localAdded.erase( filePath );
+      continue;
+    }
```
tomasMizera, line 386: If there are 200 files to upload, we would print this line 100 times in the log. How about removing all of the remaining files at once and `break` the loop?


```suggestion
    if ( fileCounter >= MAX_UPLOAD_CHANGES )
    {
      CoreUtils::log( "push " + projectFullName, QStringLiteral( "Maximum amount of changed files reached, push will be split to multiple versions" ) );
      filePath = localDiff.localAdded.erase( filePath, localDiff.localAdded.end() );
      break;
    }
```

core/merginapi.cpp
```diff
@@ -359,111 +361,123 @@ void MerginApi::preparePushPayload( const QString &projectFullName )
...
   {
-    MerginFile file = findFile( filePath, localFiles );
+    localDiff.localUpdated.clear();
+  }
+
+  for ( auto filePath = localDiff.localUpdated.begin(); filePath != localDiff.localUpdated.end(); ++filePath )
+  {
+    if ( fileCounter >= MAX_UPLOAD_CHANGES )
+    {
+      CoreUtils::log( "push " + projectFullName, QStringLiteral( "Maximum amount of changed files reached, push will be split to multiple versions" ) );
```
tomasMizera, line 426: Same comment here

core/merginapi.cpp
```diff
@@ -474,7 +488,9 @@ void MerginApi::preparePushPayload( const QString &projectFullName )
     fileCounter++;
   }
 
-  for ( QString filePath : transaction.diff.localDeleted )
+  transaction.diff = localDiff;
+
+  for ( const QString &filePath : transaction.diff.localDeleted )
```
tomasMizera, line 493: Don't we still need to `std::as_const` the array in order to prevent the potential detach?

```suggestion
  for ( const QString &filePath : std::as_const( transaction.diff.localDeleted ) )
```

tomasMizera (approved):

core/merginapi.cpp
Withalion: thanks for pointing it out!

core/merginapi.cpp
Withalion: There's actually no API like that, I solved it with simple while loop

